PR #256, opened 2015-04-30
Add tests for multiple column families in Put and Append.

AngusDavis:
This is not yet safe to merge (it will break integration tests), but I'd like to get it out there for review and so I don't forget it exists.


sduskis:
@vphani told me offline that this is safe to checkin.

sduskis:
LGTM

PR #265, opened 2015-05-02
Move ReconnectingChannel's async close to its own Executor.

AngusDavis:
Per the commit comments for shutdown() and awaitTerminated(), they can
sometimes hang. As a result, we should move close to its own
ExecutorService so that we don't block a Call from completing (which
could in turn complete channel shutdown).


sduskis:
@AngusDavis: I'm merging this.

PR #278, opened 2015-05-08
Fix comment on ReadRowsRequest.filter.

dsymonds:
If no filter is specified, all values are read, not just the most recent values.


sduskis:
Please update the protos from which these were generated.  We should not be changing these generated protos directly.

PR #279, opened 2015-05-08
Fix ReadRows and SampleRowKeys to be streaming methods.

dsymonds:
In production, these RPC methods return streams.


AngusDavis:
@sduskis - was the stream keyword removed by the proto export tools?

sduskis:
I cut and pasted files that the proto export tool.  I'm having trouble with the proto export tool now, so I can't confirm current behavior.

sduskis:
LGTM

PR #281, opened 2015-05-12
Updating to the public release of netty.

sduskis:
Also updating the source code to reflect the fact that Channel is now an abstract class rather than an interface.


kevinsi4508:
LGTM

PR #282, opened 2015-05-12
Regenerate protos

sduskis:


sduskis:
I had to update the protos because of a gRPC change.  They changed Channel from being an interface to being an abstract class.  All generated protos needed to be updated.

kevinsi4508:
LGTM

PR #283, opened 2015-05-13
Store bigtable-hbase project version in a properties file and make it…

kevinsi4508:
… accessible through BigtableConstants.java.

### DIFF
--- a/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/IntegrationTests.java
+++ b/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/IntegrationTests.java
@@ -68,6 +68,7 @@ public class IntegrationTests {
   public static final TableName TABLE_NAME;
   public static final int MAX_VERSIONS = 6;
   public static final byte[] COLUMN_FAMILY = Bytes.toBytes("test_family");
+  public static final byte[] COLUMN_FAMILY2 = Bytes.toBytes("test_family2");
   public static final Configuration BASE_CONFIGURATION = HBaseConfiguration.create();
 
   // testingUtility, connection, and configuration are provided via the connectionResource later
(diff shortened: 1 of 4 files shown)


bigtable-hbase/src/main/resources/com/google/cloud/bigtable/hbase/bigtable-hbase.properties
```diff
@@ -0,0 +1 @@
+project.version=${project.version}
```
AngusDavis: Consider making the property name more specific (bigtable.hbase.version)

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConstants.java
```diff
@@ -50,4 +55,37 @@
...
+        LOG.error("project.version not found in bigtable-hbase.properties.");
+      } else {
+        return value;
+      }
+    } catch (IOException e) {
+      LOG.error(
+          String.format("Error while trying to get project.version from bigtable-hbase.properties"),
+          e);
+    }
+    return "";
```
AngusDavis: I'd make this something that you can recognize (e.g., UNKNOWN)

bigtable-hbase/src/main/resources/com/google/cloud/bigtable/hbase/bigtable-hbase.properties
AngusDavis: Or perhaps a bigtable.hbase.ua=BigtableHbase/${project.version} Netty/${netty.version} gRPC/${grpc.version}

kevinsi4508: hm... the properties file belongs to com/google/cloud/bigtable/hbase package, do we still need to specify bigtable.hbase.version?
If we want to keep track of netty version and grpc version, maybe we should use different properties, could be in the same properties file.

sduskis: the project name and jar file are called bigtable-hbase.  We shouldn't need to specify the versions of hbase, netty and grpc, but rather the versions of our bigtable-hbase.jar, since that ought to be clear by looking at the pom file that generated the bigtable-hbase.jar.

As an aside, we also ought to ship our jar without grpc and netty.  Netty has a stable version out there and grpc's stable version is coming soon.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConstants.java
kevinsi4508: Set the default value to "dev-" version.

bigtable-hbase/src/main/resources/com/google/cloud/bigtable/hbase/bigtable-hbase.properties
kevinsi4508: done

kevinsi4508:
PTAL

sduskis:
LGTM

PR #286, opened 2015-05-14
Update User -Agent header to include bigtable-hbase project version

kevinsi4508:


bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/ChannelOptions.java
```diff
@@ -31,6 +31,7 @@
   public static class Builder {
     private Credentials credential;
     private String authority;
+    private String projectVersion;
```
sduskis: consider calling this something like userAgentName.

bigtable-grpc-interface/src/test/java/com/google/cloud/bigtable/grpc/ProjectVersionUpdaterInterceptorTest.java
```diff
@@ -0,0 +1,80 @@
...
+  private Call<MutateRowRequest, Empty> callStub;
+  @Mock
+  private Call.Listener<Empty> responseListenerStub;
+
+  private ProjectVersionUpdaterInterceptor interceptor;
+
+  @Before
+  public void setup() {
+    MockitoAnnotations.initMocks(this);
+    interceptor = new ProjectVersionUpdaterInterceptor("project_version");
```
sduskis: "project_version" should probably be a constant.

PR #288, opened 2015-05-14
Update User-Agent header to include bigtable-hbase project version.

kevinsi4508:


bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/ChannelOptions.java
```diff
@@ -180,6 +191,8 @@ public ChannelOptions build() {
    * @param credential A credential to use, may be null.
    * @param authority The authority to be passed in the HTTP/2 headers, or {@code null}
    * to use the default.
+   * @param userAgent The user-agent to be appended to HTTP/2 User-Agent header, or {@code null}
+   * to use the default.
```
sduskis: Perhaps an example here?  Something like bigtable-hbase/1.0.0 :)

sduskis:
LGTM

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/ChannelOptions.java
kevinsi4508: done

PR #291, opened 2015-05-14
Using javaAssist to throw unsupported exceptions for Admin.

sduskis:
Admin has differences in hbase 1.0 and hbase 1.1.  Bigtable doesn't support a whole bunch of operations in hbase 1.0 and non of the new hbase 1.1 methods.  Use JavaAssist to be able to auto-magically throw UnsupportedExcpetion for all of the methods we don't care about.

### DIFF
--- a/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/BigtableConnection.java
+++ b/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/BigtableConnection.java
@@ -18,7 +18,6 @@ package org.apache.hadoop.hbase.client;
 
 import java.io.Closeable;
 import java.io.IOException;
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.net.InetAddress;
 import java.util.Collections;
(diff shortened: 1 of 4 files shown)


bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/BigtableConnection.java
```diff
@@ -263,9 +269,46 @@ public RegionLocator getRegionLocator(TableName tableName) throws IOException {
...
+
   @Override
+  // Admin has different methods depending on which hbase version contains it.  Bigtable doesn't
+  // support a lot of methods in hbase Admin, since they do not apply to Bigtable's operations.
+  // for those methods, throw UnsupportedException.
   public Admin getAdmin() throws IOException {
-    return new BigtableAdmin(options, conf, this, bigtableTableAdminClient, this.disabledTables);
+    Object[] methodParams = { options, conf, this, bigtableTableAdminClient, this.disabledTables };
+    try {
+      return (Admin) ADMIN_FACTORY.create(ADMIN_CONSTRUCTOR_TYPES, methodParams, ADMIN_HANDLER);
```
kevinsi4508: This seems to work. Can you create test cases to test calls are forwarded to the super class and rejected when method is not implemented?

sduskis: We have plenty of good test cases for forwarded calls, although not a 100% coverage.  testing UnsupportedOperationException and no-ops seems like a good idea as well.

sduskis:
Carter didn't like this approach.  I'll close this PR in favor of #294

PR #294, opened 2015-05-15
Adding an abstract bigtable admin class.

sduskis:
This allows us to support hbase 1.1 and hbase 1.0.

### DIFF
--- a/bigtable-grpc-interface/pom.xml
+++ b/bigtable-grpc-interface/pom.xml
@@ -26,10 +26,6 @@ limitations under the License.
     <artifactId>bigtable-grpc-interface</artifactId>
     <packaging>jar</packaging>
 
-    <properties>
-        <netty.version>4.1.0.Beta5</netty.version>
-    </properties>
-
     <dependencies>
         <dependency>
             <groupId>${project.groupId}</groupId>
--- a/bigtable-hbase-1.0/pom.xml
+++ b/bigtable-hbase-1.0/pom.xml
@@ -1,0 +1,134 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Copyright 2014 Google Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.google.cloud.bigtable</groupId>
+        <artifactId>bigtable-client</artifactId>
+        <version>0.1.6-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>bigtable-hbase-1.0</artifactId>
+    <packaging>jar</packaging>
+
+    <properties>
+        <hbase.version>1.0.1</hbase.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.google.protobuf</groupId>
+            <artifactId>protobuf-java</artifactId>
+            <version>${protobuff-java.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>bigtable-hbase</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hbase</groupId>
+            <artifactId>hbase-client</artifactId>
+        </dependency>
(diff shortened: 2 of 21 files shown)


sduskis:
@AngusDavis: This was @carterpage's preferred approach to hbase 1.1 support.  Can you PTAL?

AngusDavis:
adding an hbase-1.0 artifact for consistency and so that only hbase-1.0 and hbase-1.1 have shade rules (bigtable-hbase would not, as an intermediary).

bigtable-grpc-interface/pom.xml
```diff
@@ -26,10 +26,6 @@ limitations under the License.
     <artifactId>bigtable-grpc-interface</artifactId>
     <packaging>jar</packaging>
 
-    <properties>
-        <netty.version>4.1.0.Beta5</netty.version>
-    </properties>
-
```
sduskis: FYI, this was moved up to the top level pom.xml.

sduskis:
Added comments and bigtable-hbase-1.0.  PTAL.

AngusDavis:
A few things - why does hbase-1.1 have an abstract admin? Shouldn't it be concrete? Or is the idea that you'll switch out abstract base classes depending on version? If that's the case, what guarantees that 1.1's will be included in the final jar and not the one defined for 1.0?

sduskis:
The abstract admin class in is indeed a bit of a "clever" semi black magic solution, but one Carter wasn't too fearful of.  The ordering is "guaranteed" by maven (I hope).  I'm planning on getting an hbase-1.1 test up and running in Jenkins once this gets merged to test out the hbase-1.1 integration, what ever it is.

I'm trying to avoid having different Connection implementations.  Perhaps we can check for the existence of an BigtableAdmin11 implementation in BigtableConnection, and if so, use that instead of BigtableAdmin.

AngusDavis:
Would it make sense to have 3 abstract admin classes, then? 
One in bigtable-hbase - this is not meant to make it into a compiled executable and has comments and code to enforce that (Throws on construction?)
One in each bigtable-hbase-1.0 and one in bigtable-hbase-1.1.

sduskis:
I'd rather have a working bigtable-hbase.

I added tests to against bigtable-hbase 1.0 and 1.1 for the existence (or non-existence in the case of hbase 1.0) of a method that ought to be in Admin in hbase 1.1.

sduskis:
PTAL

AngusDavis:
My guess is that having a 'working' bigtable-hbase will lead to confusion. If the only artifact people should use are hbase-1.0 or hbase-1.1.

Are you trying to maintain test compatibility? Would having a test version of the class in the src/test tree make sense (and work)?

sduskis:
BigtableConnection and BigtableAdmin (or atleast equivalents) are now abstract in bigtable-hbase.  Only the bigtable-hbase-1.x have concrete implementations.

sduskis:
PTAL

bigtable-hbase-1.1/src/main/java/org/apache/hadoop/hbase/client1_1/BigtableConnection.java
```diff
@@ -0,0 +1,79 @@
...
+            throw new UnsupportedOperationException("setQuota");  // TODO
+          }
+
+          @Override
+          public QuotaRetriever getQuotaRetriever(QuotaFilter filter) throws IOException {
+            throw new UnsupportedOperationException("getQuotaRetriever");  // TODO
+          }
+    };
+  }
+
```
AngusDavis: Nit: Extra whitespace

AngusDavis:
In general, looks good. Small nit on formatting (same comment applies to a couple of new classes).

A question - the package was o.a.h.hbase.client originally due to interface visibility. Does it make sense to change this to com.google.cloud (I think the interfaces are now public, but could be wrong) with this change (at least for the hbase versioned implementations).

sduskis:
Updated hte package name and fixed the spacing issue.  o.a.h.hbase.client is still needed for the abstract class, since it uses package protected classes.

bigtable-hbase-1.0/pom.xml
```diff
@@ -0,0 +1,134 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.google.cloud.bigtable</groupId>
+        <artifactId>bigtable-client</artifactId>
+        <version>0.1.6-SNAPSHOT</version>
```
mgarolera: Both clients have version 0.1.6-SNAPSHOT. This will make that both clients, 1.0 and 1.1, will send the same version to the server. We want to track the clients separately.

Maybe $artifactId-0.1.6-SNAPSHOT?

sduskis: The standard for jar name is <groupId>/<artifactId>-<version>.jar.  Pom.xml files will always refer to group + artifact Id + version.  Since artifact Id is always present, it will be confusing to duplicate artifactId in the version number.

mgarolera: I wasnt thinking about the jars, as long as they have different names Im fine with it.

I was talking about the version sent to the server on requests: https://github.com/GoogleCloudPlatform/cloud-bigtable-client/commit/eac6601610541b63b88b36f31531f23a37e8187d

We want to be able to distinguish the clients.

sduskis: Yes... we should pass along artifact id in the http headers.  I'll add another PR for that.

PR #298, opened 2015-05-20
Add logging to the BigtableConnection constructor.

AngusDavis:
This adds error-level log lines when an exception is raised from the
BigtableConnection constructor. This is intended primarily to aid in
debugging hbase shell issues.


sduskis:
LGTM

PR #304, opened 2015-05-28
Adding hbase 1.0 and 1.1 user agents.

sduskis:


bigtable-hbase-1.1/src/main/resources/com/google/cloud/bigtable/hbase/bigtable-hbase.properties
```diff
@@ -0,0 +1 @@
+bigtable.hbase.version=${project.artifactId}-${project.version}
```
kevinsi4508: if we include artifact ID, do we still want to call it version?

bigtable-hbase/src/test/resources/com/google/cloud/bigtable/hbase/bigtable-hbase.properties
```diff
@@ -0,0 +1 @@
+bigtable.hbase.version=${project.artifactId}-${project.version}
```
kevinsi4508: why this is needed?

sduskis: This ought to help us identify TestClusterAPI tests.

bigtable-hbase-1.1/src/main/resources/com/google/cloud/bigtable/hbase/bigtable-hbase.properties
sduskis: Good point.  I'll update the key.

bigtable-hbase/src/test/resources/com/google/cloud/bigtable/hbase/bigtable-hbase.properties
kevinsi4508: It is not used yet, please don't add it. This is file is not filtered as the one in main/resources

kevinsi4508:
Both bigtable-hbase-1.0/pom.xml and bigtable-hbase-1.1/pom.xml have the same artifactId and version. This mean bigtable-hbase-1.0 and bigtable-hbase-1.1 are going to send the same user_agent header. I don't think this is what we wanted.

sduskis:
I updated the code as per most of the comments.

I'm not sure what you meant by the last comment.  bigtable-hbase-1.1/pom.xml has an artifactId of  bigtable-hbase-1.1 and  bigtable-hbase-1.0/pom.xml has an artifactId of  bigtable-hbase-1.0.  What am I missing?

sduskis:
FYI, here's the results...

$ find . -name '*.properties' | xargs grep bigtable  | grep 0.1.6
./bigtable-hbase-1.0/target/classes/com/google/cloud/bigtable/hbase/bigtable-hbase.properties:bigtable.hbase.user_agent=bigtable-hbase-1.0-0.1.6-SNAPSHOT

./bigtable-hbase-1.1/target/classes/com/google/cloud/bigtable/hbase/bigtable-hbase.properties:bigtable.hbase.user_agent=bigtable-hbase-1.1-0.1.6-SNAPSHOT

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConstants.java
```diff
@@ -59,7 +59,7 @@
   /**
    * A User-Agent token to be added to User-Agent request header.
    */
-  public static final String USER_AGENT = "bigtable-hbase/" + getVersion();
+  public static final String USER_AGENT = getVersion();
```
kevinsi4508: Can you update the code to getUserAgent() and update the code there to use the right property name?

kevinsi4508:
Sorry! I was looking at parent artifactId :P

This means we have following two different artifact IDs for bigtable-hbase:
bigtable-hbase-1.0 and bigtable-hbase-1.1

and with the same version: 0.1.6-SNAPSHOT

I don't like what we have. I thought the artifact ID should stay the same and the versions are different. Two different artifact IDs with the same version is different from same artifact with two different versions. If this is what you have decided to use, I am fine with it.

sduskis:
Updated the method name.  Good catch.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConstants.java
```diff
@@ -59,13 +59,13 @@
   /**
    * A User-Agent token to be added to User-Agent request header.
    */
-  public static final String USER_AGENT = "bigtable-hbase/" + getVersion();
+  public static final String USER_AGENT = getUserAgent();
 
   /**
    * Gets current project version from bigtable-hbase.properties. Returns a default dev version with
```
kevinsi4508: Update comment.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConstants.java
```diff
@@ -76,9 +76,9 @@ private static String getVersion() {
 
       Properties properties = new Properties();
       properties.load(stream);
-      String value = properties.getProperty("bigtable.hbase.version");
+      String value = properties.getProperty("bigtable.hbase.user_agent");
       if (value == null) {
-        LOG.error("bigtable.hbase.version not found in bigtable-hbase.properties.");
+        LOG.error("bigtable.hbase.user_agent not found in bigtable-hbase.properties.");
       } else if (value.startsWith("$")){
         LOG.info("project.version token is not replaced.");
```
kevinsi4508: update log to mention ${project.artifactId}-${project.version} or just "Tokens are not replaced".

kevinsi4508:
Other than the different artifact IDs I don't like, I am fine with just PR.  I think it is doable to have different versions of bigtable-hbase in a same github project but not under the same parent project.

Please ask Angus to take a second look.
LGTM

sduskis:
Merging this.  The questions @kevinsi4508 has are about the project structure rather than the user agent header.   If we change the project structure, the proposed code in this PR ought not change.

PR #305, opened 2015-05-28
Commenting out streamWindowSize for testing purposes.

sduskis:
I'm doing this for a temporary performance test on jenkins.


AngusDavis:
This seems like the wrong way to go about this (assuming you're intending to merge this). 
Why not just spin up a VM in the same zone and run tests with a locally built JAR?

sduskis:
You're right.  I tested locally.  That was enough of a performance test to see that streamWindowSize is meaningful.  It might have even been a better test than running on jenkins.

PR #306, opened 2015-05-28
Fast array comparison in TestBasicOps

sduskis:
The Assert.assertArraysEquals takes 2 seconds because it assumes objects.  The Arrays.equals() method is a few milliseconds using bytes.


AngusDavis:
LGTM

PR #307, opened 2015-05-29
Updating project dependencies.

sduskis:


bigtable-grpc-interface/pom.xml
```diff
@@ -43,35 +43,8 @@ limitations under the License.
             <artifactId>grpc-all</artifactId>
             <version>${grpc.version}</version>
         </dependency>
-        <dependency>
-            <groupId>io.netty</groupId>
-            <artifactId>netty-all</artifactId>
-            <version>${netty.version}</version>
-        </dependency>
```
AngusDavis: We have a few code references to netty. We should maintain a direct reference and not just depend on transitive dependencies resolving properly for us.

bigtable-hbase/pom.xml
```diff
@@ -41,19 +41,12 @@ limitations under the License.
             <groupId>io.netty</groupId>
             <artifactId>netty-all</artifactId>
             <version>${netty.version}</version>
+            <scope>test</scope>
```
AngusDavis: There is netty code referenced from non-test code. I don't think test is the correct scope.

bigtable-hbase/pom.xml
```diff
@@ -81,6 +74,25 @@ limitations under the License.
...
+            <groupId>javax.validation</groupId>
+            <artifactId>validation-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hbase</groupId>
+            <artifactId>hbase-server</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.google.api-client</groupId>
+            <artifactId>google-api-client</artifactId>
```
AngusDavis: I think this is specified twice.

bigtable-hbase/pom.xml
```diff
@@ -154,22 +154,17 @@ limitations under the License.
...
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>com.google.api-client</groupId>
             <artifactId>google-api-client-jackson2</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>com.google.api-client</groupId>
-            <artifactId>google-api-client-java6</artifactId>
+            <scope>test</scope>
```
AngusDavis: Do we still reference jackson or the jackson JsonFactory in tests?

bigtable-grpc-interface/pom.xml
sduskis: mvn dependency:analyze was flaging this netty-all dependency as a not used, and grpc's more specific use of netty-codec (and something else) were used.  I'll add those in.

bigtable-hbase/pom.xml
sduskis: Maven compiled fine without this, probably because of the grpc depdencies.  I'll make those dependencies explicit.

bigtable-hbase/pom.xml
sduskis: woops.  I thought I deleted the others.  Good catch.

bigtable-hbase/pom.xml
sduskis: It doesn't look like it.

sduskis:
Closing this for now.

PR #308, opened 2015-05-29
Separate map reduce

sduskis:
Moving the hadoop jobs to a separate project.  That will help us create more streamlined dependencies for users like those in GAE


sduskis:
closing this for now.

PR #310, opened 2015-06-03
Replacing bigtable-protos-internal jar with generated java.

sduskis:
The bigtable-protos project was causing a problem with the release strategy.  The release plugin doesn't allow -SNAPSHOT dependencies, and the protos jar was a -SNAPSHOT version.  Instead of just checking in a non SNAPSHOT version, I think that it's better to check in java files generated from gradle's generateProto process.  That simplified the pom.xml and seems more transparent.


sduskis:
@kevinsi4508, can you please take a look?  I will pass this on to Angus later today.  Mostly, it would be great to get your opinion on the documentation.

bigtable-protos/README.md
```diff
@@ -54,4 +54,12 @@ Retrieved from https://github.com/google/googleapis:
...
 
+# Generating the java gRPC implementations
 
+The gRPC java generation process requires protoc 3 to create the java artifacts and gradle to drive the process.  
+
+Protoc can be installed via the instructions at: https://github.com/grpc/grpc-java#build-protobuf.  Installing it isn't always trivial; you may have to Google error messages to understand what to install next; there are plenty of answers on stackoverfolw for protobuf installation requirements.  
+
+Gradle version 2.3+ also has to be on your path; you can get it from https://gradle.org/downloads/.
+
+Once that's done, the easiest way to generate the java files is to run *gradle :generateProto*.  That will generate the java source files in target/generated-sources/main.  Manually delete src/main/generated-java/com and move  target/generated-sources/main/com to src/main/generated-java/.
```
kevinsi4508: remove the extra whitespace below:
"generated-java/com and move  target/gen"

kevinsi4508:
LGTM

AngusDavis:
My only comment is on the location of the generated code - grpc-java uses src/generated/java which I'm somewhat partial to (and seems to be supported by random StackOverflow commenters: http://stackoverflow.com/questions/22912900/location-of-generated-source-files-for-maven-directory-structure)

sduskis:
I updated the directory structure along with the pom.xml file and README.

PR #313, opened 2015-06-04
Moving some files from bigtable-hbase to bigtable-grpc-interface

sduskis:
I had to add a setter for userAgent in BigtableOptions, since BigtableConstants.USER_AGENT is specific to bigtable-hbase.


kevinsi4508:
LGTM

PR #314, opened 2015-06-05
Configure BigtableConnection in TestClusterAPI via env veriables.

sduskis:
It doesn't make much sense to use bigtable-test.xml which essentially pulls in a whole bunch of environment varaibles anyway.


bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -97,6 +97,16 @@ public void run() {
 
   private static final Set<RegionLocator> locatorCache = new CopyOnWriteArraySet<>();
 
+  private static BigtableOptions toOptions(Configuration conf) throws IOException {
+    try {
+      return BigtableOptionsFactory.fromConfiguration(conf);
+    } catch (IOException ioe) {
+      new Logger(AbstractBigtableConnection.class).error(
```
kevinsi4508: nit: use a static logger?

bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -112,14 +122,24 @@ public void run() {
...
+      boolean managed, ExecutorService pool, User user) throws IOException {
+    this(conf, toOptions(conf), managed, pool, user);
+  }
+
+  protected AbstractBigtableConnection(Configuration conf, BigtableOptions options,
+      boolean managed, ExecutorService pool, User user) throws IOException {
     this.batchPool = pool;
     this.closed = false;
     this.conf = conf;
+    this.options = options;
```
kevinsi4508: Why don't we call toOptions() here?

bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
sduskis: We only need this logger if there's a startup failure in converting the configuration to the BigtableOption.  The LOG variable is a member variable so that its log statements can show exactly which type of subclass is being used.  This usage can't use the member variable, so it needs its own Logger.  Since there's no reuse, I figured there's no benefit of maintaining a reference to this logger elsewhere.

sduskis:
I removed the changes to the classes that you previously commented on.  This PR now only makes a change in TestClusterAPI.java and removes the two xml files.

bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/TestClusterAPI.java
```diff
@@ -75,24 +74,15 @@ public void setup() throws IOException {
...
-    if (extraResources == null) {
-      Assert.fail("Please set bigtable.test.extra.resources");
-    }
-
-    InputStream resourceStream =
-        TestClusterAPI.class.getClassLoader().getResourceAsStream(extraResources);
-    if (resourceStream == null) {
-      Assert.fail(extraResources + " does not exist");
+    Configuration configuration = new Configuration();
+    for (Entry<Object, Object> entry : System.getProperties().entrySet()) {
```
kevinsi4508: setting all system properties in configuration? I prefer more explicit here on what properties we are setting.

sduskis: The ultimate goal here is to separate the configuration in BigtableOptionsFactory to not rely on a Configuration object, which is HBase specific.  The change I have in mind will be a bit more intrusive than just changing a Unit test.  Are you ok with me making a much bigger PR out of this change?

kevinsi4508: Add TODO if you plan further enhancing the code.

kevinsi4508:
LGTM

PR #315, opened 2015-06-05
Makings BigtableOptionsFactory usable without an HBase Configuration 

sduskis:
Moving BigtableOptionsFactory to bigtable-grpc so that the connection configuration can be done without having an HBase Configuration object.

This is a step towards getting a Bigtable API that's independent of hbase

### DIFF
--- a/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/TestClusterAPI.java
+++ b/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/TestClusterAPI.java
@@ -21,7 +21,6 @@ import io.grpc.Status.OperationRuntimeException;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
-import java.util.Map.Entry;
 import java.util.Random;
 import java.util.UUID;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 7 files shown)


bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/TestClusterAPI.java
```diff
@@ -67,6 +67,46 @@
...
+
+    @Override
+    public String get(String key, String defaultValue) {
+      return System.getProperty(key, defaultValue);
+    }
+
+    @Override
+    public int getInt(String key, int defaultValue) {
+      if (System.getProperties().contains(key)) {
+        return new Integer(get(key));
```
kevinsi4508: Integer.parseInt()?

kevinsi4508:
LGTM

AngusDavis:
While I don't have any particular issues with this PR - I don't think we'll want to have the BigtableOptionsFactory used outside of Hadoop. The key names are all very verbose and Hadoop-y. I'd expect non-HBase clients to instead make use of a BigtableOptions Builder of some sort.

sduskis:
@AngusDavis: I agree with you with the details.  I guess what I'm really aiming for is a way to set defaults for BigtableOptions.  Right now, the hadoopy names and default values are tightly coupled.  Let me see if I can separate those out.

PR #318, opened 2015-06-10
Checkin to allow deployments to sonatype's repos

sduskis:
see https://oss.sonatype.org/content/repositories/snapshots/com/google/cloud/bigtable/ for the result
see http://central.sonatype.org/pages/apache-maven.html for more info about the requirements

### DIFF
--- a/bigtable-hbase-1.0/pom.xml
+++ b/bigtable-hbase-1.0/pom.xml
@@ -25,6 +25,10 @@ limitations under the License.
 
     <artifactId>bigtable-hbase-1.0</artifactId>
     <packaging>jar</packaging>
+    <name>${project.groupId}:${project.artifactId}</name>
+    <description>
+       This project contains hbase 1.0 specific implementation of bigtable hbase.
+    </description>
 
     <properties>
         <hbase.version>1.0.1</hbase.version>
@@ -77,7 +81,6 @@ limitations under the License.
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-shade-plugin</artifactId>
-                <version>2.3</version>
                 <executions>
                     <execution>
                         <phase>package</phase>
--- a/bigtable-hbase/pom.xml
+++ b/bigtable-hbase/pom.xml
@@ -25,6 +25,10 @@ limitations under the License.
 
     <artifactId>bigtable-hbase</artifactId>
     <packaging>jar</packaging>
+    <name>${project.groupId}:${project.artifactId}</name>
+    <description>
+       This project contains artifacts that adapt bigtable client to work with hbase.
+    </description>
 
     <dependencies>
         <dependency>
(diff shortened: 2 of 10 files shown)


bigtable-hbase-1.0/pom.xml
```diff
@@ -66,6 +66,7 @@ limitations under the License.
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
+                <version>2.6</version>
```
AngusDavis: Can this version be specified in the parent pom in pluginManagement?

bigtable-hbase/pom.xml
```diff
@@ -44,28 +44,6 @@ limitations under the License.
         </dependency>
 
         <dependency>
-            <groupId>com.google.api-client</groupId>
```
AngusDavis: It's not obvious why this change is here - did closing the snapshot repository with these fail?

pom.xml
```diff
@@ -21,6 +21,21 @@ limitations under the License.
     <artifactId>bigtable-client</artifactId>
     <version>0.1.7-SNAPSHOT</version>
     <packaging>pom</packaging>
+    <name>${project.groupId}:${project.artifactId}</name>
+    <url>https://cloud.google.com/bigtable/</url>
+    <description>
```
AngusDavis: This seems like marketing copy instead of "A client to access Google Cloud Bigtable (http://...) from HBase".

It might also make sense to have descriptions in the submodules if the intent is for them to be referenced separately (e.g., the lightweight client in bigtable-grpc).

pom.xml
```diff
@@ -307,6 +356,55 @@ limitations under the License.
...
+                        <execution>
+                          <id>attach-javadocs</id>
+                          <goals>
+                      <goal>jar</goal>
+                        </goals>
+                      </execution>
+                    </executions>
+                </plugin>
+                <plugin>
+                    <groupId>org.sonatype.plugins</groupId>
```
AngusDavis: This is optional. mvn deploy works fine without it. It's completely up to you, though.

pom.xml
```diff
@@ -307,6 +356,55 @@ limitations under the License.
...
+                    <version>1.6.3</version>
+                    <extensions>true</extensions>
+                    <configuration>
+                        <serverId>ossrh</serverId>
+                        <nexusUrl>https://oss.sonatype.org/</nexusUrl>
+                        <autoReleaseAfterClose>true</autoReleaseAfterClose>
+                    </configuration>
+                </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
```
AngusDavis: Does this get invoked properly (it's in management and not in a top-level build)? 
Even if it does, would it make sense to add it only in to a release profile so people can mvn install without having GPG setup correctly? An example: https://github.com/GoogleCloudPlatform/bigdata-interop/blob/master/pom.xml#L117

bigtable-hbase/pom.xml
sduskis: There were 3 duplicated dependencies that mvn deploy complained about.  I had to remove them for mvn deploy to work.

pom.xml
```diff
@@ -276,6 +298,33 @@ limitations under the License.
         <tag>HEAD</tag>
     </scm>
 
+    <developers>
+       <developer>
+         <name>Angus Davis</name>
+         <email>mngusdavis@google.com</email>
+         <organization>Google</organization>
```
AngusDavis: s/mngusdavis/angusdavis/ :)

AngusDavis: Also - I don't really mind being named, but the other option is a simple:
  <developers>
    <developer>
      <organization>Google Inc.</organization>
      <organizationUrl>http://www.google.com</organizationUrl>
    </developer>
  </developers>

AngusDavis:
If I recall, an issue tracker is also required in the pom.

bigtable-hbase/pom.xml
AngusDavis: Ah, makes sense. Thanks for the clarification.

pom.xml
sduskis: A release profile makes sense for both this and sources/javadoc.  I'll make it so.

sduskis:
Updated the top pom.xml to make the javadoc and source jar plugins work as well as the signing.

I still need to update the description.  If that's all there is left, then it would be nice to get this PR merged.

sduskis:
I added more information to the top level description.  I also changed autoReleaseAfterClose to false.  PTAL.

AngusDavis:
Couple of small requests, otherwise LGTM.

sduskis:
Moved the distributionManagement closer to repositories.  Also added basic name/description to all modules.

PR #319, opened 2015-06-11
Add support for HBase PageFilters.

AngusDavis:
Add a PageFilterAdapter which will adapt PageFilter instances so long as
they appear at the top-level or as the last element of the top-most
FilterList.

Add the concept of ReadHooks. Right now, ReadHooks are for modifying a
ReadRowsRequest before sending the request to CBT.

Add a ReadOperationAdapter interface for Scan and Get adapters and move
from plain OperationAdapter to ReadOperationAdapter.

### DIFF
--- a/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestFilters.java
+++ b/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestFilters.java
@@ -18,6 +18,7 @@ package com.google.cloud.bigtable.hbase;
 import static com.google.cloud.bigtable.hbase.IntegrationTests.*;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterators;
 
 import org.apache.commons.lang.RandomStringUtils;
 import org.apache.hadoop.hbase.Cell;
@@ -39,10 +40,13 @@ import org.apache.hadoop.hbase.filter.ColumnRangeFilter;
 import org.apache.hadoop.hbase.filter.CompareFilter;
 import org.apache.hadoop.hbase.filter.CompareFilter.CompareOp;
 import org.apache.hadoop.hbase.filter.Filter;
+import org.apache.hadoop.hbase.filter.FilterList;
+import org.apache.hadoop.hbase.filter.FilterList.Operator;
 import org.apache.hadoop.hbase.filter.FirstKeyOnlyFilter;
 import org.apache.hadoop.hbase.filter.KeyOnlyFilter;
 import org.apache.hadoop.hbase.filter.MultipleColumnPrefixFilter;
 import org.apache.hadoop.hbase.filter.NullComparator;
+import org.apache.hadoop.hbase.filter.PageFilter;
 import org.apache.hadoop.hbase.filter.PrefixFilter;
 import org.apache.hadoop.hbase.filter.QualifierFilter;
 import org.apache.hadoop.hbase.filter.RandomRowFilter;
@@ -1376,6 +1380,35 @@ public class TestFilters extends AbstractTest {
     get.setFilter(regexQualFilter);
     result = table.get(get);
     Assert.assertEquals(1, result.size());
+  }
+
+  @Test
+  public void testPageFilters() throws IOException {
+    byte[][] rowKeys = dataHelper.randomData("pageFilter-", 100);
+    byte[] qualA = dataHelper.randomData("qualA");
+    byte[] value = Bytes.toBytes("Important data goes here");
+    Table table = getConnection().getTable(TABLE_NAME);
+    for (byte[] rowKey : rowKeys) {
+      Put put = new Put(rowKey).addColumn(COLUMN_FAMILY, qualA, value);
+      table.put(put);
+    }
+
+    Scan scan = new Scan(Bytes.toBytes("pageFilter-"));
+
+    PageFilter pageFilter = new PageFilter(20);
+    scan.setFilter(pageFilter);
+    try (ResultScanner scanner = table.getScanner(scan)) {
+      Assert.assertEquals(20,  Iterators.size(scanner.iterator()));
+    }
+
+    FilterList filterList = new FilterList(
+        Operator.MUST_PASS_ALL,
+        new QualifierFilter(CompareOp.EQUAL, new BinaryComparator(qualA)),
+        pageFilter);
+    scan.setFilter(filterList);
+    try (ResultScanner scanner = table.getScanner(scan)) {
+      Assert.assertEquals(20,  Iterators.size(scanner.iterator()));
+    }
   }
 
   private Result[] scanWithFilter(Table t, byte[] startRow, byte[] endRow, byte[] qual,
--- a/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
+++ b/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
@@ -39,14 +39,17 @@ import com.google.bigtable.v1.MutateRowRequest;
 import com.google.bigtable.v1.ReadModifyWriteRowRequest;
 import com.google.bigtable.v1.ReadRowsRequest;
 import com.google.cloud.bigtable.hbase.adapters.AppendAdapter;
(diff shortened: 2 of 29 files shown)


AngusDavis:
Fixed handling of FilterListAdapter#isFilterSupported for PageFilter, and in general. There was an embarrassing flipped conditional causing it to return true when it should not have. Tests were improved to cover this.

sduskis:
Is there any chance this PR can be split?

AngusDavis:
Potentially - can you elaborate on the reason you'd like a split?

sduskis:
It's just the size... It's difficult to review.

AngusDavis:
Unfortunately, a lot of the size came from mechanical changing of signatures e.g., FilterAdapterContext(Scan) -> FilterAdapterContext(Scan, ReadHooks). All tests save for FilterList, Filter, and PageFilterAdapter fall into this category.
 
I could potentially separate out adding ReadHooks, but I rather dislike adding plumbing without it actually being used somewhere and tested end to end (which ReadHooks won't be as it's only needed by PageFilter).

A recommended reading order would be ReadHooks -> PageFilterAdapter (and tests) -> FilterAdapterContext -> FilterList -> Table / BatchExecutor for ReadHooks and then the mechanical changes (almost everything else).

I'm open to suggestions on logical separation, but I'd appreciate it if you can give it a go and we can break it up if needed.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
```diff
@@ -214,9 +217,11 @@ public BatchExecutor(
    */
   ListenableFuture<com.google.bigtable.v1.Row> issueGetRequest(Get get) {
     LOG.trace("issueGetRequest(Get)");
-    ReadRowsRequest.Builder builder = getAdapter.adapt(get);
```
sduskis: Is there any reason that getAdapter.adapt(get) doesn't create a DefaultReadHooks if a parameter isn't supplied?

AngusDavis: My thinking is that it is / will be error prone (e.g., when making this change it'd have been easy to miss a call-site that should have been fixed)

sduskis:
LGTM.

The filtering infrastructure is pretty complicated.  It does quite a bit, so I don't think that there's much to do about it.  I didn't see anything wrong.  The functionality looked right.  I wasn't completely sure about the details of the tests being correct or not, but they seemed right.  The functional code looked good.  It's probably worth for me to do a more in depth review and tracing of the filtering framework.  It's worth merging this, but it may be worth a review from someone else as well.

AngusDavis:
@kevinsi4508 / @mgarolera - would one of you please self-assign and take a look when you can?

sduskis:
I updated to @mgarolera who has more bandwidth for the review.

bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestFilters.java
```diff
@@ -1378,6 +1382,35 @@ public void testQualifierFilter() throws IOException {
...
   }
 
+  @Test
+  public void testPageFilters() throws IOException {
+    byte[][] rowKeys = dataHelper.randomData("pageFilter-", 100);
+    byte[] qualA = dataHelper.randomData("qualA");
+    byte[] value = Bytes.toBytes("Important data goes here");
+    Table table = getConnection().getTable(TABLE_NAME);
+    for (byte[] rowKey : rowKeys) {
+      Put put = new Put(rowKey).addColumn(COLUMN_FAMILY, qualA, value);
```
mgarolera: Maybe interleaving rows with different qualifiers?

mgarolera:
I won't get to it today, will do on monday.

-- 

Marçal

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/ReadHooks.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Hooks for modifying a ReadRowsRequest before being sent to Cloud Bigtable.
+ *
+ * Note that it is expected that this will be extended to include post-read
+ * hooks to transform Rows when appropriate.
+ */
+public interface ReadHooks {
+
+  /**
+   * Add a Function that will modify the ReadRowsRequest before it is sent to Cloud Bigtable.
```
mgarolera: Does it support multiple calls? If so, will the lastest function composed be called first or last?

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/ReadHooks.java
```diff
@@ -0,0 +1,38 @@
...
+   */
+  void composePreSendHook(Function<ReadRowsRequest, ReadRowsRequest> newHook);
+
+  /**
+   * Apply all pre-send hooks to the given request.
+   */
+  ReadRowsRequest applyPreSendHook(ReadRowsRequest readRowsRequest);
```
mgarolera: s/applyPreSendHook/applyPreSendHooks/?

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/DefaultReadHooks.java
```diff
@@ -0,0 +1,35 @@
...
+import com.google.bigtable.v1.ReadRowsRequest;
+import com.google.common.base.Function;
+import com.google.common.base.Functions;
+
+/**
+ * Default implementation of {@link ReadHooks}.
+ */
+public class DefaultReadHooks implements ReadHooks {
+  private Function<ReadRowsRequest, ReadRowsRequest> preSendHook = Functions.identity();
+  public void composePreSendHook(Function<ReadRowsRequest, ReadRowsRequest> newHook) {
```
mgarolera: Override?

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/ReadOperationAdapter.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.google.bigtable.v1.ReadRowsRequest;
+
+import org.apache.hadoop.hbase.client.Operation;
+
+/**
+ * Interface used for Scan and Get operation adapters.
+ */
+public interface ReadOperationAdapter<T extends Operation> {
+  ReadRowsRequest.Builder adapt(T request, ReadHooks readHooks);
```
mgarolera: Explain that it configures the readhooks based on the adapted request and that they should be applied before sending the request?

bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/adapters/filters/TestPageFilterAdapter.java
```diff
@@ -0,0 +1,124 @@
...
+    FilterAdapterContext context = new FilterAdapterContext(new Scan(), hooks);
+    PageFilter pageFilter = new PageFilter(20);
+    RowFilter adaptedFilter = pageFilterAdapter.adapt(context, pageFilter);
+    Assert.assertNull("PageFilterAdapter should not return a RowFilter.", adaptedFilter);
+
+    ReadRowsRequest request = ReadRowsRequest.newBuilder().setNumRowsLimit(100).build();
+    ReadRowsRequest postHookRequest = hooks.applyPreSendHook(request);
+    Assert.assertEquals(20, postHookRequest.getNumRowsLimit());
+  }
+}
```
mgarolera: Add a test of two diferent pageFilters at the end?

mgarolera:
LGTM
As Solomon said, it looks good but complicated. Not sure how to make it simplier, but I'm not a fan of all the ReadHooks (seems too much to have an interface and composePreSendHook) and the Contexts.

sduskis:
I discussed this with @mgarolera.  We need this functionality, so let's merge this.  We'll think about how to restructure the functionality at some other point in time.

sduskis:
@AngusDavis: this PR has been sitting here for a while.  Should someone else pick it up?

PR #321, opened 2015-06-17
Fixing potential problems with resource accounting in BigtableBufferedMutator

sduskis:
Handling a case where the RPC is already complete before it gets registered.

### DIFF
--- a/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
+++ b/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
@@ -4,9 +4,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -101,51 +111,121 @@ public long getMaxHeapSize() {
...
-        wait(WAIT_MILLIS);
-      }
-      if (performedWarning) {
-        LOG.info("flush() completed");
+      cleanupCompletedCallbacks();
+      while (!activeRequests.isEmpty()) {
+        waitForCompletions();
       }
+      LOG.trace("flush() completed");
+      nextLogOrSweep = -1L;
```
kevinsi4508: move this to before while loop?

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -101,51 +111,121 @@ public long getMaxHeapSize() {
...
+    public synchronized void registerInvokedCallback(RequestAccountant callback) {
+      // There could have been a backlog of RPCs, and the RPC completed by the time the synchronized
+      // registerInvokedCallback method was finally called. In that case, simply clean up the heap
+      // size. Otherwise, keep track of the callback in the activeRequests Map.
+      if (callback.isRequestDone()) {
+        operationComplete(callback);
       } else {
-        LOG.warn("An operation completion was recieved multiple times. Your operations completed."
-            + " Please notify Google that this occurred.");
+        this.activeRequests.put(callback.operationSequenceId, callback);
```
kevinsi4508: nit: remove "this."

AngusDavis:
I'm a bit confused on the problem. In registerOperationWithHeapSize, if an InterruptedException is thrown then no pending operation will be registered and so the catch block on 350 (now 407), just needs to either throw or exit the method.

If the Channel has given us a Future, even if the future is already complete, calling addCallback will cause the future's callback to be triggered. If that's not happening, then Guava and/or Java's Futures are broken. 

If the Future is never being completed or Failed, then that's a gRPC issue and I'm not certain this will solve that problem.

sduskis:
Closing this in lieu of #322.

PR #322, opened 2015-06-18
Fixing some issues with BufferedMutator

sduskis:
Adding a separate executor for the heapSizeManager synchronized work.
Better accounting for exceptions in issueRequest, such as validation issues.

### DIFF
--- a/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
+++ b/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
@@ -4,9 +4,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -16,12 +16,12 @@
 package com.google.cloud.bigtable.hbase;
 
 import java.io.IOException;
-import java.net.InetAddress;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock.ReadLock;
@@ -36,6 +36,9 @@ import org.apache.hadoop.hbase.client.Mutation;
 import org.apache.hadoop.hbase.client.RetriesExhaustedWithDetailsException;
 import org.apache.hadoop.hbase.client.Row;
 
+import com.google.cloud.bigtable.config.BigtableOptions;
+import com.google.cloud.bigtable.config.Logger;
+import com.google.cloud.bigtable.grpc.BigtableClient;
 import com.google.cloud.bigtable.hbase.adapters.AppendAdapter;
 import com.google.cloud.bigtable.hbase.adapters.DeleteAdapter;
 import com.google.cloud.bigtable.hbase.adapters.GetAdapter;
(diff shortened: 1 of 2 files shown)


sduskis:
I addressed the globalException issues.  I moved close (although it probably doesn't make a difference).  I wasn't sure what to do about Futures.immediateFailedFuture() as per the comments.

PTAL.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -196,13 +209,8 @@ public BigtableBufferedMutator(
...
     this.exceptionListener = listener;
-    InetAddress host = null;
+    InetAddress host = options.getDataHost();
 
-    try {
-      host = options.getDataTransportOptions().getHost();
-    } catch (IOException e) {
-      LOG.warn("Could not get the host used for writes", options);
-    }
     this.host = host == null ? null : host.toString();
```
mgarolera: Can host be null? Could this be simplified?

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -147,6 +156,10 @@ public synchronized void operationComplete(long operationSequenceId) {
     public synchronized boolean hasInflightRequests() {
       return !pendingOperationsWithSize.isEmpty();
     }
+
+    public synchronized long getHeapSize() {
```
mgarolera: Should this be @VisibleForTesting?

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -347,14 +356,37 @@ private void doMutation(final Mutation mutation) throws RetriesExhaustedWithDeta
...
 
-    // TODO: Consider adding the callback in another executor for the blocking call. There are some
-    // concerns running this callback on the rpc threadpool. This callback requires locks on
-    // resources. If they cannot be acquired, the callback would block future completions.
-    Futures.addCallback(batchExecutor.issueRequest(mutation), callback);
+    ListenableFuture<? extends GeneratedMessage> request = null;
+    try {
+      request = batchExecutor.issueRequest(mutation);
+    } catch (RuntimeException e) {
+      // issueReqeust(mutation) could throw an Exception for validation issues. Remove the heapsize
```
mgarolera: issueRequest

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -347,14 +356,37 @@ private void doMutation(final Mutation mutation) throws RetriesExhaustedWithDeta
...
     }
     AccountingFutureCallback callback = new AccountingFutureCallback(mutation, sequenceId);
 
-    // TODO: Consider adding the callback in another executor for the blocking call. There are some
-    // concerns running this callback on the rpc threadpool. This callback requires locks on
-    // resources. If they cannot be acquired, the callback would block future completions.
-    Futures.addCallback(batchExecutor.issueRequest(mutation), callback);
+    ListenableFuture<? extends GeneratedMessage> request = null;
+    try {
+      request = batchExecutor.issueRequest(mutation);
```
mgarolera: Is this raising an UnsupportedOperationException? (that is the one I found reading). It is a pity that it is an unchecked exception, otherwise it would have been easier to catch this. :/ Maybe it should be added to issueRequest and OperationAdapter.adapt?

sduskis: putAdapter throws IllegalArgumentException.  mutationAdapter and DeleteAdapters throw UnsupportedOperationException.  Perhaps we should discuss checked exceptions for the Adapter framework, but that sounds like something for another time.

AngusDavis: Further, all gRPC errors (e.g., channel closed) are RuntimeException derived.

sduskis:
@mgarolera: I fixed the typo and added a bunch of @VisibleForTesting.  PTAL

sduskis:
@AngusDavis: Thanks for the pointer on immediateFailFuture.  I updated the code to use it, and update the test to account for the asynchronous processing of the exception (since the handling happens on a different thread)

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -347,14 +356,34 @@ private void doMutation(final Mutation mutation) throws RetriesExhaustedWithDeta
       // requirements are achieved.
       sequenceId = sizeManager.registerOperationWithHeapSize(mutation.heapSize());
     } catch (InterruptedException e) {
+      synchronized (globalExceptions) {
+        // Add the exception to the list of global exceptions and immediately throw a
+        // RetriesExhaustedWithDetailsException.
+        addGlobalException(mutation, e);
+        handleExceptions();
```
AngusDavis: This now handles the case of concurrent writers, but couldn't a client still specify an error handler other than the default 'throw whenever floats your boat and on a random thread'? I'm not a fan of that error handler so I imagine I would be one of those clients :)

sduskis:
@AngusDavis: Good point on the custom exception handler.  I updated the handling in doMutate() - I now return in the catch clause.  I also updated the exception comment.

AngusDavis:
LGTM. For a future change, if you make ExecutorService a constructor parameter then you can pass in a directExecutor and remove some sleep calls from tests. Or a FakeExecutorService-sort of class that allows you to execute 'ticks' to verify that it's done async properly.

PR #323, opened 2015-06-18
Removing netty-all and older netty dependencies.

sduskis:
Older versions of netty-all are exposed via some of the maven dependencies we have.  We need to have a consistent version of netty.


sduskis:
I need to fix this.  Closing for now.

PR #325, opened 2015-06-18
Excluding older guava versions in the root pom.xml.

sduskis:
Explicitely relying on guava 13 for the integration tests.

### DIFF
--- a/bigtable-hbase-1.0/pom.xml
+++ b/bigtable-hbase-1.0/pom.xml
@@ -80,6 +80,37 @@ limitations under the License.
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-enforcer-plugin</artifactId>
+                <version>1.4</version>
+                <executions>
+                    <execution>
+                        <id>enforce-banned-dependencies</id>
+                        <goals>
+                            <goal>enforce</goal>
+                        </goals>
+                        <configuration>
+                            <rules>
+                                <bannedDependencies>
+                                    <excludes>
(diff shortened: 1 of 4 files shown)


sduskis:
FYI, both the hbase minicluster and bigtable integration tests pass.

AngusDavis:
First, I agree on the java5 exclude. Second, questions:
1) Would it be better to have a direct dependency on guava ${guava.version} to override the hadoop/hbase dependency? (and keep the guava 13 in integration tests)
2) If not, it seems that having excludes might be prone to things sneaking in again in the future. Consider using: http://maven.apache.org/enforcer/enforcer-rules/bannedDependencies.html to ban guava versions less than 18 [,18.0]

sduskis:
1) I'm not sure where you mean.  We already have a dependency on guava 18.0 in bigtable-hbase.  We shade in that dependency in bigtable-hbase-1.0.  We need guava 13 in integration tests.  I'm not sure how ${guava.version} helps us in this case.  What problem does ${guava.version} solve?

2) good idea.  I'll add it to bigtable-hbase-1.0 and -1.1.

sduskis:
Oh, I see.  It makes sense to add a ${guava.version} so it's consistent between the inclusion and the bannedDependenies plugin

AngusDavis:
For 1) - if the hbase-1.0 and hbase-1.1 jars depend directly on guava, then the version they reference will win for the purposes of dependency resolution. At that point, excludes are probably not required.
LGTM, anyways. Feel free to modify or not.

sduskis:
Ugh.  I forgot to rebase -i master first.  Sorry.

PR #326, opened 2015-06-18
Adding an explicit dependency on commons-lang.

sduskis:
Currently, relying on bigtable-hbase-1.0 by itself isn't enough to pull in commons-lang.  Applications need it at runtime, otherwise they fail with ClassNotFoundException.


AngusDavis:
LGTM

PR #327, opened 2015-06-19
Allow override of bigtable-hbase version under test.

sduskis:
This will help with the "external" jobs that run in Jenkins when versions change.  A paramater override will allow us to keep a consistent version for those tests.


kevinsi4508:
LGTM

PR #329, opened 2015-06-22
Separate shading into a separate project

sduskis:
Include all dependencies as part of the shaded jar and have 0 dependencies on other projects.

### DIFF
--- a/bigtable-hbase-integration-tests/pom.xml
+++ b/bigtable-hbase-integration-tests/pom.xml
@@ -252,7 +252,6 @@ limitations under the License.
         <dependency>
             <groupId>org.mortbay.jetty.alpn</groupId>
             <artifactId>alpn-boot</artifactId>
-            <version>${alpn.version}</version>
         </dependency>
     </dependencies>
     <build>
--- a/bigtable-hbase-shaded/pom.xml
+++ b/bigtable-hbase-shaded/pom.xml
@@ -1,0 +1,157 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Copyright 2014 Google Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.google.cloud.bigtable</groupId>
+        <artifactId>bigtable-client</artifactId>
+        <version>0.1.10-SNAPSHOT</version>
+    </parent>
+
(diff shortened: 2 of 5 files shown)


AngusDavis:
Could you give a bit of context on when you expect this to be used and not used? It doesn't seem very safe in general usage so I assume it's for something very specific. If it is for something specific, perhaps changing the artifact id (e.g., bigtable-hbase-mvm)

sduskis:
The context here is for any place where we'd want to use a shaded artifact.  This should work for bdutil or mvm.

There are 2 objectives here:

1) create a jar with all of the dependencies bundled
2) create a pom with no other dependencies.

AngusDavis:
Running:
$ mvn clean package && jar tf bigtable-hbase-shaded/target/bigtable-hbase-shaded-0.1.10-SNAPSHOT.jar 
leads me to believe this will have classloader / resource issues.

How are different HBase versions handled handled?

sduskis:
RE: How are different HBase versions handled handled?

I think we should only support a single version of HBase for the shaded jar.

sduskis:
RE: leads me to believe this will have classloader / resource issues.

Can you clarify please?

AngusDavis:
OK - to start: 

Bare references to org.slf4j, org.codehaus, jcodings, and org.apache.log4j mean that anyone trying to use this shaded artifact will be required to use these versions, but we won't actually be telling them that those are the versions in use so instead they'll just start getting weird NoSuchMethod errors that they can't debug.

The generated HBase protos were generated with proto 2.5. gRPC requires proto 3. There are class incompatibilities between the two. In the present versions, we shade proto 3, include grpc and don't include HBase classes and as a result HBase classes properly load proto 2.5.

sduskis:
I see.  Good points.  I'll re-add the bigtable-hbase-1.x shaded jars.  

I still would like to support MVM and dataflow.  Something like an uber shaded jar/pom seems useful in those cases.  It would have to be for cases where hbase access wasn't a factor.  I'd still need to figure out what to do with slf3j, codehause, jconings and log4j; one possibility is package renaming.  This would be similar to hbase's shading rules: http://central.maven.org/maven2/org/apache/hbase/hbase-shaded/1.1.0.1/hbase-shaded-1.1.0.1.pom

sduskis:
I re-added the hbase-1.0 and hbase-1.1  shading.  I'm also referencing the shaded classifier from hbase-1.0 in the -shaded project.  We need to shade 2 different versions of guava/protobuf and etc.  It's a bit of a mess, but it seems like the mess is mostly due to needed multiple versions of the same jar.

PTAL.  I'm sure that there is room for improvement.

sduskis:
@lesv: This produces a pom.xml file with no dependencies and a single jar that gets imported.  PTAL

AngusDavis:
I'm not convinced creating this is a good idea, nor am I sure of when it would be a good idea to use this. If users are programming against HBase interfaces contained in HBase artifacts, they should be referencing HBase artifacts, not our artifacts. 

I think that a larger discussion of what problems this is solving would be good so that we can explore other options. Do we have bugs for each issue? I've seen a FR about creating an hbase api interface + POJO jar upstream which would be awesome (but also probably not as simple as it seems with Filters, etc), but otherwise just random emails for everything else.

lesv:
I agree w/ Angus's view, though it's fairly easy just to have a single dependency, however, if we are going to have a bunch of <exclusions> anyway, then why not switch from default scope to something else, and add all the hbase stuff back in?

sduskis:
@lesv: I'm not sure what you mean.

sduskis:
@AngusDavis: One issue is that we have to exclude too much right now from hbase-1.0 and hbase-1.1 even with the classifier=shaded.  We need a reduced dependency pom.xml when using the shaded classifier.  Right now, we either use the non-shaded jar, which brings in a version of protobuff and guava which will either be inconsistent with hbase or with bigtable.  Or, we can use the shaded jar and pull in bigtable-hbase, bigtable-grpc-interface and bigtable-protos which will conflict with the shaded jar.

sduskis:
closing this for now.  We need a different solution.

PR #330, opened 2015-06-23
ALPN should be a test scope dependency in bigtable-hbase.

sduskis:


AngusDavis:
Is there any case where it shouldn't be test scoped? If not, should this be in dependency management in the parent pom?

sduskis:
@AngusDavis: Good call.  I moved the test scope to the parent pom.xml and verified with "mvn dependency:tree"  PTAL

AngusDavis:
LGTM

PR #332, opened 2015-06-24
Moving up protobuf-java and grpc versions to the parent pom.

sduskis:
I did some of this work for the shaded jar.  This seems independently useful.


AngusDavis:
LGTM

PR #346, opened 2015-07-08
Updating pom.xml - reducing grpc dependencies and others.

sduskis:
Used "mvn dependency:analyze" to figure out which dependencies could be removed and which need to be explicitly added.  Also updated grpc dependencies to the core requirements rather than grpc-all.


AngusDavis:
LGTM just a newline nit

PR #347, opened 2015-07-08
Making the shaded artifacts the main one rather than a classifier.

sduskis:


AngusDavis:
Could you add comments on these so they don't get flipped back? More or less copying some of the points from the email thread on why it's required.

sduskis:
Comments added.

sduskis:
@AngusDavis: How do the comments look?

AngusDavis:
It would be great if you could include mention of which artifacts /  packages clash so that in a year or two when HBase upgrades proto / netty, we'll know it's OK to re-instate the non-shaded artifacts.

sduskis:
@AngusDavis: good point.  I udpates the comments.

AngusDavis:
LGTM

PR #348, opened 2015-07-10
Initializing protobuf and ssl context artifacts in parallel.

sduskis:
This should improve connection startup time by 150+ ms.

### DIFF
--- a/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
+++ b/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
@@ -16,6 +16,8 @@
 // Because MasterKeepAliveConnection is default scope, we have to use this package.  :-/
 package org.apache.hadoop.hbase.client;
 
+import io.netty.handler.ssl.SslContext;
+
 import java.io.Closeable;
 import java.io.IOException;
 import java.net.InetAddress;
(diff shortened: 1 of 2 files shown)


bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -68,6 +72,39 @@
...
+            new ThreadFactoryBuilder()
+                .setNameFormat("AbstractBigtableConnection-startup-%s")
+                .setDaemon(true)
+                .build());
+    connectionStartupExecutor.execute(new Runnable() {
+      @Override
+      public void run() {
+        // Startup of SSL_CONTEXT_FACTORY is a really expensive operation for the first operation.
+        // It could take up to 2 seconds.
+        BigtableOptions.SSL_CONTEXT_FACTORY.create();
```
mgarolera: This returns an SslContext, maybe do a similar thing like the other run() so that it is obvious we are createing an object and then discarding it. Also, is it ok to create SslContext and not use them?

bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -68,6 +72,39 @@
...
+      }
+    });
+    connectionStartupExecutor.execute(new Runnable() {
+      @Override
+      public void run() {
+        // Initializing this variable is expensive. It should happen once at startup asynchronously.
+        @SuppressWarnings("unused")
+        Object ignored = BigtableServiceGrpc.CONFIG;
```
mgarolera: Use BigtableServiceServiceDescriptor for clarity? How come this happens when assigning the variable? would it be worth asking grpc team to initialize it at startup?

sduskis:
@mgarolera: Is the current state better?

mgarolera:
LGTM

PR #349, opened 2015-07-13
Updating BigtableOptions to use a RetryOptions object

sduskis:
Also updating the SslContextFactory to use GrpcSslContexts


mgarolera:
LGTM

PR #350, opened 2015-07-13
Fixing issue with retries and sslConext

sduskis:
The Grpc SSL context code also needs to be honed.  The change from SslContext to GrpcSslContext broke the cluster tests.

### DIFF
--- a/bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
+++ b/bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
@@ -15,7 +15,6 @@
  */
 package com.google.cloud.bigtable.config;
 
-import io.grpc.transport.netty.GrpcSslContexts;
 import io.netty.channel.EventLoopGroup;
 import io.netty.handler.ssl.SslContext;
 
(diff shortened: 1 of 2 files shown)


mgarolera:
Not sure the error is related to the SslContext:

java.lang.NullPointerException: null
	at com.google.cloud.bigtable.grpc.BigtableChannels.wrapChannel(BigtableChannels.java:195)
	at com.google.cloud.bigtable.grpc.BigtableChannels.createChannel(BigtableChannels.java:112)
	at com.google.cloud.bigtable.grpc.BigtableChannels.createChannel(BigtableChannels.java:81)
	at com.google.cloud.bigtable.grpc.BigtableClusterAdminGrpcClient.createClient(BigtableClusterAdminGrpcClient.java:48)
	at com.google.cloud.bigtable.hbase.TestClusterAPI.createClusterAdminStub(TestClusterAPI.java:175)
	at com.google.cloud.bigtable.hbase.TestClusterAPI.testClusters(TestClusterAPI.java:84)

https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/6b68daebe13eb444d8a93716dd3a9bc0bb567b5c/bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableChannels.java#L195

I think there is problems when setRetryOptions() is not called, because then the builder uses null as retryOptions.

PR #352, opened 2015-07-14
Removing the shaded classifier from the tests.

sduskis:
Fixing an issue was breaking our jenkins tests.


bigtable-hbase-1.0/pom.xml
```diff
@@ -137,7 +137,6 @@ limitations under the License.
                                 cause classloading issues in some cases like web apps.
                             -->
                             <shadedArtifactAttached>false</shadedArtifactAttached>
-                            <shadedClassifierName>shaded</shadedClassifierName>
```
mgarolera: Also update comment about shaded classifier, here and other poms.

mgarolera:
I have no idea about pom files... but LGTM after the comment thing

PR #356, opened 2015-07-14
Closing tricky to close threadpools

sduskis:
There are some threadpools opened in BigtableOptionsFactory that need to be closed when a BigtableConnection is closed.  Setting Closable objects on the ChannelOptions so that the BigtableConnection can close them.  More closable objects will be added later from code that only have access to ChannelOptions making ChannelOptions a better location to keep the Closable objects.


sduskis:
moving this to another PR.

PR #358, opened 2015-07-15
Fxing warnings and TODOs

sduskis:


bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableGrpcClientOptions.java
```diff
@@ -22,16 +22,25 @@
  */
 public class BigtableGrpcClientOptions {
 
+  private static int DEFAULT_STREAMING_BUFFER_SIZE = BigtableGrpcClient.SCANNER_BUFFER_SIZE;
+  // We can timeout when reading large cells with a low value here. With a 10MB
+  // cell limit, 60 seconds allows our connection to drop to ~170kbyte/s. A 10 second
+  // timeout requires 1Mbyte/s
+  private static int DEFAULT_READ_PARTIAL_ROW_TIMEOUT_MILLIS =
```
mgarolera: 60 * 1000? I remember we were doing something like this in other PRs. Also MILLIS -> MS

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableRetriesExhaustedException.java
```diff
@@ -21,6 +21,8 @@
  * An Exception that is thrown when an operation fails, even in the face of retrues.
  */
 public class BigtableRetriesExhaustedException extends IOException {
+  private static final long serialVersionUID = 6905598607595217072L;
```
mgarolera: Why?

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/IOExceptionWithStatus.java
```diff
@@ -36,6 +37,7 @@ public IOExceptionWithStatus(String message, OperationRuntimeException cause) {
   /**
    * Status from the provided OperationRuntimeException.
    */
+  @SuppressWarnings("deprecation")
```
mgarolera: Hmm.. why? As I read it, StatusRuntimeException shouldnt trigger deprecation warnings. Which one should be used instead?

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/FuzzyRowFilterAdapter.java
```diff
@@ -72,6 +72,7 @@ private static RowFilter createSingleRowFilter(byte[] first, byte[] second) thro
       }
     }
     ByteString quotedValue = ByteString.copyFrom(baos.toByteArray());
+    quotingStream.close();
```
mgarolera: nit: Could this be done with a try with resource?

bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -220,10 +222,12 @@ protected BigtableClient getBigtableClient(
       TransportOptions dataTransportOptions,
       ChannelOptions channelOptions,
       ExecutorService executorService) {
-
     try {
-      return BigtableGrpcClient.createClient(
-          dataTransportOptions, channelOptions, executorService);
+      RetryOptions retryOptions = channelOptions.getUnaryCallRetryOptions();
+      BigtableGrpcClientOptions clientOptions =
```
mgarolera: This seems convoluted. Why is retryoptions in both channeloptions and clientoptions?

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableRetriesExhaustedException.java
sduskis: All Java exceptions are Serializable.  The JVM spits out warnings if serialVersionUID isn't there.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/IOExceptionWithStatus.java
sduskis: StatusRuntimeException is the correct class to use, and getStatus() is the right method.  I'll explain in person.  This is kind of what we have to do right now.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/FuzzyRowFilterAdapter.java
sduskis: Try with resource didn't look right in this case.  I tried it out.

bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
sduskis: It is convoluted.  I'll fix this in another PR.

sduskis:
I address the comments, but didn't actually change code.  Can you PTAL?

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/IOExceptionWithStatus.java
mgarolera: I would put some comment here on why it is ok to ignore the deprecation warning. Warnings are there for a reason, so we should have a reason to suppress them, or talk to grpc if the warning is wrong.

Asides from this and the MS/MILLIS comment, LGTM.

PR #359, opened 2015-07-15
Revamping the way the channel shutdown happens.

sduskis:
Adding additional components to the shutdown process, such as the NioEventLoopGroup.
Removing ClientCloseHandler and CloseableChannel in favor of the standard Closeable interface.
Simplified the use of ReconnectingChannel and ChannelPool in BigtableChannels


bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableChannels.java
```diff
@@ -66,9 +66,9 @@
   private static final String POST_RETRY_REPORT_ENTRY = "PostRetry";
   /** Number of threads to use to initiate retry calls */
   private static final int RETRY_THREAD_COUNT = 5;
-  
+
   /** Number of seconds to wait for a termination before trying again. */
```
mgarolera: Is this variable in seconds or millis? Update comment and value to match ms?

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableChannels.java
```diff
@@ -87,87 +87,66 @@ public static CloseableChannel createChannel(TransportOptions transportOptions,
...
    */
-  public static CloseableChannel createChannel(final TransportOptions transportOptions,
+  public static Channel createChannel(final TransportOptions transportOptions,
       final ChannelOptions channelOptions, final ExecutorService executor, boolean createPool) {
     Preconditions.checkArgument(
         transportOptions.getTransport() == TransportOptions.BigtableTransports.HTTP2_NETTY_TLS,
         "Bigtable requires the NETTY_TLS transport.");
 
-    int channelCount = channelOptions.getChannelCount();
+    int channelCount = createPool ? channelOptions.getChannelCount() : 1;
```
mgarolera: Why is createPool necessary? Shouldnt this only be controled by channelOptions.getChannelCount() ?

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableChannels.java
```diff
@@ -87,87 +87,66 @@ public static CloseableChannel createChannel(TransportOptions transportOptions,
...
-    } else {
-      return createCloseableChannel(transportOptions, executor);
+    Channel channels[] = new Channel[channelCount];
+    for (int i = 0; i < channelCount; i++) {
+      ReconnectingChannel reconnectingChannel =
+          createReconnectingChannel(transportOptions, executor, timeoutMs);
+      channelOptions.addClientCloseHandler(reconnectingChannel);
+      channels[i] = reconnectingChannel;
     }
+    Channel channel = new ChannelPool(channels);
```
mgarolera: nit: channel -> channelPool ?

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableChannels.java
```diff
@@ -87,87 +87,66 @@ public static CloseableChannel createChannel(TransportOptions transportOptions,
...
+      public Closeable createClosable(final Channel channel) {
+        return new Closeable() {
+
+          @Override
+          public void close() throws IOException {
+            ChannelImpl channelImpl = (ChannelImpl) channel;
+            channelImpl.shutdown();
+            while (!channelImpl.isTerminated()) {
+              try {
+                channelImpl.awaitTerminated(CHANNEL_TERMINATE_WAIT_MS, TimeUnit.MILLISECONDS);
```
mgarolera: awaitTerminated might return false, apparently.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableChannels.java
```diff
@@ -87,87 +87,66 @@ public static CloseableChannel createChannel(TransportOptions transportOptions,
...
-        .executor(executor)
-        .negotiationType(NegotiationType.TLS)
-        .streamWindowSize(1 << 20) // 1 MB
-        .build();
+    });
   }
 
-  private static CloseableChannel wrapChannel(ChannelOptions channelOptions,
-      ExecutorService executor, Channel channel, ClientCloseHandler onClientClose) {
+  private static Channel wrapChannel(ChannelOptions channelOptions,
```
mgarolera: Maybe document what kind of wrapping does it do on the channel?

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterAdminGrpcClient.java
```diff
@@ -101,7 +101,6 @@ public void deleteCluster(DeleteClusterRequest request) {
   
   @Override
   public void close() throws Exception {
-    channel.close();
```
mgarolera: Add a comment that things get closed behind the scenes. Does it need to be autocloseable now?

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/ChannelOptions.java
```diff
@@ -161,6 +163,8 @@ public ChannelOptions(Credentials credential,
       ScheduledExecutorService scheduledExecutorService,
       long timeoutMs,
       int channelCount) {
+    Preconditions.checkArgument(credential == null || scheduledExecutorService != null,
+      "scheduledExecutorService has to be set if credentials are supplied");
```
mgarolera: Why? because they are refreshed? could you explain that in the comment?

mgarolera:
It's scary the amount of code that we are writing that doesn't seem particularly related to Cloud Bigtable, every grpc client shouldnt have to reimplement it.
Refresh, retry, reconnects... all this should be handled by GRPC. Maybe involve Nathan in this review?

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableChannels.java
sduskis: We don't terminate the channel.  We awaitTermination for 5 seconds and then try again.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableChannels.java
sduskis: I moved this setting to Transport options.  Only the data api needs more than 1 channel in the pool.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterAdminGrpcClient.java
sduskis: The major functionality I'm adding here is channelOptions.addClientCloseHandler(..).  There were artifacts created in both BigtableChannels and BigtableOptionsFactory that needed closing; they both call addClientCloseHandler now.  AbstractBigtableConnection.close() does the closing of those artifacts

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableTableAdminGrpcClient.java
```diff
@@ -50,17 +52,14 @@ public static BigtableTableAdminClient createClient(
     return new BigtableTableAdminGrpcClient(channel);
   }
 
-  protected final CloseableChannel channel;
   private final BigtableTableServiceGrpc.BigtableTableServiceBlockingStub blockingStub;
 
   /**
    * Create a new BigtableTableAdminGrpcClient. When constructed, the client takes ownership of the
```
AngusDavis: These comments are no longer correct (the client no longer takes ownership of the Channel)

sduskis:
I updated the code based on the requests, PTAL.

Nathan is aware of the situation.  He's been consulted through out the process.

Refresh/reconnect is due to a (potentially fixed) downstream problem in the web server.  There was an hourly timeout that wasn't handled very well in the web server.  It wasn't a gRPC problem, per se.

Retry is custom logic on our end.  It's more about business logic than an RPC tier issue

ChannelPooling was due to slow individual channel performance, which may have been fixed at this point.

sduskis:
Added comments to BigtableGrpcClient and BigtableTableAdminGrpcClient about how to close the channels that are passed in.

PR #357, opened 2015-07-15
Adding some startup performance enhancements

sduskis:
Parallelizing Host lookup and credential lookup, both of which can be expensive operations.

PR #362, opened 2015-07-16
Adding some startup performance enhancements

sduskis:
Parallelizing Host lookup and credential lookup, both of which can be expensive operations.

Adding some startup performance enhancements

Parallelizing Host lookup and credential lookup, both of which can be expensive operations.

Renaming the types in getHost parameters
Adding comments around the use of retryExecutor.

This is a continuation of #357 which got messed up

### DIFF
--- a/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
+++ b/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
@@ -21,13 +21,16 @@ import io.netty.channel.nio.NioEventLoopGroup;
 import java.io.Closeable;
 import java.io.IOException;
 import java.net.InetAddress;
+import java.net.UnknownHostException;
 import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.security.GeneralSecurityException;
+import java.util.concurrent.Callable;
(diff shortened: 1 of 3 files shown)


bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
```diff
@@ -178,66 +227,49 @@ public static BigtableOptions fromConfiguration(Configuration configuration) thr
...
-    bigtableOptionsBuilder.setChannelOptions(createChannelOptions(configuration));
-    createChannelOptions(configuration).addClientCloseHandler(new Closeable() {
-      @Override
-      public void close() throws IOException {
-        elg.shutdownGracefully();
-      }
-    });
+  private static InetAddress getHost(Configuration configuration, InetAddress overrideIpAddress,
+      String key, String defaultVal, String type) throws UnknownHostException {
+    String hostName = configuration.get(key, defaultVal);
```
mgarolera: use getValue?

mgarolera:
LGTM

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
sduskis: It's slightly different logic... defaultValue and logging are different for InetAddress host lookup.

PR #367, opened 2015-07-21
Introduce cbt session

sduskis:
BigtableSession should contain all Channel creation logic, including ExecutorService and Channel creation.


bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -0,0 +1,506 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+import javax.annotation.Nullable;
+import javax.net.ssl.SSLException;
+
+/**
+ * Encapsulates the creation of Bigtable Grpc services.
```
mgarolera: I would expand this comment. It is not only creation, it handles the life of several resources that last during the whole time that bigtable is used.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -0,0 +1,506 @@
...
+        }
+      }
+    });
+    for (final String host : Arrays.asList(BigtableOptions.BIGTABLE_HOST_DEFAULT,
+      BigtableOptions.BIGTABLE_CLUSTER_ADMIN_HOST_DEFAULT,
+      BigtableOptions.BIGTABLE_CLUSTER_ADMIN_HOST_DEFAULT)) {
+      connectionStartupExecutor.execute(new Runnable() {
+        @Override
+        public void run() {
+          // The first invocation of BigtableTableServiceGrpcs.CONFIG is expensive.
```
mgarolera: This comment is wrong

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -0,0 +1,506 @@
...
+      TransportOptions.BigtableTransports.HTTP2_NETTY_TLS);
+    this.scheduledRetries = scheduledRetries;
+  }
+
+  protected TransportOptions createTransportOptions(InetAddress host) {
+    return new TransportOptions(
+        TransportOptions.BigtableTransports.HTTP2_NETTY_TLS,
+        host,
+        options.getPort(),
+        SSL_CONTEXT_FACTORY,
```
mgarolera: Maybe move SSL_CONTEXT_FACTORY to TransportOptions, and no need to configure it.

Also, I know it's future proofing, but we might want to drop the HTTP2_NETTY_TLS option, or make it not configurable, and re-add it when we need it. Right now it adds some knotting.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -0,0 +1,506 @@
...
+        elg);
+  }
+
+  public synchronized BigtableClient getDataClient() {
+    if (this.client == null) {
```
mgarolera: Why not initialize them at construction time instead of lazily on request?

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -0,0 +1,506 @@
...
+    }
+
+    return client;
+  }
+
+  public synchronized BigtableTableAdminClient getTableAdminClient() {
+    if (this.tableAdminClient == null) {
+      Channel channel = createChannel(
+        createTransportOptions(options.getTableAdminHost()),
+        options.getChannelOptions(),
```
mgarolera: batchPool is a class fields, not sure its necessary to pass it here.

bigtable-grpc-interface/src/test/java/com/google/cloud/bigtable/config/BigtableOptionsTestFactory.java
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.config;
+
+/**
+ * This class provides a back door approach to creating BigtableOptions that are not fully
```
mgarolera: Isnt this the same as using the builder and not setting anything?

mgarolera:
LGTM with some minor comments.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
sduskis: This is also used by the Cluster API test.  We don't always need Data, TableAdmin or ClusterAdmin, and creating those objects are slow.

mgarolera: Sounds good. Note that this might be moving blocking time from when the session is created to the first operation, depending when getDataClient is first called.

PR #368, opened 2015-07-21
Removing BigtableGrpcClientOption and ChannelOptions.

sduskis:
BigtableGrpcClientOption was merged into RetryOptions and ChannelOptions was merged into BigtableOptsions.


bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
```diff
@@ -132,14 +196,26 @@ private BigtableOptions(
       String projectId,
       String zone,
       String cluster,
-      ChannelOptions channelOptions) {
+      Credentials credential,
+      String authority,
+      String userAgent,
+      String callTimingReportPath,
+      String callStatusReportPath,
+      RetryOptions unaryCallRetryOptions,
```
kevinsi4508: "RetryOptions retryOptions" instead of "RetryOptions unaryCallRetryOptions"?

kevinsi4508:
LGTM

PR #369, opened 2015-07-22
Removing TransportOptions

sduskis:
Also, delaying lookup of of InetAddress until it's needed.


bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
```diff
@@ -254,18 +261,22 @@ public String getCluster() {
...
+  public String getTableAdminHost() {
     return tableAdminHost;
   }
 
-  public InetAddress getClusterAdminHost() {
+  public String getClusterAdminHost() {
     return clusterAdminHost;
   }
 
+  public String getOverrideIp() {
```
mgarolera: I think this method deserves some documentation. What does it override? ...

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -265,36 +279,34 @@ public synchronized BigtableClusterAdminClient getClusterAdminClient() {
...
    * </p>
    */
-  protected Channel createChannel(final TransportOptions transportOptions, int channelCount) {
-
-    Preconditions.checkArgument(
-      transportOptions.getTransport() == TransportOptions.BigtableTransports.HTTP2_NETTY_TLS,
-      "Bigtable requires the NETTY_TLS transport.");
-    long timeoutMs = options.getTimeoutMs();
-
+  protected Channel createChannel(String hostString, int channelCount) throws IOException {
```
mgarolera: Update doc of this method?

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -210,53 +231,46 @@ public BigtableSession(BigtableOptions options, ExecutorService batchPool, Event
...
-        createTransportOptions(options.getClusterAdminHost()),
-        1);
+      Channel channel = createChannel(options.getClusterAdminHost(), 1);
       this.clusterAdminClient = new BigtableClusterAdminGrpcClient(channel);
     }
 
     return clusterAdminClient;
   }
 
+  private InetAddress getHost(String hostName) throws IOException {
```
mgarolera: You could already create here the InetSocketAddess here, and avoid passing InetAddess to later add the port.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -265,36 +279,34 @@ public synchronized BigtableClusterAdminClient getClusterAdminClient() {
...
-            .negotiationType(NegotiationType.TLS).streamWindowSize(1 << 20) // 1 MB
-            .build();
+        try {
+          return NettyChannelBuilder
+              .forAddress(new InetSocketAddress(host, options.getPort()))
+              .sslContext(SslContext.newClientContext())
+              .eventLoopGroup(elg)
+              .executor(batchPool)
+              .negotiationType(NegotiationType.TLS)
+              .streamWindowSize(1 << 20) // 1 MB
```
mgarolera: Doesnt have to be in this PR: This 1M should probably be part of BigtableOptions. Also, why default to 1M should have some explanation.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -265,36 +279,34 @@ public synchronized BigtableClusterAdminClient getClusterAdminClient() {
...
+              .build();
+        } catch (SSLException e) {
```
mgarolera: Seems SslContext.newClientContext throws an IllegalStateException already, so maybe the SSLException catch is not needed.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
```diff
@@ -170,20 +167,14 @@ private static String getValue(final Configuration configuration, String key, St
...
   }
 
-  private static InetAddress getHost(Configuration configuration, InetAddress overrideIpAddress,
-      String key, String defaultVal, String type) throws UnknownHostException {
+  private static String getHost(Configuration configuration, String key,
+      String defaultVal, String type) {
     String hostName = configuration.get(key, defaultVal);
     Preconditions.checkArgument(
         !Strings.isNullOrEmpty(hostName),
         String.format("%s endpoint host must be supplied via %s", type, key));
```
mgarolera: Is this even possible? if it is not set we will use the default, right?

bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -271,8 +270,8 @@ public void close() {
 
   @Override
   public String toString() {
-    InetAddress dataHost = options.getDataHost();
-    InetAddress tableAdminHost = options.getTableAdminHost();
+    String dataHost = options.getDataHost();
```
mgarolera: No need to create local variables for this.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
sduskis: SSLException is an IOException not a RuntimeException.  I'll refactor the IOException to be handled higher up the chain.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
sduskis: Done

sduskis:
Comments addressed.  PTAL.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
mgarolera: "optionally adding OAuth2 support" ?

mgarolera:
LGTM

PR #371, opened 2015-07-23
Updating integration tests to have a single version

sduskis:
The release process doesn't currently work because the integration tests pom.xml allows for multiple versions to be tested.  Reverting back to the old way of having only the current version to be testable.  We switched our jenkins jobs to pull in older versions of the tests using tags in so the pom.xml flexibility is no longer needed anyway.


kevinsi4508:
LGTM

PR #375, opened 2015-07-28
Improving shutdown time and changing SslContext creation.

sduskis:
The netty EventLoopGroup takes about 2 seconds to terminate.  It also shuts down when the JVM shuts down.  Older versions didn't wait for it to terminate, so this version shouldn't either.

Using GrpcSslContexts.forClient() as per Nathan's instruction instead of SslContext.newClientContext().

### DIFF
--- a/bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -45,11 +45,13 @@ import io.grpc.Metadata;
 import io.grpc.MethodDescriptor;
 import io.grpc.auth.ClientAuthInterceptor;
 import io.grpc.stub.MetadataUtils;
+import io.grpc.transport.netty.GrpcSslContexts;
 import io.grpc.transport.netty.NegotiationType;
 import io.grpc.transport.netty.NettyChannelBuilder;
 import io.netty.channel.EventLoopGroup;
 import io.netty.channel.nio.NioEventLoopGroup;
 import io.netty.handler.ssl.SslContext;
+import io.netty.handler.ssl.SslContextBuilder;
 
 import java.io.BufferedWriter;
 import java.io.Closeable;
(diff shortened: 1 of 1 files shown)


bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -99,28 +101,15 @@
...
-        // Each one needs its own SslContext.
-        return SslContext.newClientContext();
-      } catch (SSLException e) {
-        throw new IllegalStateException("Could not create an ssl context.", e);
-      }
+  static {
+
+    if (System.getProperty("java.version").startsWith("1.7.")) {
+      // The grpc cyphers only work in JDK 1.8+.  Use the default system cyphers for JDK 1.7.
+      sslBuilder.ciphers(null);
```
nmittler: You may also want to log a warning that they should use Java 8 to get more secure ciphers.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -99,28 +101,15 @@
...
-    @Override
-    public SslContext create() {
-      try {
-        // We create multiple channels via refreshing and pooling channel implementation.
-        // Each one needs its own SslContext.
-        return SslContext.newClientContext();
-      } catch (SSLException e) {
-        throw new IllegalStateException("Could not create an ssl context.", e);
-      }
+  static {
```
kevinsi4508: nit:
Can we make it a private static method?

private final static SslContextBuilder sslBuilder = getSslContextBuilder();

private static SslContextBuilder getSslContextBuilder() {
  SslContextBuilder sslBuilder = GrpcSslContexts.forClient();
  ...
  return sslBuilder;
}

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
kevinsi4508: +1

kevinsi4508:
LGTM

PR #376, opened 2015-07-28
Adding a channel for scans.

sduskis:
There was a problem doing mutations while a scan was ongoing.  Adding a channel specifically for scans.

### DIFF
--- a/bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableGrpcClient.java
+++ b/bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableGrpcClient.java
@@ -15,7 +15,6 @@
  */
 package com.google.cloud.bigtable.grpc;
 
-import com.google.api.client.repackaged.com.google.common.annotations.VisibleForTesting;
 import com.google.bigtable.v1.BigtableServiceGrpc;
 import com.google.bigtable.v1.CheckAndMutateRowRequest;
 import com.google.bigtable.v1.CheckAndMutateRowResponse;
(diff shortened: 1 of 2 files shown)


bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableGrpcClient.java
```diff
@@ -140,15 +139,18 @@ public void onCompleted() {
...
   private final RetryOptions retryOptions;
 
   public BigtableGrpcClient(
-      Channel channel,
+      Channel generalChannel,
+      Channel scanChannel,
       ExecutorService executorService,
       RetryOptions retryOptions) {
-    this.channel = channel;
+    this.generalChannel = generalChannel;
```
kevinsi4508: how about following?
readChannel
writeChannel?

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -271,8 +271,10 @@ public void run() {
 
   public synchronized BigtableClient getDataClient() throws IOException {
     if (this.client == null) {
-      Channel channel = createChannel(options.getDataHost(), options.getChannelCount());
-      this.client = new BigtableGrpcClient(channel, batchPool, options.getRetryOptions());
+      Channel generalChannel = createChannel(options.getDataHost(), options.getChannelCount());
+      Channel scanChannel = createChannel(options.getDataHost(), 1);
```
kevinsi4508: So for read, 1 channel is enough?

sduskis: Yes.  Scans run a long time over a single channel.  We encourage a lot of mutations asynchronously, so mutations benefit from more channels.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableGrpcClient.java
sduskis: that's reasonable.

sduskis:
Renamed the channels and added a comment in BigtableGrpcClient.

kevinsi4508:
LGTM

PR #382, opened 2015-07-31
Moving more startup logic into the BigtableSession startup.

sduskis:
I think the deferred startup was causing the performance tests that were blocking the launch of 0.2.0

### DIFF
--- a/bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -61,6 +61,7 @@ import java.net.UnknownHostException;
 import java.security.GeneralSecurityException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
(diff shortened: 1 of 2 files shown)


bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -102,6 +103,13 @@
   private static final Logger LOG = new Logger(BigtableSession.class);
 //  private static final SslContextBuilder sslBuilder = createGrpcSslBuilder();
```
kevinsi4508: remove

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -102,6 +103,13 @@
...
+  private static ExecutorService connectionStartupExecutor =
+      Executors.newCachedThreadPool(
+          new ThreadFactoryBuilder()
+              .setNameFormat("BigtableSession-startup-%s")
+              .setDaemon(true)
+              .build());
+
   static {
     performWarmup();
   }
```
kevinsi4508: Remove below comment if it is not needed anymore.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
sduskis: This is not related to the current PR.  We can address it elsewhere.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
sduskis: This is not related to the current PR.  We can address it elsewhere.

kevinsi4508:
LGTM

PR #385, opened 2015-08-03
Some usability improvements to BigtableSession.

sduskis:
close() does not have 'throws Exception' in the signature
Added a new constructor for BigtableSession that only takes in a BigtableOption.
BigtableSessionTests now does not open a BigtableSession.
TestClusterAPI uses the new single parameter BigtableSession constructor.

### DIFF
--- a/bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -88,12 +88,14 @@ import javax.net.ssl.SSLException;
  */
 public class BigtableSession implements AutoCloseable {
 
+  public static final String BATCH_POOL_THREAD_NAME = "bigtable-batch-pool";
+
   /** Entry in call reports that indicates an entry is from before retries */
   private static final String PRE_RETRY_REPORT_ENTRY = "PreRetry";
   /** Entry in call reports that indicates an entry is from after retries */
   private static final String POST_RETRY_REPORT_ENTRY = "PostRetry";
-  /** Number of threads to use to initiate retry calls */
   public static final String RETRY_THREADPOOL_NAME = "bigtable-rpc-retry";
+  /** Number of threads to use to initiate retry calls */
   public static final int RETRY_THREAD_COUNT = 4;
   public static final String GRPC_EVENTLOOP_GROUP_NAME = "bigtable-grpc-elg";
   /** Number of milliseconds to wait for a termination before trying again. */
@@ -205,6 +207,14 @@ public class BigtableSession implements AutoCloseable {
     connectionStartupExecutor.shutdown();
   }
 
+  protected static ExecutorService createDefaultBatchPool(){
+    return Executors.newCachedThreadPool(
+      new ThreadFactoryBuilder()
+          .setNameFormat(BATCH_POOL_THREAD_NAME + "-%d")
+          .setDaemon(true)
+          .build());
+  }
+
   protected static EventLoopGroup createDefaultEventLoopGroup() {
     return new NioEventLoopGroup(0,
         new ThreadFactoryBuilder()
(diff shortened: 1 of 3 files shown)


bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -205,7 +207,15 @@ public void run() {
     connectionStartupExecutor.shutdown();
   }
 
-  protected static EventLoopGroup createDefaultEventLoopGroup() {
+  protected ExecutorService createDefaultBatchPool(){
+    return Executors.newCachedThreadPool(
+      new ThreadFactoryBuilder()
+          .setNameFormat(BATCH_POOL_THREAD_NAME + "-d")
```
kevinsi4508: nit: why BATCH_POOL_THREAD_NAME? and then append "-d" here?

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -213,7 +223,7 @@ protected static EventLoopGroup createDefaultEventLoopGroup() {
             .build());
   }
 
-  protected static ScheduledExecutorService createDefaultRetryExecutor() {
+  protected ScheduledExecutorService createDefaultRetryExecutor() {
```
kevinsi4508: why changed from static to non-static?

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -232,27 +242,32 @@ protected static ScheduledExecutorService createDefaultRetryExecutor() {
   private final ScheduledExecutorService scheduledRetries;
   private final List<Closeable> clientCloseHandlers = Collections
       .synchronizedList(new ArrayList<Closeable>());
-  private final ClientInterceptor clientAuthInterceptor;
+  private ClientInterceptor clientAuthInterceptor;
+
+  public BigtableSession(BigtableOptions options) throws IOException {
+    this (options, null);
```
kevinsi4508: I personally prefer to call one overloaded constructor.  example:

public BigtableSession(BigtableOptions options) throws IOException {
  this(options, null, null, null);
}

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -232,27 +242,32 @@ protected static ScheduledExecutorService createDefaultRetryExecutor() {
...
+  }
 
   public BigtableSession(BigtableOptions options, ExecutorService batchPool) throws IOException {
     this(options, batchPool, null, null);
   }
 
-  public BigtableSession(final BigtableOptions options, final ExecutorService batchPool,
+  public BigtableSession(BigtableOptions options, ExecutorService batchPool,
```
kevinsi4508: @Nullable ExecutorService

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -274,9 +317,9 @@ public Void call() throws Exception {
             return null;
           }
         });
-        clientAuthInterceptor = oauth2Interceptor;
+        this.clientAuthInterceptor = oauth2Interceptor;
       } else {
-        clientAuthInterceptor = new ClientAuthInterceptor(credentials, batchPool);
+        this.clientAuthInterceptor = new ClientAuthInterceptor(credentials, batchPool);
         clientAuthInterceptorFuture = batchPool.submit(new Callable<Void>() {
```
kevinsi4508: I don't know if we should call it clientAuthInterceptorFuture, it is used to refresh credentials.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -232,27 +242,32 @@ protected static ScheduledExecutorService createDefaultRetryExecutor() {
   private final ScheduledExecutorService scheduledRetries;
   private final List<Closeable> clientCloseHandlers = Collections
       .synchronizedList(new ArrayList<Closeable>());
-  private final ClientInterceptor clientAuthInterceptor;
+  private ClientInterceptor clientAuthInterceptor;
```
kevinsi4508: where is this used?

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
sduskis: I figured that users should be able to override this, if they want to.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
sduskis: This is used in wrapChannel.

sduskis:
I addressed the comments.  PTAL.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
kevinsi4508: but I thought we are letting user pass in ScheduledExecutorService. I prefer not allowing user to override this, up to you though.

kevinsi4508:
LGTM

PR #388, opened 2015-08-03
Reverting the change made for the scan/delete scenario.

sduskis:


kevinsi4508:
It seems to have conflicts, otherwise LGTM

PR #389, opened 2015-08-03
Minor performance improvement in RefreshingOAuth2CredentialsInterceptor.

sduskis:


bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -86,12 +85,12 @@ public HeaderCacheElement(IOException exception) {
 
     public CacheState getCacheState() {
       long now = System.currentTimeMillis();
-      if (now > expiresTimeMs) {
-        return CacheState.Expired;
-      } else if (now > staleTimeMs) {
+      if (now < staleTimeMs){
+        return CacheState.Good;
```
kevinsi4508: we return CacheState.Good when  now =< staleTimeMs && now =< expiresTimeMs, what is the reason for changing the logic here?

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -171,11 +170,18 @@ public void syncRefresh() throws IOException {
   private String getHeader() throws IOException {
     HeaderCacheElement headerCache = getCachedHeader();
     CacheState state = (headerCache == null) ? CacheState.Expired : headerCache.getCacheState();
-    if (state == CacheState.Expired) {
+    switch (state) {
+    case Good:
```
kevinsi4508: two chars indention

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/RefreshingOAuth2CredentialsInterceptor.java
sduskis: The end result shouldn't be different.  This change is to have the most common case, the Good case, require only a single check rather than 2 checks against longs.  If "now =< staleTimeMs" then "now =< expiresTimeMs" is implicit, since staleTimeMs < expiresTimeMs.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/RefreshingOAuth2CredentialsInterceptor.java
sduskis: done

sduskis:
Added the 2 char indentation

sduskis:
FYI, I did a perf test with this code, and this appears to correct the performance problems we've been seeing.

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/RefreshingOAuth2CredentialsInterceptor.java
kevinsi4508: I see, it looks like staleTimeMs < expiresTimeMs, but there is nothing in this code guarantee that is the case.
Can we at least do a check in the constructor?

kevinsi4508:
The change in the CL is for minor performance improvement and your test result indicated a bigger performance improvement than expected?

bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/RefreshingOAuth2CredentialsInterceptor.java
sduskis: I added a precondition in the constructor.

kevinsi4508:
LGTM

PR #391, opened 2015-08-06
Massive renaming in bigtable-grpc-interface

sduskis:
Renaming bigtable-grpc-interface to bigtable-client-core
Moving classes in bigtable-client-core into the following packages
 - client
 - config
 - grpc
 - scanner


bigtable-client-core/src/main/java/com/google/cloud/bigtable/client/BigtableDataClient.java
```diff
@@ -23,17 +25,16 @@
...
+import com.google.cloud.bigtable.scanner.ResultScanner;
 import com.google.common.collect.ImmutableList;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.protobuf.Empty;
 import com.google.protobuf.ServiceException;
 
-import java.util.List;
-
 /**
  * Interface to access v1 final Bigtable service methods
```
kevinsi4508: nit: Bigtable data service methods.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/client/BigtableDataGrpcClient.java
```diff
@@ -49,7 +54,7 @@
 /**
  * A gRPC client to access the v1 Bigtable service.
```
kevinsi4508: ditto Bigtable data service.

PR #394, opened 2015-08-07
Initial commit of Cloud Bigtable/Dataflow connector.

sduskis:


bigtable-hbase-dataflow/README.md
```diff
@@ -0,0 +1,3 @@
+# Bigtable Dataflow
+
+This project contains the core [Bigtable](https://cloud.google.com/bigtable/)-[Dataflow](https://cloud.google.com/dataflow/) integration.  The main class you should care about is: [CloudBigtableIO](https://github.com/GoogleCloudPlatform/cloud-bigtable-cdf/blob/master/bigtable-dataflow/src/main/java/com/google/cloud/dataflow/sdk/io/CloudBigtableIO.java).  It contains the Cloud Bigtable Sink and Source.
```
mgarolera: update link to point to the right repo

bigtable-hbase-dataflow/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+        <artifactId>bigtable-client</artifactId>
+        <version>0.2.1-SNAPSHOT</version>
+    </parent>
+
+    <groupId>com.google.cloud.bigtable.dataflow</groupId>
+    <artifactId>bigtable-hbase-dataflow</artifactId>
+    <packaging>jar</packaging>
+    <name>${project.groupId}:${project.artifactId}</name>
+    <description>
+       This project contains artifacts that adapt bigtable client to work with Google Cloud Dataflow.
```
mgarolera: bigtable -> (Cloud?) Bigtable

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
```diff
@@ -0,0 +1,101 @@
...
+
+import java.io.Serializable;
+
+import org.apache.hadoop.conf.Configuration;
+
+import com.google.cloud.bigtable.config.BigtableOptions;
+import com.google.cloud.bigtable.hbase.BigtableOptionsFactory;
+
+/**
+ * This class defines all of the information cloud bigtable client needs to connect to a user's
```
mgarolera: cloud bigtable - > Cloud Bigtable

Why does this belong to the dataflow package?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableTableConfiguration.java
```diff
@@ -0,0 +1,76 @@
...
+        options.getBigtableTable());
+  }
+
+  public static class Builder {
+    private String projectId;
+    private String zone;
+    private String cluster;
+    private String table;
+
+    public Builder withProjectId(String projectId) {
```
mgarolera: Why not extend also CloudBigtableConfiguration.Builder and have the constructor of CloudBigtableTableConfiguration take an already built CloudBigtableConfiguration and a table?

Its a lot of code repetition (withProjectId is defined 5 times) for something that is only used once in a test.

bigtable-hbase-dataflow/README.md
```diff
@@ -0,0 +1,3 @@
+# Bigtable Dataflow
```
mgarolera: This is a big PR....

Bigtable HBase Dataflow? To match the path.

On that note... This uses HBase, mention it in the documentation? Will there be a non-HBase dataflow?
Does this belong to cloud-bigtable-client repo? Why not keep it in a separate repo, since it is justa a use of the cbt client and not part of it? Also, was this developed in a separate repo? should we try to keep the commits?

Also is it worth creating integration tests? Specially since it interacts with another system.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableTableConfiguration.java
sduskis: It's the return type that gets you.  Builder is different for all incarnations of withProjectId... it's for chaining purposes.  We can probably just get rid of the Builder concept and use getters/setters.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,440 @@
...
+      public SourceWithKeys(
+          CloudBigtableScanConfiguration configuration, byte[] startKey, byte[] stopKey) {
+        this.configuration = configuration;
+        this.startKey = startKey;
+        this.stopKey = stopKey;
+      }
+
+      @Override
+      public long getEstimatedSizeBytes(PipelineOptions options) throws Exception {
+        return 0;
```
chamikaramj: Can't we estimated the size of the source here ?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,440 @@
...
+
+      private Result current;
+
+      public Reader(BoundedSource<Result> source, CloudBigtableScanConfiguration config,
+          byte[] startRow, byte[] endRow) throws IOException {
+        super();
+        this.source = source;
+        this.config = config;
+        if (startRow != null || endRow != null) {
+          this.scan = new Scan(config.getScan());
```
chamikaramj: Do any operations in the Reader constructor perform IO operations ? If so please move them to the start() method.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,440 @@
...
+
+      @Override
+      public List<? extends BoundedSource<Result>> splitIntoBundles(long arg0, PipelineOptions arg1)
+          throws Exception {
+        return Collections.singletonList(this);
+      }
+
+      @Override
+      public void validate() {
+        checkNotNullOrEmpty(configuration.getProjectId(), "project");
```
chamikaramj: Can you perform a remote (Cloud BigTable) operation to check if the given key range is actually valid ?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,440 @@
...
+      public BoundedSource<Result> getCurrentSource() {
+        return source;
+      }
+
+      @Override
+      public boolean start() throws IOException {
+        connection = new BigtableConnection(config.toHBaseConfig());
+        table = connection.getTable(TableName.valueOf(config.getTable()));
+        scanner = table.getScanner(scan);
+        return advance();
```
chamikaramj: We shouldn't do an advance() call here. The sequence of operations for the user to get the first record will be start(), advance(), getCurrent(). If we do an advance() call in the start() method the user will miss the first record.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,440 @@
...
+    public Coder<Result> getDefaultOutputCoder() {
+      return new HBaseResultCoder();
+    }
+
+    @Override
+    public List<BoundedSource<Result>> splitIntoBundles(
+        long desiredBundleSizeBytes, PipelineOptions options) throws Exception {
+      TableName tableName = TableName.valueOf(configuration.getTable());
+      try (Connection conn = new BigtableConnection(configuration.toHBaseConfig());
+          Admin admin = conn.getAdmin(); RegionLocator locator = conn.getRegionLocator(tableName)) {
```
chamikaramj: Can you document the approach you use for splitting the source here ?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,440 @@
...
+        for (int i = 0; i < startKeys.length; i++) {
+          splits.add(new SourceWithKeys(configuration, startKeys[i], endKeys[i]));
+        }
+        return splits;
+      }
+    }
+
+    @Override
+    public void validate() {
+      checkNotNullOrEmpty(configuration.getProjectId(), "project");
```
chamikaramj: Can you validate if the project, zone, cluster, and table are actually valid here ?

bigtable-hbase-dataflow/src/test/java/com/google/cloud/bigtable/dataflow/CloudBigtableIOTest.java
```diff
@@ -0,0 +1,111 @@
...
+import com.google.cloud.dataflow.sdk.Pipeline;
+import com.google.cloud.dataflow.sdk.coders.CannotProvideCoderException;
+import com.google.cloud.dataflow.sdk.coders.Coder;
+import com.google.cloud.dataflow.sdk.coders.CoderRegistry;
+import com.google.cloud.dataflow.sdk.values.TypeDescriptor;
+
+/**
+ * Tests for {@link CloudBigtableIO}.
+ */
+public class CloudBigtableIOTest {
```
chamikaramj: Please add method to test read, splitIntoBundles, and other methods in Source/Sink classes. Can you mock CloudBigTableConnection for this ? Please see DatastoreIOTest, FileBasedSourceTest for examples.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,440 @@
...
+      return 0;
+    }
+
+    @Override
+    public boolean producesSortedKeys(PipelineOptions options) throws Exception {
+      return true;
+    }
+
+    @Override
+    public com.google.cloud.dataflow.sdk.io.BoundedSource.BoundedReader<Result> createReader(
```
bjchambers: This can be BoundedSource.BoundedReader<Result>

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableOptions.java
```diff
@@ -0,0 +1,40 @@
...
+package com.google.cloud.bigtable.dataflow;
+
+import com.google.cloud.dataflow.sdk.options.PipelineOptions;
+
+/**
+ * CloudBigtableOptions is an extension of {@link PipelineOptions} the information that {@link
+ * CloudBigtableConfiguration} needs to configure a connection to Cloud Bigtable.
+ */
+public interface CloudBigtableOptions extends PipelineOptions {
+  String getBigtableProject();
```
bjchambers: Do these options need validation? Should any of them be marked required?

Can you also add a description of the interface and each of the options:

https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/bdde37492d5f79b3570a543fe6392391023dddec/sdk/src/main/java/com/google/cloud/dataflow/sdk/options/BigQueryOptions.java

bigtable-hbase-dataflow/src/test/java/com/google/cloud/bigtable/dataflow/CloudBigtableIOTest.java
```diff
@@ -0,0 +1,111 @@
...
+  public void testInitialize() throws Exception{
+    CloudBigtableIO.initializeForSink(underTest);
+    checkRegistry(Put.class);
+    checkRegistry(Delete.class);
+    checkRegistry(Mutation.class);
+
+    // Appends and Increments are not supported
+    try {
+      registry.getCoder(TypeDescriptor.of(Append.class));
+      Assert.fail("Appends should not have Coders");
```
bjchambers: Prefer smaller tests that use @Rule ExpectedException for verifying exceptions.

bigtable-hbase-dataflow/src/test/java/com/google/cloud/bigtable/dataflow/CloudBigtableScanConfigurationTest.java
```diff
@@ -0,0 +1,70 @@
...
+    CloudBigtableScanConfiguration config = new CloudBigtableScanConfiguration.Builder()
+      .withProjectId(projectId)
+      .withZone(zone)
+      .withCluster(cluster)
+      .withTable(table)
+      .withScan(new Scan(startRow, stopRow))
+      .build();
+
+    ByteArrayOutputStream bos = new ByteArrayOutputStream();
+    ObjectOutputStream oos = new ObjectOutputStream(bos);
```
bjchambers: Look at [SerializableUtils](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/bdde37492d5f79b3570a543fe6392391023dddec/sdk/src/main/java/com/google/cloud/dataflow/sdk/util/SerializableUtils.java) for helpers to test serializability.

bigtable-hbase-dataflow/src/test/java/com/google/cloud/bigtable/dataflow/HBaseMutationCoderTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.junit.Test;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+
+/**
+ * Tests for {@link HBaseMutationCoder}.
+ */
+public class HBaseMutationCoderTest {
```
bjchambers: Look at (CoderProperties)[https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/e872e0b55b131df432bdf3d34879a78dfdbba1c0/sdk/src/main/java/com/google/cloud/dataflow/sdk/testing/CoderProperties.java]

for helpers to automatically test coders.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/HBaseResultCoder.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.Serializable;
+
+/**
+ * A {@link Coder} that serializes {@link Result} objects used by the CBT {@link Source}.
+ */
+public class HBaseResultCoder extends AtomicCoder<Result> implements Serializable{
```
bjchambers: Is this encoding deterministic -- if two results are equal, will it always produce the same bytes? If not, it should have an ensureDeterministic that indicatse such.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/HBaseMutationCoder.java
```diff
@@ -0,0 +1,71 @@
...
+import com.google.cloud.dataflow.sdk.coders.AtomicCoder;
+import com.google.cloud.dataflow.sdk.coders.Coder;
+import com.google.cloud.dataflow.sdk.coders.CoderException;
+
+/**
+ * A {@link Coder} that serializes {@link Mutation} objects used by the CBT {@link Sink}. See
+ * {@link CloudBigtableIO#initializeForSink(Pipeline)}.
+ */
+
+public class HBaseMutationCoder extends AtomicCoder<Mutation> implements Serializable {
```
bjchambers: Is this encoding deterministic -- if two mutations are equal, will it always produce the same bytes? If not, it should have an ensureDeterministic that indicates such.

bigtable-hbase-dataflow/src/test/java/com/google/cloud/bigtable/dataflow/HBaseResultCoderTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.google.cloud.bigtable.dataflow.HBaseResultCoder;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+
+/**
+ * Tests for {@link HBaseResultCoder}
+ */
+public class HBaseResultCoderTest {
```
bjchambers: Same for testing here.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
jkff: Actually the code is correct - the protocol is that start() reads the first record and subsequent advance()'s read subsequent records. A record should be available via getCurrent() immediately after start(), and many readers call advance() from start() after opening the resources.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,440 @@
...
+      CloudBigtableOptions options) {
+    return write(CloudBigtableTableConfiguration.fromCBTOptions(options));
+  }
+
+  public static com.google.cloud.dataflow.sdk.io.Sink<Mutation> write(
+      CloudBigtableTableConfiguration config) {
+    return new Sink(config);
+  }
+
+  public static PTransform<PCollection<Mutation>, PDone> streamingWrite(
```
bjchambers: The streamingWrite shouldn't be a separate transform. There should be a single write() that the user provides, that "does the right thing". What is the purpose of distinguishing these? For now, it may be best to have the write PTransform select its behavior from the isStreaming pipeline option, or the bounded/unboundedness of the input PCollection, depending on what the differences in behavior are.

bigtable-hbase-dataflow/README.md
sduskis: We'll have an easier time maintaining a single repo than multiple repos.  Integration tests are a bit tricky.  I'll add a hello world job that will serve as a short term test.  We'll add more integration tests shortly.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: Not easily, yet.  We'll add that in the near term.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: No, these do not perform IO operations.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: Not easily here.  We use http2 under the covers, which requires some difficult configuration from the user's perspective.  I don't want to put the onus on the user.  I'll add an issue to have an option to do RPCs via http that would not require ALPN.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: Thanks for confirming jkff.  I copied the approach from the Dataflow SDK.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: No

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: This SourceWithKeys can only be created via a previous API call to the server.  Is there a reason to suspect that the previous call won't be OK?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: Thanks for the clarification.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: Unfortunately, not on the client side yet.  I added issue #399 for this.

bigtable-hbase-dataflow/src/test/java/com/google/cloud/bigtable/dataflow/CloudBigtableIOTest.java
sduskis: Added issue #401

bigtable-hbase-dataflow/src/test/java/com/google/cloud/bigtable/dataflow/CloudBigtableScanConfigurationTest.java
sduskis: Nice!

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
chamikaramj: I think it's good to implement validation for each sub-source, in case the keys get deleted for example. We'll be validating each sub-source before reading it. But we can add a TODO and proceed if this is not possible now.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
chamikaramj: @jkff thanks for correcting that.

bigtable-hbase-dataflow/src/test/java/com/google/cloud/bigtable/dataflow/HBaseMutationCoderTest.java
sduskis: CoderProperties required hamcrest and json dependencies.  I'm going to punt on this for now.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: I originally added a Sink, since I thought that was the idiomatic way.  I'll remove the Sink, since it seems to be superfluous.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: I'm not sure what the validation will entail here.  The code will work fine if keys get added or deleted or even if the tablet split again.  Any rows that were deleted will simply not be iterated through as part of the scan.

Perhaps I'm missing something here.  I'm not sure what the validation would be in this case, since this split is assumed to be part of a whole.  A single call to RegionLocator.getStartEndKeys() ought to be sufficient to get a breakdown of all rows.

chamikaramj: Ok, sounds good. Implementing validate() for top level source (or a TODO there) should be enough then.

bigtable-hbase-dataflow/README.md
```diff
@@ -0,0 +1,92 @@
...
+    };
+
+    CountOptions options =
+        PipelineOptionsFactory.fromArgs(args).withValidation().as(CountOptions.class);
+
+    // See the hbase hadoop job at
+    // https://github.com/apache/hbase/blob/master/hbase-server/src/main/java/org/apache/hadoop/hbase/mapreduce/RowCounter.java#L151
+    // for more ways to configure this scan.
+    Scan scan = new Scan();
+    scan.setCacheBlocks(false);
```
mgarolera: Why is this necessary? I think that in bigtable it is not piped through.

bigtable-hbase-dataflow/README.md
```diff
@@ -0,0 +1,92 @@
...
+    scan.setFilter(new FirstKeyOnlyFilter());
+
+    // CloudBigtableTableConfiguration contains the project, zone, cluster and table to connect to.
+    // You can supply an optional Scan() to filter the .
```
mgarolera: filter the.. ?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableTableConfiguration.java
mgarolera: Ahh.. true.

And is CloudBigtableConfiguration used for anything else asides from superclass of CloudBigtableTableConfiguration? I couldn't find it, so maybe we can colapse the two classes into a single one.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableRowFilterConfiguration.java
```diff
@@ -0,0 +1,95 @@
...
+
+import com.google.bigtable.v1.RowFilter;
+
+/**
+ * This class encapsulates the metadata required to create a connection to a Bigtable cluster,
+ * and a specific table therein along with a filter on the table in the form of a {@link Scan}.
+ * NOTE: {@link Scan} isn't Serializable, so this class uses HBase's {@link ProtobufUtil} as part
+ * of the {@Externalizable} implementation.
+ */
+public class CloudBigtableRowFilterConfiguration extends CloudBigtableTableConfiguration {
```
mgarolera: Is this used? Why wouldn't you use CloudBigtableScanConfiguration?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableScanConfiguration.java
```diff
@@ -0,0 +1,134 @@
...
+
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.Serializable;
+
+/**
+ * This class encapsulates the metadata required to create a connection to a Bigtable cluster,
+ * and a specific table therein along with a filter on the table in the form of a {@link Scan}.
+ * NOTE: {@link Scan} isn't Serializable, so this class uses HBase's {@link ProtobufUtil} as part
```
mgarolera: Mention that even if the Scan has start and end keys, they are ignored?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,348 @@
...
+
+    /**
+     * A {@link BoundedSource} for a Bigtable {@link Table} with a start/stop key range, along with
+     * a potential filter via a {@link Scan}.
+     */
+    private static class SourceWithKeys extends BoundedSource<Result> {
+      private static final long serialVersionUID = 2561066007121040429L;
+
+      private final CloudBigtableScanConfiguration configuration;
+      private final byte[] startKey;
```
mgarolera: You use both startKey or startRow. I think we can use everywhere startRow, which is what the scanner uses. Or is Key a dataflow concept? would be nice to stick with one.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,348 @@
...
+        return true;
+      }
+
+      @Override
+      public Coder<Result> getDefaultOutputCoder() {
+        return new HBaseResultCoder();
+      }
+
+      /**
+       * This does not actually split.  It just returns this.  This ought to be fixed.
```
mgarolera: Add TODO? mention what is blocking it if anything?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,348 @@
...
+
+      @Override
+      public BoundedSource.BoundedReader<Result> createReader(PipelineOptions options)
+          throws IOException {
+        return new CloudBigtableIO.Source.Reader(this, configuration, startKey, stopKey);
+      }
+    }
+
+    /**
+     * Reads rows for a specific {@link Table}, potentially filted by a Scan and/or a start/stop key
```
mgarolera: "filted" ?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,348 @@
...
+    @Override
+    public Coder<Result> getDefaultOutputCoder() {
+      return new HBaseResultCoder();
+    }
+
+    /**
+     * Splits the table based on keys that belong to "regions" a.k.a. tablets via the HBase
+     * {@link RegionLocator} interface which calls
+     * {@link BigtableService#sampleRowKeys(com.google.bigtable.v1.SampleRowKeysRequest, io.grpc.stub.StreamObserver)
+     * under the covers}.  Each BoundedResource will correspond to a single region.
```
mgarolera: Mention that desiredBundleSizeBytes is ignored?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,348 @@
...
+      private ResultScanner scanner;
+      private Table table;
+      private Result current;
+
+      public Reader(BoundedSource<Result> source, CloudBigtableScanConfiguration config,
+          byte[] startRow, byte[] endRow) throws IOException {
+        super();
+        this.source = source;
+        this.config = config;
+        if (startRow != null || endRow != null) {
```
mgarolera: nit: Maybe it will be clearer if this is split into two constructors instead of using nulls? One for a reader bounded with start and end rows, and one for an.. unbounded reader?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,348 @@
...
+   *
+   * @return the original Pipeline.
+   */
+  @SuppressWarnings({"rawtypes", "unchecked"})
+  public static Pipeline initializeForWrite(Pipeline p) {
+    // This enables the serialization of various Mutation types in the pipeline.
+    CoderRegistry registry = p.getCoderRegistry();
+    Coder coder = new HBaseMutationCoder();
+
+    // MutationCoder only supports Puts and Deletes. It will throw exceptions for Increment
```
mgarolera: How strict is the requirement for idempotency? Puts are idempotent only if they have a timestamp set.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,348 @@
...
+      @Override
+      public long getEstimatedSizeBytes(PipelineOptions options) throws Exception {
+        // TODO(sduskis): Hook up the right size for the source.  The Cloud Bigtable API has the
+        // capability to get the size, but it's not currently exposed via HBase. (issue #400)
+        return 0;
+      }
+
+      @Override
+      public boolean producesSortedKeys(PipelineOptions options) throws Exception {
+        return true;
```
mgarolera: nit: HBase Scans support setReversed. Bigtable doesn't support it and ignores (ugh) that value.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: idempotence is relative.  Increment and Append will never be idempotent.  Puts will at least be logically idempotent in most cases.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: The separate Reader constructor is a good idea.

Unbounded reader has a different meaning than "really big".  Unbounded is more about changes over time, which in the case of Bigtable would be updates to the table as opposed to reading a huge table.  We don't currently have a way to get updates in Cloud Bigtable.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: Key is an HBase concept.  I'll change everything to use row.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableScanConfiguration.java
sduskis: I'll fix this.  I'll copy the HBase TableInputFormatBase logic for start/end keys.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
mgarolera: Do users set the timestamp by default? you don't do it in the example, which makes it not idempotent. I assume the idempotency requirement is because dataflow might retry. Do we know how often? Are users ok with seeing repeated entries at different timestamps?

At the very least this deserves a TODO.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
```diff
@@ -0,0 +1,102 @@
...
+  public String getClusterId() {
+    return clusterId;
+  }
+
+  public BigtableOptions toBigtableOptions() {
+    return new BigtableOptions.Builder()
+      .setProjectId(getProjectId())
+      .setZoneId(getZoneId())
+      .setClusterId(getClusterId())
+      .setUserAgent("CloudBigtableGrpcIO")
```
mgarolera: Maybe mention dataflow? Why IO?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,551 @@
...
+        request.setTableName(tableName.toString());
+
+        List<SampleRowKeysResponse> sampleRowKeys =
+            session.getDataClient().sampleRowKeys(request.build());
+        byte[] startKey = HConstants.EMPTY_START_ROW;
+        long lastOffset = 0;
+        for (SampleRowKeysResponse response : sampleRowKeys) {
+          byte[] endKey = response.getRowKey().toByteArray();
+          // Avoid empty regions.
+          if (Bytes.equals(startKey, endKey) && startKey.length > 0) {
```
mgarolera: Is this possible? also, whystartKey.length > 0? Maybe ==HConstants.EMPTY_START_ROW?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,551 @@
...
+    public long getEstimatedSizeBytes(PipelineOptions options) throws Exception {
+      BigtableOptions bigtableOptions = configuration.toBigtableOptions();
+      long totalEstimatedSizeBytes = 0;
+
+      Scan scan = configuration.getScan();
+      byte[] scanStartKey = scan.getStartRow();
+      byte[] scanEndKey = scan.getStopRow();
+
+      List<SourceWithKeys> splits = new ArrayList<>();
+      try (BigtableSession session = new BigtableSession(bigtableOptions)) {
```
mgarolera: Make this a separate function returning  List<SampleRowKeysResponse>?

+    try (BigtableSession session = new BigtableSession(bigtableOptions)) {
+        BigtableClusterName clusterName = bigtableOptions.getClusterName();
+        BigtableTableName tableName = clusterName.toTableName(configuration.getTableId());
+        SampleRowKeysRequest.Builder request = SampleRowKeysRequest.newBuilder();
+        request.setTableName(tableName.toString());
+        List<SampleRowKeysResponse> sampleRowKeys =
+            session.getDataClient().sampleRowKeys(request.build());

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,551 @@
...
+      return true;
+    }
+
+    @Override
+    public BoundedSource.BoundedReader<Result> createReader(PipelineOptions options)
+        throws IOException {
+      return new Source.Reader(this, configuration, configuration.getScan());
+    }
+
+    private void split(List<SourceWithKeys> results, long regionSize,
```
mgarolera: Return a list of SourceWithKeys instead of appending to a parameter?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,551 @@
...
+      BigtableOptions bigtableOptions = configuration.toBigtableOptions();
+      try (BigtableSession session = new BigtableSession(bigtableOptions)) {
+        BigtableClusterName clusterName = bigtableOptions.getClusterName();
+        BigtableTableName tableName = clusterName.toTableName(configuration.getTableId());
+        SampleRowKeysRequest.Builder request = SampleRowKeysRequest.newBuilder();
+        request.setTableName(tableName.toString());
+
+        List<SampleRowKeysResponse> sampleRowKeys =
+            session.getDataClient().sampleRowKeys(request.build());
+        byte[] startKey = HConstants.EMPTY_START_ROW;
```
mgarolera: lastStartKey?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,551 @@
...
+                    ? endKey
+                    : scanEndKey;
+            split(splits, offset - lastOffset, desiredBundleSizeBytes, splitStart, splitStop);
+          }
+          lastOffset = offset;
+          startKey = endKey;
+        }
+        // Create one last region if the last region doesn't reach the end or there are no regions.
+        byte[] endKey = HConstants.EMPTY_END_ROW;
+        if (!Bytes.equals(startKey, endKey)) {
```
mgarolera: What if scanEndKey > Empty_end_row?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -0,0 +1,551 @@
...
+          byte[] endKey = response.getRowKey().toByteArray();
+          // Avoid empty regions.
+          if (Bytes.equals(startKey, endKey) && startKey.length > 0) {
+            continue;
+          }
+
+          long offset = response.getOffsetBytes();
+          // Get all the start/end key ranges that match the user supplied Scan.  See
+          // https://github.com/apache/hbase/blob/master/hbase-server/src/main/java/org/apache/hadoop/hbase/mapreduce/TableInputFormatBase.java#L298
+          // for more info.
```
mgarolera: There isn't more info in that link.
I find the code hard to read. Comparing with HConstants.EMPTY_START_ROW and HConstants.EMPTY_END_ROW might make it clearer.

What do you think of:
if ( scanStartKey == HConstants.EMPTY_START_ROW ||
    Bytes.compareTo(scanStartKey, startKey) < 0)
  splitStart = startKey;
else
  splitStart = scanStartKey;

// similar for endkey

if (Bytes.compareTo(splitStart, splitEnd) || splitEnd == HConstants.EMPTY_END_ROW)  {
  split(splits, offset - lastOffset, desiredBundleSizeBytes, splitStart, splitStop)
}

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: It turns out that we need ALPN configured on the user side for Sources.  I'll add validation that the connection can be created and the table exists before we launch dataflow jobs.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: Validating here as well.

bigtable-hbase-dataflow/README.md
sduskis: I'll remove it.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableRowFilterConfiguration.java
sduskis: I'll remove this.  This was used as part of the CloudBigtableGrpcIO class, and should not be exposed here.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableScanConfiguration.java
sduskis: I fixed the start/end key issue.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: Removed the comment.  The code does split now.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: desiredBundleSizeBytes is no longer ignored.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: I added more comments around how to make Puts idempotent by setting the timestamp.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
sduskis: Updated.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: Empty byte arrays are not necessarily ==.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: Good idea.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: ok

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: currentStartKey = lastEndKey.  lastStartKey doesn't seem to make sense here.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
sduskis: good point.

bigtable-hbase-dataflow/src/test/java/com/google/cloud/bigtable/dataflow/CloudBigtableIOIntegrationTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+  public static final byte[] COLUMN_FAMILY = Bytes.toBytes("test_family");
+  public static final byte[] QUALIFIER1 = Bytes.toBytes("qualifier1");
+
+  private static String projectId = System.getProperty(BIGTABLE_PROJECT_KEY);
+  private static String zoneId = System.getProperty(BIGTABLE_ZONE_KEY);
+  private static String clusterId = System.getProperty(BIGTABLE_CLUSTER_KEY);
+
+  public static TableName newTestTableName() {
+    return TableName.valueOf("test_table-" + UUID.randomUUID().toString());
```
mgarolera: its useful to specify which test created the table in the name.

bigtable-hbase-dataflow/src/test/java/com/google/cloud/bigtable/dataflow/CloudBigtableIOIntegrationTest.java
```diff
@@ -0,0 +1,145 @@
...
+            new CloudBigtableIO.Source(new CloudBigtableScanConfiguration(
+                projectId, zoneId, clusterId, tableName.getQualifierAsString(), new Scan()));
+        long estimatedSizeBytes = source.getEstimatedSizeBytes(null);
+        SampleRowKeysResponse lastSample = sampleRowKeys.get(sampleRowKeys.size() - 1);
+        Assert.assertEquals(lastSample.getOffsetBytes(), estimatedSizeBytes);
+
+        List<SourceWithKeys> bundles = source.splitIntoBundles(estimatedSizeBytes / 2 + 1, null);
+        // This will be a small table with no splits, so we return HConstants.EMPTY_START_ROW
+        // which can't be split.
+        Assert.assertEquals(sampleRowKeys.size() * 2 - 1, bundles.size());
```
mgarolera: Might be clearer if you actually check that sampleRowKeys.size()  and bundles.size() are 1. Also a check on the  estimatedSizeBytes will make things clearer.

bigtable-hbase-dataflow/src/test/java/com/google/cloud/bigtable/dataflow/CloudBigtableIOIntegrationTest.java
```diff
@@ -0,0 +1,145 @@
...
+            new CloudBigtableIO.Source(new CloudBigtableScanConfiguration(
+                projectId, zoneId, clusterId, tableName.getQualifierAsString(), new Scan()));
+        long estimatedSizeBytes = source.getEstimatedSizeBytes(null);
+        SampleRowKeysResponse lastSample = sampleRowKeys.get(sampleRowKeys.size() - 1);
+        Assert.assertEquals(lastSample.getOffsetBytes(), estimatedSizeBytes);
+
+        List<SourceWithKeys> bundles =
+            source.splitIntoBundles(sampleRowKeys.get(0).getOffsetBytes() / 2, null);
+        // The last sample includes the EMPTY_END_ROW key, which cannot be split.  
+        Assert.assertEquals(sampleRowKeys.size() * 2 - 1, bundles.size());
```
mgarolera: Check that you have the expected number of keys and bundles?

mgarolera:
LGTM

bigtable-hbase-dataflow/src/test/java/com/google/cloud/bigtable/dataflow/CloudBigtableIOIntegrationTest.java
sduskis: It's difficult to figure out the right number of bundles, since it depends on the number of splits, which isn't easy to determine.  I added a check for key count.

Issue #393, opened 2015-08-07
BigtableSession closes provided ExecutorService on #close() which is prone to deadlocks


udoprog:
I'm attempting to incorporate BigtableSession into a service which uses a shared ExecutorService for common tasks and establishes multiple bigtable backends.

Currently I am required to create an ExecutorService per BigtableSession instance because the `BigtableSession#close()` method aggressively attempts to assert that the provided ExecutorService has been shut down.

I initially ended up queuing up a task to close the `BigtableSession` on the same thread pool, causing `BigtableSession#close()` to either deadlock (or wait for a long time) on a [`ExecutorService#isTerminated()`](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java#L474) loop (see test below).

I believe the either **1)** behavior should be modified so that `BigtableSession` does not shutdown a user-provided `ExecutorService`, or **2)** it does not allow the user to provide a thread pool at all.

## Test Case

This test case always deadlocks for me.

```java
public class BigtableSessionTest {
    public static final String USER_AGENT = "test";

    public static void main(String[] args) throws Exception {
        CredentialOptions credentials = /* ... */;
        final MySession builder = /* ... */;

        final ExecutorService executorService = Executors.newFixedThreadPool(4);

        final Future<Void> future = executorService.submit(new Callable<Void>() {
            @Override
            public Void call() throws Exception {
                // attempt to close on the same thread pool...
                builder.build(executorService).close();
                return null;
            }
        });

        try {
            future.get(30, TimeUnit.SECONDS);
        } catch (TimeoutException e) {
            System.out.println("Timeout :(");
            e.printStackTrace();
        }

        final ExecutorService executorService2 = Executors.newFixedThreadPool(4);

        final BigtableSession session = builder.build(executorService2);

        System.out.println("Main closing...");
        session.close();
        System.out.println("Closed without problems!");

        System.out.println("bye, bye");
        System.exit(0);
    }

    static class MySession {
        final String project;
        final String zone;
        final String cluster;
        final CredentialOptions credentials;

        public MySession(final String project, final String zone, final String cluster, CredentialOptions credentials) {
            this.project = project;
            this.zone = zone;
            this.cluster = cluster;
            this.credentials = credentials;
        }

        public BigtableSession build(final ExecutorService executorService) throws IOException {
[...]

Issue #392, opened 2015-08-07
BigtableSession deadlocks when given a thread pool smaller than 3


udoprog:
BigtableSession currently deadlocks when given a thread pool with less than 3 available live threads.

I can always reproduce this.

## Test Case

```java
public class BigtableSessionTest {
    public static void main(String[] args) throws Exception {
        CredentialOptions credentials = /* .. */;
        final MySession session = /* .. */;

        final ExecutorService tasks = Executors.newFixedThreadPool(1);

        for (int count = 1; count <= 8; count++) {
            final ExecutorService executorService = Executors.newFixedThreadPool(count);
            final CountDownLatch latch = new CountDownLatch(1);

            final AtomicReference<Throwable> errored = new AtomicReference<Throwable>();

            // thread that builds and closes the BigtableSession...
            final Thread thread = new Thread() {
                public void run() {
                    try {
                        session.build(executorService).close();
                    } catch (Exception e) {
                        errored.set(e);
                    }

                    latch.countDown();
                };
            };

            System.out.println(String.format("%d: starting...", count));
            thread.start();

            if (!latch.await(1, TimeUnit.SECONDS)) {
                System.out.println(String.format("%d: seems to have deadlocked :(", count));
                thread.interrupt();
            }

            // wait one more time in case of deadlock...
            if (!latch.await(1, TimeUnit.SECONDS)) {
                System.out.println(String.format("%d: did not recover, exiting D:", count));
                break;
            }

            if (errored.get() != null) {
                System.out.println(String.format("%d: Task failed:", count));
                errored.get().printStackTrace(System.out);
            } else {
                System.out.println(String.format("%d: Task OK", count));
            }
        }

        System.out.println("bye, bye");
        System.exit(0);
    }

    static class MySession {
        public static final String USER_AGENT = "test";

        final String project;
        final String zone;
        final String cluster;
        final CredentialOptions credentials;

        public MySession(final String project, final String zone, final String cluster, CredentialOptions credentials) {
            this.project = project;
            this.zone = zone;
            this.cluster = cluster;
            this.credentials = credentials;
        }

        public BigtableSession build(final ExecutorService executorService) throws IOException {
            final BigtableOptions options = new BigtableOptions.Builder().setProjectId(project).setZoneId(zone)
                    .setClusterId(cluster).setUserAgent(USER_AGENT).setCredentialOptions(credentials).build();

            return new BigtableSession(options, executorService);
        }
    }
}
```

[...]

PR #396, opened 2015-08-10
Addressing issues related to BigtableSession startup and shutdown

sduskis:
Fixing issue #392 by removing the parallelism in creating individual ChannelImpls.  That didn't add much startup performance anyway.
Fixing issue #393 by adding a flag that checks if the batchPool is user supplied, and if so, don't close it when the session is closed.


bigtable-grpc-interface/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -130,6 +130,7 @@
 //    return sslBuilder;
 //  }
 
+  @SuppressWarnings("deprecation")
```
kevinsi4508: Can you document more on why this is deprecated, and what is the alternative?

kevinsi4508:
LGTM

PR #397, opened 2015-08-10
Massive renaming in bigtable-grpc-interface

sduskis:
Renaming bigtable-grpc-interface to bigtable-client-core
Moving classes in bigtable-client-core into the following packages
 - client
 - config
 - grpc
 - scanner


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableTableName.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.google.cloud.bigtable.naming;
+package com.google.cloud.bigtable.grpc;
```
kevinsi4508: I don't think BigtableTableName.java has anything to do with grpc.

kevinsi4508:
LGTM

PR #404, opened 2015-08-14
Update the number of rows need to be returned on rescan

kevinsi4508:
Update the number of rows need to be returned by removing the number of rows that are already read from the original page limit.


kevinsi4508:
ping

sduskis:
We need to discuss this in person.  I don't think I can do a review of this CL without that.

kevinsi4508:
Sure, we can discuss this in person. BTW, this PR is a fix for a bug discovered when working on WhileMatchFilter.

kevinsi4508:
As discussed offline. I updated the code to keep track of row count and not resetting it. Plus we check if rescan would result in the number rows read over the page limit.
PTAL

sduskis:
LGTM (after squashing commits)

PR #405, opened 2015-08-17
Initial version of WhileMatchFilter adapter.

kevinsi4508:
- Create WhileMatchFilter adapter
- This adapter is not yet used as the integration work is not done


sduskis:
LGTM

PR #410, opened 2015-08-19
Using GrpcSslContexts.

sduskis:
The gRPC team suggested this solution, and it is confirmed to work with Dataflow.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -50,11 +50,13 @@ import io.grpc.ClientInterceptor;
 import io.grpc.ClientInterceptors;
 import io.grpc.MethodDescriptor;
 import io.grpc.auth.ClientAuthInterceptor;
+import io.grpc.transport.netty.GrpcSslContexts;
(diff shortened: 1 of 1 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -122,26 +124,14 @@
...
-//    }
-//    return sslBuilder;
-//  }
-
-  // The deprecation is caused by SslContext.newClientContext(). We won't need the @SuppressWarnings
-  // once we get GrpcSslContexts to work properly.
-  @SuppressWarnings("deprecation")
+  private static SslContextBuilder createGrpcSslBuilder() {
+    SslContextBuilder sslBuilder = GrpcSslContexts.forClient();
+    sslBuilder.ciphers(null);
```
kevinsi4508: I thought setting ciphers to null is for JDK 1.7.*. Is it still true?

sduskis: @nmittler suggested that it should always be set to null after an issue was found with the Dataflow environment.

kevinsi4508:
LGTM

PR #415, opened 2015-08-19
Enforce only one WhileMatchFilter can be used.

kevinsi4508:
Enforce only one WhileMatchFilter can be used.


bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/WhileMatchFilterAdapter.java
```diff
@@ -75,6 +75,9 @@ public WhileMatchFilterAdapter(FilterAdapter subFilterAdapter) {
    */
   @Override
   public RowFilter adapt(FilterAdapterContext context, WhileMatchFilter filter) throws IOException {
+    checkArgument(
+        context.getNumberOfWhileMatchFilters() == 0,
```
mgarolera: Maybe make it a get/set and check that get returns null? its kind of weird to allow a list but enforce a maximum of 1.

kevinsi4508: The intention is keeping track of all WMFs been added. I would say checking size for enforcing size limit makes more sense in this case.

mgarolera:
LGTM

PR #420, opened 2015-08-20
Upgrading to grpc-java 0.8.0

sduskis:
The proto generation process changed.  See the README.md for the process, but the main difference is that the proto generation process now uses maven.  I had some problems generating both the service java files and the data proto java files at the same time, so now it's a two step prcess requiring different directories.
There were some API differences in grpc-java that had to be addressed as well.


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -156,24 +153,6 @@ public void run() {
     connectionStartupExecutor.execute(new Runnable() {
       @Override
       public void run() {
-        // The first invocation of BigtableServiceGrpc.CONFIG is expensive.
```
kevinsi4508: Don't we need the first call to data and table service? Is this change part of the PR?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/CallCompletionStatusInterceptor.java
```diff
@@ -149,16 +152,16 @@ public void start(Listener<ResponseT> responseListener, Metadata.Headers headers
   }
 
   @Override
-  public <ReqT, RespT> CompletionStatusGatheringCall<ReqT, RespT> interceptCall(
-      MethodDescriptor<ReqT, RespT> methodDescriptor, Channel channel) {
-    return wrapCall(methodDescriptor, channel.newCall(methodDescriptor));
+  public <RequestT, ResponseT> CompletionStatusGatheringCall<RequestT, ResponseT> interceptCall(
+      MethodDescriptor<RequestT, ResponseT> method, CallOptions callOptions, Channel next) {
```
kevinsi4508: why "Channel next" instead of "Channel channel"?

bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -251,9 +251,13 @@ public HTableDescriptor getTableDescriptor(TableName tableName)
 
     try {
       return tableAdapter.adapt(bigtableTableAdminClient.getTable(request));
+    } catch (StatusRuntimeException e) {
+      if (e.getStatus().getCode() == Status.NOT_FOUND.getCode()) {
```
kevinsi4508: cannot just compare enum? e.getStatus() == Status.NOT_FOUND?

bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/TestClusterAPI.java
```diff
@@ -134,19 +133,15 @@ private void countTables(Admin admin, int expectedCount) throws IOException {
...
     try {
       return client.getCluster(request);
-    } catch (UncheckedExecutionException e) {
-      if (e.getCause() != null && e.getCause() instanceof StatusRuntimeException) {
-        Status status = ((StatusRuntimeException) e.getCause()).getStatus();
-        if (status.getCode() == Status.NOT_FOUND.getCode()) {
-          return null;
-        }
+    } catch (StatusRuntimeException e) {
+      Status status = e.getStatus();
```
kevinsi4508: if (e.getStatus() == Status.NOT_FOUND) {
  return null;
}

sduskis:
I've been testing this with JDK8.  JDK7 breaks with the new changes.

bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
sduskis: Status is a class, not an enum: https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/Status.java#L62

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/CallCompletionStatusInterceptor.java
sduskis: That was the name that the gRPC team chose as the name, and I used eclipse to auto-generate the new signature.

https://github.com/grpc/grpc-java/blob/572f4332a75367e0d0e1cb096198839b68787670/core/src/main/java/io/grpc/ClientInterceptor.java#L61

bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
sduskis: The are not actually ==.  A status can have a whole bunch of stuff in it that will make it not ==...

see https://github.com/grpc/grpc-java/blob/572f4332a75367e0d0e1cb096198839b68787670/core/src/main/java/io/grpc/Status.java#L353

kevinsi4508: fine with checking code too, thought not found status should be the static instance: public static final Status NOT_FOUND.

kevinsi4508:
LGTM

bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
sduskis: I thought so too.  Tests showed that I was wrong.

sduskis:
The problem with JDK7 was related to the version of ALPN I was using.  I'm going to merge this.

PR #422, opened 2015-08-21
Refactorying scan methods in BigtableDataGrpcClient.java

sduskis:


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -56,6 +56,16 @@
  */
 public class BigtableDataGrpcClient implements BigtableDataClient {
 
+  private class ImmutableListFunction<T> implements Function<List<T>, ImmutableList<T>> {
```
kevinsi4508: why do we need this? ImmutableList.copy should do, no?

kevinsi4508: This is generic usage, maybe name it that?
ImmutableListCopier? ImmutableListCopierFunction?

List<T> items instead of List<T> sampleRowKeysResponses

kevinsi4508: private static class?

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableTable.java
```diff
@@ -97,6 +99,22 @@
       new BigtableResultScannerAdapter(ROW_ADAPTER);
   public static final GetAdapter GET_ADAPTER = new GetAdapter(SCAN_ADAPTER);
 
+  // ReadHooks don't make sense from conditional mutations. If any filter attempts to make use of
+  // them (which they shouldn't since we built the filter), throw an exception.
+  private static final ReadHooks THROW_IF_USED_READ_HOOKS = new ReadHooks() {
```
kevinsi4508: nit: UNSUPPORTED_READ_HOOKS?

kevinsi4508:
LGTM

PR #423, opened 2015-08-21
Add BigtableWhileMatchResultScannerAdapter

kevinsi4508:
Add BigtableWhileMatchResultScannerAdapter to be used when a WhileMatchFilter is used.
We only support single WhileMatchFilter.


sduskis:
LGTM after squash

PR #427, opened 2015-08-24
Check if a WhileMatchFilter is used.

kevinsi4508:
- Adapt WhileMatchFilter
- Add integration tests for WhileMatchFilter
- Check if WhileMatchFilter is used before using scan adapter


sduskis:
LGTM.  Please merge #426 first.

PR #428, opened 2015-08-24
Perform retry on failure in readRowsAsync().

kevinsi4508:


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -291,6 +302,65 @@ public void run() {
...
+    private ListenableFuture<List<Row>> backOffAndRetry(Throwable cause) throws IOException,
+        ScanRetriesExhaustedException {
+      long nextBackOff = currentBackoff.nextBackOffMillis();
+      if (nextBackOff == BackOff.STOP) {
+        LOG.warn("RetriesExhausted: ", cause);
+        throw new ScanRetriesExhaustedException("Exhausted streaming retries.", cause);
+      }
+
+      sleep(nextBackOff);
+      return BigtableDataGrpcClient.this.doReadRowsAsync(request);
```
sduskis: We should probably do this for sample keys as well, but at some future date.

sduskis:
LGTM

PR #430, opened 2015-08-26
Refactor retry read async and update sampleRawKeysAysnc to retry on exception

kevinsi4508:
Refactor retry read async so that it can be used for both read rows async and sample keys aysnc call.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataClient.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataClient.java
@@ -80,8 +80,7 @@ public interface BigtableDataClient {
    * Sample row keys from a table, returning a Future that will complete when the sampling has
    * completed.
    */
-  ListenableFuture<ImmutableList<SampleRowKeysResponse>> sampleRowKeysAsync(
-      SampleRowKeysRequest request);
+  ListenableFuture<List<SampleRowKeysResponse>> sampleRowKeysAsync(SampleRowKeysRequest request);
 
   /**
    * Perform a scan over rows.
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
@@ -68,9 +68,9 @@ import java.util.concurrent.ExecutorService;
  */
 public class BigtableDataGrpcClient implements BigtableDataClient {
 
-  private static class ImmutableListFunction<T> implements Function<List<T>, ImmutableList<T>> {
+  private static class ImmutableListFunction<T> implements Function<List<T>, List<T>> {
     @Override
-    public ImmutableList<T> apply(List<T> list) {
+    public List<T> apply(List<T> list) {
       return ImmutableList.copyOf(list);
     }
   }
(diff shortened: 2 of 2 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataClient.java
```diff
@@ -80,8 +80,7 @@ CheckAndMutateRowResponse checkAndMutateRow(CheckAndMutateRowRequest request)
    * Sample row keys from a table, returning a Future that will complete when the sampling has
    * completed.
    */
-  ListenableFuture<ImmutableList<SampleRowKeysResponse>> sampleRowKeysAsync(
-      SampleRowKeysRequest request);
+  ListenableFuture<List<SampleRowKeysResponse>> sampleRowKeysAsync(SampleRowKeysRequest request);
```
sduskis: Why are you doing this change?

kevinsi4508: The response is set as List<RESPONSE> in readAsync().
ListenableFuture<List<RESPONSE>> readyAsync(REQUEST request);

Usually we don't have to expose ImmutableList even though the list is immutable.

sduskis: ok

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -247,19 +247,31 @@ public Row readModifyWriteRow(ReadModifyWriteRowRequest request) {
...
-    CollectingStreamObserver<SampleRowKeysResponse> responseBuffer =
-        new CollectingStreamObserver<>(call);
-    ClientCalls.asyncServerStreamingCall(
-        call,
-        request,
-        responseBuffer);
-    return Futures.transform(
-        responseBuffer.getResponseCompleteFuture(),
-        IMMUTABLE_LIST_FUNCTION);
+    ReadAsync<SampleRowKeysRequest, SampleRowKeysResponse> readAsync =
```
sduskis: Can we make this a member variable?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -247,19 +247,31 @@ public Row readModifyWriteRow(ReadModifyWriteRowRequest request) {
...
+        ClientCalls.asyncServerStreamingCall(
+            call,
+            request,
+            responseBuffer);
+        return Futures.transform(
+            responseBuffer.getResponseCompleteFuture(),
+            IMMUTABLE_LIST_FUNCTION);
+      }};
+
+    if (retryOptions.enableRetries()) {
```
sduskis: Can the 

if (enableretries) {
  ..
 Futures.withFallback(readAsyc.. ...);
} else {
  return readAsyc.readAsyc...
}

be a method that both readSampleRowKeys and readRows use?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
kevinsi4508: done

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
kevinsi4508: done.

sduskis:
LGTM

PR #431, opened 2015-08-26
Making RowMerger a top level object.

sduskis:
RowMerger seems to be useful outside of StreamingBigtableResultScanner.  StreamingBigtableResultScanner may be going away in favor of a different solution.  Pulling out RowMerger into a separate file to prepare for that eventuality.


kevinsi4508:
LGTM, please add tests in a separate PR.

PR #433, opened 2015-08-26
Removing ugly stacktrace from BigtableSession startup.

sduskis:
If ALPN is not configured, any class that references BigtableSession will currently cause a big stackTrace that's scary, but not all that useful.  The stack trace occurs when ALPN is not configured and the BigtableSession async sslContext initialization logic is invoked.  Remove the async stack trace; there will be a synchronous stack trace for user code if they actually care about creating a BigtableSession and ALPN is not configured.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -141,12 +141,18 @@ public class BigtableSession implements AutoCloseable {
         // Create a throw away object in order to speed up the creation of the first
         // BigtableConnection which uses SslContexts under the covers.
         try {
+          Class.forName("org.eclipse.jetty.alpn.ALPN");
+        } catch (ClassNotFoundException e1) {
+          LOG.warn(
+            "Could not asynchronously create the ssl context, since ALPN is not installed.");
+          return;
+        }
+        try {
           // We create multiple channels via refreshing and pooling channel implementation.
           // Each one needs its own SslContext.
-          @SuppressWarnings("unused")
-          SslContext warmup = createSslContext();
+          createSslContext();
         } catch (SSLException e) {
-          throw new IllegalStateException("Could not create an ssl context.", e);
+          LOG.warn("Could not asynchronously create the ssl context", e);
         }
       }
     });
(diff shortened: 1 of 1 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -141,12 +141,19 @@ public void run() {
...
+          return;
+        }
+        try {
           // We create multiple channels via refreshing and pooling channel implementation.
           // Each one needs its own SslContext.
           @SuppressWarnings("unused")
           SslContext warmup = createSslContext();
         } catch (SSLException e) {
-          throw new IllegalStateException("Could not create an ssl context.", e);
+          LOG.warn("Could not asynchronously the ssl context", e);
```
kevinsi4508: Could not asynchronously create the ssl context

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -141,12 +141,19 @@ public void run() {
         // Create a throw away object in order to speed up the creation of the first
         // BigtableConnection which uses SslContexts under the covers.
         try {
+          Class.forName("org.eclipse.jetty.alpn.ALPN");
+        } catch (ClassNotFoundException e1) {
+          LOG.warn(
+            "Could not asynchronously the ssl context, since ALPN is not installed.");
```
kevinsi4508: Could not asynchronously create the ssl context

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -141,12 +141,19 @@ public void run() {
...
+          return;
+        }
+        try {
           // We create multiple channels via refreshing and pooling channel implementation.
           // Each one needs its own SslContext.
           @SuppressWarnings("unused")
           SslContext warmup = createSslContext();
```
kevinsi4508: "createSslContext();" instead of "SslContext warmup = createSslContext();" ?

kevinsi4508:
nits. LGTM

sduskis:
Fixed the nits.  Merging.

PR #434, opened 2015-08-26
Add random suffix in the rowkey prefix in WhileMatchFilter tests.

kevinsi4508:
Add random suffix in the rowkey prefix in WhileMatchFilter tests.


sduskis:
LGTM

PR #436, opened 2015-08-27
Reverting back to gRPC version 0.7.1.

sduskis:
We're having a problem with 0.8.0 which seems to not be closing the batchPool executor in AbstractBigtableConnection properly.


bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/TestClusterAPI.java
```diff
@@ -137,6 +138,15 @@ private Cluster getCluster(BigtableClusterAdminClient client, String clusterName
     GetClusterRequest request = GetClusterRequest.newBuilder().setName(clusterName).build();
     try {
       return client.getCluster(request);
+    } catch (UncheckedExecutionException e) {
+      if (e.getCause() != null && e.getCause() instanceof StatusRuntimeException) {
```
kevinsi4508: I think u can do without "e.getCause() != null". If the cause is null, instanceof will be false.

bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/TestClusterAPI.java
```diff
@@ -137,6 +138,15 @@ private Cluster getCluster(BigtableClusterAdminClient client, String clusterName
...
+        Status status = ((StatusRuntimeException) e.getCause()).getStatus();
+        if (status.getCode() == Status.NOT_FOUND.getCode()) {
+          return null;
+        }
+      }
+      e.printStackTrace();
```
kevinsi4508: remove?

kevinsi4508:
LGTM

bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/TestClusterAPI.java
sduskis: I'm on with this code in the cluster api test.

PR #438, opened 2015-08-27
Dedup cells with same timestamp in RowAdapter.

kevinsi4508:
Dedup cells with same timestamp in RowAdapter.

### DIFF
--- a/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestFilters.java
+++ b/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestFilters.java
@@ -1001,7 +1001,7 @@ public class TestFilters extends AbstractTest {
   public void testWhileMatchFilter_simple() throws IOException {
     String rowKeyPrefix = dataHelper.randomString("wmf-simple-");
     byte[] qualA = dataHelper.randomData("qualA");
-    Table table = addDataForWhileMatchFilterTest(rowKeyPrefix, qualA);
+    Table table = addDataForTesting(rowKeyPrefix, qualA);
 
     ByteArrayComparable rowValue2Comparable = new BinaryComparator(Bytes.toBytes("12"));
     ValueFilter valueFilter =
@@ -1018,7 +1018,7 @@ public class TestFilters extends AbstractTest {
   public void testWhileMatchFilter_singleChained() throws IOException {    
     String rowKeyPrefix = dataHelper.randomString("wmf-sc-");
     byte[] qualA = dataHelper.randomData("qualA");
-    Table table = addDataForWhileMatchFilterTest(rowKeyPrefix, qualA);
+    Table table = addDataForTesting(rowKeyPrefix, qualA);
 
     ByteArrayComparable valueComparable = new BinaryComparator(String.valueOf(2).getBytes());
     SingleColumnValueFilter valueFilter = new SingleColumnValueFilter(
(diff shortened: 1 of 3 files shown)


bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/RowAdapter.java
```diff
@@ -54,17 +59,20 @@ public Result adaptResponse(Row response) {
...
-          KeyValue keyValue = new KeyValue(
-              rowKey,
-              familyNameBytes,
-              columnQualifier,
-              hbaseTimestamp,
-              cell.getValue().toByteArray());
-
-          hbaseCells.add(keyValue);
+          Long cellTimestamp = cell.getTimestampMicros();
+          if (!cellTimestamps.contains(cellTimestamp)) {
```
sduskis: What if there are different family/columns with the same cell timestamp?  I think this would only return one of those cells, when we should be returning 2.  We should also add a test for that condition for different family/column, same family/different column, different family/same column.

kevinsi4508: The set is cleared for each column.
updated test.

sduskis: Are we guaranteed that the columns will be ordered correctly?

kevinsi4508: There won't be duplicate columns, confirmed with server side code.

bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestFilters.java
```diff
@@ -1548,6 +1536,29 @@ public void testPageFilters() throws IOException {
...
+    FilterList filterList = new FilterList(
+        Operator.MUST_PASS_ONE,
+        new QualifierFilter(CompareOp.EQUAL, new BinaryComparator(qualA)),
+        prefixFilter);
+    Scan scan = new Scan(Bytes.toBytes(rowKeyPrefix));
+    scan.setFilter(filterList);
+
+    try (ResultScanner scanner = table.getScanner(scan)) {
+      Iterator<Result> iterator = scanner.iterator();
+      while (iterator.hasNext()) {
```
sduskis: why not?
for(Result result : scanner)

bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestFilters.java
```diff
@@ -1555,4 +1566,16 @@ public void testPageFilters() throws IOException {
...
   }
+
+  private Table addDataForTesting(String rowKeyPrefix, byte[] qualA)
+      throws IOException {
+    Table table = getConnection().getTable(TABLE_NAME);
+    for (int i = 0; i < 100; i++) {
+      String indexStr = String.valueOf(i);
+      byte[] rowKey = Bytes.toBytes(rowKeyPrefix + indexStr);
+      Put put = new Put(rowKey).addColumn(COLUMN_FAMILY, qualA, Bytes.toBytes(indexStr));
+      table.put(put);
```
sduskis: Please do a table.put(List<Put>) instead.

bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestFilters.java
kevinsi4508: done

bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestFilters.java
kevinsi4508: done.

sduskis:
LGTM

PR #439, opened 2015-08-28
Changing the order in which we close things.

sduskis:
Channels should be closed before we close ExecutorServices that the channels use.


sduskis:
This is a precursor to upgrading to gRPC 0.8.0.  I confirmed that this sequence works with 0.8.0

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -426,18 +429,28 @@ public void close() throws IOException {
...
           return null;
         }
-      });
+      }));
     }
+    try {
+      Futures.allAsList(closingChannelsFutures).get();
+    } catch (InterruptedException e) {
+      Thread.interrupted();
+      throw new IOException("Interrupted while waiting for close", e);
```
kevinsi4508: nit: Interrupted while waiting for channels to be closed

kevinsi4508:
LGTM

PR #440, opened 2015-08-28
Using grpc 0.8.0 again.

sduskis:
#439 fixes the problems we had with closing connections in gRPC 0.8.0.  Using 0.8.0 again.


kevinsi4508:
LGTM

PR #441, opened 2015-08-30
Update testInterleaveNoDuplicateCells() to use two ColumnPrefixFilters.

kevinsi4508:
Update testInterleaveNoDuplicateCells() to use two ColumnPrefixFilters instead. QualifierFilter includes rows with empty qualifier therefore not a good choice for this test.


sduskis:
LGTM

PR #444, opened 2015-08-31
Throwing an exception if a Channel doesn't close.

sduskis:


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -414,13 +414,14 @@ public Closeable createClosable(final Channel channel) {
...
             channelImpl.shutdown();
-            while (!channelImpl.isTerminated()) {
-              try {
-                channelImpl.awaitTermination(CHANNEL_TERMINATE_WAIT_MS, TimeUnit.MILLISECONDS);
-              } catch (InterruptedException e) {
-                Thread.interrupted();
-                throw new IOException("Interrupted while sleeping for close", e);
-              }
+            try {
+              channelImpl.awaitTermination(CHANNEL_TERMINATE_WAIT_MS, TimeUnit.MILLISECONDS);
```
kevinsi4508: We don't want to retry if it times out?

sduskis: We need to stop at some point.  10 seconds seems fair.  We shouldn't continue indefinitely.  Do you have a better suggestion?

kevinsi4508: How about configurable n tries?

sduskis: n tries isn't all that helpful.  awaitTermination waits until the channel is terminated.  Any configuration ought to be in seconds rather than tries.  Tries really isn't all that meaningful in this case.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -414,13 +414,14 @@ public Closeable createClosable(final Channel channel) {
...
+            } catch (InterruptedException e) {
+              Thread.interrupted();
+              throw new IOException("Interrupted while sleeping for close", e);
+            }
+            if (!channelImpl.isTerminated()) {
+              throw new IOException("Could not close the channel.  ");
```
kevinsi4508: nit: Could not close the channel after x seconds.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
kevinsi4508: Sounds good to me.

kevinsi4508:
LGTM

Issue #447, opened 2015-09-01
SingleValueColumnFilter unit testing


reagere:
Hello,

I've made a JUnit test of SingleValueColumnFilter of Long values in a Scan using Bigtable.

Only EQUAL works, for LESS I have 10 values, for GREATER I have 0 values.

I am using the HBase 1.1.1 and Bigtable 0.2.1 with maven.

Thank you for the help.

```java
package com.accengage.globulus.test.segment;

import java.io.IOException;
import java.util.HashMap;
import java.util.HashSet;
import java.util.Map;
import java.util.Set;
import java.util.UUID;

import org.apache.hadoop.hbase.HColumnDescriptor;
import org.apache.hadoop.hbase.HTableDescriptor;
import org.apache.hadoop.hbase.TableName;
import org.apache.hadoop.hbase.client.Admin;
import org.apache.hadoop.hbase.client.Connection;
import org.apache.hadoop.hbase.client.ConnectionFactory;
import org.apache.hadoop.hbase.client.Put;
import org.apache.hadoop.hbase.client.Result;
import org.apache.hadoop.hbase.client.ResultScanner;
import org.apache.hadoop.hbase.client.Scan;
import org.apache.hadoop.hbase.client.Table;
import org.apache.hadoop.hbase.filter.CompareFilter.CompareOp;
import org.apache.hadoop.hbase.filter.SingleColumnValueFilter;
import org.apache.hadoop.hbase.util.Bytes;
import org.junit.AfterClass;
import org.junit.Assert;
import org.junit.BeforeClass;
import org.junit.Test;

import io.grpc.StatusRuntimeException;

public class SupInfEqDiffHbaseTest {

	private static Connection connection;
	private static TableName tableName = TableName.valueOf("testBBO");
	private static Table table;
	private static Admin admin;
	private static int nValue = 10;
	private static Map<String, Long> result = new HashMap<>();

	@AfterClass
	public static void clean() throws IOException {
		SupInfEqDiffHbaseTest.deleteTestTable();
		Assert.assertFalse(SupInfEqDiffHbaseTest.admin.tableExists(SupInfEqDiffHbaseTest.tableName));
		SupInfEqDiffHbaseTest.admin.close();
		SupInfEqDiffHbaseTest.connection.close();
		connection.close();
	}

	private static void deleteTestTable() throws IOException {
[...]

PR #454, opened 2015-09-02
Removing regular expression escaping from ValueFilterAdapter

sduskis:
This fixes Issue #447, which shows incorrect behavior for SingleValueColumnFilter.  ValueFilterAdapter is used by SingleValueColumnFilterAdapter, so this fixes behavior for both ValueFilter and SingleValueColumnFilter.
Added a unit test for the scenarios in #447.

### DIFF
--- a/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestSingleColumnValueFilter.java
+++ b/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestSingleColumnValueFilter.java
@@ -1,0 +1,185 @@
+/*
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.bigtable.hbase;
+
+import static com.google.cloud.bigtable.hbase.IntegrationTests.COLUMN_FAMILY;
+import static com.google.cloud.bigtable.hbase.IntegrationTests.TABLE_NAME;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+
+import org.apache.hadoop.hbase.client.Put;
+import org.apache.hadoop.hbase.client.Result;
+import org.apache.hadoop.hbase.client.ResultScanner;
+import org.apache.hadoop.hbase.client.Scan;
+import org.apache.hadoop.hbase.client.Table;
+import org.apache.hadoop.hbase.filter.CompareFilter.CompareOp;
+import org.apache.hadoop.hbase.filter.SingleColumnValueFilter;
+import org.apache.hadoop.hbase.util.Bytes;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class TestSingleColumnValueFilter extends AbstractTest {
+
+  static final int count = 10;
+  final byte[] QUALIFIER = dataHelper.randomData("TestSingleColumnValueFilter");
+
+  private Map<String, Long> result = new HashMap<>();
+  private Table table;
+  private boolean added = false;
+
+  @Before
+  public void fillTable() throws IOException {
+    table = getConnection().getTable(TABLE_NAME);
+    if (!added) {
+      result.clear();
+      List<Put> puts = new ArrayList<>();
+      for (long i = 0; i < count; i++) {
+        final UUID rowKey = UUID.randomUUID();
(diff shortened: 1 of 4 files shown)


bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestSingleColumnValueFilter.java
```diff
@@ -0,0 +1,234 @@
...
+import com.google.common.collect.Iterators;
+
+public class TestSingleColumnValueFilter extends AbstractTest {
+
+  byte[] qual = dataHelper.randomData("testSingleValueFilterCompOps");
+  static final int count = 10;
+  static final byte[] COLUMN_FAMILY = Bytes.toBytes("cd");
+  static final byte[] QUALIFIER = Bytes.toBytes("value");
+
+  TableName tableName = IntegrationTests.newTestTableName();
```
kevinsi4508: Is this a shared table? Can we use a new table?

bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestSingleColumnValueFilter.java
```diff
@@ -0,0 +1,234 @@
...
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+import com.google.common.collect.Iterators;
+
+public class TestSingleColumnValueFilter extends AbstractTest {
+
+  byte[] qual = dataHelper.randomData("testSingleValueFilterCompOps");
```
kevinsi4508: private static final if possible.

bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestSingleColumnValueFilter.java
```diff
@@ -0,0 +1,234 @@
...
+    }
+    table.close();
+    table = null;
+    result.clear();
+  }
+
+  @Test
+  public void testSingleValueFilterAscii() throws IOException {
+    // Add {, }, and @ to make sure that they do not need to be encoded
+    byte[] rowKeyA = dataHelper.randomData("KeyA{");
```
kevinsi4508: Shouldn't adding data be part of test setup @Before?

bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestSingleColumnValueFilter.java
```diff
@@ -0,0 +1,234 @@
...
+    byte[] rowKeyB = dataHelper.randomData("KeyB}");
+    byte[] rowKeyC = dataHelper.randomData("KeyC@");
+    byte[] qualValueA = dataHelper.randomData("ValueA{");
+    byte[] qualValueB = dataHelper.randomData("ValueB}");
+    byte[] qualValueC = dataHelper.randomData("ValueC@");
+    try (Table table = getConnection().getTable(TABLE_NAME)) {
+      table.put(new Put(rowKeyA).addColumn(COLUMN_FAMILY, qual, qualValueA));
+      table.put(new Put(rowKeyB).addColumn(COLUMN_FAMILY, qual, qualValueC));
+      table.put(new Put(rowKeyC).addColumn(COLUMN_FAMILY, qual, qualValueC));
+      assertSizeOfSCVF(table, 1, CompareOp.EQUAL, qualValueA);
```
kevinsi4508: Add comments and explain why we expect 1 (same for the rest of assertions).

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/ValueFilterAdapter.java
```diff
@@ -65,27 +63,23 @@ public FilterSupportStatus isFilterSupported(
 
   private RowFilter adaptBinaryComparator(
       CompareOp compareOp, BinaryComparator comparator) throws IOException {
-    byte[] comparatorValue = comparator.getValue();
-    ByteArrayOutputStream baos =
-        new ByteArrayOutputStream(comparatorValue.length * 2);
-    ReaderExpressionHelper.writeQuotedRegularExpression(baos, comparatorValue);
```
kevinsi4508: writeQuotedRegularExpression() is used to escape special characters. I wonder why this is not needed now. Notice that writeQuotedRegularExpression() is used in other places too. If this is a problem, we need to remove it all together.

kevinsi4508:
We do want to escape the characters for equality comparison.

bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestSingleColumnValueFilter.java
sduskis: I'll move this to TestFilters.  This is a stand alone test condition that doesn't interact with the other @Test methods.

kevinsi4508: I think it is better to have setup data in @Before, then have test cases:
1. testEqual
2. testGreater
3. testLess
...

kevinsi4508: Up to you, it is a nit.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/ValueFilterAdapter.java
sduskis: moved regex escaping to the EQUALS case only

kevinsi4508:
LGTM

PR #455, opened 2015-09-02
Set max message size limit to 256 MB to match the server limit.

kevinsi4508:
Set max message size limit to 256 MB to match the server limit instead of using default 100 MB.


sduskis:
LGTM

PR #462, opened 2015-09-03
Introducing BigtableAsyncExecutor

sduskis:
BigtableAsyncExecutor is a grpc version of BigtableBufferedMutator.  It should be usable by people who want to use BigtableSession without the HBase API.  Moving BigtableBufferedMutator and BatchExecutor to use it.

If this is too big of a commit, it could probably be broken up.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BigtableAsyncExecutor.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BigtableAsyncExecutor.java
@@ -1,0 +1,286 @@
+/*
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BigtableAsyncExecutor.java
```diff
@@ -0,0 +1,286 @@
...
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import com.google.protobuf.Empty;
+import com.google.protobuf.GeneratedMessage;
+
+/**
+ * Performs asynchronous mutations and reads w
+ */
+// TODO: Cleanup the interface so that @VisibleForTesting can be reduced.
+public class BigtableAsyncExecutor implements Closeable {
```
kevinsi4508: Any way u can do a git mv or something, so that we can compare this with the original file where the logic is copy from?

kevinsi4508:
Can you list the changes and see if we can break the PR?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BigtableAsyncExecutor.java
sduskis: Let me see if git cp works.

sduskis:
I'm working through a different approach that would break this down to more steps.

PR #473, opened 2015-09-08
Adding a test for RetryingRpcFutureFallback.

sduskis:


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/RetryingRpcFutureFallback.java
```diff
@@ -37,6 +37,12 @@
  */
 public class RetryingRpcFutureFallback<RequestT, ResponseT> implements FutureFallback<ResponseT> {
 
+  public static <RequestT, ResponseT> RetryingRpcFutureFallback<RequestT, ResponseT> create(
```
kevinsi4508: You want to make the constructor private since you want to use static factory method?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/RetryOptions.java
```diff
@@ -125,6 +129,14 @@ public Builder setReadPartialRowTimeoutMillis(int timeout) {
     }
 
     /**
+     * Set a {@link NanoClock} for testing purposes.
+     */
```
kevinsi4508: @VisibleForTesting?

bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/async/RetryingRpcFutureFallbackTest.java
```diff
@@ -0,0 +1,103 @@
...
+  private RetryingRpcFutureFallback underTest;
+
+  @Mock
+  private RetryableRpc readAsync;
+
+  @Rule
+  public ExpectedException expectedException = ExpectedException.none();
+
+  @Mock
+  private NanoClock nanoClock;
```
kevinsi4508: Adding this NanoClock for the sake of this test seems overkill.
Can't we use MaxElaspedBackoffMillis to test backoff logic?

sduskis: I tried something, but the test wasn't at all clear in terms of intent.  I tried something a bit different.  How do you like RetryOptions.createBackoff()?

kevinsi4508: If there is no other way, then this is fine.
I was thinking of setting MaxElaspedBackoffMillis to short period of time and then let retry expires quickly.

sduskis: I see.  You're suggesting not mimicking the process, but rather making the process short.

There's always another way.  The NanoClock mechanism seems to be the Backoff framwork's approach to testing.  I was trying to be idiomatic more than anything else.  Is there a reason why you don't like the Backoff framework's approach?

kevinsi4508: The only reason I don't like using NanoClock is introducing something we don't use in prod running code. If we do have a reason to use NanoClock other than for testing purpose then it is fine.

Adding NanoClock at test time is better, but we may not be testing the same code running in production though.

What is the purpose of this test? If we only want to test the backoff logic, we can do following:

1. test backoff and finally succeeded
2. test backoff multiple times and failed at the end

I am OK with what we have for now.

kevinsi4508:
LGTM

PR #479, opened 2015-09-10
Allow logging for RetriesExhaustedWithDetailsException

derjust:
providing a static method to log RetriesExhaustedWithDetailsException -
so user code which hits this exception (like context.output()) or
`finishBundle` itself logs it - unwrapping the causes


derjust:
Logical got introduced via https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/481 - no need for this dedicated PR anymore

Issue #484, opened 2015-09-16
Setting DataHost is not possible in DataFlow


derjust:
Setting the `DataHost` is not possible in DataFlow:
`CloudBigtableConfiguration` and derived class do not provide such a field.
The methods 
`com.google.cloud.bigtable.dataflow.CloudBigtableConfiguration.toBigtableOptions()` and `com.google.cloud.bigtable.dataflow.CloudBigtableConfiguration.toHBaseConfig()` do not set the required property. 
Even adding it there seems to be ignored.

Only changing `com.google.cloud.bigtable.config.BigtableOptions.BIGTABLE_DATA_HOST_DEFAULT` has the desired effect - which requires a binary release of a custom build of this library just to use a different data host (i.e. for testing or load testing)

sduskis:
What's the context that you're trying to set the host in?  Is it a read or write?

derjust:
Both :-)
We discovered it while testing a pure PUT/Ingestion workflow.
But a Scan/PUT workflow is around the corner, too.

sduskis:
In most cases, overriding data host shouldn't be required, so I think that we need to use a custom solution here.

You probably need to create a subclass of CloudBigtableConfiguration, or more likley CloudBigtableScanConfiguration, and override toHBaseConfig().

Instead of calling writeToTable(), we'll try this:
 new CloudBigtableWriteTransform<>(new CloudBigtableMultiTableWriteFn([YourConfigObject]));

I'll add a new writeTo() method that will allow you to do that, but it's also something you can do yourself.

You should also pass YourConfigObject which subclasses CloudBigtableScanConfiguration into the read(CloudBigtableScanConfiguration config) method.

sduskis:
Closing this.  The solution in #487 was deemed to be a better alternative than my suggestion here.

PR #487, opened 2015-09-16
Allowing Dataflow users to provide additional connection configuration.

sduskis:
This should address the configuration requirements of issue #484 to set the datahost.

### DIFF
--- a/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
+++ b/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
@@ -15,10 +15,15 @@
  */
 package com.google.cloud.bigtable.dataflow;
 
+import java.io.IOException;
 import java.io.Serializable;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Map.Entry;
 
 import org.apache.hadoop.conf.Configuration;
 
+import com.google.api.client.repackaged.com.google.common.base.Preconditions;
 import com.google.cloud.bigtable.config.BigtableOptions;
 import com.google.cloud.bigtable.hbase.BigtableOptionsFactory;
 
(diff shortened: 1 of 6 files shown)


bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
```diff
@@ -94,10 +105,12 @@ public CloudBigtableConfiguration build() {
    * @param zoneId The zone where the cluster is located.
    * @param clusterId The cluster ID for the cluster.
```
kevinsi4508: document new parameter

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
```diff
@@ -146,6 +154,9 @@ public Configuration toHBaseConfig() {
     config.set(BigtableOptionsFactory.PROJECT_ID_KEY, getProjectId());
     config.set(BigtableOptionsFactory.ZONE_KEY, getZoneId());
     config.set(BigtableOptionsFactory.CLUSTER_KEY, getClusterId());
+    for (Entry<String, String> entry : additionalConfiguration.entrySet()) {
```
kevinsi4508: If additionalConfiguration contains config for CLUSTER_KEY let say, it overrides "config.set(BigtableOptionsFactory.CLUSTER_KEY, getClusterId());" Please document the order.

kevinsi4508:
What is the reason for exposing generic configuration such as Map<String, String>?

derjust:
Just as I was wading through the code yesterday: What about `com.google.cloud.bigtable.dataflow.CloudBigtableConfiguration.toHBaseConfig()`? As it has this `addtionalProperties`, shouldn't that applied there, too?

sduskis:
@kevinsi4508: updated as per your comments.  We're exposing a more generic construct because the likelihood of any configuration change is as likely as the datahost, which is to say not likely, but possibly needed.  PTAL

sduskis:
@derjust: Can you clarify your question?

If you're asking if CloudBigtableConfiguration.toHBaseConfig() was updated, then the answer is: yes, it was updated.

derjust:
@sduskis ah found it. From what I can envision as a consumer of the library I think the request from #484 is covered in a better/more flexible way :+1:

kevinsi4508:
LGTM

PR #488, opened 2015-09-16
Changing CloudBigtableIO write related classes to be public.

sduskis:
I spoke offline with @derjust.   He needed access to some of the inaccessible internal classes in Cloud Bigtable, and I would venture to say that others will need it as well.  Changing  CloudBigtableSingleTableWriteFn, CloudBigtableMultiTableWriteFn and CloudBigtableWriteTransform to be public rather than package private.


derjust:
:+1 :)

sduskis:
This is a low risk change.  I'm merging.

kevinsi4508:
LGTM, consider moving them out to individual files.

sduskis:
@kevinsi4508: The Dataflow SDK does the One-Big-File-Per-Integration class structure, which is why I did it as well.  I agree with you that it would be nicer to have multiple files.  This structure will be used in a similar class in the DataflowSDK in the near future, so I figured that we'll keep a similar structure to the DataflowSDK version.

PR #490, opened 2015-09-22
Revert "Revert "Set max message size limit to 256 MB to match the ser…

kevinsi4508:
Reverts GoogleCloudPlatform/cloud-bigtable-client#456
Set max message size to 256 MB as we are using gRPC 0.9.0 now/


sduskis:
LGTM

PR #491, opened 2015-09-24
Removing log4 transitive dependencies from bigtable-hbase-1.*

sduskis:
This change has been prompted by a request from a user.  This change was inspired by https://github.com/derjust/cloud-bigtable-client/commit/c775b09b7a9a64344910f17a679afbc42bccfb77

### DIFF
--- a/bigtable-hbase-1.0/pom.xml
+++ b/bigtable-hbase-1.0/pom.xml
@@ -52,6 +52,10 @@ limitations under the License.
             <groupId>org.apache.hbase</groupId>
             <artifactId>hbase-common</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>log4j-over-slf4j</artifactId>
+        </dependency>
     </dependencies>
     <build>
         <resources>
@@ -91,6 +95,8 @@ limitations under the License.
                                         <exclude>io.grpc:grpc-all:*</exclude>
                                         <exclude>io.netty:netty-all:*</exclude>
                                         <exclude>io.netty:netty:*</exclude>
+                                        <exclude>log4j:log4j:*</exclude>
+                                        <exclude>org.slf4j:lf4j-log4j12:*</exclude>
                                     </excludes>
                                     <includes>
                                         <include>com.google.guava:guava:${guava.version}</include>
@@ -124,11 +130,11 @@ limitations under the License.
 
                                 Using shadedArtifactAttached=true causes some tricky issues.  If =true is on,
                                 then the main artifact isn't usable, because of the incompatibility issues in the artifacts found
-                                in the relocations section.  If HBase updates its dependencies to compatible versions, then 
+                                in the relocations section.  If HBase updates its dependencies to compatible versions, then
                                 this project will not need to relocate and include those dependencies in a shaded jar.
 
-                                As an example, the start up of hbase requires guava, and using a newer version of guava causes 
-                                startup issues.  If an older version of guava is in the classpath, then HBase classes will load 
+                                As an example, the start up of hbase requires guava, and using a newer version of guava causes
+                                startup issues.  If an older version of guava is in the classpath, then HBase classes will load
                                 fine, but CBT will have runtime issues.
 
                                 Additionally, if shadedArtifactAttached=true there are even ssues using the shaded classifier.
--- a/pom.xml
+++ b/pom.xml
@@ -126,6 +126,13 @@ limitations under the License.
             </dependency>
 
             <dependency>
+                <groupId>org.slf4j</groupId>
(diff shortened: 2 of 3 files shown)


bigtable-hbase-1.0/pom.xml
```diff
@@ -52,6 +52,10 @@ limitations under the License.
             <groupId>org.apache.hbase</groupId>
             <artifactId>hbase-common</artifactId>
         </dependency>
+        <dependency>
```
kevinsi4508: Will this exclude transient dependencies, or we need to exclude them?
https://github.com/derjust/cloud-bigtable-client/commit/c775b09b7a9a64344910f17a679afbc42bccfb77

Just want to confirm.

derjust: Err no. 
See high level comment above

derjust:
I think that is not working as intendet:

```
mvn dependency:tree |grep log4j
[INFO] |  |  \- log4j:log4j:jar:1.2.17:compile
[INFO] |  |  \- org.slf4j:slf4j-log4j12:jar:1.6.1:compile
[INFO] |  +- log4j:log4j:jar:1.2.17:compile
[INFO] |  +- org.slf4j:slf4j-log4j12:jar:1.7.5:compile
[INFO] \- org.slf4j:log4j-over-slf4j:jar:1.7.5:runtime
[INFO] |  +- log4j:log4j:jar:1.2.17:compile
[INFO] |  +- org.slf4j:slf4j-log4j12:jar:1.7.5:compile
[INFO] \- org.slf4j:log4j-over-slf4j:jar:1.7.5:runtime
[INFO] |  |  |  \- log4j:log4j:jar:1.2.17:compile
[INFO] |  |  |  \- org.slf4j:slf4j-log4j12:jar:1.6.1:compile
[INFO] |  \- org.slf4j:log4j-over-slf4j:jar:1.7.5:runtime
[INFO] |  +- log4j:log4j:jar:1.2.17:compile
[INFO] |  +- org.slf4j:slf4j-log4j12:jar:1.7.5:compile
[INFO] |  \- org.slf4j:log4j-over-slf4j:jar:1.7.5:runtime
[INFO] |  +- log4j:log4j:jar:1.2.17:compile
[INFO] |  +- org.slf4j:slf4j-log4j12:jar:1.7.5:compile
```

My approach in derjust@c775b09 was 
* Exclude from the Hbase dependency `slf4j-log4j12` 
* Explicitly also exclude from the Hbase dependency `log4j` as it is a direct `compile` dependency, too
* Add as new dependency `log4j-over-slf4j` as `runtime` dependency to have at runtime log4j API calls to succeed from within the Hbase code

Does that makes sense?

sduskis:
I'm not sure why my top level pom.xml exclusions didn't make it into the PR.  That's where the exclusions ought to be managed to avoid duplication.

sduskis:
I updated the root pom.xml.  It now looks like this:

[INFO] |  +- log4j:log4j:jar:1.2.17:provided
[INFO] |  +- org.slf4j:slf4j-log4j12:jar:1.7.5:provided
[INFO] \- org.slf4j:log4j-over-slf4j:jar:1.7.5:runtime
[INFO] |  +- log4j:log4j:jar:1.2.17:provided
[INFO] |  +- org.slf4j:slf4j-log4j12:jar:1.7.5:provided
[INFO] \- org.slf4j:log4j-over-slf4j:jar:1.7.5:runtime
[INFO] |  \- org.slf4j:log4j-over-slf4j:jar:1.7.5:runtime
[INFO] |  +- log4j:log4j:jar:1.2.17:compile
[INFO] |  +- org.slf4j:slf4j-log4j12:jar:1.7.5:compile
[INFO] |  \- org.slf4j:log4j-over-slf4j:jar:1.7.5:runtime
[INFO] |  +- log4j:log4j:jar:1.2.17:compile
[INFO] |  +- org.slf4j:slf4j-log4j12:jar:1.7.5:compile

pom.xml
```diff
@@ -367,6 +374,14 @@ limitations under the License.
                         <groupId>com.google.guava</groupId>
                         <artifactId>guava</artifactId>
                     </exclusion>
+                    <exclusion>
```
kevinsi4508: exclude in hadoop-common too?

AngusDavis:
Is there any way to ensure that these don't make it back in via the enforcer plugin or something similar?

derjust:
LGTM 
Opening an issue to add an enforcer rule http://maven.apache.org/enforcer/enforcer-rules/bannedDependencies.html?

sduskis:
I was adding the enforcer rule and the hadoop-common exclusion before the comment.  I checked it in.  PTAL.

kevinsi4508:
LGTM

PR #492, opened 2015-09-24
Adding a test to make sure Puts with timestamps are retryable.

sduskis:


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -122,13 +121,18 @@
...
-    SslContextBuilder sslBuilder = GrpcSslContexts.forClient();
-    sslBuilder.ciphers(null);
-    return sslBuilder;
+    try {
+      SslContextBuilder sslBuilder = GrpcSslContexts.forClient();
+      sslBuilder.ciphers(null);
+      return sslBuilder;
+    } catch (IllegalArgumentException e) {
+      LOG.error("Could not initialize GrpcSslContexts", e);
+      return null;
```
kevinsi4508: How is this null value handled when it is returned from here?

sduskis:
I updated the BigtableSession sslBuilder handling to be a bit more in line with what we actually need.  We needed to bypass the sslBuilder creation until we needed it to create a BigtableSession rather than at class loading time.  There were failing unit tests that were testing a different aspect of Bigtable Session, but were failing on the creation of the sslBuilder, since ALPN was not present.  The new getSslBuilder() method defers the construction of the sslBuilder for that purpose.

PTAL

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -127,8 +126,16 @@ private static SslContextBuilder createGrpcSslBuilder() {
...
+    if (sslBuilder == null) {
+      sslBuilder = createGrpcSslBuilder();
+    }
+    return sslBuilder;
+  }
+
   private static SslContext createSslContext() throws SSLException {
-     return sslBuilder.build();
+    SslContextBuilder builder = getSslBuilder();
+    return builder == null ? null : builder.build();
```
kevinsi4508: can builder even be null?

sduskis: You're right.  I'll inline it.

sduskis:
I inlined getSslBuilder().

bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/adapters/TestPutAdapter.java
```diff
@@ -175,4 +178,23 @@ public void testEmptyPut() {
...
+    byte[] family1 = dataHelper.randomData("f1");
+    byte[] qualifier1 = dataHelper.randomData("qual1");
+    byte[] value1 = dataHelper.randomData("v1");
+
+    Put hbasePut = new Put(row, System.currentTimeMillis());
+    hbasePut.addColumn(family1, qualifier1, value1);
+    MutateRowRequest.Builder rowMutationBuilder = adapter.adapt(hbasePut);
+    MutateRowRequest request = rowMutationBuilder.build();
+    Predicate predicate =
+        BigtableSession.METHODS_TO_RETRY_MAP.get(BigtableServiceGrpc.METHOD_MUTATE_ROW);
```
kevinsi4508: can you verify all predicates in the map and with different inputs?

sduskis: Not in TestPutAdapter.  Other methods take different input.

kevinsi4508:
LGTM

derjust:
Highly promoting here to add a large notice in the README.md/java docs that it is (nearly) always preferably to use `new Put(..., timestamp)` over `new Put(...)` - From the (HBase?) API docs it just reads that overriding the timestamp has only an impact on the versioning and not around retry/reliability.

PR #493, opened 2015-09-24
Fixing CloudBigtableIO writer logging.

sduskis:
This was requested by a user: https://github.com/derjust/cloud-bigtable-client/commit/7d42db301d26f1f74a3480b97d6ed5809c0af080

### DIFF
--- a/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
+++ b/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
@@ -23,7 +23,6 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.NoSuchElementException;
-import java.util.UUID;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -636,16 +636,12 @@ public void onException(RetriesExhaustedWithDetailsException exception,
...
-      String failEventId = UUID.randomUUID().toString();
-      String message =
-          String.format("For context %s: eventId: %s occured during bulk writing.", context,
-            failEventId);
-      LOG.warn(message, exception);
-      for (int i = 0; i < causes.size(); i++) {
-        LOG.warn(String.format("context %s: eventId %s excepition %d", context, failEventId, i),
-          causes.get(i));
-      }
+      // RetriesExhaustedWithDetailsException has its pre-JDK1.7 semantics
```
kevinsi4508: Hard to understand this comment, can you rewrite it in simpler way? Focus on the difference between #getCause() and #getExhaustiveDescription().

derjust: The old generate single logging events for each `Caused by` exception. In a distributed environment where this code might be hit in parallel on multiple machines that makes it quite hard to collect the related exceptions together.
`getExhaustiveDescription` logs all `Caused by` exceptions in one logging event, including also details like rowkey and server address.

kevinsi4508: Makes sense, please update the comment to something like "logging all Causes" or move comment all together.

sduskis:
I updated the comment.  It's a lot simpler and more direct now.  PTAL

kevinsi4508:
LGTM

PR #494, opened 2015-09-24
Adding a toBuilder for all bigtable/dataflow Configuration objects.

sduskis:
This is something that @derjust was asking for.

### DIFF
--- a/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
+++ b/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
@@ -17,6 +17,7 @@ package com.google.cloud.bigtable.dataflow;
 
 import java.io.IOException;
 import java.io.Serializable;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -24,7 +25,8 @@ import java.util.Map.Entry;
 import org.apache.commons.lang.builder.EqualsBuilder;
 import org.apache.hadoop.conf.Configuration;
 
-import com.google.api.client.repackaged.com.google.common.base.Preconditions;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableMap;
 import com.google.cloud.bigtable.config.BigtableOptions;
 import com.google.cloud.bigtable.hbase.BigtableOptionsFactory;
 
@@ -45,6 +47,17 @@ public class CloudBigtableConfiguration implements Serializable {
     protected String zoneId;
     protected String clusterId;
     protected Map<String, String> additionalConfiguration = new HashMap<>();
+
+    public Builder() {
+    }
+
+    protected Builder(Map<String, String> configuration) {
+      this.additionalConfiguration.putAll(configuration);
+
+      this.projectId = this.additionalConfiguration.remove(BigtableOptionsFactory.PROJECT_ID_KEY);
+      this.zoneId = this.additionalConfiguration.remove(BigtableOptionsFactory.ZONE_KEY);
+      this.clusterId = this.additionalConfiguration.remove(BigtableOptionsFactory.CLUSTER_KEY);
+    }
 
     /**
      * Specifies the project ID for the Cloud Bigtable cluster.
@@ -88,24 +101,6 @@ public class CloudBigtableConfiguration implements Serializable {
     }
 
     /**
-     * Used for creating a new {@link Builder} from a {@link CloudBigtableConfiguration} or its
-     * subclasses.
-     */
-    protected T setConfiguration(Map<String, String> configuration) {
-      this.additionalConfiguration.putAll(configuration);
-
-      this.projectId = configuration.get(BigtableOptionsFactory.PROJECT_ID_KEY);
-      this.zoneId = configuration.get(BigtableOptionsFactory.ZONE_KEY);
-      this.clusterId = configuration.get(BigtableOptionsFactory.CLUSTER_KEY);
-
-      this.additionalConfiguration.remove(BigtableOptionsFactory.PROJECT_ID_KEY);
-      this.additionalConfiguration.remove(BigtableOptionsFactory.ZONE_KEY);
-      this.additionalConfiguration.remove(BigtableOptionsFactory.CLUSTER_KEY);
-
-      return (T) this;
-    }
-
-    /**
      * Builds the {@link CloudBigtableConfiguration}.
      * @return The new {@link CloudBigtableConfiguration}.
      */
@@ -119,7 +114,7 @@ public class CloudBigtableConfiguration implements Serializable {
 }
 
   // Not final due to serialization of CloudBigtableScanConfiguration.
-  protected Map<String, String> configuration;
+  private Map<String, String> configuration;
 
   // Used for serialization of CloudBigtableScanConfiguration.
   CloudBigtableConfiguration() {
@@ -198,9 +193,16 @@ public class CloudBigtableConfiguration implements Serializable {
    * Creates a new Builder from the configuration
    * @return A new {@link Builder}
    */
-  @SuppressWarnings("rawtypes")
+  @SuppressWarnings({ "rawtypes", "unchecked" })
   public Builder toBuilder() {
-    return new Builder<Builder<?>>().setConfiguration(configuration);
+    return new Builder(getConfiguration());
+  }
+
+  /**
+   * Returns an unmodifiable copy of the configuration map.
+   */
+  protected Map<String, String> getConfiguration() {
+    return ImmutableMap.copyOf(configuration);
   }
 
   public boolean equals(Object obj) {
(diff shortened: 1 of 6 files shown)


bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableScanConfiguration.java
```diff
@@ -155,4 +172,12 @@ public CloudBigtableScanConfiguration(String projectId, String zone, String clus
   public Scan getScan() {
     return serializableScan.scan;
   }
+
+  @Override
+  public Builder toBuilder() {
```
kevinsi4508: super.toBuilder().withTableId(tableId)...

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
```diff
@@ -169,4 +193,17 @@ public Configuration toHBaseConfig() {
...
     return config;
   }
+
+  /**
+   * Creates a new Builder from the configuration
+   * @return A new {@link Builder}
+   */
+  @SuppressWarnings("rawtypes")
+  public Builder toBuilder() {
+    return new Builder<Builder<?>>().setConfiguration(configuration);
```
kevinsi4508: In general, not a good idea to pass a mutable map around. ImmutableMap.copOf()?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableScanConfiguration.java
sduskis: I was thinking about doing that, but it's not practical.  Each subclass creates a different implementation of Builder.  super.toBuilder() would not have a withScan() method in it, since eventually it would return a CloudBigtableTableConfiguration.Builder instead of a CloudBigtableScanConfiguration.Builder.... it would be worse, if CloudBigtableTableConfiguration.toBuilder() returned super.toBuilder(), because then it wouldn't even have a withTableId in it since it would return a CloudBigtableConfiguration.Builder...

This whole thing is a bit convoluted, but I don't know if there's a better solution.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
```diff
@@ -87,13 +88,35 @@ public T withConfiguration(String key, String value) {
     }
 
     /**
+     * Used for creating a new {@link Builder} from a {@link CloudBigtableConfiguration} or its
+     * subclasses.
+     */
+    protected T setConfiguration(Map<String, String> configuration) {
+      this.additionalConfiguration.putAll(configuration);
```
kevinsi4508: how about this?
this.additionalConfiguration.putAll(configuration);
this.projectId = additionalConfiguration.remove(BigtableOptionsFactory.PROJECT_ID_KEY);
this.zoneId = additionalConfiguration.remove(BigtableOptionsFactory.ZONE_KEY);
this.clusterId = additionalConfiguration.remove(BigtableOptionsFactory.CLUSTER_KEY);

sduskis: +1

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
sduskis: +1

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
```diff
@@ -115,14 +138,15 @@ public CloudBigtableConfiguration build() {
...
+    setValue(BigtableOptionsFactory.PROJECT_ID_KEY, projectId, "Project ID");
+    setValue(BigtableOptionsFactory.ZONE_KEY, zoneId, "Zone ID");
+    setValue(BigtableOptionsFactory.CLUSTER_KEY, clusterId, "Cluster ID");
   }
 
-  private void setValue(String key, String value) {
+  private void setValue(String key, String value, String type) {
     Preconditions.checkArgument(!configuration.containsKey(key),
       String.format("%s was set twice", key));
+    Preconditions.checkArgument(value != null, String.format("%s must be set.", type));
```
kevinsi4508: Why can't we use key instead of type here?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableScanConfiguration.java
```diff
@@ -102,7 +102,11 @@ public SerializableScan(Scan scan) {
     }
 
     private void writeObject(ObjectOutputStream out) throws IOException {
-      out.writeObject(Base64.encodeBytes(ProtobufUtil.toScan(scan).toByteArray()));
+      out.writeObject(toEncodedString());
+    }
+
+    protected String toEncodedString() throws IOException {
```
kevinsi4508: nit: private?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableScanConfiguration.java
```diff
@@ -116,6 +120,19 @@ private static Scan toScan(String scanStr) throws IOException {
         throw new IOException("Could not deserialize the Scan.", ipbe);
       }
     }
+
+    @Override
+    public boolean equals(Object obj) {
+      if (obj == null || !(obj instanceof SerializableScan)) {
```
kevinsi4508: don't need to check null. "null instanceof SerializableScan" returns false.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableScanConfiguration.java
kevinsi4508: Different Builder classes, I see.
.setConfiguration(configuration) looks like constructor to me then.
but I am OK with what you have.

sduskis: Good point.  I'll move it to the constructor.

bigtable-hbase-dataflow/src/test/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfigurationTest.java
```diff
@@ -0,0 +1,91 @@
...
+            Collections.<String, String> emptyMap());
+
+    // Test CloudBigtableConfiguration that should be equal.
+    Assert.assertEquals(underTest1, underTest2);
+
+    // Test CloudBigtableConfiguration with different additionalConfigurations are not equal.
+    Assert.assertNotEquals(underTest1, underTest3);
+
+    // Test CloudBigtableConfiguration with different ProjectId are not equal.
+    Assert.assertNotEquals(underTest1, underTest4);
```
kevinsi4508: Create a test that compare two configs with additional properties are equal

bigtable-hbase-dataflow/src/test/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfigurationTest.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  @Test
+  public void testToBuilder() {
+    CloudBigtableConfiguration underTest =
+        new CloudBigtableConfiguration(PROJECT, ZONE, CLUSTER, Collections.singletonMap("somekey",
+          "somevalue"));
+    CloudBigtableConfiguration copy = underTest.toBuilder().build();
+    Assert.assertTrue(underTest != copy);
```
kevinsi4508: nit: assertNotSame()?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
sduskis: Type is more understandable from a user perspective in this case.  They have to set the projectId, zoneId or clusterId via a -- dataflow parameter (i.e. --projectId).  The key is only pertinent if a user is a more advanced user and sets a withConfiguration(key, value).  A user of the basic setup wouldn't know what a "google.bigtable.project.id" is, but would understand "Project ID."

sduskis:
Addressed all the comments in the last commit.  PTAL

kevinsi4508:
LGTM

PR #495, opened 2015-09-25
Fixing a flaky test in TestAsyncExecutor.

sduskis:
We had an instance where Thread.sleep(10) didn't work, and the test failed.  Hardening the test and making its intent clearer.


bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/async/TestAsyncExecutor.java
```diff
@@ -186,10 +189,14 @@ public Void call() throws Exception {
           return null;
         }
       });
-      Thread.sleep(10);
-      Assert.assertFalse(newRpcInvoked.get());
+      try {
+        future.get(100, TimeUnit.MILLISECONDS);
```
kevinsi4508: I believe I have asked about whether this would make the test flaky :D

kevinsi4508:
LGTM

PR #496, opened 2015-09-25
Removing the Generics from the Dataflow *Configuration classes.

sduskis:
The Generics in the Builder hierarchy causes all sorts of issues.  Removing them required some additional verbosity in overridden methods.  I think that the change will increase framework code, but reduce complexity from the user's perspective.

### DIFF
--- a/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableScanConfiguration.java
+++ b/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableScanConfiguration.java
@@ -67,7 +67,7 @@ public class CloudBigtableScanConfiguration extends CloudBigtableTableConfigurat
   /**
    * Builds a {@link CloudBigtableScanConfiguration}.
    */
-  public static class Builder extends CloudBigtableTableConfiguration.Builder<Builder> {
+  public static class Builder extends CloudBigtableTableConfiguration.Builder {
     protected Scan scan = new Scan();
 
     public Builder() {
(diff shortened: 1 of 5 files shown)


bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableScanConfiguration.java
```diff
@@ -87,6 +87,50 @@ public Builder withScan(Scan scan) {
...
+    /*
+     * Overrides of the CloudBigtableConfiguration.Builder's with*() so that they return
+     * CloudBigtableScanConfiguration.Builder.
+     */
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Builder withProjectId(String projectId) {
+      return (Builder) super.withProjectId(projectId);
```
kevinsi4508: @Override
public Builder withProjectId(String projectId) {
  super.withProjectId(projectId);
  return this;
}

kevinsi4508: This is no better than what we currently have.

sduskis: We ran into a bizarre problem with Generics.  Builder<?> != Builder in some cases; I don't know what kind of JVM quirk caused that, but Builder vs. Builder<?> caused some compilation problems fora  user.   This change fixes that problem.

kevinsi4508: Got it. There may be a better way to do this, but may not worth the time to pursue further.
I prefer following as casting is dangerous. It is up to you though.

{
super.withProjectId(projectId);
return this;
}

kevinsi4508:
LGTM

PR #498, opened 2015-09-29
Switch from Pipeline Options to Dataflow Pipeline Options

waprin:
Currently you can't use CloudBigtableOptions and use flags like setStreaming() since those are contained in StreamingOptions. This forces you to make extra classes to do the same thing. You could have BigtableOptions just extend StreamingOptions but DataflowPipelineOptions seems to contain a lot of useful additional options along with it.


sduskis:
LGTM

PR #500, opened 2015-10-01
Maven based switching of ALPN versions depending on JDK version.

sduskis:
The top pom.xml file detects JVM version and sets the correct ALPN version for testing purposes.


sduskis:
@derjust: This change is more about our internal testing.  We use the latest JDK 7 for now and JDK 8.  Our ALPN dependency has a test scope, so it won't affect our users too much, unless they're trying to do some development on the client; if ALPN is an issue in that case, one can still override the default behavior with -Dalpn.version={some other version} in the command line.

kevinsi4508:
LGTM

PR #501, opened 2015-10-02
Using OpenSSL for testing instead of alpn-boot jar

sduskis:
See https://github.com/grpc/grpc-java/blob/master/SECURITY.md for more details
See https://github.com/grpc/grpc-java/commit/a99e97791f3160af5f6e91b67c6bbc17ae29be36 for a change related to the grpc testing envrionment

We need to set up our testing environment with OpenSSL 1.0.2 before checking this in.

DO NOT CHECK IN

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -53,8 +53,10 @@ import io.grpc.netty.NegotiationType;
 import io.grpc.netty.NettyChannelBuilder;
 import io.netty.channel.EventLoopGroup;
 import io.netty.channel.nio.NioEventLoopGroup;
+import io.netty.handler.ssl.OpenSsl;
 import io.netty.handler.ssl.SslContext;
 import io.netty.handler.ssl.SslContextBuilder;
+import io.netty.handler.ssl.SslProvider;
 
 import java.io.BufferedWriter;
 import java.io.Closeable;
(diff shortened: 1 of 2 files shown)


kevinsi4508:
remember to remove the following warning:

BigtableSession:51 - Could not asynchronously create the ssl context, since ALPN is not installed

sduskis:
@kevinsi4508 and I spoke off line about logging which mode gRPC (alpn-boot vs. tcnactive) is working with.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -120,21 +122,16 @@
...
-
-  private static synchronized SslContextBuilder getSslBuilder() {
+  private synchronized static SslContext createSslContext() throws SSLException {
     if (sslBuilder == null) {
-      sslBuilder = createGrpcSslBuilder();
+      sslBuilder = GrpcSslContexts.forClient().ciphers(null);
+      if (OpenSsl.isAvailable()) {
+        LOG.info("gRPC is using the OpenSSL provider");
+      } else {
+        LOG.info("gRPC is using the JDK provder");
```
kevinsi4508: is that when JDK provider is used, we need to check org.eclipse.jetty.alpn.ALPN ?

kevinsi4508: gRPC is using the JDK provider

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -120,21 +122,16 @@
...
-    sslBuilder.ciphers(null);
-    return sslBuilder;
-  }
-
-  private static synchronized SslContextBuilder getSslBuilder() {
+  private synchronized static SslContext createSslContext() throws SSLException {
     if (sslBuilder == null) {
-      sslBuilder = createGrpcSslBuilder();
+      sslBuilder = GrpcSslContexts.forClient().ciphers(null);
+      if (OpenSsl.isAvailable()) {
```
kevinsi4508: Move this OpenSsl.isAvailable() before GrpcSslContexts.forClient().ciphers(null) because if it may fail. If it does fail, we want to see the log.

sduskis:
Closing this in favor of other PRs that are more targeted at what we actually need.

Issue #506, opened 2015-10-09
Allow user-specification of TLS provider


cmccoy:
#501 adds support for using OpenSSL for TLS instead the JDK (via Netty ALPN).  The current PR tries OpenSSL first, falling back to JDK if OpenSSL support isn't available. 

It would be helpful to have a configuration value to specify the TLS transport which should be used, raising an exception if it's unavailable (rather than falling back on a slower method).

sduskis:
gRPC configuration was malleable enough to make this work.  We ought to have this available some time today.

sduskis:
By default, tcnative is chose if both are configured.  Any code we write is going to be more confusing than test code managing the inclusion of the tcnative jar.

PR #508, opened 2015-10-12
Logging use of alpn-boot vs. tcnative jar.

sduskis:
There are a few issues related to using the tcnative jar.  This one is about logging the use of tcnative or alpn-boot.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -53,6 +53,7 @@ import io.grpc.netty.NegotiationType;
 import io.grpc.netty.NettyChannelBuilder;
 import io.netty.channel.EventLoopGroup;
 import io.netty.channel.nio.NioEventLoopGroup;
+import io.netty.handler.ssl.OpenSsl;
 import io.netty.handler.ssl.SslContext;
 import io.netty.handler.ssl.SslContextBuilder;
 
@@ -120,21 +121,16 @@ public class BigtableSession implements AutoCloseable {
     performWarmup();
   }
 
-  private static SslContextBuilder createGrpcSslBuilder() {
-    SslContextBuilder sslBuilder = GrpcSslContexts.forClient();
-    sslBuilder.ciphers(null);
-    return sslBuilder;
-  }
-
-  private static synchronized SslContextBuilder getSslBuilder() {
+  private synchronized static SslContext createSslContext() throws SSLException {
     if (sslBuilder == null) {
-      sslBuilder = createGrpcSslBuilder();
+      sslBuilder = GrpcSslContexts.forClient().ciphers(null);
+      if (OpenSsl.isAvailable()) {
+        LOG.info("gRPC is using the OpenSSL provider (tcnactive jar)");
+      } else {
+        LOG.info("gRPC is using the JDK provider (alpn-boot jar)");
+      }
     }
-    return sslBuilder;
-  }
-
-  private static SslContext createSslContext() throws SSLException {
-    return getSslBuilder().build();
+    return sslBuilder.build();
   }
 
   private static void performWarmup() {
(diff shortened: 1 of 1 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -120,21 +121,16 @@
...
-
-  private static synchronized SslContextBuilder getSslBuilder() {
+  private synchronized static SslContext createSslContext() throws SSLException {
     if (sslBuilder == null) {
-      sslBuilder = createGrpcSslBuilder();
+      sslBuilder = GrpcSslContexts.forClient().ciphers(null);
+      if (OpenSsl.isAvailable()) {
+        LOG.info("gRPC is using the OpenSSL provider (tcnactive jar)");
+      } else {
+        LOG.info("gRPC is using the JDK provder (alpn-boot jar)");
```
kevinsi4508: "JDK provider"

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -120,21 +121,16 @@
...
-    return sslBuilder;
-  }
-
-  private static synchronized SslContextBuilder getSslBuilder() {
+  private synchronized static SslContext createSslContext() throws SSLException {
     if (sslBuilder == null) {
-      sslBuilder = createGrpcSslBuilder();
+      sslBuilder = GrpcSslContexts.forClient().ciphers(null);
+      if (OpenSsl.isAvailable()) {
+        LOG.info("gRPC is using the OpenSSL provider (tcnactive jar)");
```
kevinsi4508: Do we really want to log this multiple times? If not, we can log in performWarmup().

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -146,12 +142,14 @@ public void run() {
...
-          LOG.warn(
-            "Could not asynchronously create the ssl context, since ALPN is not installed.");
-          return;
+        if (!OpenSsl.isAvailable()) {
+          try {
+            Class.forName("org.eclipse.jetty.alpn.ALPN");
+          } catch (ClassNotFoundException e1) {
+            LOG.warn(
+              "Could not asynchronously create the ssl context, since ALPN is not installed.");
+            return;
```
kevinsi4508: Do we want to return here? if we continue, we may get more detailed error message.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
sduskis: There's a request to have a parameter of some sort to use a programatic configuration of openSSL vs. boot jar (#506).  I think per connection is better for that case.  If you're OK with it, let's keep it here pending the fix for #506

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
sduskis: Fixed.  Thanks.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
sduskis: I was fine with the original logic.  I think that we saw that the gRPC infrastructure's error message was actually less informative in earlier versions of gRPC.  Regardless, I'd rather not change the logic for the alpn-boot case in the tcnative PRs, if that's OK with you.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
kevinsi4508: Yes, I am fine with it. Per connection is OK.

kevinsi4508:
LGTM

PR #509, opened 2015-10-12
Adding a tcnative option for integration testing.

sduskis:
Moving the alpn-boot dependency to the two existing bigtable integration test profiles.  Adding a new profile, bigtableIntegrationTcnativeTest, that uses tcnative instead of alpn.

### DIFF
--- a/bigtable-hbase-integration-tests/pom.xml
+++ b/bigtable-hbase-integration-tests/pom.xml
@@ -47,6 +47,10 @@ limitations under the License.
                     <artifactId>${google.bigtable.project.under.test}</artifactId>
                     <version>${project.version}</version>
                 </dependency>
+                <dependency>
+                    <groupId>org.mortbay.jetty.alpn</groupId>
+                    <artifactId>alpn-boot</artifactId>
+                </dependency>
             </dependencies>
             <build>
                 <plugins>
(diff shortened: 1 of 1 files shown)


bigtable-hbase-integration-tests/pom.xml
```diff
@@ -47,6 +47,10 @@ limitations under the License.
                     <artifactId>${google.bigtable.project.under.test}</artifactId>
                     <version>${project.version}</version>
                 </dependency>
+              <dependency>
```
kevinsi4508: indent

bigtable-hbase-integration-tests/pom.xml
```diff
@@ -98,6 +152,10 @@ limitations under the License.
...
```
kevinsi4508: indent

sduskis: done.

kevinsi4508: Do we want GapTcnativeTest?

kevinsi4508:
LGTM

bigtable-hbase-integration-tests/pom.xml
sduskis: I don't think so.  We want to test to make sure that both alpn-boot and tcnative work with NoKnownGap tests.  The Gap test is an orthogonal concern.

PR #510, opened 2015-10-12
Better handling of ALPN configuration at startup, plus debug info.

sduskis:
gRPC provides some information about the state of ALPN configuration.  Using that information for BigtableSession startup as well as CloudBigtableIO client side validation.

NOTE: This checkin contains some code that will be used for a short period of time on our Jenkins server to diagnose the tcnative issues we're having there.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -124,13 +124,37 @@ public class BigtableSession implements AutoCloseable {
   private synchronized static SslContext createSslContext() throws SSLException {
     if (sslBuilder == null) {
       sslBuilder = GrpcSslContexts.forClient().ciphers(null);
-      if (OpenSsl.isAvailable()) {
-        LOG.info("gRPC is using the OpenSSL provider (tcnactive jar)");
+      // gRPC uses tcnative / OpenSsl by default, if it's available.  It defaults to alpn-boot
+      // if tcnative is not in the classpath.
+      if (OpenSsl.isAlpnSupported()) {
+        LOG.info("gRPC is using the OpenSSL provider (tcnactive jar - Open Ssl version: %s)",
+          OpenSsl.versionString());
       } else {
-        LOG.info("gRPC is using the JDK provider (alpn-boot jar)");
+        if (isJettyAlpnConfigured()) {
+          // gRPC uses jetty ALPN as a backup to tcnative.
+          LOG.info("gRPC is using the JDK provider (alpn-boot jar)");
+        } else {
+          LOG.info("gRPC cannot be configured.  Neither OpenSsl nor Alpn are available.");
+        }
       }
     }
     return sslBuilder.build();
+  }
+
+  public static boolean isAlpnProviderEnabled() {
+    return OpenSsl.isAlpnSupported() || isJettyAlpnConfigured();
+  }
+
+  /**
+   * Indicates whether or not the Jetty ALPN jar is installed in the boot classloader.
+   */
+  private static boolean isJettyAlpnConfigured() {
+    try {
+      Class.forName("org.eclipse.jetty.alpn.ALPN", true, null);
+      return true;
+    } catch (ClassNotFoundException e) {
+      return false;
+    }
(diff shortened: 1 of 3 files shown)


kevinsi4508:
What tcnative issues we currently have?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -124,10 +124,17 @@
...
       if (OpenSsl.isAvailable()) {
-        LOG.info("gRPC is using the OpenSSL provider (tcnactive jar)");
+        LOG.info("gRPC is using the OpenSSL provider (tcnactive jar - Open Ssl version: {})",
+          OpenSsl.versionString());
       } else {
-        LOG.info("gRPC is using the JDK provider (alpn-boot jar)");
+        if (OpenSsl.isAlpnSupported()) {
+          LOG.info("gRPC is using the JDK provider (alpn-boot jar)");
+        } else {
+          LOG.info("gRPC can't be configured");
```
kevinsi4508: Can we say something like ...., both OpenSsl and Alpn are not supported?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -142,22 +149,15 @@ public void run() {
         // The first invocation of createSslContext() is expensive.
         // Create a throw away object in order to speed up the creation of the first
         // BigtableConnection which uses SslContexts under the covers.
-        if (!OpenSsl.isAvailable()) {
+        if (isAlpnConfigured()) {
```
kevinsi4508: what happened here?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -599,5 +605,9 @@ public void run() {
 
     Runtime.getRuntime().addShutdownHook(reportingThread);
   }
+
+  public static boolean isAlpnConfigured() {
+    return OpenSsl.isAvailable() || OpenSsl.isAlpnSupported();
```
kevinsi4508: Not sure what this means? By default if OpenSsl.isAvailable() == true, OpenSsl is used, not Alpn.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
sduskis: We should not try to do warmup if neither alpn-boot or tcnative is available

kevinsi4508: Sorry, isAlpnConfigured() confused me. how about isSslProviderAvailable()?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -842,17 +842,19 @@ private static void validate(CloudBigtableConfiguration configuration, String ta
     if (tableId != null) {
       checkNotNullOrEmpty(tableId, "tableid");
     }
-    try (BigtableConnection conn = new BigtableConnection(configuration.toHBaseConfig());
-        Admin admin = conn.getAdmin()) {
-      if (tableId != null) {
-        Preconditions.checkState(admin.tableExists(TableName.valueOf(tableId)));
-      } else {
-        admin.listTableNames();
+    if (BigtableSession.isAlpnConfigured()) {
```
kevinsi4508: Don't we want to log something if no Ssl provider is availabe?

kevinsi4508:
LGTM, please address some comments.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
sduskis: ALPN is the protocol.  It's supported either through tcnative or alpn-boot jar.

PR #514, opened 2015-10-13
Move CloudBigtableIO.writeToTable() test case from CloudBigtableIOTes…

kevinsi4508:
Move CloudBigtableIO.writeToTable() test case from CloudBigtableIOTest to CloudBigtableIOIntegrationTest.

### DIFF
--- a/bigtable-hbase-dataflow/src/test/java/com/google/cloud/bigtable/dataflow/CloudBigtableIOTest.java
+++ b/bigtable-hbase-dataflow/src/test/java/com/google/cloud/bigtable/dataflow/CloudBigtableIOTest.java
@@ -20,8 +20,6 @@ import static org.junit.Assert.assertNotNull;
 import static org.mockito.Matchers.any;
 import static org.mockito.Mockito.when;
 
-import java.util.Collections;
-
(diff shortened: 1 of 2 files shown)


bigtable-hbase-dataflow/src/test/java/com/google/cloud/bigtable/dataflow/CloudBigtableIOTest.java
```diff
@@ -81,8 +70,6 @@ public void testInitialize() throws Exception{
     checkRegistry(Put.class);
```
sduskis: Can you please move this into the Integration test?  That seems like a better fit.

kevinsi4508: Coder registry is mocked. I think we should keep it in unit testing.
"when(underTest.getCoderRegistry()).thenReturn(registry);"

sduskis: ok

sduskis:
LGTM

PR #515, opened 2015-10-14
Removing noisy exception logging from the test process.

sduskis:
Our logging for problems during retries and async behavior is great for the users.  Testing those expected problems internally shouldn't trigger the logging.  This PR is to remove the noise from maven testing so that meaningful exceptions are easier to spot.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/RetryingRpcFutureFallback.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/RetryingRpcFutureFallback.java
@@ -80,7 +80,6 @@ public class RetryingRpcFutureFallback<RequestT, ResponseT> implements FutureFal
     }
     long nextBackOff = currentBackoff.nextBackOffMillis();
     if (nextBackOff == BackOff.STOP) {
-      LOG.warn("RetriesExhausted: ", cause);
       throw new ScanRetriesExhaustedException("Exhausted streaming retries.", cause);
     }
 
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
@@ -4,9 +4,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -27,8 +27,6 @@ import java.io.IOException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
-
-import javax.annotation.Nullable;
 
 import com.google.api.client.util.BackOff;
 import com.google.api.client.util.Clock;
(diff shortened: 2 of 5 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/RetryingRpcFutureFallback.java
```diff
@@ -80,7 +80,6 @@ private RetryingRpcFutureFallback(RetryOptions retryOptions, RequestT request,
     }
     long nextBackOff = currentBackoff.nextBackOffMillis();
     if (nextBackOff == BackOff.STOP) {
-      LOG.warn("RetriesExhausted: ", cause);
       throw new ScanRetriesExhaustedException("Exhausted streaming retries.", cause);
```
sduskis: NOTE: The underlying cause of the retries exhausted is included in the ScanRetriesExhaustedException.  It should not be logged here, since the user can figure out how to handle the exception, including logging.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -263,45 +266,63 @@ boolean doRefresh() {
    *          retry exhaustion logic.
    * @return HeaderCacheElement containing either a valid {@link AccessToken} or an exception.
```
kevinsi4508: Update doc

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -263,45 +266,63 @@ boolean doRefresh() {
...
-        // initialize backoff.
-        backoff = retryOptions.createBackoff();
-      }
-
-      // Given the backoff, either sleep for a short duration, or terminate if the backoff has
-      // reached its configured timeout limit.
+  protected HeaderCacheElement refreshCredentialsWithRetry() {
+    BackOff backoff = null;
+    while (true) {
+      // Performs a return when:
```
kevinsi4508: Move this to method doc?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -263,45 +266,63 @@ boolean doRefresh() {
...
+        warnLog("Got an unexpected exception while trying to refresh google credentials.", e);
+        return new HeaderCacheElement(new IOException("Could not read headers", e));
       }
-    } catch (Exception e) {
-      LOG.warn("Got an unexpected exception while trying to refresh google credentials.", e);
-      return new HeaderCacheElement(new IOException("Could not read headers", e));
+    }
+  }
+
+  private void infoLog(String message) {
```
kevinsi4508: I am fine with this, would be nice if we have something like this in "bigtable/config/Logger.java".

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResumingStreamingResultScanner.java
```diff
@@ -97,10 +101,14 @@ public Row next() throws IOException {
 
         return result;
       } catch (ScanTimeoutException rte) {
-        LOG.warn("ReadTimeoutException: ", rte);
+        if (logWarningsBeforeRetry) {
```
kevinsi4508: ditto here, if we have it in Logger would be nice.

kevinsi4508:
discussed offline. Please add some comment on the flags that we use to suppress the log message in testing.

kevinsi4508:
LGTM

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
sduskis: Done

sduskis:
The mock logger was a good idea.  I updated the code and the tests.

kevinsi4508:
LGTM

PR #516, opened 2015-10-15
Adding ability to dual configure alpn-boot and tcnative

sduskis:
Our performance test team wants to dual configure tcnative and alpn-boot and configure the implementation.  With this runtime change, it was also ideal to simply theconfiguration of our integration-testing pom.xml.

see #506

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
@@ -18,7 +18,10 @@ package com.google.cloud.bigtable.config;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.api.client.util.Strings;
 import com.google.cloud.bigtable.grpc.BigtableClusterName;
+import com.google.cloud.bigtable.grpc.BigtableSession;
 import com.google.common.base.Preconditions;
+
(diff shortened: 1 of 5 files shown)


kevinsi4508:
The description is not clear to me. We want to flag control the switching between tcnative and alpn-boot instead of using maven profile, correct?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
```diff
@@ -71,6 +74,7 @@
     private String callStatusReportPath;
     private String callTimingReportPath;
 
+    private boolean forceJdkTlsProvider = false;
```
kevinsi4508: prefer explicit configuration values of { JDK, OPENSSL }, then map them to SslProvider.JDK and SslProvider.OPENSSL.

sduskis: We can't use SslProvider directly.  In most cases, we want what ever the default configuration that gRPC provides.  When forceJdkTlsProvider is turned on, we have to do more than just override the default SslProvider... the logic is a bit more complex than that.

sduskis:
We discussed this offline.  We will not merge this change, since it's more confusing to do things this way than simply including or not including a tcnative jar in the classpath in test code.

PR #518, opened 2015-10-15
Removing CallCompletionStatusInterceptor.

sduskis:
It was really helpful when we first started this project.  It's not used anymore, and doesn't seem to be helpful now.  It's getting in the way of some other things we need to do.


bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
```diff
@@ -53,7 +50,6 @@
   public static final String PROJECT_ID_KEY = "google.bigtable.project.id";
   public static final String CLUSTER_KEY = "google.bigtable.cluster.name";
   public static final String ZONE_KEY = "google.bigtable.zone.name";
-  public static final String CALL_REPORT_DIRECTORY_KEY = "google.bigtable.call.report.directory";
```
kevinsi4508: Can you also remove reference to "google.bigtable.call.report.directory"?
https://cs.corp.google.com/#search/&q=%22google.bigtable.call.report.directory%22&type=cs

sduskis: Done.  PTAL

kevinsi4508:
LGTM

PR #520, opened 2015-10-16
Moving duplicated RetryOptions test build logic into a utility.

sduskis:
We use a nanoClock override for the BackoffBuilder in our unit tests.  We have two copies of nasty looking, but useful code.  I found a third use for it, and figured that I'd centralize the logic before adding the third test.


kevinsi4508:
LGTM

PR #523, opened 2015-10-19
Removing some unnecessary transitive dependencies from the hbase-client

sduskis:


bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
```diff
@@ -203,6 +199,10 @@ public Builder toBuilder() {
   }
 
   public boolean equals(Object obj) {
-    return EqualsBuilder.reflectionEquals(this, obj);
+    if (obj == null || obj.getClass() != this.getClass()) {
```
kevinsi4508: obj instanceof CloudBigtableConfiguration?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableScanConfiguration.java
```diff
@@ -230,6 +231,13 @@ public Scan getScan() {
   }
 
   @Override
+  public boolean equals(Object obj) {
+    return super.equals(obj)
+        && Objects
+            .equals(serializableScan, ((CloudBigtableScanConfiguration) obj).serializableScan);
```
kevinsi4508: do u want to check if obj is null, similar to CloudBigtableConfiguration.java

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
sduskis: If it's not the same class (i.e. if objs or this is a different subclass), then the objects are not equals.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableScanConfiguration.java
sduskis: is it worth doing even though super.equals(obj) does it?

kevinsi4508: missed super.equals(obj). That is fine.

kevinsi4508:
LGTM

PR #527, opened 2015-10-21
Fixing the deleting while scanning issue.

sduskis:
A scan prevents other operations from being done on a channel.  Reserve a channel specifically for the scan, and let the other operations proceed on the pool.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
@@ -16,7 +16,6 @@
 package com.google.cloud.bigtable.grpc;
 
 import io.grpc.CallOptions;
-import io.grpc.Channel;
 import io.grpc.ClientCall;
 import io.grpc.MethodDescriptor;
 import io.grpc.stub.StreamObserver;
@@ -42,6 +41,8 @@ import com.google.cloud.bigtable.config.RetryOptions;
 import com.google.cloud.bigtable.grpc.async.BigtableAsyncUtilities;
 import com.google.cloud.bigtable.grpc.async.RetryableRpc;
 import com.google.cloud.bigtable.grpc.io.CancellationToken;
+import com.google.cloud.bigtable.grpc.io.ChannelPool;
+import com.google.cloud.bigtable.grpc.io.ChannelPool.PooledChannel;
 import com.google.cloud.bigtable.grpc.io.ClientCallService;
 import com.google.cloud.bigtable.grpc.io.RetryingCall;
 import com.google.cloud.bigtable.grpc.scanner.BigtableResultScannerFactory;
@@ -117,7 +118,7 @@ public class BigtableDataGrpcClient implements BigtableDataClient {
     return true;
   }
 
-  private final Channel channel;
+  private final ChannelPool channelPool;
 
   private final ScheduledExecutorService retryExecutorService;
   private final ExecutorService executorService;
@@ -136,30 +137,31 @@ public class BigtableDataGrpcClient implements BigtableDataClient {
   private ClientCallService clientCallService;
 
   public BigtableDataGrpcClient(
-    Channel channel,
-    ExecutorService executorService,
-    ScheduledExecutorService retryExecutorService,
+      ChannelPool channelPool,
+      ExecutorService executorService,
+      ScheduledExecutorService retryExecutorService,
     RetryOptions retryOptions) {
-    this(channel, executorService, retryExecutorService, retryOptions, ClientCallService.DEFAULT);
+    this(channelPool, executorService, retryExecutorService, retryOptions,
+        ClientCallService.DEFAULT);
   }
 
   @VisibleForTesting
   BigtableDataGrpcClient(
-      Channel channel,
+      ChannelPool channelPool,
       ExecutorService executorService,
       ScheduledExecutorService retryExecutorService,
       RetryOptions retryOptions,
       ClientCallService clientCallService) {
-    this.channel = channel;
+    this.channelPool = channelPool;
     this.executorService = executorService;
     this.retryOptions = retryOptions;
     this.clientCallService = clientCallService;
     this.retryExecutorService = retryExecutorService;
 
     this.sampleRowKeysAsync =
-        BigtableAsyncUtilities.createSampleRowKeyAsyncReader(this.channel, clientCallService);
+        BigtableAsyncUtilities.createSampleRowKeyAsyncReader(this.channelPool, clientCallService);
     this.readRowsAsync =
-        BigtableAsyncUtilities.createRowKeyAysncReader(this.channel, clientCallService);
+        BigtableAsyncUtilities.createRowKeyAysncReader(this.channelPool, clientCallService);
   }
 
   @Override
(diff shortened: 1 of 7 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -136,30 +137,30 @@ private static final boolean allCellsHaveTimestamps(Iterable<Mutation> mutations
   private ClientCallService clientCallService;
 
   public BigtableDataGrpcClient(
-    Channel channel,
-    ExecutorService executorService,
-    ScheduledExecutorService retryExecutorService,
+      ChannelPool channel,
```
kevinsi4508: nit:  ChannelPool channelPool

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -33,7 +35,35 @@
...
+        MethodDescriptor<RequestT, ResponseT> methodDescriptor, CallOptions callOptions) {
+      return createWrappedCall(methodDescriptor, callOptions, delegate);
+    }
+
+    @Override
+    public String authority() {
+      return delegate.authority();
+    }
+
+    public synchronized void returnToPool() {
```
kevinsi4508: this synchronized on WrappedChannel class, but returnChannel() synchronized on ChannelPool

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
sduskis: done

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
sduskis: I synchronize returnToPool() so that checking and setting WrappedChannel.returned is atomic.  Channel.returnChannel is also synchronized.

kevinsi4508: Got it. thanks!

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -62,8 +100,44 @@ public ChannelPool(Channel[] channels, List<HeaderInterceptor> headerInterceptor
...
   @Override
   public String authority() {
     return channels[0].authority();
   }
+
+  public synchronized WrappedChannel reserveChannel() {
+    Channel reserved;
+    boolean returned = false;
+    if (channels.length == 1) {
+      // Always leave one connection in the pool.
```
kevinsi4508: This logic is a bit weird. Can you move this comment to the method doc?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -62,8 +100,44 @@ public ChannelPool(Channel[] channels, List<HeaderInterceptor> headerInterceptor
...
+      reserved = channels[0];
+      returned = true;
+    } else {
```
kevinsi4508: I guess the size of the channels can never be 0.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -62,8 +100,44 @@ public ChannelPool(Channel[] channels, List<HeaderInterceptor> headerInterceptor
...
+      reserved = channels[channels.length - 1];
+      Channel[] newChannelArray = new Channel[channels.length - 1];
+      System.arraycopy(channels, 0, newChannelArray, 0, channels.length - 1);
```
kevinsi4508: Would it be easier if we use List instead of an array?

bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/io/ChannelPoolTest.java
```diff
@@ -31,35 +33,62 @@
...
-  @Mock
-  private ClientCall.Listener responseListenerStub;
-  @Mock
-  private HeaderInterceptor interceptor;
-
   @Test
   public void testInterceptorIsCalled() throws Exception {
-    MockitoAnnotations.initMocks(this);
-    when(channel.newCall(any(MethodDescriptor.class), any(CallOptions.class))).thenReturn(
+    Channel channel = mock(Channel.class);
```
kevinsi4508: Why can't we mock and initialize the channel pool at Setup step?

bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/io/ChannelPoolTest.java
```diff
@@ -31,35 +33,62 @@
...
+  }
+  
+  @Test
+  public void testReserveReturnl() {
+    ChannelPool pool = new ChannelPool(new Channel[] { null, null }, null);
+    assertEquals(2, pool.size());
+    WrappedChannel reserved = pool.reserveChannel();
+    assertEquals(1, pool.size());
+    reserved.returnToPool();
+    assertEquals(2, pool.size());
```
kevinsi4508: Ideally, we should check the channels in the pool. I am fine with this.

bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/io/ChannelPoolTest.java
```diff
@@ -31,35 +33,62 @@
...
+    pool.newCall(descriptor, CallOptions.DEFAULT);
+    verify(channel1, times(1)).newCall(same(descriptor), same(CallOptions.DEFAULT));
+    verify(channel2, times(0)).newCall(same(descriptor), same(CallOptions.DEFAULT));
+    pool.newCall(descriptor, CallOptions.DEFAULT);
+    verify(channel1, times(1)).newCall(same(descriptor), same(CallOptions.DEFAULT));
+    verify(channel2, times(1)).newCall(same(descriptor), same(CallOptions.DEFAULT));
+  }
+  
+  @Test
+  public void testReserveReturnl() {
```
kevinsi4508: testReserveReturn?

bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/io/ChannelPoolTest.java
```diff
@@ -31,35 +33,62 @@
...
+    WrappedChannel reserved = pool.reserveChannel();
+    assertEquals(1, pool.size());
+    reserved.returnToPool();
+    assertEquals(2, pool.size());
+  }
+
+  @Test
+  public void testReserveNeverExhaustsPool() {
+    ChannelPool pool = new ChannelPool(new Channel[] { null, null }, null);
+    for (int i = 0; i < 10; i++) {
```
kevinsi4508: nit:
pool.reserveChannel();
for (int i = 0; i < 10; i++) {
  assertEquals(1, pool.size());
}

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -33,7 +35,36 @@
 
   protected static final Logger log = Logger.getLogger(ChannelPool.class.getName());
 
-  private final Channel[] channels;
+  public class WrappedChannel extends Channel {
```
kevinsi4508: ReservableChannel or PooledChannel? since it has returnToPool()

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/StreamingBigtableResultScanner.java
```diff
@@ -124,10 +125,12 @@ public void addResult(ReadRowsResponse response) {
   }
 
   public void setError(Throwable error) {
+    reservedChannel.returnToPool();
```
kevinsi4508: The way we are reserving and returning channel is kind of scary. If we make a mistake somewhere and not returning channels, that will result in memory leak and keeping too many channels open.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/StreamingBigtableResultScanner.java
```diff
@@ -97,17 +98,17 @@ public ResponseQueueReader(
   private final CancellationToken cancellationToken;
   private final BlockingQueue<ResultQueueEntry<ReadRowsResponse>> resultQueue;
   private final ResponseQueueReader responseQueueReader;
+  private final WrappedChannel reservedChannel;
```
kevinsi4508: where is this channel used, other than returnToPool()?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
sduskis: Correct

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
sduskis: That might have an impact on the normal use case of ChannelPool.newCall().  We want that use case to be as fast as possible.

bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/io/ChannelPoolTest.java
sduskis: There are different requirements for mocking objects in different test cases.  We can either mock all combinations for every test, or specify the objects we want to mock for a specific test case.  I opted for the latter.

bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/io/ChannelPoolTest.java
sduskis: What do you mean?

bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/io/ChannelPoolTest.java
sduskis: done

sduskis:
All comments were addressed. PTAL.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/StreamingBigtableResultScanner.java
sduskis: The channel will get closed when the BigtableSession closes.  We have hooks for that.  We don't actually create new Channel in the pool.  It's more likely that if we have lots of scans that somehow lose channels, that we'll go down from 4 channels to 1, and we'll have a potential performance hit.

kevinsi4508:
LGTM

PR #528, opened 2015-10-21
Implementing Admin.getTableDescriptorsByTableName.

sduskis:


bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestListTables.java
```diff
@@ -85,8 +88,34 @@ public void testTableNames() throws IOException {
         Assert.assertTrue(ArrayUtils.contains(tableList, tableName2));
       }
 
-      checkColumnFamilies(admin, tableName2);
+      {
+        TableName[] tableList = admin.listTableNames(Pattern.compile("list_table1-.*"));
```
kevinsi4508: tableNameList or nameList or tableNames

bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestListTables.java
```diff
@@ -85,8 +88,34 @@ public void testTableNames() throws IOException {
...
+        Assert.assertTrue(ArrayUtils.contains(tableList, tableName1));
+        Assert.assertFalse(ArrayUtils.contains(tableList, tableName2));
```
kevinsi4508: if we are expecting only one element in the array.
we should do:
1. check size == 1
2. [0] == expect

or 
some util that asserts array has exactly has one element == expected

sduskis: Failures in our integration tests often leave behind unexpected tables.  I don't want one test to fail for a valid reason and then subsequent tests to fail because the tables were left over.

kevinsi4508: Are you saying that we cannot expect the result to be a single element array?

sduskis: We have different names for the tables because we add a random UUID as a suffix.  We're retrieving based on a pattern with a shared prefix, but verify that the specific table is in the list.

kevinsi4508: You just explained what the code is doing, I understand that.
Anyway, we can have a separate discussion about this. Let's not delay this PR because of this.

kevinsi4508:
LGTM

PR #536, opened 2015-10-22
Reducing the number of threads required by a connection.

sduskis:
We used 256 following HBase's example.  HBase uses those threads to perform asynchronous RPC calls.  We use gRPC which uses nio for the long running work.  We use threads only to handle the object returned by the RPC.  We need way fewer threads, at least in theory.  This change is to change the startup number of threads only, but the thread pool will increase to 256 if necessary.

### DIFF
--- a/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
+++ b/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
@@ -29,6 +29,7 @@ import com.google.cloud.bigtable.hbase.BigtableRegionLocator;
 import com.google.cloud.bigtable.hbase.BigtableTable;
 import com.google.cloud.bigtable.hbase.adapters.HBaseRequestAdapter;
 import com.google.common.base.MoreObjects;
+import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.google.common.util.concurrent.MoreExecutors;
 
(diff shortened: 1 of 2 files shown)


bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -324,6 +325,14 @@ private ExecutorService getBatchPool() {
     return this.batchPool;
   }
 
+  private int getMinThreads() {
```
kevinsi4508: min threads or core threads?

bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -304,6 +304,7 @@ private ExecutorService getBatchPool() {
       // shared HTable thread executor not yet initialized
       synchronized (this) {
         if (batchPool == null) {
+          int minThreads = getMinThreads();
```
kevinsi4508: check minThreads <= maxThreads?

sduskis: I used Math.min to solve this problem. I don't know that it's worth spending that much time and effort to do better, but I will if you think it's important.

kevinsi4508: Math.min looks good!

kevinsi4508:
LGTM

Issue #535, opened 2015-10-22
ResultScannerBlocking


benjumanji:
request for method:

Could we add a `bool mightBlock()` to the ResultScanner?

Why? I would like to be able to wrap the scanner with something like a `Spool[T]` from twitter util which is essentially a linked list with the tail as `Future[T]`. The trick to not making this ungodly efficient is to only make a new task to be deferred only when one needs to. An example: 

https://github.com/websudos/phantom/blob/develop/phantom-dsl/src/main/scala/com/websudos/phantom/iteratee/ResultSpool.scala

(NB: I am actually the author of that code see [here](https://github.com/websudos/phantom/pull/191/files))

I think for the current impl backed by a blocking queue this is as simple as adding a method that calls `isEmpty()` on the underlying queue. I'd be happy to contribute a patch if you guys are ok with the idea in principle.

sduskis:
+1 on this.  This will help with some other issues we're having relating to long running scans.

PR #537, opened 2015-10-22
Implement available on RowScanner

benjumanji:
As referenced in #535:

I changed my mind a little, obviously `mayBlock()` is just `available() == 0`. I can add both, I just figured the more general method is more interesting.

It compiles and tests seem to pass, so that's nice. What I'm not sure about is how to test _my_ implementation. I think it's trivially correct given the calls to the underlying collections, but tests would be nice.

I'll have a think about it tomorrow. I also checked the impl of `size()` for LinkedBlockingQueue's: it's just a [read on an atomic int](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/util/concurrent/LinkedBlockingQueue.java#LinkedBlockingQueue.size%28%29), at least in openjdk, so shouldn't be too costly.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResultScanner.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResultScanner.java
@@ -4,9 +4,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -35,4 +35,10 @@ public interface ResultScanner<T> extends Closeable {
    * @param count The number of rows to read.
    */
   T[] next(int count) throws IOException;
+
+  /**
+   * Check number of rows immediately available. Calls to {@link #next()} will not block on network for at least
+   * n results.
+   */
+  int available();
 }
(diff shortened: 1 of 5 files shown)


benjumanji:
I signed it, make me a winner!

sduskis:
Would it make more sense in your case to simply make a new ClientCall to BigtableServiceGrpc.METHOD_READ_ROWS with your own StreamObserver?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResultScanner.java
```diff
@@ -35,4 +35,10 @@
    * @param count The number of rows to read.
    */
   T[] next(int count) throws IOException;
+
+  /**
+   * Check number of rows immediately available. Calls to {@next()} will not block on network for at least
```
sduskis: {@next()} probably should be {@link #next()}

sduskis:
Please add a test to StreamingBigtableResultScannerTest

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResultScanner.java
benjumanji: Done.

benjumanji:
Yeah, probably going in at the stream observer is the better approach, I'm just being a little lazy. I still need to write my ultimate scala wrapper<sup>tm</sup>  . I'll add a test tomorrow.

benjumanji:
Test added, seems a bit cheesy. PTAL.

sduskis:
LGTM.  Please squash the commits, and I'll merge.  Sorry about the cheesiness.  It will help us maintain the feature in the long term.

To squash commits: https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/CONTRIBUTING.md#squashing-commits

benjumanji:
Done!

PR #539, opened 2015-10-23
Improving performance through zero copy bytes mechanism.

sduskis:
Adding BigtableZeroCopyByteString to Get/Put/Row/Scan adapters.
Improved managing of families in RowMerger.

### DIFF
--- a/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/GetAdapter.java
+++ b/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/GetAdapter.java
@@ -17,10 +17,10 @@ package com.google.cloud.bigtable.hbase.adapters;
 
 import com.google.bigtable.v1.ReadRowsRequest;
 import com.google.bigtable.v1.ReadRowsRequest.Builder;
-import com.google.protobuf.ByteString;
 
 import org.apache.hadoop.hbase.client.Get;
 import org.apache.hadoop.hbase.client.Scan;
+import org.apache.hadoop.hbase.util.ByteStringer;
 
 /**
  * A Get adapter that transform the Get into a ReadRowsRequest using the proto-based
@@ -39,6 +39,6 @@ public class GetAdapter implements ReadOperationAdapter<Get> {
     scanAdapter.throwIfUnsupportedScan(operationAsScan);
     return ReadRowsRequest.newBuilder()
         .setFilter(scanAdapter.buildFilter(operationAsScan, readHooks))
-        .setRowKey(ByteString.copyFrom(operation.getRow()));
+        .setRowKey(ByteStringer.wrap(operation.getRow()));
   }
 }
--- a/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/ScanAdapter.java
+++ b/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/ScanAdapter.java
@@ -31,9 +31,9 @@ import com.google.protobuf.ByteString;
 
 import org.apache.hadoop.hbase.client.Scan;
 import org.apache.hadoop.hbase.io.TimeRange;
+import org.apache.hadoop.hbase.util.ByteStringer;
 
 import java.io.IOException;
-import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.NavigableSet;
 
@@ -90,13 +90,15 @@ public class ScanAdapter implements ReadOperationAdapter<Scan> {
   @Override
   public Builder adapt(Scan scan, ReadHooks readHooks) {
     throwIfUnsupportedScan(scan);
+    // For gets, startRow == stopRow.  There's no need to create a new ByteString for stopRow
+    RowFilter filter = buildFilter(scan, readHooks);
 
     return ReadRowsRequest.newBuilder()
-        .setFilter(buildFilter(scan, readHooks))
+        .setFilter(filter)
         .setRowRange(
             RowRange.newBuilder()
-                .setStartKey(ByteString.copyFrom(scan.getStartRow()))
-                .setEndKey(ByteString.copyFrom(scan.getStopRow())));
+                .setStartKey(ByteStringer.wrap(scan.getStartRow()))
+                .setEndKey(ByteStringer.wrap(scan.getStopRow())));
   }
 
   private static byte[] quoteRegex(byte[] unquoted)  {
@@ -120,15 +122,14 @@ public class ScanAdapter implements ReadOperationAdapter<Scan> {
   private RowFilter createColumnQualifierFilter(byte[] unquotedQualifier) {
     return RowFilter.newBuilder()
         .setColumnQualifierRegexFilter(
-            ByteString.copyFrom(
-                quoteRegex(unquotedQualifier)))
+          ByteStringer.wrap(quoteRegex(unquotedQualifier)))
         .build();
   }
 
   private RowFilter createFamilyFilter(byte[] familyName) {
     return RowFilter.newBuilder()
-        .setFamilyNameRegexFilter(
-            new String(quoteRegex(familyName), StandardCharsets.UTF_8))
+        .setFamilyNameRegexFilterBytes(
+            ByteStringer.wrap(quoteRegex(familyName)))
         .build();
   }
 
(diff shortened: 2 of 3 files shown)


kevinsi4508:
Sorry! I am not very familiar with BigtableZeroCopyByteString. Can you explain a little more on how it would help improving the performance?
Is "Improved managing of families in RowMerger." related to BigtableZeroCopyByteString? If not, can you break it to 2 PRs? Thanks!

sduskis:
BigtableZeroCopyByteString is a performance enhancement that eliminates new byte[] creation from Protobuf ByteStrings.  It's somewhat hacky, but will work in our hbase case.  It's dangerous when the protobuf will be used in arbitrary ways and the byte[] can be manipulated.  By default, Protobuf will create a new byte[] and copy an existing byte[] into it for safety purposes.  Here's the uses of hbase's version: https://github.com/apache/hbase/search?utf8=%E2%9C%93&q=ByteStringer&type=Code

I'll submit the RowMerger thing separately.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RowMerger.java
```diff
@@ -107,28 +112,54 @@ public boolean isRowCommitted() {
...
   public Row buildRow() {
     Preconditions.checkState(committed,
-        "Cannot build a Row object if we have not yet encountered a COMMIT_ROW chunk.");
-    Row.Builder currentRowBuilder = Row.newBuilder();
-    currentRowBuilder.setKey(currentRowKey);
-    for (Family.Builder builder : familyMap.values()) {
-      currentRowBuilder.addFamilies(builder.build());
+      "Cannot build a Row object if we have not yet encountered a COMMIT_ROW chunk.");
+    Row.Builder rowBuilder = Row.newBuilder().setKey(currentRowKey);
+    if (!families.isEmpty()) {
```
kevinsi4508: I agree that we should merge families at the end, but why looping through families multiple times?
Can we do following?
Map<String, <Family.Builder> family_builder_map ...
for (Family family : families) {
  Family.Builder builder = family_builder_map.get(family.getName());
  if (builder == null) {
    builder = Family.newBuilder().setName(family.getName());
  }
  builder.addFamilies(family.getColumnsList());
}

bigtable-client-core/src/test/java/com/google/protobuf/BigtableZeroCopyByteString.java
```diff
@@ -0,0 +1,65 @@
...
+ * <p>
+ * Without this protobufs would force us to copy every single byte array out
+ * of the objects de-serialized from the wire (which already do one copy, on
+ * top of the copies the JVM does to go from kernel buffer to C buffer and
+ * from C buffer to JVM buffer).
+ * <p>
+ * See HBase's HBaseZeroCopyByteString for more information.
+ */
+@SuppressWarnings("serial")
+public class BigtableZeroCopyByteString  extends LiteralByteString {
```
kevinsi4508: this is in: "bigtable-client-core/src/test/java/com/google/protobuf/BigtableZeroCopyByteString.java"

bigtable-client-core/src/test/java/com/google/protobuf/BigtableZeroCopyByteString.java
```diff
@@ -0,0 +1,65 @@
...
+  // Gotten from AsyncHBase code base with permission.
+  /** Private constructor so this class cannot be instantiated. */
+  private BigtableZeroCopyByteString() {
+    super(null);
+    throw new UnsupportedOperationException("Should never be here.");
```
kevinsi4508: this tells me that BigtableZeroCopyByteString should not  extends LiteralByteString.

bigtable-client-core/src/test/java/com/google/protobuf/BigtableZeroCopyByteString.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  /**
+   * Wraps a byte array in a {@link ByteString} without copying it.
+   */
+  public static ByteString wrap(final byte[] array) {
+    return new LiteralByteString(array);
```
kevinsi4508: Where is this LiteralByteString class?

bigtable-client-core/src/test/java/com/google/protobuf/BigtableZeroCopyByteString.java
```diff
@@ -0,0 +1,65 @@
...
+  /** Private constructor so this class cannot be instantiated. */
+  private BigtableZeroCopyByteString() {
+    super(null);
+    throw new UnsupportedOperationException("Should never be here.");
+  }
+
+  /**
+   * Wraps a byte array in a {@link ByteString} without copying it.
+   */
+  public static ByteString wrap(final byte[] array) {
```
kevinsi4508: toByteString()?

bigtable-client-core/src/test/java/com/google/protobuf/BigtableZeroCopyByteString.java
```diff
@@ -0,0 +1,65 @@
...
+  // TODO:
+  // ZeroCopyLiteralByteString.wrap(this.buf, 0, this.count);
+  /**
+   * Extracts the byte array from the given {@link ByteString} without copy.
+   * @param buf A buffer from which to extract the array. This buffer must be actually an instance
+   *          of a {@code LiteralByteString}.
+   */
+  public static byte[] zeroCopyGetBytes(final ByteString buf) {
+    if (buf instanceof LiteralByteString) {
+      return ((LiteralByteString) buf).bytes;
```
kevinsi4508: This is very dangerous. The representation of bytes in subclass of LiteralByteString can be different. It could be substring of a byte[] in BoundedByteString, in that case, we should not return the whole byte[].

bigtable-client-core/src/test/java/com/google/protobuf/BigtableZeroCopyByteString.java
sduskis: It's a class in com.google.protobuf.  It has package level scope.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/ScanAdapter.java
```diff
@@ -90,13 +90,22 @@ public RowFilter buildFilter(Scan scan, ReadHooks hooks) {
   @Override
   public Builder adapt(Scan scan, ReadHooks readHooks) {
     throwIfUnsupportedScan(scan);
+    byte[] startRowBytes = scan.getStartRow();
+    byte[] stopRowBytes = scan.getStopRow();
+
+    ByteString startRow = ByteStringer.wrap(startRowBytes);
+    // For gets, startRow == stopRow.  There's no need to create a new ByteString for stopRow
+    ByteString stopRow =
+        startRowBytes == stopRowBytes ? startRow : ByteStringer.wrap(stopRowBytes);
```
kevinsi4508: compare object instead of reference?

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/GetAdapter.java
```diff
@@ -39,6 +39,6 @@ public Builder adapt(Get operation, ReadHooks readHooks) {
     scanAdapter.throwIfUnsupportedScan(operationAsScan);
     return ReadRowsRequest.newBuilder()
         .setFilter(scanAdapter.buildFilter(operationAsScan, readHooks))
-        .setRowKey(ByteString.copyFrom(operation.getRow()));
+        .setRowKey(ByteStringer.wrap(operation.getRow()));
```
kevinsi4508: missing ByteStringer class?

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/ScanAdapter.java
sduskis: reference is important here... Get operations use the same reference for start/stop.  If we do comparisons, we hit a performance issue, so I'd rather not do that.  I'll document the intent.

kevinsi4508: Yes, do document it. expecting same reference, etc.

kevinsi4508:
LGTM

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/GetAdapter.java
sduskis: It's an HBase class.

PR #544, opened 2015-10-26
Use existing pools in BufferedMutator.

manolama:
This modifies AbstractBigtableConnection.getBufferedMutator() to use the executor service given in the params if present. Otherwise it will use the client's batch pool. Previously it created a separate thread pool per mutator that could result in a ton of unecessary threads.
Also start a unit test class for the AbstractBigTableConnection and rename the existing TestBigtableConnection to TestBigtableConnectionImplementation (since it extends the abstract class).

### DIFF
--- a/bigtable-hbase/src/test/java/org/apache/hadoop/hbase/client/TestBigtableConnection.java
+++ b/bigtable-hbase/src/test/java/org/apache/hadoop/hbase/client/TestBigtableConnection.java
@@ -1,0 +1,170 @@
+/*
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.hbase.client;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertSame;
+
+import java.util.concurrent.ExecutorService;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.AbstractBigtableConnection;
+import org.apache.hadoop.hbase.client.Connection;
+import org.junit.Before;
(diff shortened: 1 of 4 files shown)


manolama:
BTW, what's Google's policy on using PowerMockito in unit tests for projects? It makes it a lot easier to mock and test some things without having to create ``@VisibleForTesting`` methods.

manolama:
Also any thoughts on moving ``ACTIVE_BUFFERED_MUTATORS`` to a member object instead of a static global? I think it'd be nicer in that we can iterate and close the buffered mutators when the connection is closed.

sduskis:
Re: using the We can't use the existing pool for the BufferedMutator type management operations.  The pool ought to be reserved for the gRPC operations.  Did you see problems?

sduskis:
In terms of PowerMockito, I think it's ok given my limited research into it.

sduskis:
We need a variant of ACTIVE_BUFFERED_MUTATORS that's static.  We need to give meaningful logs when the JVM is shut down.  We've followed the lead of HBase.  They don't close mutators or tables.

manolama:
Thanks! Yeah I saw a thread explosion on my host writing around 3k "puts" per second where we climbed up to around 300 threads. How about I just tweak it so that if the pool is there then we use it, otherwise create one with a better name?

sduskis:
Ah, ok.  I think we can rejigger the BufferedMutator logic to use a single thread.  Please don't use the batch pool for this.

manolama:
> Ah, ok. I think we can rejigger the BufferedMutator logic to use a single thread. Please don't use the batch pool for this.

Cool, take a look at this update in the mean time and see if it's ok please.

sduskis:
I removed the need for any ExecutorService in BigtableBufferedMutator in #548.  Does that suffice?

bigtable-hbase/src/test/java/org/apache/hadoop/hbase/client/TestBigtableConnection.java
```diff
@@ -0,0 +1,170 @@
...
+import com.google.cloud.bigtable.grpc.BigtableDataClient;
+import com.google.cloud.bigtable.grpc.BigtableSession;
+import com.google.cloud.bigtable.grpc.async.AsyncExecutor;
+import com.google.cloud.bigtable.hbase.BigtableBufferedMutator;
+import com.google.cloud.bigtable.hbase.BigtableOptionsFactory;
+import com.google.cloud.bigtable.hbase.TestBigtableConnectionImplementation;
+
+@RunWith(JUnit4.class)
+public class TestBigtableConnection {
+  private static final String PROJECT_ID = "unittestpid";
```
sduskis: We can't check anything in with specific projects/clusters.

manolama:
Ah ok, #548 is definitely better, thanks!

And to note before I forget, you may want to give the option to provide an optional executor for RPC callbacks in ``AsyncExecutor.call()`` so that long running callbacks don't tie up a channel's thread. I've seen this in AsyncHBase and I'm trying a few fixes.

PR #546, opened 2015-10-27
Adding zeroCopy operation for RowAdapter.

sduskis:


bigtable-hbase/src/main/java/com/google/protobuf/BigtableZeroCopyByteStringUtil.java
```diff
@@ -0,0 +1,35 @@
...
+public class BigtableZeroCopyByteStringUtil {
+
+  /**
+   * Extracts the byte array from the given {@link ByteString} without copy.
+   * @param buf A buffer from which to extract the array.  This buffer must be
+   * actually an instance of a {@code LiteralByteString}.
+   */
+  public static byte[] zeroCopyGetBytes(final ByteString buf) {
+    if (buf.getClass() == LiteralByteString.class) {
+      return ((LiteralByteString) buf).bytes;
```
kevinsi4508: where do we use LiteralByteString for row key, qualifier, and cell value?

sduskis: That happens under the covers in Protobuf classes.

bigtable-client-core/src/main/java/com/google/protobuf/BigtableZeroCopyByteString.java
```diff
@@ -33,19 +33,16 @@ public static ByteString wrap(final byte[] array, int offset, int length) {
...
-  // ZeroCopyLiteralByteString.wrap(this.buf, 0, this.count);
-
   /**
    * Extracts the byte array from the given {@link ByteString} without copy.
    * @param buf A buffer from which to extract the array.  This buffer must be
    * actually an instance of a {@code LiteralByteString}.
    */
   public static byte[] zeroCopyGetBytes(final ByteString buf) {
-    if (buf instanceof LiteralByteString) {
+    if (buf.getClass() == LiteralByteString.class) {
```
kevinsi4508: What about BoundedByteString? Do we care?

bigtable-client-core/src/main/java/com/google/protobuf/BigtableZeroCopyByteString.java
```diff
@@ -33,19 +33,16 @@ public static ByteString wrap(final byte[] array, int offset, int length) {
     return new BoundedByteString(array, offset, length);
   }
 
-  // TODO:
-  // ZeroCopyLiteralByteString.wrap(this.buf, 0, this.count);
-
   /**
    * Extracts the byte array from the given {@link ByteString} without copy.
    * @param buf A buffer from which to extract the array.  This buffer must be
    * actually an instance of a {@code LiteralByteString}.
```
kevinsi4508: cannot be subclass of LiteralByteString, e.g. BoundedByteString

sduskis: toByteArray() will be called on any implementation other than LiteralByteString, included BoundedByteString.

bigtable-client-core/src/main/java/com/google/protobuf/BigtableZeroCopyByteString.java
sduskis: toByteArray() will be called on any implementation other than LiteralByteString, included BoundedByteString

bigtable-client-core/src/main/java/com/google/protobuf/BigtableZeroCopyByteString.java
```diff
@@ -8,10 +8,10 @@
  * of the objects de-serialized from the wire (which already do one copy, on
  * top of the copies the JVM does to go from kernel buffer to C buffer and
  * from C buffer to JVM buffer).
- *
- * @since 0.96.1
  */
 public final class BigtableZeroCopyByteString extends LiteralByteString {
```
kevinsi4508: I am confused. Why this class extends LiteralByteString? I asked this question before.

bigtable-client-core/src/main/java/com/google/protobuf/BigtableZeroCopyByteString.java
kevinsi4508: understood, please update the method doc.

bigtable-client-core/src/main/java/com/google/protobuf/BigtableZeroCopyByteString.java
sduskis: This class has the basic function of efficient copying of byte[] to and from protobufs.  

Protobufs use ByteString to wrapy byte[]. We need an efficient way to write byte[] into our protobuf objects, and that's where our BigtableZeroCopyByteString comes in.  We can't instantiate a LiteralByteString directly, since by default, protobuf doesn't allow you to create the more efficient LiteralByteString directly; rather, it makes you copy the entire byte[] via ByteString.copyOf() which copies the byte[] and then constructs a new LiteralByteString.

We need new BigtableZeroCopyByteString when writing to protobuf objects and zeroCopyGetBytes when reading from protobuf objects.

sduskis: You're right.  It doesn't.  I copied the code from HBase, and I didn't pay too much attention.  Good catch.

bigtable-client-core/src/main/java/com/google/protobuf/BigtableZeroCopyByteStringUtil.java
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  /**
+   * Extracts the byte array from the given {@link ByteString} without copy.
+   * @param buf A buffer from which to extract the array. This buffer must be an instance of a
+   *          {@code LiteralByteString} in order to be efficient. {@link ByteString#toByteArray()}
+   *          will be called All other implementations, including subclasses like
+   *          {@link BoundedByteString}
+   */
+  public static byte[] zeroCopyGetBytes(final ByteString buf) {
```
kevinsi4508: possible zero copy :D this fine.

kevinsi4508:
LGTM

PR #547, opened 2015-10-27
Calculating milliseconds more efficiently in RowAdapter.

sduskis:


bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/RowAdapter.java
```diff
@@ -32,6 +32,9 @@
  * Adapt a bigtable.v1.Row to an hbase client Result.
  */
 public class RowAdapter implements ResponseAdapter<Row, Result> {
+  static final int TIME_CONVERSION_UNIT = (int) BigtableConstants.BIGTABLE_TIMEUNIT.convert(1,
```
kevinsi4508: Make a note that this only works because BIGTABLE_TIMEUNIT is smaller than HBASE_TIMEUNIT, otherwise we will get 0.

Why do we need to cast to int, why not leave it as long?

sduskis: ints are more efficient in terms of java multiplication

sduskis: Updated back to long.

kevinsi4508:
Please a comment to explain why this conversion unit works.
something like: Make a note that this only works because BIGTABLE_TIMEUNIT is smaller than HBASE_TIMEUNIT, otherwise we will get 0.

LGTM

PR #549, opened 2015-10-28
Using an extra call.request() in streaming readRows.

sduskis:
This should improve throughput.


AngusDavis:
When the blocking queue was added, there was no flow control from gRPC. There is now (in the form of call.request). Remove the blocking queue and replace it with a concurrent queue and every time a partial response is read in the scan side, invoke call.request(N) (or something similar).

Wouldn't you want the initial buffer size to be the maximum number of messages you want buffered before the flow-control window is closed and then remove the automatic call.request(1) from the observer?

sduskis:
Closing this in favor of a different solution, as per Angus.

PR #550, opened 2015-10-28
Using ByteStringer.wrap in more places.

sduskis:
Also removing ByteStringer.wrap() that takes in an array, offset and size; ByteString.copy() is more efficient in that case.


AngusDavis:
What are the safety guarantees now? Before wrapping, the guarantee was that as soon as the Table methods returned, any inputs were no longer needed and could be modified  (e.g., batch calls w/ callback, scan, etc). I don't believe that is now the case. I'm not certain about the HBase client guarantees and this may be compliant, but wanted to bring it up.

PR #551, opened 2015-10-29
Improving use of ClientCall.request() for scanning.

sduskis:
Calling ClientCall.request(capacity) to request the maximum number of requests from the server at the beginning of a scan.  Calling clientCall.request(1) for every response that's handled by ResponseQueueReader so that the server can send an additional response, since there is now more room in the queue.
Also removing a lot of the tests from StreamingBigtableResultScannerTest.  It no longer needs to test the functionality enacpsulated by a ResponseQueueReader.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
@@ -17,8 +17,9 @@ package com.google.cloud.bigtable.grpc;
 
 import io.grpc.CallOptions;
 import io.grpc.ClientCall;
+import io.grpc.Metadata;
 import io.grpc.MethodDescriptor;
-import io.grpc.stub.StreamObserver;
+import io.grpc.Status;
 
 import java.util.List;
 import java.util.concurrent.ExecutorService;
@@ -46,6 +47,7 @@ import com.google.cloud.bigtable.grpc.io.ChannelPool.PooledChannel;
 import com.google.cloud.bigtable.grpc.io.ClientCallService;
 import com.google.cloud.bigtable.grpc.io.RetryingCall;
 import com.google.cloud.bigtable.grpc.scanner.BigtableResultScannerFactory;
+import com.google.cloud.bigtable.grpc.scanner.ResponseQueueReader;
 import com.google.cloud.bigtable.grpc.scanner.ResultScanner;
 import com.google.cloud.bigtable.grpc.scanner.ResumingStreamingResultScanner;
 import com.google.cloud.bigtable.grpc.scanner.StreamingBigtableResultScanner;
@@ -60,33 +62,6 @@ import com.google.protobuf.ServiceException;
  * A gRPC client to access the v1 Bigtable service.
  */
 public class BigtableDataGrpcClient implements BigtableDataClient {
-
-  /**
-   * A StreamObserver that sends results to a StreamingBigtableResultScanner.
-   */
-  public static class ReadRowsStreamObserver
-      implements StreamObserver<ReadRowsResponse> {
-    private final StreamingBigtableResultScanner scanner;
-
-    public ReadRowsStreamObserver(StreamingBigtableResultScanner scanner) {
-      this.scanner = scanner;
-    }
-
-    @Override
-    public void onNext(ReadRowsResponse readTableResponse) {
-      scanner.addResult(readTableResponse);
-    }
-
-    @Override
-    public void onError(Throwable throwable) {
-      scanner.setError(throwable);
-    }
-
-    @Override
-    public void onCompleted() {
-      scanner.complete();
-    }
-  }
 
   @VisibleForTesting
   public static final Predicate<MutateRowRequest> IS_RETRYABLE_MUTATION =
@@ -267,18 +242,42 @@ public class BigtableDataGrpcClient implements BigtableDataClient {
       }
     }, executorService);
 
+    int timeout = retryOptions.getReadPartialRowTimeoutMillis();
+
+    int streamingBufferSize = retryOptions.getStreamingBufferSize();
+    int batchRequestSize = streamingBufferSize / 2;
+    ResponseQueueReader responseQueueReader =
+        new ResponseQueueReader(timeout, streamingBufferSize, batchRequestSize,
+            batchRequestSize, readRowsCall);
+
     StreamingBigtableResultScanner resultScanner =
-        new StreamingBigtableResultScanner(
-          reservedChannel,
-          retryOptions.getStreamingBufferSize(),
-          retryOptions.getReadPartialRowTimeoutMillis(),
-          cancellationToken);
+        new StreamingBigtableResultScanner(reservedChannel, responseQueueReader, cancellationToken);
 
     clientCallService.asyncServerStreamingCall(
         readRowsCall,
         request,
-        new ReadRowsStreamObserver(resultScanner));
+        batchRequestSize,
+        createClientCallListener(resultScanner));
 
     return resultScanner;
+  }
+
+  private ClientCall.Listener<ReadRowsResponse> createClientCallListener(
+      final StreamingBigtableResultScanner resultScanner) {
+    return new ClientCall.Listener<ReadRowsResponse>() {
+      @Override
+      public void onMessage(ReadRowsResponse readRowResponse) {
+        resultScanner.addResult(readRowResponse);
+      }
+
+      @Override
+      public void onClose(Status status, Metadata trailers) {
+        if (status.isOk()) {
+          resultScanner.complete();
+        } else {
+          resultScanner.setError(status.asRuntimeException());
+        }
+      }
+    };
   }
 }
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ClientCallService.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ClientCallService.java
@@ -4,9 +4,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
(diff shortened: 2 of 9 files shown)


sduskis:
FWIW, I did a simple test with this change and with the current state.  The current state got up to 35K rows per second.  This change got upwards of 50K.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -267,18 +242,40 @@ public void run() {
       }
     }, executorService);
 
+    int bufferCapacity = retryOptions.getStreamingBufferSize();
```
kevinsi4508: nit: move this down to clientCallService.asyncServerStreamingCall() where it is used.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ClientCallService.java
```diff
@@ -32,6 +35,8 @@
   <ReqT, RespT> void asyncServerStreamingCall(ClientCall<ReqT, RespT> call,
```
kevinsi4508: Is this still needed?

sduskis: Yup.  It's used elsewhere, specifically for sample row key streaming.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResponseQueueReader.java
```diff
@@ -86,9 +88,12 @@ public synchronized Row getNextMergedRow() throws IOException {
     if (queueEntry == null) {
       throw new ScanTimeoutException("Timeout while merging responses.");
     }
+    if (!queueEntry.isCompletionMarker()) {
+      call.request(1);
```
kevinsi4508: this means, when the capacity is N and the response consumption is quick we will have about ~N active requests. Could it be a problem? Also at the end, we will have N -1 requests return no result.

sduskis:
The gRPC team suggested that calling ClientCall.request() from a different thread, like this PR does, could cause performance problems.  Closing this PR and trying again.

sduskis:
@kevinsi4508, @AngusDavis, Nathan and I came up with an enhancement for this change.  We need to retrieve data in batch.

sduskis:
I updated the code as per our offline discussion.  The ResponseQueueReader now batches requests instead of sending clientCall.request(1).

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -267,18 +242,40 @@ public void run() {
...
 
     return resultScanner;
   }
+
+  private ClientCall.Listener<ReadRowsResponse> createClientCallListener(
+      final StreamingBigtableResultScanner resultScanner) {
+    return new ClientCall.Listener<ReadRowsResponse>() {
+      @Override
+      public void onMessage(ReadRowsResponse readRowResponse) {
+        resultScanner.addResult(readRowResponse);
```
kevinsi4508: instead of what u have, can we do this:

  private ClientCall.Listener<ReadRowsResponse> createClientCallListener(
      final StreamingBigtableResultScanner resultScanner, final ClientCall<?, ReadRowsResponse> call) {
    return new ClientCall.Listener<ReadRowsResponse>() {
      @Override
      public void onMessage(ReadRowsResponse readRowResponse) {
        resultScanner.addResult(readRowResponse);
        // if (not end of result) {
        //   call.request(x) where x = capacity - available;
        // }
        call.request(x);
      }

sduskis: Not really.  The idea is to maintain this:
  queueSize + outstanding requests > capacity - call request size.

When that is false, request more.

In onMessage, queueSize is incremented by 1 and outstanding requests is decremented by 1 making the sum the same as it was before.

It would seem to me that we should only request more when objects are popped off the queue.

kevinsi4508: First, can this be done in onMessage()? I believe it is the right place to have this logic.
Second, I think it is OK to request x = capacity - available.
If x increases, that means user thread is hungry for more, so we should increase the batch size.
It starts with 1, as user threads consumes more quickly, x increases too to match the consumption speed.

The response will be blocked if the queue is full, it will automatically stop further request. We could keep track of pending request size here too, but I don't think it is necessary.

sduskis: I think we need to request x when:
(capacity - (queue size + outstanding requests) < x) && !(finished)

As you said, the most likely place for that formula to change is when the user thread is hungry for more.  I would think that the most likely place to determine if a "user thread is hungry for more" is on the user thread, not the producer/onMessage thread.

Consider this scenario.  What happens if onMessage is called enough times to fill the queue before the user thread can pull things off?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ClientCallService.java
```diff
@@ -44,29 +49,52 @@
...
+      call.start(listener, new Metadata());
+      call.request(1);
+      try {
+        call.sendMessage(request);
+        call.halfClose();
+      } catch (Throwable t) {
+        call.cancel();
+        throw Throwables.propagate(t);
+      }
+      call.request(initialRequestCount - 1);
```
kevinsi4508: Can we make sure initialRequestCount - 1 >= 1?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -267,18 +242,42 @@ public void run() {
       }
     }, executorService);
 
+    int timeout = retryOptions.getReadPartialRowTimeoutMillis();
+
+    int streamingBufferSize = retryOptions.getStreamingBufferSize();
+    int batchRequestSize = streamingBufferSize / 2;
+    ResponseQueueReader responseQueueReader =
+        new ResponseQueueReader(timeout, streamingBufferSize, streamingBufferSize,
```
kevinsi4508: why setting outstandingRequestCount to streamingBufferSize?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResponseQueueReader.java
```diff
@@ -66,16 +80,25 @@ public synchronized Row getNextMergedRow() throws IOException {
...
 
   private ResultQueueEntry<ReadRowsResponse> getNext() throws IOException {
+    int unprocessedRequestCount = outstandingRequestCount.get() + resultQueue.size();
+
+    // If there are currently less than or equal to half of the capacityCap, then ask gRPC to
+    // request more results in a batch. Batch requests are more efficient that reading one at
+    // a time.
+    while (!completionMarkerFound.get()
+        && unprocessedRequestCount <= capacityCap - batchRequestSize) {
+      call.request(batchRequestSize);
```
kevinsi4508: I feel like we should still make a request if !completionMarkerFound.get().

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResponseQueueReader.java
```diff
@@ -66,16 +80,25 @@ public synchronized Row getNextMergedRow() throws IOException {
...
+      "End of stream marker encountered while merging a row.");
+    Preconditions.checkState(lastResponseProcessed,
+      "Should only exit merge loop with by returning a complete Row or hitting end of stream.");
     return null;
   }
 
   private ResultQueueEntry<ReadRowsResponse> getNext() throws IOException {
+    int unprocessedRequestCount = outstandingRequestCount.get() + resultQueue.size();
+
+    // If there are currently less than or equal to half of the capacityCap, then ask gRPC to
```
kevinsi4508: Update doc

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResponseQueueReader.java
```diff
@@ -66,16 +80,25 @@ public synchronized Row getNextMergedRow() throws IOException {
...
-        "Should only exit merge loop with by returning a complete Row or hitting end of stream.");
+    Preconditions.checkState(builder == null,
+      "End of stream marker encountered while merging a row.");
+    Preconditions.checkState(lastResponseProcessed,
+      "Should only exit merge loop with by returning a complete Row or hitting end of stream.");
     return null;
   }
 
   private ResultQueueEntry<ReadRowsResponse> getNext() throws IOException {
+    int unprocessedRequestCount = outstandingRequestCount.get() + resultQueue.size();
```
kevinsi4508: either get rid of this variable or use a better name.

bigtable-hbase-integration-tests/pom.xml
```diff
@@ -339,6 +339,14 @@ limitations under the License.
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
```
kevinsi4508: Is this part of the PR? for what reason?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResponseQueueReader.java
sduskis: if completionMarkerFound, it means that the server indicated that there is no more data.  There's no need to read more.

kevinsi4508: when !completionMarkerFound.get(), request(1).

sduskis:

kevinsi4508:
LGTM

PR #552, opened 2015-10-29
Removing ByteStringer.wrap() around partial byte arrays.

sduskis:
The ByteString.copyFrom() turns out to be more efficient.  Removing the ByteStringer.wrap(byte[], int, int) and replacing its uses with BytString.copyFrom(byte[], int, int).


kevinsi4508:
LGTM

PR #553, opened 2015-10-29
Using an extra call.request() in streaming readRows.

sduskis:
This should improve throughput.

gRPC suggested that this is a better way than the method proposed in #549 by @AngusDavis.  We need to call clientCall.request() on the same thread that the Listener handles the client call.


sduskis:
retrying #551

PR #555, opened 2015-11-02
Using gRPC 0.10.0-SNAPSHOT for forTarget().

sduskis:
forTarget() is more abstract that forAddress(), since it will allow gRPC to reconnect when the network topology changes.


sduskis:
This didn't fix the problem.  Closing.

PR #557, opened 2015-11-02
Fixing issue with integration test project pom.xml.

sduskis:
Currently, the integration test project will not work well with eclipse
due to the default behavior of the eclipse plugin (project based
dependency) not working well with our shading plugin.  Also, cleaning up
old properties from bigtable-test.xml


kevinsi4508:
LGTM

PR #561, opened 2015-11-03
Replacing HBase's KeyValue Cell implementation with a new one.

sduskis:
KeyValue works great for transporting HBase data back and forth, since they use it as both a runtime mechanism and an efficient transport mechanism.  They write the KeyValue object once, and then use it many times.  It's very efficient when used that way.  It's not as effective for us, since we don't use it as a transport mechanism.  This change seems to reduce CPU cost by a significant percentage from microseconds to hundreds of nanoseconds per row.

### DIFF
--- a/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/RowAdapter.java
+++ b/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/RowAdapter.java
@@ -64,7 +64,7 @@ public class RowAdapter implements ResponseAdapter<Row, Result> {
           // cells are deduped unintentionally here. On the other hand, if we don't dedup them,
           // HBase will treat them as duplicates.
           long hbaseTimestamp = cell.getTimestampMicros() / TIME_CONVERSION_UNIT;
-          KeyValue keyValue = new KeyValue(
+          RowCell keyValue = new RowCell(
               rowKey,
               familyNameBytes,
               columnQualifier,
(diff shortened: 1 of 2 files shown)


kevinsi4508:
Still not sure what is the difference between KeyValue and RowCell. Inventing something like RowCell makes me nervous.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/RowAdapter.java
```diff
@@ -64,7 +64,7 @@ public Result adaptResponse(Row response) {
           // cells are deduped unintentionally here. On the other hand, if we don't dedup them,
           // HBase will treat them as duplicates.
           long hbaseTimestamp = cell.getTimestampMicros() / TIME_CONVERSION_UNIT;
-          KeyValue keyValue = new KeyValue(
+          RowCell keyValue = new RowCell(
```
kevinsi4508: does it work well with KeyValue.COMPARATOR?

sduskis: Yup.  KeyValue.COMPARATOR uses the Cell interface.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/RowCell.java
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  @Override
+  public int getValueLength() {
+    return this.valueArray.length;
+  }
+
+  @Override
+  public byte[] getTagsArray() {
+    return new byte[0];
```
kevinsi4508: return HConstants.EMPTY_BYTE_ARRAY; ?

kevinsi4508:
LGTM

PR #565, opened 2015-11-05
Using a batch size of 1 for Gets.

sduskis:
Gets will likely have better performance with smaller batch sizes.  Long
scans will likely have better performance with larger ones.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
@@ -232,10 +232,27 @@ public class BigtableDataGrpcClient implements BigtableDataClient {
   private ResultScanner<Row> streamRows(ReadRowsRequest request) {
     boolean isGet = request.getTargetCase() == ReadRowsRequest.TargetCase.ROW_KEY;
 
+    int streamingBufferSize;
+
+    int batchRequestSize;
+
     // If it's a long running scan, reserve the pool. There is a gRPC bug that makes the Channel
-    // unavailable for other streams. Reserve a channel for now. This can probably be removed once
-    // we use gRPC 0.10.0. See https://github.com/grpc/grpc-java/issues/1175 for more details
-    Channel channel = isGet ? channelPool : channelPool.reserveChannel();
+    // unavailable for other streams. Reserve a channel for now. Reservations can probably be
+    // removed once we use gRPC 0.10.0. See https://github.com/grpc/grpc-java/issues/1175 for more
+    // details
+    Channel channel;
+
+    if (isGet) {
+      // Batch request size is more performant with a value of 1 for single row gets, while a higher
+      // number is more performant for scanning
+      batchRequestSize = 1;
+      streamingBufferSize = 10;
+      channel = channelPool;
+    } else {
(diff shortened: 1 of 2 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -232,10 +232,27 @@ public Row readModifyWriteRow(ReadModifyWriteRowRequest request) {
   private ResultScanner<Row> streamRows(ReadRowsRequest request) {
     boolean isGet = request.getTargetCase() == ReadRowsRequest.TargetCase.ROW_KEY;
 
+    int streamingBufferSize = retryOptions.getStreamingBufferSize();
+
+    // Batch request size is more performant with a value of 1 for single row gets, while a higher
```
kevinsi4508: Move this comment to "batchRequestSize = 1;" ?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -232,10 +232,27 @@ public Row readModifyWriteRow(ReadModifyWriteRowRequest request) {
   private ResultScanner<Row> streamRows(ReadRowsRequest request) {
     boolean isGet = request.getTargetCase() == ReadRowsRequest.TargetCase.ROW_KEY;
 
+    int streamingBufferSize = retryOptions.getStreamingBufferSize();
```
kevinsi4508: Do we want to initialize streamingBufferSize here? we don't for batchRequestSize and channel.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -232,10 +232,27 @@ public Row readModifyWriteRow(ReadModifyWriteRowRequest request) {
...
-    // we use gRPC 0.10.0. See https://github.com/grpc/grpc-java/issues/1175 for more details
-    Channel channel = isGet ? channelPool : channelPool.reserveChannel();
+    // unavailable for other streams. Reserve a channel for now. Reservations can probably be
+    // removed once we use gRPC 0.10.0. See https://github.com/grpc/grpc-java/issues/1175 for more
+    // details
+    Channel channel;
+
+    if (isGet) {
+      batchRequestSize = 1;
+      streamingBufferSize = 10;
```
kevinsi4508: why 10 for Get?

kevinsi4508:
LGTM please address some comments

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
sduskis: Each result represents 1 column family.  There may be more than one column family per row.  10 < 32 which was there before.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
sduskis: Fixed.

PR #566, opened 2015-11-06
(Potentially) Fixing an auth test related problem.

sduskis:
The Backoff logic is really difficult to test given that it has so much randomness built in.  I built in a bigger buffer for randomness so that the tests will be less flaky.


kevinsi4508:
Do you know what is causing the flakiness? Where is the bigger buffer, the max_end is the same as before: startTime + maxElaspedBackoffMillis * 2, no?

sduskis:
Woops.  I thought I updated it to "startTime + maxElaspedBackoffMillis * 10".  Sorry about that.  Just submitted the 2 -> 10 update.

kevinsi4508:
LGTM

PR #567, opened 2015-11-10
Using ByteStringer where appropriate and removing places that are not.

sduskis:


bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/ScanAdapter.java
```diff
@@ -96,8 +97,8 @@ public Builder adapt(Scan scan, ReadHooks readHooks) {
         .setFilter(filter)
         .setRowRange(
             RowRange.newBuilder()
-                .setStartKey(ByteStringer.wrap(scan.getStartRow()))
-                .setEndKey(ByteStringer.wrap(scan.getStopRow())));
+                .setStartKey(ByteString.copyFrom(scan.getStartRow()))
+                .setEndKey(ByteString.copyFrom((scan.getStopRow()))));
```
kevinsi4508: ByteString.copyFrom(scan.getStopRow ?

kevinsi4508:
What is the difference between ByteStringer.wrap() and ByteString.copyFrom() again?

sduskis:
ByteStringer.wrap() uses the original byte[] and ByteString.copyFrom() makes a new copy.

kevinsi4508:
How do we know when to use ByteStringer.wrap() and when to use ByteString.copyFrom()? It is not obvious to me by looking at those codes.

sduskis:
It's safer to use  ByteString.copyFrom(), but it comes at a performance cost.  We should only use  ByteStringer.wrap() when we can guarantee that the underlying byte[] will not be manipulated while we're using it in an RPC.

kevinsi4508:
What you said makes sense. let me take a look and see if I can figure out byte[] is not going to be changed at places where we use ByteStringer.wrap().

kevinsi4508:
The places that we use ByteStringer.wrap() seem to be wrapping local byte[]s.
LGTM

PR #568, opened 2015-11-10
Using the bigtable ByteStringer instead of the hbase version.

sduskis:


sduskis:
This broke the build.  Committing without review.

PR #569, opened 2015-11-11
Attempting to fix a classpath problem in managed vm examples.

sduskis:
There were some conflicts between the changes in our pom.xml files and
the expectation of the
https://github.com/GoogleCloudPlatform/cloud-bigtable-examples/tree/master/java/managed-vm-gae
pom.xml and the updates we made between cloud-bigtable-client 0.2.1 and
0.2.2.  This change seems to correct some of the issues.

### DIFF
--- a/bigtable-hbase-1.0/pom.xml
+++ b/bigtable-hbase-1.0/pom.xml
@@ -53,8 +53,8 @@ limitations under the License.
             <artifactId>hbase-common</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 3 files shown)


kevinsi4508:
If there is a conflict, why can't we fix the config in managed-vm-gae?
Updating cloud-bigtable-client means the expectation in managed-vm-gae is correct.

I am OK with the change, hope fixing this won't break others.

kevinsi4508:
LGTM

PR #572, opened 2015-11-17
Fixing a long standing problem with Table batch operations.

sduskis:
Batching was configured incorrectly in BigtableTable.  We set number of RPCs as the heap size.  This PR fixes that problem.


kevinsi4508:
LGTM

PR #573, opened 2015-11-19
Reducing scary warnings on scanner retries.

sduskis:
Minor performance improvement (40 nanoseconds) by not constructing Backoffs in ResumingStreamingResultScanner and calling reset on every row.
Comment spelling correction in RetryOptions.


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResumingStreamingResultScanner.java
```diff
@@ -136,11 +129,12 @@ public int available() {
    */
   private void backOffAndRetry(IOException cause) throws IOException,
       ScanRetriesExhaustedException {
+    if(currentBackoff == null) {
```
kevinsi4508: leave a space between "if" and "(" => "if (currentBackoff"

kevinsi4508:
LGTM

PR #574, opened 2015-11-20
Using a ReadWriteLock instead of synchronized in ChannelPool.

sduskis:
This could potentially improve performance for buffered mutator and the like.


sduskis:
This actually slowed things down.  Closing.

PR #575, opened 2015-11-20
Create a simple put-get-delete integration test.

kevinsi4508:
Create a simple put-get-delete integration test. We are going to send alert if this test fails.


sduskis:
LGTM.  This is a great idea.

PR #576, opened 2015-11-22
Making BigtableTable exception logging to be consistent.

sduskis:


bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableTable.java
```diff
@@ -460,17 +408,21 @@ public Result increment(Increment increment) throws IOException {
...
-              "increment",
-              options.getProjectId(),
-              tableName.getQualifierAsString(),
-              increment.getRow()),
-          e);
+      throw logAndCreateIOException("increment", increment.getRow(), e);
     }
   }
 
+  private IOException logAndCreateIOException(String type, byte[] row, Throwable e) {
```
kevinsi4508: nit: "Throwable t" instead of "Throwable e"?

kevinsi4508:
LGTM

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableTable.java
sduskis: Fixed.

PR #577, opened 2015-11-23
Properly consider empty gc expression as keeping all the versions.

mgarolera:
Is not supported by HBase and will cause IllegalArgumentException. This
is better than falling back to hbase.column.max.version which is 1 by
default, which makes all the versions except the first one invisible.

### DIFF
--- a/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/adapters/TestColumnDescriptorAdapter.java
+++ b/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/adapters/TestColumnDescriptorAdapter.java
@@ -191,7 +191,10 @@ public class TestColumnDescriptorAdapter {
 
   @Test
   public void testBlankExpression(){
-    HColumnDescriptor result = adapter.adapt("family", asColumnFamily(""));
-    Assert.assertEquals(new HColumnDescriptor("family"), result);
+    // An empty gc expression means keep all versions, which is not supported by HBase.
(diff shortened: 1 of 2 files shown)


bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/adapters/TestColumnDescriptorAdapter.java
```diff
@@ -191,7 +191,10 @@ public void badExpressionLessThan(){
 
   @Test
   public void testBlankExpression(){
-    HColumnDescriptor result = adapter.adapt("family", asColumnFamily(""));
-    Assert.assertEquals(new HColumnDescriptor("family"), result);
+    // An empty gc expression means keep all versions, which is not supported by HBase.
+    expectedException.expectMessage("Maximum versions must be positive");
+    expectedException.expect(IllegalArgumentException.class);
```
sduskis: Should this be an error case, or should we allow 0 = infinite when creating a column family?

mgarolera: Apparently HBase doesn't support it, so we can't pipe things through.
https://github.com/apache/hbase/blob/master/hbase-client/src/main/java/org/apache/hadoop/hbase/HColumnDescriptor.java#L507

sduskis:
LGTM

PR #578, opened 2015-11-24
Replacing the ChannelPool's array with a List.

sduskis:
It's just as performant and makes for a much clearer codebase.


kevinsi4508:
LGTM

PR #579, opened 2015-11-24
Replacing Locks with synchronized in ReconnectingChannel.

sduskis:
synchronized is microseconds faster than Locks under load.  It also produces clearer code.


kevinsi4508:
LGTM

PR #584, opened 2015-11-26
Renaming io.grpc as part of the shading process.

### LINKED ISSUE #582: BigTable should repackage io.grpc ?
It appears that the hbase bigtable jar is shipping a version of the io.grpc.protobuf.ProtoUtils
that's been shaded to operate on com.google.bigtable.repackaged.com.google.protobuf.*
messages.

This makes it impossible to build applications that are using both BigTable and GRPC (since our own classes need a version of ProtoUtils that works om com.google.protobuf.Messages)

My impression is that BigTable's ProtoUtils should be repackaged too to a private package?

sduskis:
This fixes #582

### DIFF
--- a/bigtable-hbase-1.0/pom.xml
+++ b/bigtable-hbase-1.0/pom.xml
@@ -145,6 +145,7 @@ limitations under the License.
                             <createDependencyReducedPom>true</createDependencyReducedPom>
                             <artifactSet>
                                 <includes>
+                                    <include>com.fasterxml.jackson.core:*</include>
                                     <include>com.google.cloud.bigtable:bigtable-protos</include>
                                     <include>com.google.cloud.bigtable:bigtable-client-core</include>
                                     <include>com.google.cloud.bigtable:bigtable-hbase</include>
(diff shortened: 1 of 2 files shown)


bigtable-hbase-1.0/pom.xml
```diff
@@ -155,13 +156,24 @@ limitations under the License.
...
-                                    <include>com.fasterxml.jackson.core:*</include>
                                     <include>io.netty:*</include>
                                     <include>io.grpc:*</include>
                                 </includes>
                             </artifactSet>
                             <relocations>
                                 <relocation>
+                                    <pattern>com.fasterxml</pattern>
+                                    <shadedPattern>com.google.bigtable.repackaged.com.faster.xml</shadedPattern>
+                                </relocation>
```
kevinsi4508: com.twitter is added. Does it mean that CBT and HBase are using two different incompatible versions?

sduskis: It means that com.twitter is an import that may possibly be incompatible with external users of gRPC.

kevinsi4508:
LGTM

PR #591, opened 2015-12-10
Log Jetty ALPN and OpenSSL status.

cmccoy:
Also log the OpenSSL unavailability cause when neither Jetty ALPN nor OpenSSL
could be loaded.


sduskis:
LGYM

sduskis:
LGTM

PR #595, opened 2015-12-11
Mention how the Mutations are encoded in the descriptio of HBaseMutationCoder.

mgarolera:


bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/HBaseMutationCoder.java
```diff
@@ -27,14 +27,13 @@
...
 
-import com.google.api.services.dataflow.model.Sink;
 import com.google.cloud.dataflow.sdk.Pipeline;
 import com.google.cloud.dataflow.sdk.coders.AtomicCoder;
 import com.google.cloud.dataflow.sdk.coders.Coder;
 import com.google.cloud.dataflow.sdk.coders.CoderException;
 
 /**
- * A {@link Coder} that serializes {@link Mutation} objects used by the Cloud Bigtable {@link Sink}.
+ * A {@link Coder} that serializes HBase {@link Mutation} objects using Protocol Buffers.
```
kevinsi4508: "serialiszes and deserializes" or "encode and decode ... to and from byte streams"

kevinsi4508: "using {@link ProtobufUtil}"? "using Protocol Buffers" doesn't seem to make sense.

mgarolera: Done.

mgarolera:
PTAL

kevinsi4508:
LGTM (wait for Solomon and Jeff's review)

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/HBaseMutationCoder.java
```diff
@@ -27,14 +27,14 @@
...
-import com.google.api.services.dataflow.model.Sink;
 import com.google.cloud.dataflow.sdk.Pipeline;
 import com.google.cloud.dataflow.sdk.coders.AtomicCoder;
 import com.google.cloud.dataflow.sdk.coders.Coder;
 import com.google.cloud.dataflow.sdk.coders.CoderException;
 
 /**
- * A {@link Coder} that serializes {@link Mutation} objects used by the Cloud Bigtable {@link Sink}.
+ * A {@link Coder} that serializes and deserializes the HBase {@link Mutation} objects
+ * using {@link ProtobufUtil}.
```
sduskis: ProtobufUtil is an internal HBase tool.  I don't think it will be meaningful to users.

Perhaps we could write something to the affect of: "When Dataflow notices a slowdown in executing Puts and Deletes, it will send those Puts and Deletes to be processed on another Dataflow machine.  This class handles the serialization and desrialization of the Puts and Deletes."

kevinsi4508: More context is good, but still doesn't tell more on how serialization and desiralization are done/implmented.

sduskis: I'm not sure that the implementation is relevant in the Javadoc.  I would think that a user would care why they need class and where they'd use it.  Implementation details seem secondary.

kevinsi4508: I didn't mean how it is implemented, but how this implementation behave is relevant.

sduskis:
LGTM

kevinsi4508:
Jeff, please also take a look.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/HBaseMutationCoder.java
hegemonic: I agree with Solomon. I don't think users benefit from knowing that this implementation uses ProtobufUtil.

hegemonic:
See my comment.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/HBaseMutationCoder.java
kevinsi4508: I am OK with not mentioning ProtoBufUtil. @hegemonic  Are you OK with putting more context as Solomon suggested? A potential problem with the context could be that it could be changed later, and what are are saying here could be outdated later.

"When Dataflow notices a slowdown in executing Puts and Deletes, it will send those Puts and Deletes to be processed on another Dataflow machine. This class handles the serialization and desrialization of the Puts and Deletes."

sduskis: The mai reason for a serializer is the reason I mentioned.  We could probably get more info from the Dataflow team.

kevinsi4508: I am even OK with this:

"A {@link Coder} that serializes and deserializes the HBase {@link Mutation} objects."

I am pretty sure there are other reasons to use this class.

sduskis: Can you please elaborate on where else this would be used?

kevinsi4508: I don't have enough context to comment on how this is used. My understand is serialization and deserialization of HBase mutation seems to be a generic operation that can be utilized whenever serialization and deserialization is needed.

sduskis: That's true.  That's the existing serialization/deserialization functions from the HBase code base exist.  Given my talks with the Dataflow team, this Coder is needed for the specific case that I mentioned.

kevinsi4508: Sounds good, it is worth mentioning it here then.

kevinsi4508:
close this PR as I can't update it.
replaced with: https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/604

PR #599, opened 2015-12-15
Fixing a problem with alpn configuration exception detection.

sduskis:
A user got a NoClassDefFoundError instead of a ClassNotFoundException when trying to configure openSSL. I think it's due to the alpn jar being on the classpath, but not the bootclasspath.


lesv:
LGTM

PR #601, opened 2015-12-16
Updating various dependencies to newest possible artifact.

sduskis:


bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/BigtableWhileMatchResultScannerAdapter.java
```diff
@@ -69,6 +69,14 @@ public void close() {
           throw Throwables.propagate(ioe);
         }
       }
+
+      /**
+       * This is an HBase concept that was added in hbase 1.0.2.  It's not relevent for Cloud
+       * Bigtable.  It will not be called from the hbase code and should not be called by the user.
+       */
+      public boolean renewLease() {
+        return false;
```
kevinsi4508: I am not sure if this is the right implementation.

If we want this to work with existing hbase code, we need an implementation that corresponds to Bigtable behavior.

If Bigtable always allow scan lease to be renewed, we should always return true.

If we do not allow user to call this method, we should throw an exception.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/BigtableResultScannerAdapter.java
```diff
@@ -56,6 +56,10 @@ public void close() {
           throw Throwables.propagate(ioe);
         }
       }
+
+      public boolean renewLease() {
```
kevinsi4508: Add doc
Add @Override

pom.xml
```diff
@@ -53,20 +53,22 @@ limitations under the License.
...
+        <mockito.version>1.10.19</mockito.version>
         <protobuff-java.version>3.0.0-beta-1</protobuff-java.version>
         <grpc.version>0.9.0</grpc.version>
-        <google.api.client.version>1.19.0</google.api.client.version>
+        <google.api.client.version>1.21.0</google.api.client.version>
         <!-- For Netty HTTP2/TLS negotiation -->
         <alpn.jdk7.version>7.1.3.v20150130</alpn.jdk7.version>
         <alpn.jdk8.version>8.1.3.v20150130</alpn.jdk8.version>
         <netty.version>4.1.0.Beta6</netty.version>
         <guava.version>18.0</guava.version>
```
kevinsi4508: remove this

kevinsi4508:
Make sure integration tests run successfully.

pom.xml
sduskis: Can we please do that in another PR?

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/BigtableResultScannerAdapter.java
sduskis: I specifically don't want to add @Override, since I want this to be backwards compatible with hbase 1.0.1 and 1.1.1.

sduskis:
I addressed the comments, and re-ran the integration tests.  PTAL.

kevinsi4508:
LGTM

PR #602, opened 2015-12-17
Making BigtableOptions serializable, and adding equals and toString().

sduskis:


bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CredentialOptions.java
```diff
@@ -133,6 +137,15 @@ public String getServiceAccount() {
...
       return keyFile;
     }
+
+    @Override
+    public boolean equals(Object obj) {
+      if (obj == null || obj.getClass() != P12CredentialOptions.class) {
+        return false;
+      }
+      P12CredentialOptions other = (P12CredentialOptions) obj;
+      return Objects.equal(keyFile, other.keyFile);
```
kevinsi4508: also check serviceAccount?

kevinsi4508:
A minor thing, otherwise LGTMed.

sduskis:
Fixed the equals method.  Merging.

PR #606, opened 2015-12-21
Better handling of dataflow verification issues.

sduskis:
Table not found issues will now propogate to the user.

Also adding a test log4j.properties file to the dataflow project.

### DIFF
--- a/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
+++ b/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
@@ -829,15 +829,14 @@ public class CloudBigtableIO {
     checkNotNullOrEmpty(configuration.getZoneId(), "zoneId");
     checkNotNullOrEmpty(configuration.getClusterId(), "clusterId");
     if (tableId != null) {
-      checkNotNullOrEmpty(tableId, "tableid");
       if (BigtableSession.isAlpnProviderEnabled()) {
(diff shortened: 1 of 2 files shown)


bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -829,15 +829,14 @@ private static void validate(CloudBigtableConfiguration configuration, String ta
...
             Admin admin = conn.getAdmin()) {
-          Preconditions.checkState(admin.tableExists(TableName.valueOf(tableId)), "Table "
-              + tableId + " does not exist.  This dataflow operation could not be run.");
-        } catch (IOException | IllegalArgumentException | ExceptionInInitializerError e) {
-          LOG.error(String.format("Could not validate that the table exists: %s (%s)", e.getClass()
-              .getName(), e.getMessage()), e);
+          Preconditions.checkState(
+              admin.tableExists(TableName.valueOf(tableId)),
+              "Table " + tableId + " does not exist.  This dataflow operation could not be run.");
+        } catch (IOException e) {
```
kevinsi4508: I may have asked this question before. When table is not found, what exception we will get here?

sduskis: Preconditions.checkState throws IllegalStateException.

kevinsi4508: This means we never caught this exception before. Would this PR change anything?

PR #610, opened 2015-12-22
Adding BuilderOptions.toBuilder() to allow further configuration.

sduskis:


bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
```diff
@@ -72,6 +72,24 @@
...
+      this.userAgent = original.userAgent;
+      this.dataHost = original.dataHost;
+      this.tableAdminHost = original.tableAdminHost;
+      this.clusterAdminHost = original.clusterAdminHost;
+      this.port = original.port;
+      this.overrideIp = original.overrideIp;
+      this.credentialOptions = original.credentialOptions;
+      this.retryOptions = original.retryOptions;
+      this.timeoutMs = original.timeoutMs;
+    }
```
gamorris: dataChannelCount?

sduskis: Good catch.  I added it.  I also noticed a discrepancy between the builder and the BigtableOptions naming convention (dataChannelCount vs. channelCount).

sduskis:
Updated.  PTAL.

gamorris:
lgtm

PR #611, opened 2015-12-22
Moving the BigtableOptions checks into BigtableSession.

sduskis:
There are cases where BigtableOptions may be partially configured.  We
should defer checking for a complete BigtableOptions until we actually
use it.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
@@ -15,15 +15,15 @@
  */
 package com.google.cloud.bigtable.config;
 
-import com.google.common.annotations.VisibleForTesting;
-import com.google.api.client.util.Objects;
-import com.google.api.client.util.Strings;
-import com.google.cloud.bigtable.grpc.BigtableClusterName;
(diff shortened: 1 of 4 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
```diff
@@ -206,14 +205,6 @@ private BigtableOptions(
       RetryOptions retryOptions,
       int timeoutMs,
       int channelCount) {
-    Preconditions.checkArgument(
-        !Strings.isNullOrEmpty(projectId), "ProjectId must not be empty or null.");
```
gamorris: These are all still used below to create the BigtableClusterName. It looks like that constructor also has the same checks in it. In that sense, I'm not sure this really changes much?

Would it make sense to remove project id, zone id, cluster id as direct members of the BigtableOptions class and have all users of those go through BigtableClusterName? (That is probably a bigger change...)

sduskis: The BigtableClusterName is no longer created unless all of the values are set.  I added some tests to make sure that the expected behavior happens.

gamorris:
lgtm

Issue #605, opened 2015-12-19
Some guava classes are not repackaged (version 0.2.2)


yaitskov:
I use guava 18.0 and
it has com/google/thirdparty/publicsuffix/PublicSuffixPatterns.class

the client has different version which doesn't have EXACT public static field
which is used by other part of guava.
So I get NoSuchFieldException.

sduskis:
The PR with the change was merged.  I also deployed a 0.2.3-SNAPSHOT with this change.  We should have an official release early in the new year.

yaitskov:
0.2.3 snapshot still has these classes. I don't understand what was changed otherwise.

https://oss.sonatype.org/content/repositories/snapshots/com/google/cloud/bigtable/bigtable-hbase-1.1/0.2.3-SNAPSHOT/bigtable-hbase-1.1-0.2.3-20151222.192823-12.jar

```
  drwxrwxrwx         0  18-Nov-2015  11:15:56  com/google/thirdparty/
  drwxrwxrwx         0  18-Nov-2015  11:15:56  com/google/thirdparty/publicsuffix/
  -rw-rw-rw-     46600  18-Nov-2015  11:15:56  com/google/thirdparty/publicsuffix/PublicSuffixPatterns.class
  -rw-rw-rw-      2437  18-Nov-2015  11:15:56  com/google/thirdparty/publicsuffix/PublicSuffixType.class
  -rw-rw-rw-      4266  18-Nov-2015  11:15:56  com/google/thirdparty/publicsuffix/TrieParser.class
```

sduskis:
I redeployed.  It looks better now.

PR #612, opened 2015-12-22
Shading com.google.thirdparty.

sduskis:
com.google.thirdparty is part of the guava jar that until now was
only partially renamed.  This was causing a user a problem in #605.


kevinsi4508:
LGTM

PR #615, opened 2015-12-28
Making an often called ChannelPool method synchronous on changes only.

sduskis:
This should improve performance for BufferedMutator operations.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
@@ -70,26 +70,30 @@ public class ChannelPool extends Channel {
     }
   }
 
-  private List<Channel> channels;
+  private final List<Channel> channels;
   private final AtomicInteger requestCount = new AtomicInteger();
   private final List<HeaderInterceptor> headerInterceptors;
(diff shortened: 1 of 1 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -75,21 +75,25 @@ public synchronized void returnToPool() {
   private final List<HeaderInterceptor> headerInterceptors;
   private ChannelFactory factory;
   private int reservedChannelCount = 0;
+  private AtomicInteger totalSize = new AtomicInteger();
 
   public ChannelPool(List<HeaderInterceptor> headerInterceptors, ChannelFactory factory)
       throws IOException {
     this.channels = new ArrayList<>();
     channels.add(factory.create());
+    totalSize.set(1);
```
kevinsi4508: nit: new AtomicInteger(1); ?

kevinsi4508:
LGTM

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
sduskis: Done.

PR #614, opened 2015-12-28
Fixing some issues with HeapSizeManager.

sduskis:
waitUntilAllOperationsAreDone() now uses a counter to check tha all operations are done instead of using the compleiton map that's used for registerOperationWithHeapSize().

PR #617, opened 2015-12-29
Fixing some issues with HeapSizeManager.

sduskis:
Using a wait/notify mechanism for HeapSizeManager flushes that waits for all completions to happen rather than reacting to each completion, which causes thread thrashing.

This is the same code as #614.  Something bad happened on the github side with all of the forced updates.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/HeapSizeManager.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/HeapSizeManager.java
@@ -16,14 +16,18 @@
 package com.google.cloud.bigtable.grpc.async;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.LinkedBlockingDeque;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
 import com.google.cloud.bigtable.config.Logger;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
@@ -38,21 +42,28 @@ public class HeapSizeManager {
   // Flush is not properly synchronized with respect to waiting. It will never exit
   // improperly, but it might wait more than it has to. Setting this to a low value ensures
   // that improper waiting is minimal.
-  private static final long FINISH_WAIT_MILLIS = 250;
-  private static final long REGISTER_WAIT_MILLIS = 5;
+  private static final long FINISH_WAIT_MILLIS = 100;
+  private static final long REGISTER_WAIT_MILLIS = 10;
 
   // In flush, wait up to this number of milliseconds without any operations completing.  If
   // this amount of time goes by without any updates, flush will log a warning.  Flush()
   // will still wait to complete.
-  private static final long INTERVAL_NO_SUCCESS_WARNING = 300000;
+  private static final long INTERVAL_NO_SUCCESS_WARNING_MS = 60000;
+
   private final long maxHeapSize;
   private final int maxInFlightRpcs;
-  private final Map<Long, Long> pendingOperationsWithSize = new HashMap<>();
-  private final LinkedBlockingDeque<Long> completedOperationIds = new LinkedBlockingDeque<>();
 
-  private long currentWriteBufferSize = 0;
   private AtomicLong operationSequenceGenerator = new AtomicLong();
+
+  private final Map<Long, Long> pendingOperationsWithSize = new HashMap<>();
+  private final AtomicInteger pendingOperationCount = new AtomicInteger();
+
+  private final LinkedBlockingQueue<Long> completedOperationIds = new LinkedBlockingQueue<>();
+  private long currentWriteBufferSize = 0;
   private long lastOperationChange = System.currentTimeMillis();
+
+  private AtomicBoolean isFlushing = new AtomicBoolean(false);
+  private AtomicBoolean isFull = new AtomicBoolean(false);
 
   public HeapSizeManager(long maxHeapSize, int maxInflightRpcs) {
     this.maxHeapSize = maxHeapSize;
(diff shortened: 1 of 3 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/HeapSizeManager.java
```diff
@@ -63,61 +73,74 @@ public long getMaxHeapSize() {
...
-            + "There are still %d operations in progress.", lastUpdated,
-          pendingOperationsWithSize.size());
+      if (!performedWarning && requiresWarning()) {
+        warn();
         performedWarning = true;
-        waitForCompletions(FINISH_WAIT_MILLIS);
       }
+      wait(FINISH_WAIT_MILLIS);
     }
     if (performedWarning) {
```
kevinsi4508: When performedWarning is false, "flush()" could also be completed, correct?

sduskis: If performedWarning is false, we don't necessarily want to log that flush() is done.  We only want to performedWarning if something seems wrong.  This follow up of "flush() complete" will show that the situation corrected itself.  We usually don't actually want to LOG.info() when flush completes... we only want to do it in unusual circumstances.

kevinsi4508: Got it. When I look at the logs, sometimes I see "flush() completed", sometimes I don't. It is bit weird. Maybe this "flush() completed with warning" ?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/HeapSizeManager.java
```diff
@@ -63,61 +73,74 @@ public long getMaxHeapSize() {
...
+      lastOperationChange = System.currentTimeMillis();
+    }
+  }
+
+  private void markOperationComplete(Long operationSequenceId) {
+    Long heapSize = pendingOperationsWithSize.remove(operationSequenceId);
+    if (heapSize != null) {
+      currentWriteBufferSize -= heapSize;
+    } else {
+      LOG.warn("An operation completion was recieved multiple times. Your operations completed."
```
kevinsi4508: Your suggestion sounds good: "An operation was reported to be complete multiple times. It's likely due to a communication glitch and a retry."

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/HeapSizeManager.java
```diff
@@ -63,61 +73,74 @@ public long getMaxHeapSize() {
     return maxHeapSize;
   }
 
-  public synchronized void waitUntilAllOperationsAreDone() throws InterruptedException {
+  public synchronized void flush() throws InterruptedException {
+    isFlushing.set(true);
     boolean performedWarning = false;
-    while(!pendingOperationsWithSize.isEmpty()) {
+    while (true) {
```
kevinsi4508: Can we do this?

while (!pendingOperationsWithSize.isEmpty()) {
  cleanupFinishedOperations();
  if (!performedWarning && requiresWarning()) {
    warn();
    performedWarning = true;
  }
  wait(FINISH_WAIT_MILLIS);
}

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/HeapSizeManager.java
sduskis: The warning should have been evident from the previous log statement.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/HeapSizeManager.java
sduskis: I prefer the cleanup before the break.

   cleanupFinishedOperations();
    while (!pendingOperationsWithSize.isEmpty()) {
      if (!performedWarning && requiresWarning()) {
        warn();
        performedWarning = true;
      }
      wait(FINISH_WAIT_MILLIS);
      cleanupFinishedOperations();
    }

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/HeapSizeManager.java
```diff
@@ -63,61 +73,74 @@ public long getMaxHeapSize() {
...
       }
+      wait(FINISH_WAIT_MILLIS);
     }
     if (performedWarning) {
       LOG.info("flush() completed");
     }
+    isFlushing.set(false);
+  }
+
+  private boolean requiresWarning() {
```
kevinsi4508: nit: This method is not obvious to me at first. Can you add comment that says Returns {code true} if the last successful completion is awhile ago.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/HeapSizeManager.java
```diff
@@ -183,5 +189,10 @@ public void onFailure(Throwable t) {
 
   public void markCanBeCompleted(Long id) {
     completedOperationIds.offer(id);
+    if (this.isFlushing.get() && completedOperationIds.size() == pendingOperationsWithSize.size()) {
```
kevinsi4508: sorry! can you explain a bit on why isFlushing has to be true and "ompletedOperationIds.size() == pendingOperationsWithSize.size()", so that we should notify thread waiting on "this"?

sduskis:
I updated the comments.  PTAL.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/HeapSizeManager.java
kevinsi4508: hm.. why an etra cleanupFinishedOperations() before (!pendingOperationsWithSize.isEmpty()) ? I thought if pendingOperationsWithSize.isEmpty(), we don't need to do anything.

"while (!pendingOperationsWithSize.isEmpty())"  not "while (pendingOperationsWithSize.isEmpty())"

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/HeapSizeManager.java
kevinsi4508: I understand "this.isFlushing.get()" means some thread is in flush().
What does it mean when completedOperationIds.size() == pendingOperationsWithSize.size()?

kevinsi4508:
More comments. Did you update the code?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/HeapSizeManager.java
```diff
@@ -63,61 +76,77 @@ public long getMaxHeapSize() {
...
     }
+    isFlushing.set(false);
+  }
+
+  /**
+   * @return {@code true} if the last successful completion is awhile ago (longer than {@code
+   *         INTERVAL_NO_SUCCESS_WARNING_MS} ms ago.
+   */
+  private boolean requiresWarning() {
+    return lastOperationChange + INTERVAL_NO_SUCCESS_WARNING_MS < System.currentTimeMillis();
```
kevinsi4508: lastOperationChange doesn't represent last successful completion because we also update this ts on register.

kevinsi4508:
Add some test if not already there:

e.g.
T1: register op1
T2: register op2
T3: register op3 (full, blocked), op1 done
T4: flush (blocked), op2 done
T5: op3 done (notify flush), flush returns

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/HeapSizeManager.java
```diff
@@ -38,22 +41,32 @@
...
   private final Map<Long, Long> pendingOperationsWithSize = new HashMap<>();
-  private final LinkedBlockingDeque<Long> completedOperationIds = new LinkedBlockingDeque<>();
+  private final AtomicInteger pendingOperationCount = new AtomicInteger();
 
+  private final LinkedBlockingDeque<Long> completedOperationIds = new LinkedBlockingDeque<>();
   private long currentWriteBufferSize = 0;
-  private AtomicLong operationSequenceGenerator = new AtomicLong();
   private long lastOperationChange = System.currentTimeMillis();
 
+  private AtomicBoolean isFlushing = new AtomicBoolean(false);
```
kevinsi4508: nit: this can be final

PR #620, opened 2016-01-04
Improving ChannelPool performance under load

sduskis:
ChannelPool's logic can be replicated with an AtomicReference.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
@@ -15,11 +15,7 @@
  */
 package com.google.cloud.bigtable.grpc.io;
 
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.logging.Logger;
+import com.google.common.collect.ImmutableList;
 
 import io.grpc.CallOptions;
 import io.grpc.Channel;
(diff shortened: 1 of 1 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -70,30 +75,42 @@ public synchronized void returnToPool() {
...
+      synchronized(this){
+        if (totalSize.get() < capacity) {
+          ImmutableList.Builder<Channel> channelsBuilder = ImmutableList.builder();
+          List<Channel> current = channels.get();
+          channelsBuilder.addAll(current);
+          int unreservedCapacity = capacity - reservedChannelCount;
+          for (int i = current.size(); i < unreservedCapacity; i++) {
+            channelsBuilder.add(factory.create());
+          }
+          channels.set(channelsBuilder.build());
```
kevinsi4508: why do you think using an ImmutableList is faster? It requires building and making a copy.

sduskis: Building an immutableList is definitely slower than modifying a List.  I'm not concerned about the difference, though, since this happens infrequently.  The point is to make access of the List consistent for the getNextChannel() method which gets called a few orders of magnitude more often than any of the methods that build a new ImmutableList.  This change allowed me to remove the synchronization from getNextChannel() since we're now guaranteed to have a consistent list to get a size on and then do a get() on.

ensureChannelCount's expansion logic should happen once or twice for an entire application.  reserveChannel() may be called infrequently for scans.  getNextChannel() gets called for every operation.

kevinsi4508: It seems to me that we don't expose "channels" directly, we don't need to make it immutable. It makes the code less readable.

It is up to you.

kevinsi4508:
LGTM

PR #621, opened 2016-01-04
Fixing a bug in HeapSizeManager.flush().

sduskis:
flush() is supposed to wait until all requests are done.  There was a wait method being called from flush() in the wrong place, so flush() didn't actually wait.  Fixing that issue.


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/HeapSizeManager.java
```diff
@@ -77,8 +74,9 @@ public synchronized void flush() throws InterruptedException {
             + "There are still %d operations in progress.", lastUpdated,
```
gamorris: unrelated nit: this log message needs an extra space at the +

sduskis: Done.

gamorris:
lgtm

PR #623, opened 2016-01-07
Removing @VisibleForTesting from BigtableBufferedMutator.

sduskis:
The method previous marked as @VisibleForTesting is used by
AbstractBigtableConnection to keep track of the state of
BigtableBufferedMutator.  Since the method is for more than just
testing, it makes sense to remove the annotation.

Also removing throws clauses where they are unnecessary in BufferedMutator classes.


gamorris:
lgtm

PR #625, opened 2016-01-11
Adding better default configuration for bigtable buffered mutator.

sduskis:
Also logging settings in bigtable buffered mutator.

The implementation:
- default channel count: 2.5 * available processors (i.e. 4 CPUs would mean 10 channels)
- default rpc count: 50 * channel count -- this wasn't changed
- default max heap: 10% of available memory.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
@@ -39,11 +39,18 @@ public class BigtableOptions implements Serializable {
   public static final String BIGTABLE_DATA_HOST_DEFAULT = "bigtable.googleapis.com";
   public static final int DEFAULT_BIGTABLE_PORT = 443;
 
-  public static final int BIGTABLE_DATA_CHANNEL_COUNT_DEFAULT = 6;
+  public static final int BIGTABLE_DATA_CHANNEL_COUNT_DEFAULT = getDefaultDataChannelCount();
   public static final int BIGTABLE_CHANNEL_TIMEOUT_MS_DEFAULT =
       (int) TimeUnit.MILLISECONDS.convert(30, TimeUnit.MINUTES);
 
   private static final Logger LOG = new Logger(BigtableOptions.class);
+
+  private static int getDefaultDataChannelCount() {
+    // 10 Channels seemed to work well on a 4 CPU machine, and this seems to scale well for higher
+    // CPU machines. Use no more than 250 Channels by default.
+    int availableProcessors = Runtime.getRuntime().availableProcessors();
+    return (int) Math.min(250, Math.max(1, Math.ceil(availableProcessors * 2.5d)));
+  }
 
   /**
    * A mutable builder for BigtableConnectionOptions.
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AsyncExecutor.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AsyncExecutor.java
@@ -42,8 +42,9 @@ public class AsyncExecutor {
   public static final int MAX_INFLIGHT_RPCS_DEFAULT = 50;
 
   // This is the maximum accumulated size of uncompleted requests that we allow before throttling.
-  // Default to 32MB.
-  public static final long ASYNC_MUTATOR_MAX_MEMORY_DEFAULT = 16 * 2097152;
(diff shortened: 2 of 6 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
```diff
@@ -39,12 +39,19 @@
...
   public static final int BIGTABLE_CHANNEL_TIMEOUT_MS_DEFAULT =
       (int) TimeUnit.MILLISECONDS.convert(30, TimeUnit.MINUTES);
 
   private static final Logger LOG = new Logger(BigtableOptions.class);
 
+  private static int getDefaultDataChannelCount() {
+    // 10 Channels seemed to work well on a 4 CPU machine, and this seems to scale well for higher
+    // CPU machines
+    int availableProcessors = Runtime.getRuntime().availableProcessors();
+    return (int) Math.max(1, Math.ceil(availableProcessors * 2.5d));
```
vphani: availableProcessors() isn't supposed to return 0, but I guess you don't trust that, and you are checking that we have atleast 1 channel -  in the same light do you also want some upper bound check as well? Say no more than 250 channels ?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AsyncExecutor.java
```diff
@@ -42,8 +42,9 @@
   public static final int MAX_INFLIGHT_RPCS_DEFAULT = 50;
 
   // This is the maximum accumulated size of uncompleted requests that we allow before throttling.
-  // Default to 32MB.
-  public static final long ASYNC_MUTATOR_MAX_MEMORY_DEFAULT = 16 * 2097152;
+  // Default to 10% of available memory.
+  public static final long ASYNC_MUTATOR_MAX_MEMORY_DEFAULT =
+      (long) (Runtime.getRuntime().maxMemory() * 0.1d);
```
vphani: "Returns the maximum amount of memory that the Java virtual machine will attempt to use. If there is no inherent limit then the value Long.MAX_VALUE will be returned."

Do you want to protect against the Long.MAX_VALUE case - have some sensible boundary limits.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AsyncExecutor.java
```diff
@@ -134,6 +135,7 @@ public void flush() throws IOException {
       sizeManager.flush();
     } catch (InterruptedException e) {
       Thread.currentThread().interrupt();
+      throw new IOException("flush() was interrupted before it could complete", e);
```
vphani: nit: This doesn't seem like it should belong in in PR

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -170,7 +185,7 @@ public void mutate(final Mutation mutation) throws IOException {
     }
   }
 
-  private void doMutation(final Mutation mutation) throws RetriesExhaustedWithDetailsException {
+  private void doMutation(final Mutation mutation) {
```
vphani: This change needs another PR?

bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -193,17 +204,6 @@ private HBaseRequestAdapter createAdapter(TableName tableName) {
     return new HBaseRequestAdapter(options.getClusterName(), tableName, conf);
   }
 
-  @Override
-  public BufferedMutator getBufferedMutator(TableName tableName) throws IOException {
```
vphani: nit: This seems to have just moved around, causing a diff, was there a reason for the move?

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
sduskis: Master doesn't show the "throws RetriesExhaustedWithDetailsException".  I'm not sure why this is showing up in the diff.  Very strange.

sduskis:
I updated the PR.  PTAL.

vphani:
LGTM

PR #626, opened 2016-01-11
BigtableBufferdMutator rpc calls are done asynchronously

sduskis:
The adapter from hbase to bigtable proto can take a microsecond or so,
depending on the size.  The initiation of the RPC can also take some
time.  This PR does those two things in worker threads rather than
running in the main thread.  This increases throughput and better
utilizes machine resources.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
@@ -37,11 +37,12 @@ public class BigtableOptions implements Serializable {
   public static final String BIGTABLE_CLUSTER_ADMIN_HOST_DEFAULT =
       "bigtableclusteradmin.googleapis.com";
   public static final String BIGTABLE_DATA_HOST_DEFAULT = "bigtable.googleapis.com";
-  public static final int DEFAULT_BIGTABLE_PORT = 443;
+  public static final int BIGTABLE_PORT_DEFAULT = 443;
 
   public static final int BIGTABLE_DATA_CHANNEL_COUNT_DEFAULT = getDefaultDataChannelCount();
   public static final int BIGTABLE_CHANNEL_TIMEOUT_MS_DEFAULT =
       (int) TimeUnit.MILLISECONDS.convert(30, TimeUnit.MINUTES);
+  public static final int BIGTABLE_ASYNC_MUTATOR_COUNT_DEFAULT = 2;
 
   private static final Logger LOG = new Logger(BigtableOptions.class);
 
@@ -56,7 +57,7 @@ public class BigtableOptions implements Serializable {
    * A mutable builder for BigtableConnectionOptions.
    */
   public static class Builder {
-     // Configuration that a user is required to set.
+    // Configuration that a user is required to set.
     private String projectId;
     private String zoneId;
     private String clusterId;
@@ -66,7 +67,7 @@ public class BigtableOptions implements Serializable {
     private String dataHost = BIGTABLE_DATA_HOST_DEFAULT;
     private String tableAdminHost = BIGTABLE_TABLE_ADMIN_HOST_DEFAULT;
     private String clusterAdminHost = BIGTABLE_CLUSTER_ADMIN_HOST_DEFAULT;
-    private int port = DEFAULT_BIGTABLE_PORT;
+    private int port = BIGTABLE_PORT_DEFAULT;
     private String overrideIp;
 
     // The default credentials get credential from well known locations, such as the GCE
@@ -78,6 +79,7 @@ public class BigtableOptions implements Serializable {
     private RetryOptions retryOptions = new RetryOptions.Builder().build();
     private int timeoutMs = BIGTABLE_CHANNEL_TIMEOUT_MS_DEFAULT;
     private int dataChannelCount = BIGTABLE_DATA_CHANNEL_COUNT_DEFAULT;
+    private int asyncMutatorCount = BIGTABLE_ASYNC_MUTATOR_COUNT_DEFAULT;
 
     public Builder() {
     }
@@ -96,6 +98,7 @@ public class BigtableOptions implements Serializable {
       this.retryOptions = original.retryOptions;
       this.timeoutMs = original.timeoutMs;
       this.dataChannelCount = original.dataChannelCount;
+      this.asyncMutatorCount = original.asyncMutatorCount;
     }
 
     public Builder setTableAdminHost(String tableAdminHost) {
@@ -167,6 +170,13 @@ public class BigtableOptions implements Serializable {
       return this;
     }
 
+    public Builder setAsyncMutatorWorkerCount(int asyncMutatorCount) {
+      Preconditions.checkArgument(
+          asyncMutatorCount > 0, "asyncMutatorCount must be greater than 0.");
+      this.asyncMutatorCount = asyncMutatorCount;
+      return this;
+    }
+
     public BigtableOptions build() {
       return new BigtableOptions(
           clusterAdminHost,
@@ -181,7 +191,8 @@ public class BigtableOptions implements Serializable {
           userAgent,
           retryOptions,
           timeoutMs,
-          dataChannelCount);
+          dataChannelCount,
+          asyncMutatorCount);
     }
   }
 
@@ -199,6 +210,7 @@ public class BigtableOptions implements Serializable {
   private final int timeoutMs;
   private final int dataChannelCount;
   private final BigtableClusterName clusterName;
+  private final int asyncMutatorCount;
 
   @VisibleForTesting
   BigtableOptions() {
@@ -216,6 +228,7 @@ public class BigtableOptions implements Serializable {
       timeoutMs = 0;
       dataChannelCount = 1;
       clusterName = null;
+      asyncMutatorCount = 1;
   }
 
   private BigtableOptions(
@@ -231,7 +244,8 @@ public class BigtableOptions implements Serializable {
       String userAgent,
       RetryOptions retryOptions,
       int timeoutMs,
-      int channelCount) {
+      int channelCount,
+      int asyncMutatorCount) {
     Preconditions.checkArgument(channelCount > 0, "Channel count has to be at least 1.");
     Preconditions.checkArgument(timeoutMs >= -1,
       "ChannelTimeoutMs has to be positive, or -1 for none.");
@@ -249,6 +263,7 @@ public class BigtableOptions implements Serializable {
     this.retryOptions = retryOptions;
     this.timeoutMs = timeoutMs;
     this.dataChannelCount = channelCount;
+    this.asyncMutatorCount = asyncMutatorCount;
 
     if (!Strings.isNullOrEmpty(projectId)
         && !Strings.isNullOrEmpty(zoneId)
@@ -343,6 +358,10 @@ public class BigtableOptions implements Serializable {
     return clusterName;
   }
 
+  public int getAsyncMutatorCount() {
+    return asyncMutatorCount;
+  }
+
   @Override
   public boolean equals(Object obj) {
     if (obj == null || obj.getClass() != BigtableOptions.class) {
@@ -352,8 +371,10 @@ public class BigtableOptions implements Serializable {
       return true;
     }
     BigtableOptions other = (BigtableOptions) obj;
-    return (port == other.port) && (timeoutMs == other.timeoutMs)
+    return (port == other.port)
+        && (timeoutMs == other.timeoutMs)
         && (dataChannelCount == other.dataChannelCount)
+        && (asyncMutatorCount == other.asyncMutatorCount)
         && Objects.equal(clusterAdminHost, other.clusterAdminHost)
         && Objects.equal(tableAdminHost, other.tableAdminHost)
         && Objects.equal(dataHost, other.dataHost)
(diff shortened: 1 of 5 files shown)


bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -81,20 +102,97 @@
...
+    @Override
+    public void run() {
+      activeMutationWorkers.incrementAndGet();
+      try {
+        while (!closed) {
+          try {
+            MutationOperation operation =
+                mutationsToBeSent.poll(MUTATION_TO_BE_SENT_WAIT_MS, TimeUnit.MILLISECONDS);
+            // The operation can be null if a timeout occurs.
+            if (operation != null) {
```
kevinsi4508: Shouldn't this thread wait and be notified when new mutation is added to mutationsToBeSent?

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -81,20 +102,97 @@
...
+                  issueRequest(operation.mutation, operation.operationId);
+              ExceptionCallback callback = new ExceptionCallback(operation.mutation);
+              Futures.addCallback(request, callback);
+            }
+          } catch (InterruptedException e) {
+            Thread.interrupted();
+            LOG.info("Interrupted.  Shutting down mutationRunnable.");
+            break;
+          } catch (Exception e) {
+            LOG.error("Exception in mutation worker", e);
```
kevinsi4508: shouldn't we break here as well?

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -81,20 +102,97 @@
...
+            // The operation can be null if a timeout occurs.
+            if (operation != null) {
+              ListenableFuture<? extends GeneratedMessage> request =
+                  issueRequest(operation.mutation, operation.operationId);
+              ExceptionCallback callback = new ExceptionCallback(operation.mutation);
+              Futures.addCallback(request, callback);
+            }
+          } catch (InterruptedException e) {
+            Thread.interrupted();
+            LOG.info("Interrupted.  Shutting down mutationRunnable.");
```
kevinsi4508: remove extra white space

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
weiminyu: mutationsToBeSent.poll does this under the hood. The blockingqueue implementation does not use busy wait.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
weiminyu: Interrupted Exception is a signal to quit thus the break. Other exceptions shouldn't cause worker to quit.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
sduskis: @weiminyu is correct.  .poll waits until something is available or the timeout occurs.  If the timeout occurs, it returns null.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
sduskis: @weiminyu is correct.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
sduskis: Done.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
kevinsi4508: Understood, It just seems odd if there is no mutation, but X number of worker threads keep waking up and wast CPU time. set longer wait time perhaps?

sduskis: BufferedMutator is usually used in cases where there are a huge number of mutations, a flush and a close.  I would guess that a buffered mutator usually isn't idle.  The thought on the 100 ms wait was that if there is an idle buffered mutator, it should check periodically for close events.

The usual case of 2 threads waking up every 100 ms doesn't seem like a huge over head to me.  I'll confirm with a manual test that opens a buffered mutator and doesn't do anything with it to see the behavior.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
```diff
@@ -160,6 +163,11 @@ public Builder setTimeoutMs(int timeoutMs) {
       return this;
     }
 
+    public Builder setAsyncMutatorWorkerCount(int asyncMutatorCount) {
+      this.asyncMutatorCount = asyncMutatorCount;
```
kevinsi4508: nit: check asyncMutatorCount > 0 ?

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
sduskis: We discussed a change offline for an isCloseMarker operation that notifies the worker that it needs to shut down.

sduskis:
The comments were addressed.  PTAL.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -61,6 +73,26 @@
     }
   }
 
+  private static class MutationOperation {
+    final Mutation mutation;
+    final long operationId;
+    final boolean isCloseMarker;
+
+    public MutationOperation(boolean isCloseMarker) {
```
kevinsi4508: nit:
static MutationOperation CloseMarker() {
  MutationOperation(x, y, true);
}

public MutationOperation(Mutation mutation, long operationId) {
  MutationOperation(mutation, operationId, false);
}

private MutationOperation(Mutation mutation, long operationId, boolean isCloseMarker) {
       this.mutation = mutation;
       this.operationId = operationId;
       this.isCloseMarker = isCloseMarker;
     }

kevinsi4508: In fact, make it a final static MutationOperation CLOSE_MARKER;

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -81,20 +113,101 @@
...
+      activeMutationWorkers.incrementAndGet();
+      try {
+        while (!closed) {
+          try {
+            MutationOperation operation =
+                mutationsToBeSent.poll(MUTATION_TO_BE_SENT_WAIT_MS, TimeUnit.MILLISECONDS);
+            // The operation can be null if a timeout occurs.
+            if (operation != null) {
+              if (operation.isCloseMarker) {
+                mutationsToBeSent.add(operation);
```
kevinsi4508: We add the close marker back to let other worker thread know?

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -150,6 +267,26 @@ public void mutate(List<? extends Mutation> mutations) throws IOException {
...
+   * object to cloud bigtable proto and the async call both take time (microseconds worth) that
+   * could be parallelized, or at least removed from the user's thread.
+   */
+  private void offer(Mutation mutation) throws IOException {
+    if (activeMutationWorkers.get() == 0) {
+      initializeAsyncMutators();
+    }
+    long operationId;
+    try {
+      operationId = heapSizeManager.registerOperationWithHeapSize(mutation.heapSize());
```
kevinsi4508: if we don't need "operationId" outside of try{}, move both "long operationId;" and "mutationsToBeSent.add(new MutationOperation(mutation, operationId));" into try {}?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AsyncExecutor.java
```diff
@@ -93,6 +93,26 @@ public AsyncExecutor(BigtableDataClient client, HeapSizeManager heapSizeManager)
     this.sizeManager = heapSizeManager;
   }
 
+  public ListenableFuture<Empty> mutateRowAsync(MutateRowRequest request, long operationId)
```
kevinsi4508: Add some documentation if some super users really want to use this.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
kevinsi4508: Add doc?

kevinsi4508:
Please address some nits.
LGTM

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
weiminyu: Could you elaborate the reason for a close marker? It is added right before executorservice.shutdown() which would interrupt the thread.

sduskis: From BlockingQueue javadoc: 

``A BlockingQueue does not intrinsically support any kind of "close" or "shutdown" operation to indicate that no more items will be added. The needs and usage of such features tend to be implementation-dependent. For example, a common tactic is for producers to insert special end-of-stream or poison objects, that are interpreted accordingly when taken by consumers. ```

https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/BlockingQueue.html

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
sduskis: I changed the implementation to add a CLOSE_MARKER per active mutation worker.  The current implementation may not work depending on the order of executor.shutdown() and the re-add of the CLOSE_MARKER.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -82,7 +121,57 @@
...
+  /**
+   * This {@link Runnable} pulls a mutation from {@link #mutationsToBeSent}, and calls {{@link
+   * #issueRequest(Mutation, long)}.
+   */
+  private final Runnable mutationWorker = new Runnable() {
+    @Override
+    public void run() {
+      activeMutationWorkers.incrementAndGet();
+      try {
+        while (!closed) {
```
weiminyu: change declaration of closed to 'private volatile boolean closed;' no that writer and reader are on different threads?

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
kevinsi4508: +1

weiminyu: I'm still not convinced that a CLOSE_MARKER is needed. Once mutator is closed no more mutations may be added. If the mutation_worker's main loop is changed to the following it seems to me that it would work
while (!closed || !mutationsToSend.isEmpty() ) {
   try {
     poll and send
   }  catch (InterruptedException e) {
     Thread.interrupted();
     // do not break;
   } catch (Exception e) {
    // as before
  } 
}

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
sduskis: I'm going to remove the while (!closed) now that we have the CLOSED_MARKER.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -102,14 +195,29 @@ public BigtableBufferedMutator(
...
+
+  private void initializeAsyncMutators() {
+    for (int i = activeMutationWorkers.get(); i < options.getAsyncMutatorCount(); i++) {
+      executorService.submit(mutationWorker);
+    }
   }
 
   @Override
   public void close() throws IOException {
+    flush();
```
weiminyu: This does not flush the mutationsToBeSent queue.

sduskis: There's a subtle set of interactions here.  There's an expectation that if heapSizeManager.flush() is complete, mutationsToBeSent queue will be empty.  I can walk you through the details in person.

PR #627, opened 2016-01-11
Adding java version to user agent header.

sduskis:


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -296,7 +296,11 @@ public Credentials call() throws IOException {
         (scheduledRetries == null) ? createDefaultRetryExecutor() : scheduledRetries;
 
     Builder<HeaderInterceptor> headerInterceptorBuilder = new ImmutableList.Builder<>();
-    headerInterceptorBuilder.add(new UserAgentInterceptor(options.getUserAgent()));
+
+    String javaVersion = "jdk" + System.getProperty("java.specification.version");
```
kevinsi4508: set/update user agent here:
public static final String USER_AGENT = getUserAgent();
in  BigtableConstants.java ?

sduskis:
I moved the logic to get the java version to BigtableConstants.  PTAL.

kevinsi4508:
LGTM

PR #628, opened 2016-01-13
Cleaning up warnings.

sduskis:
There are a whole bunch of declared Exceptions that were not thrown.  There were some javadoc issues fixed.  Some methods were deprecated in hbase, but not marked as @Deprected in Bigtable.

### DIFF
--- a/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
+++ b/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
@@ -105,6 +105,20 @@ public abstract class AbstractBigtableConnection implements Connection, Closeabl
     this(conf, false, null, null);
   }
 
+  /**
+   * The constructor called from {@link ConnectionFactory#createConnection(Configuration)} and
+   * in its many forms via reflection with this specific signature.
+   *
+   * @param conf The configuration for this channel. See {@link BigtableOptionsFactory} for more details.
+   * @param managed This should always be false. It's an artifact of old HBase behavior.
+   * @param pool An {@link ExecutorService} to run HBase/Bigtable object conversions on. The RPCs
+   *             themselves run via NIO, and not on a waiting thread
+   * @param user This is an artifact of HBase which Cloud Bigtable ignores. User information is
+   * captured in the Credentials configuration in conf.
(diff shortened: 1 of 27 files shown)


bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -105,6 +105,20 @@ public AbstractBigtableConnection(Configuration conf) throws IOException {
     this(conf, false, null, null);
   }
 
+  /**
+   * Default constructor called from {@link ConnectionFactory#createConnection(Configuration)} and
+   * in its many forms via reflection.
+   * 
+   * @param conf The configuration of 
+   * @param managed This should always be false. It's an artifact of old HBase behavior.
```
kevinsi4508: can we get rid of "managed"?

bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -105,6 +105,20 @@ public AbstractBigtableConnection(Configuration conf) throws IOException {
     this(conf, false, null, null);
   }
 
+  /**
+   * Default constructor called from {@link ConnectionFactory#createConnection(Configuration)} and
+   * in its many forms via reflection.
+   * 
+   * @param conf The configuration of
```
kevinsi4508: the configuration of ...

bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
sduskis: Unfortunately, not.  ConnectionFactory.createConnection(Configuration) uses reflection and looks for this signature.

kevinsi4508: That is what I thought. LGTM.

sduskis: FWIW, I updated the docs to "reflect" that this signature is mandated by HBase.

PR #631, opened 2016-01-19
Moving the BigtableBufferedMutator offer() method.

sduskis:
Stylistically, I'd like to have the two public mutate() methods near each other.


bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -300,6 +281,26 @@ public void mutate(final Mutation mutation) throws IOException {
     }
   }
```
kevinsi4508: nit: remove an extra new line.

kevinsi4508:
LGTM

Issue #630, opened 2016-01-19
DataFlow CloudBigtableIO leaks threads


danielnorberg:
We have a DataFlow job that's structurally very similar to the [WordCount](https://cloud.google.com/dataflow/examples/wordcount-example#UnboundedData) example, reading from pubsub, producing per X minute windowed counts that are written to CBT.

We're using `0.2.3-SNAPSHOT` of `bigtable-hbase-dataflow` and `bigtable-hbase-1.1` in order to work around https://github.com/GoogleCloudPlatform/cloud-bigtable-client/issues/613.

When deploying our job late last week we discovered that it was failing with the DF workers unable to create more threads.

```
java.lang.OutOfMemoryError: unable to create new native thread
	at java.lang.Thread.start0(Native Method)
	at java.lang.Thread.start(Thread.java:714)
	at java.util.concurrent.ThreadPoolExecutor.addWorker(ThreadPoolExecutor.java:950)
	at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1368)
	at java.util.concurrent.AbstractExecutorService.submit(AbstractExecutorService.java:112)
	at com.google.cloud.bigtable.hbase.BigtableBufferedMutator.initializeAsyncMutators(BigtableBufferedMutator.java:206)
	at com.google.cloud.bigtable.hbase.BigtableBufferedMutator.<init>(BigtableBufferedMutator.java:201)
	at com.google.cloud.bigtable.hbase.BigtableBufferedMutator.<init>(BigtableBufferedMutator.java:174)
	at org.apache.hadoop.hbase.client.AbstractBigtableConnection$2.<init>(AbstractBigtableConnection.java:193)
	at org.apache.hadoop.hbase.client.AbstractBigtableConnection.getBufferedMutator(AbstractBigtableConnection.java:192)
	at com.google.cloud.bigtable.dataflow.CloudBigtableIO$CloudBigtableSingleTableWriteFn.getBufferedMutator(CloudBigtableIO.java:615)
	at com.google.cloud.bigtable.dataflow.CloudBigtableIO$CloudBigtableSingleTableWriteFn.processElement(CloudBigtableIO.java:640)
```

We SSH'd to one of the DF worker hosts and saw that the worker JVM process had some 32k+ threads. We attempted to redeploy our job a few times, each time hitting the same issue. 

[...]

PR #633, opened 2016-01-19
Turning off BufferedMutator workers when no work is present.

sduskis:
We have too many running threads in some Dataflow scenarios, as per: #630.  This is a step towards reducing the number of active Threads.

### DIFF
--- a/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
+++ b/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
@@ -142,16 +142,14 @@ public class BigtableBufferedMutator implements BufferedMutator {
             MutationOperation operation =
                 mutationsToBeSent.poll(MUTATION_TO_BE_SENT_WAIT_MS, TimeUnit.MILLISECONDS);
             // The operation can be null if a timeout occurs.
-            if (operation != null) {
-              if (operation.isCloseMarker) {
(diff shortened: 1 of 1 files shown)


kevinsi4508:
There could be a problem here:
if (activeMutationWorkers.get() == 0) {
      initializeAsyncMutators();
 }

If this is called multiple times, and no worker threads have started yet, we could create so many worker threads.

sduskis:
Good point. I'll synchronize that section.

weiminyu:
Following up on Kevin's comment, I think there are two more issues:
1. there is a race between the count==0 check and the worker thread's exit. Might be better to move initializeAsyncMutator() after mutationsToBeSent()
2. the number of active workers may stuck below the configured asyncMutatorWorker count until flush is called.

sduskis:
@weiminyu:
1. I agree that there's an issue.  There's going to be a bigger issue if heapSizeManager is full before a new call to .registerOperationWithHeapSize(); in such a case registerOperationWithHeapSize() blocks until an asynchronous operation completes.  Those operations could block forever activeMutationWorkers.get()==0.  There's a trade off here, but I think I'd prefer it if the check happens before .registerOperationWithHeapSize(). 

2. I'll change the code to:
      if (activeMutationWorkers.get() < options.getAsyncMutatorCount()) {
        initializeAsyncMutators();
      }

sduskis:
I addressed all of the comments.  PTAL.

kevinsi4508:
LGTM

PR #634, opened 2016-01-20
Using a single Connection for CloudBigtableConnectionPool.

sduskis:
Moving the BigtableBufferedMutator offer() method.  This should also fix some of the issues in #630.  Each Connection opens up multiple channgles

### DIFF
--- a/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConnectionPool.java
+++ b/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConnectionPool.java
@@ -16,11 +16,8 @@
 package com.google.cloud.bigtable.dataflow;
 
 import java.io.IOException;
-import java.util.Iterator;
-import java.util.Set;
-import java.util.concurrent.Callable;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
+import java.util.HashMap;
+import java.util.Map;
 
 import org.apache.hadoop.conf.Configuration;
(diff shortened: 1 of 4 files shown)


bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConnectionPool.java
```diff
@@ -113,46 +76,20 @@ public PoolEntry getConnection(Configuration config) throws IOException {
...
-  }
-
-  public synchronized void returnConnection(PoolEntry entry) {
-    if (entry.isExpired()) {
-      closeAsynchronously(entry);
-    } else {
-      connections.put(entry.getKey(), entry);
-    }
+  @SuppressWarnings("unused")
+  protected Connection createConnection(Configuration config, String key) throws IOException {
```
kevinsi4508: remove "sting key"?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConnectionPool.java
```diff
@@ -51,21 +44,10 @@
   public static class PoolEntry {
     private final String key;
```
kevinsi4508: Do we still need "key" or "PoolEntry" itself?

sduskis:
I removed PoolEntry.  I removed the key variable from createConnection.  I kept createConnection so that the test can override it.

kevinsi4508:
LGTM

PR #635, opened 2016-01-20
Removing the buffered mutator "worker" behavior from dataflow.

### LINKED ISSUE #630: DataFlow CloudBigtableIO leaks threads

sduskis:
This is the quickest way to fix #630.  It will have some potentially
negative performance impacts on dataflow writes compared to what
performance could be.  That said, there are other issues with
dataflow/bigtable that need solving in bigtable/dataflow that need
solving first.

### DIFF
--- a/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
+++ b/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
@@ -178,6 +178,13 @@ public class CloudBigtableConfiguration implements Serializable {
    */
   public Configuration toHBaseConfig() {
     Configuration config = new Configuration();
+
+    // This setting can potentially decrease performance for large scale writes. However, this
+    // setting prevents problems that occur when streaming Sources, such as PubSub, are used.
+    // To override this behavior, call:
+    //    Builder.withConfiguration(BigtableOptionsFactory.BIGTABLE_ASYNC_MUTATOR_COUNT_KEY, 
+    //                              BigtableOptions.BIGTABLE_ASYNC_MUTATOR_COUNT_DEFAULT);
+    config.set(BigtableOptionsFactory.BIGTABLE_ASYNC_MUTATOR_COUNT_KEY, "0");
     for (Entry<String, String> entry : configuration.entrySet()) {
       config.set(entry.getKey(), entry.getValue());
     }
--- a/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
+++ b/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
@@ -72,7 +72,7 @@ public class BigtableBufferedMutator implements BufferedMutator {
     }
   }
 
-  private static class MutationOperation {
+  private class MutationOperation {
     final Mutation mutation;
     final long operationId;
     final boolean isCloseMarker;
@@ -82,17 +82,28 @@ public class BigtableBufferedMutator implements BufferedMutator {
       this.operationId = operationId;
       this.isCloseMarker = isClosedMarker;
     }
+    
+    public void run() {
+      ListenableFuture<? extends GeneratedMessage> request =
+          issueRequest(mutation, operationId);
+      ExceptionCallback callback = new ExceptionCallback(mutation);
+      Futures.addCallback(request, callback);
+    }
   }
 
-  private static ExecutorService createDefaultExecutorService() {
-    return Executors.newCachedThreadPool(
-            new ThreadFactoryBuilder()
-                .setNameFormat("BigtableBufferedMutator-worker-%s")
-                .setDaemon(true)
-                .build());
+  private static ExecutorService createExecutorService(BigtableOptions options) {
+    if (options.getAsyncMutatorCount() > 0) {
+      return Executors.newCachedThreadPool(
+              new ThreadFactoryBuilder()
+                  .setNameFormat("BigtableBufferedMutator-worker-%s")
+                  .setDaemon(true)
+                  .build());
+    } else {
+      return null;
+    }
   }
 
-  private final static MutationOperation CLOSE_MARKER = new MutationOperation(null, -1, true);
+  private final MutationOperation CLOSE_MARKER = new MutationOperation(null, -1, true);
   private final Configuration configuration;
 
   /**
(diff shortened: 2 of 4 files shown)


bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
```diff
@@ -178,6 +178,13 @@ public BigtableOptions toBigtableOptions() throws IOException {
    */
   public Configuration toHBaseConfig() {
     Configuration config = new Configuration();
+
```
kevinsi4508: Instead of this, why don't we set BIGTABLE_ASYNC_MUTATOR_COUNT_DEFAULT to 0?

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -137,7 +144,6 @@ public void run() {
       activeMutationWorkers.incrementAndGet();
       try {
         while (!executorService.isShutdown()) {
-          Mutation mutation = null;
```
kevinsi4508: Why is this removed?

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -202,7 +198,9 @@ public BigtableBufferedMutator(
         heapSizeManager.getMaxInFlightRpcs());
     this.options = options;
     this.heapSizeManager = heapSizeManager;
-    this.executorService = executorService;
+    this.executorService =
```
kevinsi4508: This is a bit weird, this depends on "executorService" & "AsyncMutatorCount".
Why don't we always depend on AsyncMutatorCount. It it simpler:

public BigtableBufferedMutator(...)
  this(... options.getAsyncMutatorCount() > 0 ? createDefaultExecutorService() : null);
}

@VisibleForTesting
public BigtableBufferedMutator(...)
  ...
  this.executorService = executorService;
}

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -145,22 +151,13 @@ public void run() {
             if (operation == null || operation.isCloseMarker) {
               break;
             }
-            mutation = operation.mutation;
-            ListenableFuture<? extends GeneratedMessage> request =
-                issueRequest(operation.mutation, operation.operationId);
-            ExceptionCallback callback = new ExceptionCallback(operation.mutation);
-            Futures.addCallback(request, callback);
+            operation.run();
```
kevinsi4508: Update doc to reflect this change.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
sduskis: I wanted to turn off this behavior for Dataflow by default, but keep it on everywhere else.

kevinsi4508: Ok, please note a note here then.
Also, the override should use value: BIGTABLE_ASYNC_MUTATOR_COUNT_DEFAULT
not BigtableOptions.BIGTABLE_DATA_CHANNEL_COUNT_DEFAULT

sduskis: Oh.  You mean the comment.  Woops.

sduskis: Fixed.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
sduskis: it turned out to be superfluous.  issueRequest() handles the Exception case already.  This logic will never be invoked.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
sduskis: done.

kevinsi4508:
LGTM

PR #636, opened 2016-01-21
Using a single executor for buffered mutator per connection.

sduskis:
This should be the last PR for #630.

### DIFF
--- a/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
+++ b/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
@@ -19,7 +19,6 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -46,11 +45,9 @@ import com.google.cloud.bigtable.grpc.BigtableDataClient;
 import com.google.cloud.bigtable.grpc.async.AsyncExecutor;
 import com.google.cloud.bigtable.grpc.async.HeapSizeManager;
 import com.google.cloud.bigtable.hbase.adapters.HBaseRequestAdapter;
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
-import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.google.protobuf.GeneratedMessage;
 
 /**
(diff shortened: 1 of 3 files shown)


bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -91,7 +90,7 @@ public void run() {
     }
   }
 
-  private static ExecutorService createExecutorService(BigtableOptions options) {
+  public static ExecutorService createExecutorService(BigtableOptions options) {
```
kevinsi4508: createBufferedMutatorExecutorService
and add some doc

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -178,18 +178,18 @@ public BigtableBufferedMutator(
       BufferedMutator.ExceptionListener listener,
       HeapSizeManager heapSizeManager) {
     this(client, adapter, configuration, options, listener, heapSizeManager,
-        createExecutorService(options));
+        createExecutorService(options), true);
   }
 
-  @VisibleForTesting
-  BigtableBufferedMutator(
+  public BigtableBufferedMutator(
```
kevinsi4508: Is this constructor for testing only? if yes, annotate it with @VisibleForTesting

sduskis: We're actually only using this constructor.  I'll remove the other one.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
sduskis: I'll move this to AbstractBigtableConnection and the test.

bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -184,6 +185,10 @@ public BufferedMutator getBufferedMutator(BufferedMutatorParams params) throws I
 
     final long id = SEQUENCE_GENERATOR.incrementAndGet();
 
+    if (this.bufferedMutatorExecutorService == null) {
+      this.bufferedMutatorExecutorService = BigtableBufferedMutator.createExecutorService(options);
```
kevinsi4508: this is created, but never used (except .shutdown())?

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
sduskis: Done

sduskis: I'll add docs here.  Isn't BigtableBufferedMutator. createBufferedMutatorExecutorService() redundant?

bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
sduskis: I fixed that.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
kevinsi4508: You are right, "ExecutorService" alone is kind of generic. "WorkerExecutorService"?

kevinsi4508:
LGTM

Issue #638, opened 2016-01-22
Threads hang on put in closed/closing/opening


yaitskov:
I have 3 threads stuck with the same stack trace.
Version is 0.2.2

I launched process inserting data into 1 table over 1 connection with 6 threads.
Right after that I truncated table with hbase shell.
I saw in log complains that table is not found.

```
exception #1548 from bigtable.googleapis.com for \x00\x04\x1A\x8F\x00\x00\x01H\x06\xD0}\x80A\x95
io.grpc.StatusRuntimeException: NOT_FOUND: Failed to read table: table1
        at io.grpc.Status.asRuntimeException(Status.java:430)
        at io.grpc.stub.ClientCalls$StreamObserverToCallListenerAdapter.onClose(ClientCalls.java:266)
        at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$3.run(ClientCallImpl.java:320)
        at io.grpc.internal.SerializingExecutor$TaskRunner.run(SerializingExecutor.java:154)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
        at java.lang.Thread.run(Thread.java:745)
exception #1549 from bigtable.googleapis.com for \x00\x04\x1A\x8F\x00\x00\x01H\x06\xD0}\x80EU
io.grpc.StatusRuntimeException: UNKNOWN
        at io.grpc.Status.asRuntimeException(Status.java:430)
        at io.grpc.stub.ClientCalls$StreamObserverToCallListenerAdapter.onClose(ClientCalls.java:266)
        at io.grpc.ClientInterceptors$CheckedForwardingClientCall.start(ClientInterceptors.java:172)
        at io.grpc.stub.ClientCalls.startCall(ClientCalls.java:193)
        at io.grpc.stub.ClientCalls.asyncUnaryRequestCall(ClientCalls.java:173)
        at io.grpc.stub.ClientCalls.asyncUnaryRequestCall(ClientCalls.java:163)
        at io.grpc.stub.ClientCalls.asyncUnaryCall(ClientCalls.java:68)
        at com.google.cloud.bigtable.grpc.io.ClientCallService$1.listenableAsyncCall(ClientCallService.java:94)
[...]

PR #639, opened 2016-01-22
Making sure BatchExecutor doesn't hang when a connection is closed.

sduskis:
If a connection is closed, async executor will throw an exception, but only after it registered in the HeapSizeManager; the operation doesn't unregister in that case.  That leads to the client hanging forever.  Returning a future that immediately fails solves this problem.  The failure informs the FutureCallable from the HeapSizeManager that an error occured and that the operation can be marked as complete.
This fixes a problem a user found in #638 .

### DIFF
--- a/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestBatch.java
+++ b/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestBatch.java
@@ -15,11 +15,20 @@
  */
 package com.google.cloud.bigtable.hbase;
 
-import static com.google.cloud.bigtable.hbase.IntegrationTests.*;
+import static com.google.cloud.bigtable.hbase.IntegrationTests.COLUMN_FAMILY;
+import static com.google.cloud.bigtable.hbase.IntegrationTests.TABLE_NAME;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Random;
 
 import org.apache.commons.lang.ArrayUtils;
 import org.apache.hadoop.hbase.CellUtil;
 import org.apache.hadoop.hbase.client.Append;
+import org.apache.hadoop.hbase.client.Connection;
+import org.apache.hadoop.hbase.client.ConnectionFactory;
 import org.apache.hadoop.hbase.client.Delete;
 import org.apache.hadoop.hbase.client.Get;
 import org.apache.hadoop.hbase.client.Increment;
@@ -32,13 +41,10 @@ import org.apache.hadoop.hbase.client.Table;
 import org.apache.hadoop.hbase.regionserver.NoSuchColumnFamilyException;
 import org.apache.hadoop.hbase.util.Bytes;
 import org.junit.Assert;
+import org.junit.Rule;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
-
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Random;
+import org.junit.rules.ExpectedException;
 
 public class TestBatch extends AbstractTest {
   /**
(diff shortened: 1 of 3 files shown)


bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestBatch.java
```diff
@@ -366,4 +375,12 @@ public void testBatchGets() throws Exception {
 
     table.close();
   }
+
+  public void testBatchDoesntHang() throws Exception {
+    Connection closedConnection = ConnectionFactory.createConnection(IntegrationTests.getConfiguration());
+    Table table = closedConnection.getTable(TABLE_NAME);
+    closedConnection.close();
+    thrown.expect(RetriesExhaustedWithDetailsException.class);
```
kevinsi4508: Can we verify the cause of the exception? connection closed related.

sduskis: I guess so.  The point here is that the exception is thrown and the put doesn't hang forever.

kevinsi4508: The reason I prefer checking the cause (even it is IllegalStateException) is that we set up a precondition that will eventually force the code to throw some exception, we want to to verify the exception is thrown from those async calls not somewhere else.

For example:

batch{
throw e1; // unexpected exception that could be caught in our test code.
try {
  async  aclls
} catch {
  throw e2; // expected exception to be caught in our test code.
}

} // batch

If you believe checking IllegalStateException could be confusing, it is something we have fix, not a reason for us not to check it.

kevinsi4508:
The test part is only concern, otherwise LGTMed.

bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestBatch.java
sduskis: The reason not to check for a specific type of exception HERE is that it's not part of this specific test.  If we want to observe specific types of behavior for .close() exceptions, I would ask that we do it somewhere else.  This test is specifically made to test .put() batch operation not specific .close() behavior.

sduskis: If the specifics of the exception on .close() change, it should not require changes of Batch testing, which is what we're testing here.  Getting into the specific details of the .close() exception here would require future changes.  I'd like to avoid those future changes and test specific behavior in unit tests, if that's ok.

sduskis: Of note, we'd have to make sure that both Cloud Bigtable and HBase both show the same underlying behavior.

kevinsi4508:
Let's make it a nit. In general, I want to make sure the exception is the exact exception was set up, but not absolutely needed.

LGTM

sduskis:
It turns out that the test wasn't actually executing before.  I had to add @Test to make it run.  The test currently fails, so I'm going to fix that.  This will require another review once that's done.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
```diff
@@ -155,33 +156,47 @@ public BatchExecutor(
...
     LOG.trace("issueRowRequest(Row, Batch.Callback, Object[], index");
-    ListenableFuture<? extends GeneratedMessage> future = issueAsyncRequest(row);
-
     SettableFuture<Result> resultFuture = SettableFuture.create();
-    results[index] = null;
     RpcResultFutureCallback<T> futureCallback =
         new RpcResultFutureCallback<T>(row, callback, index, results, resultFuture);
-    Futures.addCallback(future, futureCallback, service);
+    results[index] = null;
+    if (service.isShutdown()) {
```
sduskis: Kevin, can you please checkout this change?   This is where the failure will happen.

kevinsi4508: This looks good.
Is there a unit test where we test this behavior?

if (closed)
  expect ConnectionClosedException
else
  expect async all

sduskis:
I added a unit test, and updated the integration test with more details.  I got rid of ConnectionClosedException because it's not an HBase Exception; HBase uses an IOException, so I think it's appropriate that we do as well.

sduskis:
I think that things look a lot better now with the additional testing.

kevinsi4508:
LGTM

PR #640, opened 2016-01-22
This is a performance fix for cases with many open Connections.

sduskis:
We only need to load credentials once for all connections if we're doing default credential lookups.  This PR caches the value and reuses it across connections.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -55,7 +55,9 @@ import com.google.common.base.Preconditions;
 import com.google.api.client.util.Strings;
 import com.google.cloud.bigtable.config.BigtableOptions;
 import com.google.cloud.bigtable.config.CredentialFactory;
+import com.google.cloud.bigtable.config.CredentialOptions;
 import com.google.cloud.bigtable.config.Logger;
+import com.google.cloud.bigtable.config.RetryOptions;
 import com.google.cloud.bigtable.grpc.io.ChannelPool;
 import com.google.cloud.bigtable.grpc.io.CredentialInterceptorCache;
 import com.google.cloud.bigtable.grpc.io.HeaderInterceptor;
@@ -67,7 +69,6 @@ import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.ListeningExecutorService;
 import com.google.common.util.concurrent.MoreExecutors;
-import com.google.common.util.concurrent.ThreadFactoryBuilder;
 
 /**
  * <p>Encapsulates the creation of Bigtable Grpc services.</p>
@@ -150,11 +151,7 @@ public class BigtableSession implements AutoCloseable {
   private static void performWarmup() {
     // Initialize some core dependencies in parallel.  This can speed up startup by 150+ ms.
     ExecutorService connectionStartupExecutor =
-        Executors.newCachedThreadPool(
-            new ThreadFactoryBuilder()
-                .setNameFormat("BigtableSession-startup-%s")
-                .setDaemon(true)
-                .build());
+        Executors.newCachedThreadPool(createThreadFactory("BigtableSession-startup"));
 
     connectionStartupExecutor.execute(new Runnable() {
       @Override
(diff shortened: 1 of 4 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -279,35 +281,14 @@ public BigtableSession(BigtableOptions options, @Nullable ExecutorService batchP
...
-      Preconditions.checkState(credentials instanceof OAuth2Credentials, String.format(
-        "Credentials must be an instance of OAuth2Credentials, but got %s.", credentials.getClass()
-            .getName()));
-      RefreshingOAuth2CredentialsInterceptor oauth2Interceptor =
-          new RefreshingOAuth2CredentialsInterceptor(this.batchPool,
-              (OAuth2Credentials) credentials, this.options.getRetryOptions());
-      oauth2Interceptor.asyncRefresh();
-      headerInterceptorBuilder.add(oauth2Interceptor);
-    }
+    get(credentialsFuture, "Could not initialize credentials");
```
mgarolera: Comment why the get but ignore the result?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -330,6 +311,58 @@ public BigtableTableAdminClient call() throws Exception {
...
+      protected HeaderInterceptor getCredentialsInterceptor() throws IOException, GeneralSecurityException {
+        CredentialOptions credentialOptions = BigtableSession.this.options.getCredentialOptions();
+
+        // Default credentials is the most likely CredentialType. It's also the only CredentialType
+        // that can be safely cached.
+        boolean isDefaultCredentials =
+            credentialOptions.getCredentialType() == CredentialType.DefaultCredentials;
+
+        synchronized (CREDENTIALS_REFRESH_EXECUTOR_SERVICE) {
+          if (isDefaultCredentials && defaultCredentialOauth != null) {
```
mgarolera: Move this out of synchronized?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -330,6 +311,58 @@ public BigtableTableAdminClient call() throws Exception {
...
+            return null;
+          }
+          Preconditions.checkState(credentials instanceof OAuth2Credentials,
+            String.format("Credentials must be an instance of OAuth2Credentials, but got %s.",
+              credentials.getClass().getName()));
+          RefreshingOAuth2CredentialsInterceptor oauth2Interceptor =
+              new RefreshingOAuth2CredentialsInterceptor(CREDENTIALS_REFRESH_EXECUTOR_SERVICE,
+                  (OAuth2Credentials) credentials,
+                  BigtableSession.this.options.getRetryOptions());
+          oauth2Interceptor.asyncRefresh();
```
mgarolera: Explain why we trigger asyncRefresh after creation?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
sduskis: Correct.  I wanted to have only a single defaultCredentialsOauth.

mgarolera: Why not synchronize on defaultCredentialOauth instead?

sduskis:
I moved the credentials caching and interceptor creation to a new class.  I also added comments to BigtableSession.  PTAL.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -16,6 +16,8 @@
 
 package com.google.cloud.bigtable.grpc;
 
+import static com.google.cloud.bigtable.util.ThreadPoolUtil.createThreadFactory;
```
mgarolera: Why not import ThreadPoolUtil

and then use ThreadPoolUtil.createThreadFactory?

sduskis: Done.  I don't have a preference in this case.

mgarolera:
LGTM

sduskis:
I meant to squash commits first.  Woops.

PR #641, opened 2016-01-24
Dataflow configs should not load xml files by default.

sduskis:
Loading the xml files causes look ups in .jar files for .xml files. That causes problems if it's done too often.  Dataflow doesn't have those files anyway, so don't look them up to avoid locking issues found in #630.

### DIFF
--- a/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConfiguration.java
+++ b/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConfiguration.java
@@ -44,7 +44,7 @@ public class BigtableConfiguration {
   }
 
   public static Configuration configure(String projectId, String zoneName, String clusterName) {
-    Configuration config = new Configuration();
+    Configuration config = new Configuration(false);
     config.set(BigtableOptionsFactory.PROJECT_ID_KEY, projectId);
     config.set(BigtableOptionsFactory.ZONE_KEY, zoneName);
     config.set(BigtableOptionsFactory.CLUSTER_KEY, clusterName);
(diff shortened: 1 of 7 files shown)


bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConfiguration.java
```diff
@@ -44,7 +44,7 @@
   }
 
   public static Configuration configure(String projectId, String zoneName, String clusterName) {
-    Configuration config = new Configuration();
```
kevinsi4508: This is not dataflow specific. Are you sure we want to disable loading default?
I don't know the difference, just want to make sure this change is OK.

sduskis: A user observed that there are lots of hanging threads in https://github.com/GoogleCloudPlatform/cloud-bigtable-client/issues/630#issuecomment-174023573.  It indicates that new Configuration() tries to find XML files that may be in jars.  That's expensive in terms of time and locking behavior.

Our intended default behavior in dataflow is to have programmatic configuration of the connection and to avoid lookups in xml files, since XML files are harder to set up for this situation.

See https://hadoop.apache.org/docs/r2.6.1/api/org/apache/hadoop/conf/Configuration.html#Configuration(boolean) for more details.

kevinsi4508: my question was is this configure() in BigtableConfiguration.java for dataflow only?

sduskis: Gotcha.  We needed a programmatic way to configure Dataflow, and this was the way we did it.  It may make sense to discuss additional programmatic idioms for bigtable, but I hope that such discussions are out of scope for this specific change.

kevinsi4508: No discussion needed, I just wanted a confirmation from you,

kevinsi4508:
LGTM

PR #645, opened 2016-01-26
Limiting the number of Dataflow splits to 4,000.

sduskis:
That was the recommendation of the dataflow team.

### DIFF
--- a/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
+++ b/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
@@ -144,6 +144,7 @@ public class CloudBigtableIO {
   static class Source extends BoundedSource<Result> {
     private static final long serialVersionUID = -5580115943635114126L;
     private static final Logger SOURCE_LOG = LoggerFactory.getLogger(Source.class);
+    private static final long MAX_SPLIT_COUNT = 4_000;
 
     /**
      * Configuration for a Cloud Bigtable connection, a table, and an optional scan.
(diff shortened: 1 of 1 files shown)


bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -381,6 +382,17 @@ public boolean start() throws IOException {
     @Override
     public List<SourceWithKeys> splitIntoBundles(
         long desiredBundleSizeBytes, PipelineOptions options) throws Exception {
+      // Update the desiredBundleSizeBytes in order to limit the
+      // number of splits to maximumNumberOfSplits.
```
vphani: clarification: what is maximumNumberOfSplits ?

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -381,6 +382,17 @@ public boolean start() throws IOException {
...
+      long sizeEstimate = getEstimatedSizeBytes(options);
+      desiredBundleSizeBytes = Math.max(sizeEstimate / MAX_SPLIT_COUNT, desiredBundleSizeBytes);
```
vphani: I see a few concerns, nothing major but atleast we need to document this (is a major concern if we should NEVER return more than MAX_SPLIT_COUNT to dataflow):
- This will still return at a minimum number of tables which could be > MAX_SPLIT_COUNT
- You are using the number of splits as (in split() function) as splitCount = (int) Math.ceil((double) (regionSize) / (double) (desiredBundleSizeBytes)); since it is the ceiling we will always get higher number of splits, say regionSize/desiredBundleSize is 0.75 then we will consistently return 2 splits per region and we will overflow MAX_SPLIT_SIZE.

sduskis: The hard limit is 20MB worth of serialized Source objects.  4,000 is a rough estimate.

Factoring in a 1/.75 inflation (33%), that leaves each serialized source to be a max of 3,750 bytes.  I did a quick test with a basic source.  It serialized to about 1K.  I had to work hard to get a complex enough FilterList that made it near 2K.

I think we're pretty safe at 4,000 as a really rough estimate.

sduskis: @vphani raised a concern about row key size, which can be up to 64KB each.  We'll have to note that and fix that scenario in the future.

PR #648, opened 2016-01-27
Using a shared set of thread pools for multiple connections

sduskis:
There are cases where users create multiple connections in the same VM.  They either did it for (hopefully) legacy performance issues or this was done through a complex distributed job, such as a dataflow job.  Either way, each connection doesn't need its own thread pool; the thread pools are auto-expanding, so if more threads are needed, they will automatically be allocated.  Shared thread pools will be much more efficient in such cases.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
@@ -16,7 +16,6 @@
 package com.google.cloud.bigtable.grpc;
 
 import io.grpc.CallOptions;
-import io.grpc.Channel;
 import io.grpc.ClientCall;
 import io.grpc.Metadata;
 import io.grpc.MethodDescriptor;
@@ -47,7 +46,6 @@ import com.google.cloud.bigtable.grpc.async.BigtableAsyncUtilities;
 import com.google.cloud.bigtable.grpc.async.RetryableRpc;
 import com.google.cloud.bigtable.grpc.io.CancellationToken;
 import com.google.cloud.bigtable.grpc.io.ChannelPool;
-import com.google.cloud.bigtable.grpc.io.ChannelPool.PooledChannel;
 import com.google.cloud.bigtable.grpc.io.ClientCallService;
 import com.google.cloud.bigtable.grpc.io.RetryingCall;
 import com.google.cloud.bigtable.grpc.scanner.BigtableResultScannerFactory;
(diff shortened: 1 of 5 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -256,53 +233,60 @@ public BigtableSession(BigtableOptions options, @Nullable ExecutorService batchP
...
+    try {
+      // Looking up Credentials takes time. Creating the retry executor and the EventLoopGroup don't
+      // take as long, but still take time. Get the credentials on one thread, and start up the elg
+      // and scheduledRetries thread pools on another thread.
+      Future<Void> credentialsFuture =
+          initializerExecutor.submit(createCredentialsCallback(headerInterceptorBuilder));
+
+      BigtableSessionThreadPools.Builder threadPoolsBuilder =
+          new BigtableSessionThreadPools.Builder();
+      if (elg != null) {
```
mgarolera: Are these nullable parameters ever used? would be nice to be able to drop them. It reduces functionaily, but it could simplify code. There would be no need for the BigtableSessionThreadPools and we could use  BigtableSessionSharedThreadPools directly

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSessionThreadPools.java
```diff
@@ -0,0 +1,171 @@
...
+    this.terminateScheduledRetries = terminateScheduledRetries;
+    this.decrementFromSharedThreadPools = decrementFromSharedThreadPools;
+  }
+
+  public ExecutorService getBatchPool() {
+    return batchPool;
+  }
+
+  public EventLoopGroup getElg() {
+    return elg;
```
mgarolera: Why not leave null in elg if its not specified in the constructor. And in the accessor return this.elg or shared.getElg(). This way you dont need the bools, you just close this.elg if its not null.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSessionSharedThreadPools.java
```diff
@@ -0,0 +1,110 @@
...
+  private BigtableSessionSharedThreadPools() {
+  }
+
+  /**
+   * Increment the counter. This will also allocate resource if they have not been previously
+   * initialized.
+   */
+  public synchronized void incrementCounter() {
+    int usageCount = sessionCount.incrementAndGet();
+    if (usageCount == 1) {
```
mgarolera: Why not create them at the constructor? is it necessary to close them?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
sduskis: Some users might be using them.

sduskis: We're going to deprecate this constructor, and close the elg and retry pool right away.  Users should not  be using this.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSessionThreadPools.java
sduskis: I'm getting rid of BigtableSessionThreadPools.  We only need the shared class now.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSessionSharedThreadPools.java
sduskis: If they aren't used anymore, closing them is a nice to have.

sduskis:
I removed all of the user supplied threadpool management.  I still do want to keep the cleanup process, if that's OK.

PR #651, opened 2016-01-28
Removing "reservation" concept from ChannelPool.

sduskis:
We "reserved" a channel for scans due to a gRPC bug in 0.9.0 and below which disallowed multiple requests on a channel that was streaming. We upgrade gRPC versions and no longer require the hacky approach.
Adding a test to make sure that scan/deletes work.

### DIFF
--- a/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClientTests.java
+++ b/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClientTests.java
@@ -23,9 +23,6 @@ import static org.mockito.Matchers.same;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
-import io.grpc.CallOptions;
-import io.grpc.ClientCall;
-import io.grpc.MethodDescriptor;
 
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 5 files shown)


bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClientTests.java
```diff
@@ -93,8 +94,8 @@ public void setup() {
         options, clientCallService);
     when(channelPool.newCall(any(MethodDescriptor.class), any(CallOptions.class))).thenReturn(
       clientCall);
-    when(channelPool.reserveChannel()).thenReturn(pooledChannel);
-    when(pooledChannel.newCall(any(MethodDescriptor.class), any(CallOptions.class))).thenReturn(
+//    when(channelPool.reserveChannel()).thenReturn(pooledChannel);
```
kevinsi4508: remove

bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestScan.java
```diff
@@ -167,4 +164,55 @@ public void test100ResultsInScanner() throws IOException {
     table.delete(deletes);
     table.close();
   }
+
+  @Test
+  /**
+   *
```
kevinsi4508: The test looks good. Can you document here a bit, something like "Tests row deletion while iterating through a scan result."

kevinsi4508:
LGTM

bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClientTests.java
sduskis: done.

bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestScan.java
sduskis: done.

PR #652, opened 2016-01-29
Removing ReconnectingChannel now that the channels automatically reconnect

sduskis:
I'm running a long running test right now.  It ran for an hour so far, and I saw a reconnect happen automatically.  It looks like the problems we previously saw with gRPC and the server side were fixed.  We no longer need our reconnection hack.  I'm going to let my test run for 12+ hours before merging the PR.


sduskis:
It's been running for 15 hours without a problem.

kevinsi4508:
The code looks good. How do you do testing to make sure the problem we have seen is no longer there?

sduskis:
I do a single put every second in a loop.  I ran that in a GCE shell script for over 15 hours before the ssh session dropped.  Every hour, there was a notification that there was a gRPC level connection termination and reconnect.  All puts completed and no errors were thrown in those 15 hours.

kevinsi4508:
Can you also confirm with gRPC team that the fix is in the version we are using? or you already have done it?

sduskis:
This wasn't a fix for gRPC, but rather for the networking stack.  We'll have to figure out who to speak to about this.

sduskis:
We discussed this offline.  We're going to release the 0.2.3 client and then merge this PR.

PR #653, opened 2016-01-29
Using a shared set of thread pools across multiple connections

sduskis:
There are cases where users create multiple connections.  They either did it for (hopefully) legacy performance issues or this was done through a complex distributed job, such as a dataflow job.  Either way, each connection doesn't need its own thread pool; the thread pools are auto-expanding, so if more threads are needed, they will automatically be allocated.  Shared thread pools will be much more efficient in such cases.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -22,7 +22,6 @@ import io.grpc.netty.GrpcSslContexts;
 import io.grpc.netty.NegotiationType;
 import io.grpc.netty.NettyChannelBuilder;
 import io.netty.channel.EventLoopGroup;
-import io.netty.channel.nio.NioEventLoopGroup;
 import io.netty.handler.ssl.OpenSsl;
 import io.netty.handler.ssl.SslContext;
 import io.netty.handler.ssl.SslContextBuilder;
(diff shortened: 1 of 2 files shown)


sduskis:
I'm not sure what happened to the previous PR.  I think I messed things up a bit.  I removed any reference to user based thread pools, and all of the management is done by us.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -200,109 +194,106 @@ public void run() {
...
-  }
-
-  public BigtableSession(BigtableOptions options, @Nullable ExecutorService batchPool,
-      @Nullable EventLoopGroup elg, @Nullable ScheduledExecutorService scheduledRetries)
-      throws IOException {
     Preconditions.checkArgument(
         !Strings.isNullOrEmpty(options.getProjectId()), PROJECT_ID_EMPTY_OR_NULL);
-    Preconditions.checkArgument(
-        !Strings.isNullOrEmpty(options.getZoneId()), ZONE_ID_EMPTY_OR_NULL);
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(options.getZoneId()), ZONE_ID_EMPTY_OR_NULL);
```
mgarolera: nit: Not sure why this changed here but I slightly prefer it the old way, matching the other checkts. Up to you.

sduskis:
I updated the PR. I removed the increment/decrement logic.  The threadpools will stay around until the JVM dies. Users can customize the threadpooling by creating their own BigtableSessionSharedThreadPools implementation.

mgarolera:
LGTM, asides from the setInstance we discussed.

sduskis:
I removed setInstance. Merging.

PR #654, opened 2016-01-29
Updating bigtable data protos.  Also updating the cluster API protos.

sduskis:


bigtable-protos/src/main/proto/google/bigtable/admin/cluster/v1/bigtable_cluster_data.proto
```diff
@@ -91,4 +86,8 @@ enum StorageType {
 
   // Data will be stored in SSD, providing low and consistent latencies.
   STORAGE_SSD = 1;
+
```
dmmcerlean: Are we ready to expose this yet?

dmmcerlean:
LGTM except for the STORAGE_HDD comment

sduskis:
As per @vphani offline, we're ready to expose STORAGE_HDD.

PR #659, opened 2016-02-02
Removing override Ip and supporting host based lookups.

sduskis:
gRPC 0.10+ introduced a method of configuration via host strings rather than addresses.  That allows changes in the ip address to be updated at run time.  We lose the ability to override the host, but that hasn't been used in a long time.  If we need that feature again, we'll add it back in the future.

@kevinsi4508, @BrendanGleason would like this to get this feature into 0.2.3.  I'm going to hold off the release until we get this right.


kevinsi4508:
LGTM

PR #661, opened 2016-02-03
Fix integration test setup for dataflow based sequence file importer.

weiminyu:
importer integration test did not set dataflow's project id, which causes tests to fail on jenkins.


sduskis:
LGTM

PR #666, opened 2016-02-05
Fixing a bigtable to hbase translation issue.

sduskis:
Go Clients default to no garbage collection policy. A project breaks the hbase shell 'list' command

### DIFF
--- a/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/ColumnDescriptorAdapter.java
+++ b/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/ColumnDescriptorAdapter.java
@@ -170,7 +170,9 @@ public class ColumnDescriptorAdapter {
     if (buffer.length() != 0) {
       buffer.append(" || ");
     }
-    buffer.append(String.format("(version() > %s)", maxVersions));
+    if (maxVersions != Integer.MAX_VALUE) {
+      buffer.append(String.format("(version() > %s)", maxVersions));
+    }
 
     return buffer.toString();
   }
@@ -194,8 +196,9 @@ public class ColumnDescriptorAdapter {
   private static void convertGarbageCollectionExpression(String gcExpression,
       HColumnDescriptor columnDescriptor) {
     if (Strings.isNullOrEmpty(gcExpression)) {
-      // No GC Expression means keep all versions, MaxVersions = 0.
-      columnDescriptor.setMaxVersions(0);
+      // No GC expression in Bigtable means keeping all versions. HBase doesn't support keeping all
+      // version, so we translate it by setting the max version to Integer.MAX_VALUE.
+      columnDescriptor.setMaxVersions(Integer.MAX_VALUE);
       return;
     }
     String maxVersionExpression = null;
--- a/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/adapters/TestColumnDescriptorAdapter.java
+++ b/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/adapters/TestColumnDescriptorAdapter.java
@@ -191,10 +191,8 @@ public class TestColumnDescriptorAdapter {
 
   @Test
   public void testBlankExpression(){
-    // An empty gc expression means keep all versions, which is not supported by HBase.
-    expectedException.expectMessage("Maximum versions must be positive");
-    expectedException.expect(IllegalArgumentException.class);
-
-    adapter.adapt("family", asColumnFamily(""));
+    HColumnDescriptor descriptor = adapter.adapt("family", asColumnFamily(""));
+    Assert.assertEquals(Integer.MAX_VALUE, descriptor.getMaxVersions());
+    Assert.assertEquals("", ColumnDescriptorAdapter.buildGarbageCollectionExpression(descriptor));
   }
 }


bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/ColumnDescriptorAdapter.java
```diff
@@ -194,8 +194,9 @@ public static String buildGarbageCollectionExpression(HColumnDescriptor columnDe
   private static void convertGarbageCollectionExpression(String gcExpression,
       HColumnDescriptor columnDescriptor) {
     if (Strings.isNullOrEmpty(gcExpression)) {
-      // No GC Expression means keep all versions, MaxVersions = 0.
-      columnDescriptor.setMaxVersions(0);
+      // No GC Expression means keep all versions. HBase doesn't support this, but this is the
+      // default Bigtable behavior.
```
kevinsi4508: What is the default behavior? If it is default, why do we need to set anything?

sduskis: The default behavior of hbase is 1 version.  The default behavior of bigtable is infinite versions.  If we don't set anything, the user will see a maximum of 1 version, when the reality is infinite versions.

sduskis: See #577 for the introduction of the problem.

sduskis:
@mgarolera, I'm making you the reviewer, because essentially, I need to revert your PR.  I'd like your input on this.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/ColumnDescriptorAdapter.java
kevinsi4508: Discussed with Marcal. I understand what this is now.
- There are two type of adapters; one translates from HBase to Bigtable and one translates from Bigtable  to HBase; mixing them in the same package/file confused me.

- The change here basically makes the correctly translation from Bigtable (no gc set) to HBase (all versions). We need to make sure the reverse is correct: HBase (all versions) -> Bigtable (no gc set).

- a unit test that test the translation is reversible would be nice.

sduskis:
I added a change to make sure that "" converts to and back from a ColumnDescriptorAdapter.  PTAL.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/ColumnDescriptorAdapter.java
```diff
@@ -194,8 +196,9 @@ public static String buildGarbageCollectionExpression(HColumnDescriptor columnDe
   private static void convertGarbageCollectionExpression(String gcExpression,
       HColumnDescriptor columnDescriptor) {
     if (Strings.isNullOrEmpty(gcExpression)) {
-      // No GC Expression means keep all versions, MaxVersions = 0.
-      columnDescriptor.setMaxVersions(0);
+      // No GC Expression means keep all versions. HBase doesn't support this, but this is the
```
kevinsi4508: nit: No GC expression in Bigtable means keeping all versions. HBase doesn't support keeping all version, so we translate it by setting the max version to Integer.MAX_VALUE.

sduskis: Done.

bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/ColumnDescriptorAdapter.java
```diff
@@ -170,7 +170,9 @@ public static String buildGarbageCollectionExpression(HColumnDescriptor columnDe
     if (buffer.length() != 0) {
       buffer.append(" || ");
     }
-    buffer.append(String.format("(version() > %s)", maxVersions));
+    if (maxVersions != Integer.MAX_VALUE) {
```
kevinsi4508: nit: 
add comment: When maxVersions is set to Integer.MAX_VALUE, it means we want to keep all versions. By not providing GC expression, Bigtable keeps all versions.

bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/adapters/TestColumnDescriptorAdapter.java
```diff
@@ -191,10 +191,8 @@ public void badExpressionLessThan(){
 
   @Test
   public void testBlankExpression(){
```
kevinsi4508: testKeepAllVersions()  // to complete the conversion loop
1. blank gc expression, convert to HBase metadata
   verify that HBase max version is set to Integer.MAX_VALUE
2. convert the result from #1 to gc expression
  verify empty gc expression  
3. convert the result for #2 to HBase metadata again
  verify that the max version is still set to Integer.MAX_VALUE

kevinsi4508:
LGTM

bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/adapters/TestColumnDescriptorAdapter.java
sduskis: I have 1 and 2.  The result from #2 is just an empty string, so I don't think that #3 is necessary.

Renaming the test method and comiting.

PR #674, opened 2016-02-16
Using boring ssl for our tcnative tests.

sduskis:
This PR uses the newly available Boring ssl statically linked jar.

The tcnative test is invoked via:

mvn clean integration-test -PbigtableIntegrationTcnativeTest -Dgoogle.bigtable.project.id={PROJECT_ID} -Dgoogle.bigtable.zone.name={ZONE} -Dgoogle.bigtable.cluster.name={CLUSTER}

### DIFF
--- a/bigtable-hbase-integration-tests/pom.xml
+++ b/bigtable-hbase-integration-tests/pom.xml
@@ -104,8 +104,8 @@ limitations under the License.
                 </dependency>
                 <dependency>
                     <groupId>io.netty</groupId>
-                    <artifactId>netty-tcnative</artifactId>
-                    <version>1.1.33.Fork9</version>
+                    <artifactId>netty-tcnative-boringssl-static</artifactId>
+                    <version>1.1.33.Fork13</version>
                     <classifier>${os.detected.classifier}</classifier>
                 </dependency>
             </dependencies>


bigtable-hbase-integration-tests/pom.xml
```diff
@@ -104,8 +104,8 @@ limitations under the License.
                 </dependency>
                 <dependency>
                     <groupId>io.netty</groupId>
-                    <artifactId>netty-tcnative</artifactId>
-                    <version>1.1.33.Fork9</version>
+                    <artifactId>netty-tcnative-boringssl-static</artifactId>
+                    <version>1.1.33.Fork13</version>
                     <classifier>${os.detected.classifier}</classifier>
```
lesv: Is the classifier still needed?  I don't think it should be for what we were told.

sduskis: Classifier is still needed.  There is an binary library in each of the jars that's built specifically for each OS.  The major difference between boringssl and openssl is the fact that we don't need an additional installation of the openssl binary; the OS specific jar is always needed.

lesv:
Ah - I thought they had moved to a single library that figured out where you were running.  :(
LGTM

PR #675, opened 2016-02-17
Fixing a subtle bug in the dataflow CloudBigtableConnectionPool.

sduskis:
Users can use the pool in an unintended way, and close the connection. That creates issues if there are other Dataflow jobs on the same VM. Make sure that closing doesn't actually do anything.

### DIFF
--- a/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConnectionPool.java
+++ b/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConnectionPool.java
@@ -43,6 +43,15 @@ public class CloudBigtableConnectionPool {
   public CloudBigtableConnectionPool() {
   }
 
+  /**
+   * Gets a shared connection where the cluster name from the config is the key.
+   *
+   * <p>NOTE: Do not call close() on the connection, since it's shared.
+   *
+   * @param config
+   * @return
+   * @throws IOException
+   */
   public Connection getConnection(Configuration config) throws IOException {
     String key = BigtableOptionsFactory.fromConfiguration(config).getClusterName().toString();
     return getConnection(config, key);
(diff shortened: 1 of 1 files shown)


bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConnectionPool.java
```diff
@@ -60,6 +60,13 @@ protected synchronized Connection getConnection(Configuration config, String key
 
   @VisibleForTesting
   protected Connection createConnection(Configuration config) throws IOException {
-    return new BigtableConnection(config);
+    return new BigtableConnection(config) {
+      @Override
+      public void close() throws IOException {
+        // Users should not actually close the shared connection. Make sure that if a user does call
```
derjust: Just a thought: Add a WARN logging to let the user know that he is doing something wrong

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConnectionPool.java
```diff
@@ -60,6 +60,13 @@ protected synchronized Connection getConnection(Configuration config, String key
 
   @VisibleForTesting
   protected Connection createConnection(Configuration config) throws IOException {
```
derjust: I would love to see at least a comment on `getConnection` telling the user (developer) that he does not need to close the connection. Otherwise (out of usual practise) the user might be tempted/assume that he has do close the connection at the end of its DoFn or during `finsihBundle`

kevinsi4508: +1
comment on "public Connection getConnection" at the minimum.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConnectionPool.java
kevinsi4508: I don't know about printing warning. no-op seems to be fine.

sduskis:
I added comments.  PTAL.

kevinsi4508:
LGTM

PR #676, opened 2016-02-17
Removing the default batchPool from AbstractBigtableConnection.

sduskis:
The Abstract batchPool wasn't used effectively since the change in #653. This will hopefully reduce the number of threads in Pub/Sub Dataflow jobs.


bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -138,11 +134,7 @@ protected AbstractBigtableConnection(Configuration conf, boolean managed, Execut
     this.batchPool = pool;
```
mgarolera: It's a bit confusing that we use pool and batchPool.

mgarolera:
LGTM

PR #672, opened 2016-02-16
Creating a multi-Result Source.

sduskis:
This helps users who want to get multiple rows per batch so that they can do further operations on that array of rows.

PR #677, opened 2016-02-17
Some refactoring of CloudBigtableIO objects.

sduskis:
Moving the Reader class out of the Source class, so there's less nesting.
Using superclasses in tests and CloudBigtableIO to reduce leaking internals.
Adding createSourceWithKeys() instead of using new.

All of these changes are steps to getting to #672

### DIFF
--- a/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
+++ b/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
@@ -23,7 +23,6 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.NoSuchElementException;
-import java.util.concurrent.atomic.AtomicLong;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


bigtable-hbase-dataflow/src/test/java/com/google/cloud/bigtable/dataflow/CloudBigtableIOIntegrationTest.java
```diff
@@ -224,7 +225,7 @@ public void testEstimatedAndSplitForSmallTable() throws Exception {
         CloudBigtableScanConfiguration config =
             new CloudBigtableScanConfiguration(projectId, zoneId, clusterId,
                 tableName.getQualifierAsString(), new Scan());
-        CloudBigtableIO.Source source = new CloudBigtableIO.Source(config);
+        CloudBigtableIO.Source source = (Source) CloudBigtableIO.read(config);
```
mgarolera: Why the change from new Source to .read and cast?

sduskis: This does indeed sound like a code smell.  The idea of CloudBigtableIO.read() should just be to get the generic Dataflow object, but we're reaching under the covers here.

We're actually trying to test the sample row key processing, and need to reach under the covers to get to it.  I should really move the logic and this test to separate classes.  I'll add a TODO for this PR.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -566,13 +508,78 @@ public boolean producesSortedKeys(PipelineOptions options) throws Exception {
...
+        List<BoundedSource<Result>> result = new ArrayList<>();
         for (int i = 0; i < splitCount; i++) {
-          result.add(new SourceWithKeys(splitKeys[i], splitKeys[i + 1], regionSize));
+          result.add(createSourceWithKeys(splitKeys[i], splitKeys[i + 1], regionSize));
         }
         return result;
       }
     }
+
+    public BoundedSource<Result> createSourceWithKeys(byte[] startKey, byte[] stopKey, long size) {
```
kevinsi4508: package default, for testing only?

kevinsi4508: if not, can it be public static?

kevinsi4508:
LGTM

PR #682, opened 2016-02-18
ChannelPool Should create new channels in parallel.

sduskis:
Creating Channels in parallel is faster than creating them serially.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
@@ -15,8 +15,13 @@
  */
 package com.google.cloud.bigtable.grpc.io;
 
+import com.google.cloud.bigtable.grpc.BigtableSessionSharedThreadPools;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListeningExecutorService;
+import com.google.common.util.concurrent.MoreExecutors;
 
 import io.grpc.CallOptions;
 import io.grpc.Channel;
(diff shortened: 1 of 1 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -77,19 +85,49 @@ public ChannelPool(List<HeaderInterceptor> headerInterceptors, ChannelFactory fa
...
+   * @throws IOException if an exception occurred during Channel creation.
+   */
+  protected List<Channel> createNewChannels(int newChannelCount) throws IOException {
+    ExecutorService executorService = BigtableSessionSharedThreadPools.getInstance().getBatchThreadPool();
+    ListeningExecutorService listeningExecutor = MoreExecutors.listeningDecorator(executorService);
+
+    Callable<Channel> createChannelCallable = new Callable<Channel>() {
+      @Override
+      public Channel call() throws Exception {
+        return factory.create();
```
mgarolera: Could you add a comment with how long does it take to create a channel? In general all these optimizations should have a sense of how much time they save at the time they were implemented. They might become obsolete at some point and we will not know if we can clean them up.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -77,19 +85,49 @@ public ChannelPool(List<HeaderInterceptor> headerInterceptors, ChannelFactory fa
...
+      }
+    };
+
+    List<ListenableFuture<Channel>> connectionFutures = new ArrayList<>(newChannelCount);
```
mgarolera: These are not connections

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -77,19 +85,49 @@ public ChannelPool(List<HeaderInterceptor> headerInterceptors, ChannelFactory fa
...
-            newChannelList.add(factory.create());
-          }
-          setChannels(newChannelList);
+          channels.set(
+              ImmutableList.<Channel> builder()
+                  .addAll(channels.get())
+                  .addAll(createNewChannels(capacity - totalSize.get()))
+                  .build()
+          );
           totalSize.set(capacity);
```
mgarolera: when is totalSize different from channels.size()?  Should ChannelPool constructor take a capacity parameter? Or at least initialize with ensureChannelCount(1)

sduskis:
This didn't have the expected performance improvements.  Closing this.

PR #683, opened 2016-02-18
Moving Channel closing responsibilities to ChannelPool

sduskis:
Also making sure that ChannelPool has the responsibility of closing the underlying channels.
Making BigtableSession a Closable instead of an AutoClosable.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -16,8 +16,7 @@
 
 package com.google.cloud.bigtable.grpc;
 
-import io.grpc.Channel;
-import io.grpc.internal.ManagedChannelImpl;
+import io.grpc.ManagedChannel;
 import io.grpc.netty.GrpcSslContexts;
 import io.grpc.netty.NegotiationType;
 import io.grpc.netty.NettyChannelBuilder;
@@ -61,10 +60,6 @@ import com.google.cloud.bigtable.grpc.io.UserAgentInterceptor;
 import com.google.cloud.bigtable.util.ThreadPoolUtil;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableList.Builder;
-import com.google.common.util.concurrent.Futures;
-import com.google.common.util.concurrent.ListenableFuture;
-import com.google.common.util.concurrent.ListeningExecutorService;
-import com.google.common.util.concurrent.MoreExecutors;
 
 /**
  * <p>Encapsulates the creation of Bigtable Grpc services.</p>
@@ -77,7 +72,7 @@ import com.google.common.util.concurrent.MoreExecutors;
  *   <li> Close anything above that needs to be closed (ExecutorService, CahnnelImpls)
  * </ol>
  */
-public class BigtableSession implements AutoCloseable {
+public class BigtableSession implements Closeable {
 
   private static final Logger LOG = new Logger(BigtableSession.class);
   private static SslContextBuilder sslBuilder;
@@ -197,8 +192,8 @@ public class BigtableSession implements AutoCloseable {
   private BigtableClusterAdminClient clusterAdminClient;
 
   private final BigtableOptions options;
-  private final List<Closeable> clientCloseHandlers = Collections
-      .synchronizedList(new ArrayList<Closeable>());
+  private final List<ManagedChannel> managedChannels = Collections
+      .synchronizedList(new ArrayList<ManagedChannel>());
   private final ImmutableList<HeaderInterceptor> headerInterceptors;
 
   public BigtableSession(BigtableOptions options) throws IOException {
@@ -243,9 +238,18 @@ public class BigtableSession implements AutoCloseable {
 
       headerInterceptorBuilder.add(new UserAgentInterceptor(options.getUserAgent()));
       headerInterceptors = headerInterceptorBuilder.build();
+      ChannelPool dataChannel = new ChannelPool(headerInterceptors, new ChannelPool.ChannelFactory() {
+        @Override
+        public ManagedChannel create() throws IOException {
+          return createNettyChannel(BigtableSession.this.options.getDataHost());
+        }
+      });
+      managedChannels.add(dataChannel);
+      BigtableSessionSharedThreadPools sharedPools = BigtableSessionSharedThreadPools.getInstance();
 
       // More often than not, users want the dataClient. Create a new one in the constructor.
-      this.dataClient = initializeDataClient();
+      this.dataClient = new BigtableDataGrpcClient(dataChannel, sharedPools.getBatchThreadPool(),
+      sharedPools.getRetryExecutor(), options);
 
       // Defer the creation of both the tableAdminClient and clusterAdminClient until we need them.
     } finally {
@@ -306,18 +310,6 @@ public class BigtableSession implements AutoCloseable {
     };
   }
 
-  private BigtableDataClient initializeDataClient() throws IOException {
-    ChannelPool dataChannel = createChannel(options.getDataHost());
-    BigtableSessionSharedThreadPools sharedPools = BigtableSessionSharedThreadPools.getInstance();
-    return new BigtableDataGrpcClient(dataChannel, sharedPools.getBatchThreadPool(),
-        sharedPools.getRetryExecutor(), options);
-  }
-
-  private BigtableTableAdminClient initializeAdminClient() throws IOException {
-    Channel channel = createChannel(options.getTableAdminHost());
-    return new BigtableTableAdminGrpcClient(channel);
-  }
-
   private static <T> T get(Future<T> future, String errorMessage) throws IOException {
     try {
       return future.get();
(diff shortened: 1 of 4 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -371,22 +369,15 @@ public synchronized BigtableClusterAdminClient getClusterAdminClient() throws IO
   protected ChannelPool createChannel(final String hostString) throws IOException {
```
mgarolera: rename createChannel to createChannelPool

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -399,34 +390,18 @@ protected Channel createNettyChannel(final String host) throws IOException {
...
+    }
     List<ListenableFuture<Void>> closingChannelsFutures = new ArrayList<>();
-    ListeningExecutorService listenableBatchPool = MoreExecutors
-        .listeningDecorator(BigtableSessionSharedThreadPools.getInstance().getBatchThreadPool());
-    for (final Closeable clientCloseHandler : clientCloseHandlers) {
-      closingChannelsFutures.add(listenableBatchPool.submit(createCallable(clientCloseHandler)));
+    ListeningExecutorService listenableBatchPool =
+        BigtableSessionSharedThreadPools.getInstance().getListeningExecutorService();
+    for (ChannelPool channelPool : channelPools) {
+      closingChannelsFutures.add(listenableBatchPool.submit(createCallable(channelPool)));
```
mgarolera: createCallable -> createCloseCallable

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSessionSharedThreadPools.java
```diff
@@ -81,6 +83,10 @@ public ExecutorService getBatchThreadPool() {
     return batchThreadPool;
   }
 
+  public ListeningExecutorService getListeningExecutorService() {
```
mgarolera: Not sure this belongs to BigtableSessionSharedThreadPools, it doesn't really return any new object, just another flavor of the same object as getBatchThreadPool().

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -166,4 +169,56 @@ public int size() {
...
+      @Override
+      public Void call() throws Exception {
+        close(channel);
+        return null;
+      }
+    };
+  }
+
+  protected static void close(Channel channel) throws IOException {
+    if (!(channel instanceof ManagedChannel)) {
```
mgarolera: Why not used ManagedChannel everywhere? no need to do the casting.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -166,4 +169,56 @@ public int size() {
...
+    ManagedChannel channelImpl = (ManagedChannel) channel;
+    channelImpl.shutdown();
+    int timeoutMs = 10000;
+    try {
+      channelImpl.awaitTermination(timeoutMs, TimeUnit.MILLISECONDS);
+    } catch (InterruptedException e) {
+      throw new IOException("Interrupted while sleeping for close", e);
+    }
+    if (!channelImpl.isTerminated()) {
+      // Sometimes, gRPC channels don't close properly. We cannot explain why that happens,
```
mgarolera: Yuck! Would be nice to point to the corresponding GRPC issue/conversation.

bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/io/ChannelPoolTest.java
```diff
@@ -17,25 +17,25 @@
 
 import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.same;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
mgarolera: I think we normally try to avoid this. Import Mockito and reference Mockito.mock

sduskis: I removed the .*, but kept the import static.  For mockito, it's common practice.  For example, grpc does this quite a bit: https://github.com/grpc/grpc-java/search?utf8=%E2%9C%93&q=import+static+mockito

sduskis:
@mgarolera I took your advice and changed uses of Channel to ManagedChannel.  ChannelPool implements ManagedPool as well now.  That helped remove some hairy concurrency logic in BigtableSession's close logic.  PTAL.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -388,44 +375,45 @@ protected Channel createNettyChannel(final String host) throws IOException {
...
-
-  @Override
-  public void close() throws IOException {
-    List<ListenableFuture<Void>> closingChannelsFutures = new ArrayList<>();
-    ListeningExecutorService listenableBatchPool = MoreExecutors
-        .listeningDecorator(BigtableSessionSharedThreadPools.getInstance().getBatchThreadPool());
-    for (final Closeable clientCloseHandler : clientCloseHandlers) {
-      closingChannelsFutures.add(listenableBatchPool.submit(createCallable(clientCloseHandler)));
+    for (ManagedChannel channel : managedChannels) {
+      long awaitTime = endTimeNanos - System.nanoTime();
```
mgarolera: awaitTime -> awaitTimeNanos

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -388,44 +375,45 @@ protected Channel createNettyChannel(final String host) throws IOException {
...
+        // NOTE: We haven't seen this problem since removing the RefreshingChannel
+        LOG.info("Could not close the channel after 10 seconds.");
+        break;
       }
-    };
+    }
+    for (Iterator<ManagedChannel> iterator = managedChannels.iterator(); iterator.hasNext();) {
+      ManagedChannel channel = iterator.next();
+      if (channel.isTerminated()) {
+        iterator.remove();
```
mgarolera: Could we merge both for loops? What is the point of removing managedChannels one by one? could we empty it at the end? why leave the not terminated ones? they wont be used by anyone later on.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -22,41 +31,33 @@
...
   public interface ChannelFactory {
-    Channel create() throws IOException;
+    ManagedChannel create() throws IOException;
   }
 
-  private final AtomicReference<ImmutableList<Channel>> channels = new AtomicReference<>();
+  private final AtomicReference<ImmutableList<ManagedChannel>> channels = new AtomicReference<>();
   private final AtomicInteger requestCount = new AtomicInteger();
   private final ImmutableList<HeaderInterceptor> headerInterceptors;
   private final AtomicInteger totalSize;
```
mgarolera: I thought I had commented it before, but I can't find the comment.
Why keep totalSize around? isn't it channels.size()?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -22,41 +31,33 @@
...
   private final ChannelFactory factory;
   private final String authority;
-  
-  private final AtomicBoolean closed = new AtomicBoolean(false);
+
+  private boolean shutdown = false;
```
mgarolera: nit: name it shuttingdown? when it is shutdown=true it is not true that the pool is shut down.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
sduskis: They are shutdown, so you're right.  We don't need to keep them around.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
sduskis: Doh.  I missed this before.  totalSize was needed with a previous implementation of ChannelPool.  It's not needed anymore.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
sduskis: It should be shutdown.  All of the underlying channels were shutdown() and should all return true for isShutdown().  I'll change isShutdown() to return this.shutdown.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -166,4 +165,62 @@ public int size() {
...
+      if (!channel.isShutdown()) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  @Override
+  public synchronized boolean isTerminated() {
+    List<ManagedChannel> c = new ArrayList<>(channels.get());
```
mgarolera: I don't think there is a need to clean up the channels list. From the name, isTerminated shouldn't change the state of the object.

mgarolera:
LGTM

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
sduskis: fair enough.  We don't really need to clean the list.

PR #684, opened 2016-02-18
Adding some performance logging

sduskis:
The types of logging added:
- How many writes there were and how long they took
- How many records were read in a scan and how long it took
- If there is a zone mismatch between bigtable and dataflow.


sduskis:
I updated the code.  PTAL

PR #693, opened 2016-02-25
Report number of mutations done by dataflow.

mgarolera:


bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -736,9 +738,14 @@ public static Pipeline initializeForWrite(Pipeline p) {
...
     private final String tableName;
 
+    // Stats
+    private final Aggregator<Long, Long> mutations;
+
     public CloudBigtableSingleTableBufferedWriteFn(CloudBigtableTableConfiguration config) {
       super(config);
       this.tableName = config.getTableId();
+
+      mutations = createAggregator("mutations", new Sum.SumLongFn());
```
garye: this.mutations for consistency (same comment for each Fn)

mgarolera: Done. PTAL

-- 

Marçal

sduskis:
+1 on aggregators.  We should do this for the other Writer DoFns, the Reader and the Source splitting.  The performance logging in AbstractCloudBigtableTableDoFn should be removed.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
garye: Cool, thanks

PR #696, opened 2016-02-26
Allow IP address to override host name in BigtableSession.

garye:
Resolve host name to IP address for dataflow jobs to work around gRPC issue.

### DIFF
--- a/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/TestBigtableSession.java
+++ b/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/TestBigtableSession.java
@@ -40,7 +40,7 @@ public class TestBigtableSession {
           .setZoneId(zoneId)
           .setClusterId(clusterId)
           .setUserAgent(userAgent)
-          .build(), null, null, null);
+          .build());
   }
 
   @Rule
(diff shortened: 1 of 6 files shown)


bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/TestBigtableSession.java
```diff
@@ -74,4 +74,6 @@ public void testNoUserAgentBigtableOptions() throws IOException {
     createSession(PROJECT_ID, ZONE_ID, CLUSTER_ID, null);
   }
 
+
```
sduskis: nit: please remove the blank lines.

sduskis:
LGTM, except for the nit.

mgarolera:
Do we need IP override?
Having a configuration parameter to in BigtableSession.java do
      builder.forAddress(new InetSocketAddress(InetAddress.getByName(host), options.getPort()))
instead of 
      builder.forAddress(host, options.getPort());

should be enough and be more hermetic.
I was thinking about having a "DoHostResolutionAtStartup" parameter.

garye:
I think ideally we wouldn't want to force anyone to add a new parameter to their config since this is a temporary workaround that we will be removing when possible.

garye:
Point taken - it's potentially a bit invasive to solve a particular problem. As per offline discussion let's go with this so we can move forward with a snapshot release and circle back to see if the overrides will prove to be generally useful once the automatic override for dataflow goes away.

PR #697, opened 2016-02-26
Address whitespace issue from review. Throw IOException in unit tests. 

garye:
.


garye:
Merging to fix the build.

PR #701, opened 2016-03-01
Add logging for retries

mgarolera:


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RetryListener.java
```diff
@@ -21,30 +21,32 @@
...
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
 /**
  * A ClientCall.Listener that listens for lower-level ClientCall failures and if detected, attempts
  * to retry the call if the call supports retrying.
  */
 class RetryListener<RequestT, ResponseT> extends ClientCall.Listener<ResponseT> {
+  private static final Log LOG = LogFactory.getLog(RetryListener.class);
```
sduskis: We use our own Logger class.  Can you use that instead for consistency?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RetryListener.java
```diff
@@ -62,10 +64,11 @@ public void onHeaders(Metadata headers) {
   @Override
   public void onClose(Status status, Metadata trailers) {
     if (isRetriableStatus(status.getCode())
-        && isRetriableCall
         && !stateSignalledToListener) {
       if (retryingCall.retryCallAfterBackoff(payload, requestHeaders, this)) {
-        // We are retrying
+        // A retryable error.
+        failedCount += 1;
+        LOG.info("Retrying failed call. Failure #" + failedCount + ", got: " + status, status.getCause());
```
sduskis: This should probably be a Logger.warn.  If the getCause is a throwable, that should be the first argument.

mgarolera: Since there is nothing actionable about this logging I would prefer to keep it as info, not warning. One could argue it should be debug...

Seems throwable is the second argument:
https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/Logger.java#L55

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RetryListener.java
mgarolera: Done.

sduskis:
LGTM.  Please squash the commits first.

PR #705, opened 2016-03-02
Adding bulk mutation and read rpc size and count configuration.

sduskis:


bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
```diff
@@ -44,6 +45,19 @@
...
+   * and starting the next bulk call. Defaults to 1 MB.
+   */
+  public static final long BIGTABLE_MAX_BULK_RPC_SIZE_BYTES_DEFAULT = 1 << 20;
+
+  /**
+   * This describes the maximum number of individual mutation requests to bundle in a single bulk
+   * mutation RPC before sending it to the server and starting the next bulk call.
+   * The server has a maximum of 100,000.
+   */
+  public static final int BIGTABLE_MAX_BULK_REQUEST_COUNT_DEFAULT = 1_000;
```
mgarolera: nit: I like the idea of prefix being meaningful for constants. BIGTABLE + BULK_MUTATE + ( MAX_MUTATION_COUNT | MAX_REQUEST_BYTES ) + DEFAULT

It is not clear why one thing is referred as RPC and the other as REQUEST

bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
```diff
@@ -181,6 +199,16 @@ public Builder setUseBulkApi(boolean useBulkApi) {
       return this;
     }
 
+    public Builder setBulkMaxRpcCount(int bulkMaxRpcCount) {
```
mgarolera: Be consistent with the ordering. Use Count before Size everywhere or the other way around.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
```diff
@@ -37,22 +37,25 @@
...
+  private final long maxRpcSize;
+  private final int maxRpcCount;
+  
   private long size = 0l;
 
-  public BulkMutation(String tableName) {
+  public BulkMutation(String tableName, int maxRpcCount, long maxRpcSize) {
     this.builder = MutateRowsRequest.newBuilder().setTableName(tableName);
-    this.size = MutateRowsRequest.newBuilder().setTableName(tableName).build().getSerializedSize();
+    this.size = tableName.length() + 2;
```
mgarolera: size -> byteSize aproxByteSize?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
```diff
@@ -37,22 +37,25 @@
...
+    this.maxRpcSize = maxRpcSize;
```
mgarolera: consistent ordering of args and fields

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
```diff
@@ -71,7 +74,7 @@ public BulkMutation(String tableName) {
   }
 
   public boolean isFull() {
-    return size >= MAX_SIZE || futures.size() >= MAX_COUNT;
+    return size >= maxRpcSize || futures.size() >= maxRpcCount;
```
mgarolera: I'm tempted to say that the fullness logic dont fully belong in the BulkMutation. There is nothing preventing consumers of this class to keep adding mutations even after fullness. Maybe the class should just report a rpc size and a rpc count, and the consumer should check the fullness and decide what to do?

sduskis: I'm renaming this to reachedRequestedMax() which is hopefully more indicative of the intent of the method.  It's more like the user sets a threshold that can be ignored by the user, if they want to.  I'd rather not copy the logic to two places if I can avoid it.

sduskis:

mgarolera:
LGTM

PR #710, opened 2016-03-03
Caching the validity of a cloud bigtable configurtion.

sduskis:
Currently, the validation logic is run for every split that dataflow does in read mode.  That means that BigtableAdmin.tableExists() gets called often.  That means that in turn, list tables gets called often.  I've seen it happen up to 15x per second in some cases.  That strains the server unnecessarily.  The caching of the validation will tremendously reduce the load on listTables.


mgarolera:
Hmm...

Should we move the tableExists check out of validate completely? Other implementations of validate that I've seen seem to be just a bunch of Preconditions.checks

And after all, the table can be deleted any time, making a valid configuration suddently invalid.

If the user hasn't properly configured the table they will quickly see when things fail with NOT_FOUND.+

If we really want a clearer error message we can add a separate tableExists check, separate from dataflow's validation of configuration. Or maybe NOT_FOUND should be make clearer?

Thoughts?

sduskis:
I removed the table exists check.  I agree with you that the users will see a failure pretty quickly from the job, and that's enough.  PTAL

mgarolera:
LGTM
I like deleted code :)

PR #712, opened 2016-03-03
Updating RetryingRpcFunction to retry multiple times.

sduskis:
RetryingRpcFunction's retries returned a future that wasn't retried.  Only the first failure caused a retry, but was not exponentially backed off.  Added a cycle of retries.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/RetryingRpcFunction.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/RetryingRpcFunction.java
@@ -16,13 +16,13 @@
 package com.google.cloud.bigtable.grpc.async;
 
 import java.io.IOException;
-
+import java.util.concurrent.ExecutorService;
 
 import com.google.api.client.util.BackOff;
 import com.google.api.client.util.Sleeper;
 import com.google.cloud.bigtable.config.Logger;
 import com.google.cloud.bigtable.config.RetryOptions;
-import com.google.cloud.bigtable.grpc.scanner.ScanRetriesExhaustedException;
+import com.google.cloud.bigtable.grpc.scanner.BigtableRetriesExhaustedException;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.AsyncFunction;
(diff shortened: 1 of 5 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/RetryingRpcFunction.java
```diff
@@ -38,8 +38,10 @@
     implements AsyncFunction<StatusRuntimeException, ResponseT> {
 
   public static <RequestT, ResponseT> RetryingRpcFunction<RequestT, ResponseT> create(
-      RetryOptions retryOptions, RequestT request, BigtableAsyncRpc<RequestT, ResponseT> retryableRpc) {
-    return new RetryingRpcFunction<RequestT, ResponseT>(retryOptions, request, retryableRpc);
+      RetryOptions retryOptions, RequestT request,
+      BigtableAsyncRpc<RequestT, ResponseT> retryableRpc, ExecutorService executorService) {
+    return new RetryingRpcFunction<RequestT, ResponseT>(retryOptions, request, retryableRpc,
```
mgarolera: Why are we using create() vs a constructor?

sduskis: I'm not sure.  I removed the create() and switched to the constructor.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/RetryingRpcFunction.java
```diff
@@ -83,11 +80,27 @@ private RetryingRpcFunction(RetryOptions retryOptions, RequestT request,
       throw new ScanRetriesExhaustedException("Exhausted streaming retries.", cause);
     }
 
+    // TODO: Should this be a
```
mgarolera: a... what?

mgarolera:
LGTM.
It is kind of crazy that our tests don't catch retryable rpcs not retrying more than once. Could we add a test for it?

sduskis:
I added a test for OK, 5 failures and then recover, and complete failure.

PR #713, opened 2016-03-04
Improve retry logic after scan timeouts.

garye:


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResumingStreamingResultScanner.java
```diff
@@ -94,27 +95,48 @@ public Row next() throws IOException {
...
+          new TimeoutRetryBackOff(retryOptions.getMaxScanTimeoutRetries(), BackOff.ZERO_BACKOFF);
+    }
+    // Reset the error backoff in case we encountered this timeout after an error.
+    // Otherwise, we will likely have already exceeded the max elapsed time for the backoff
+    // and won't retry after the next error.
+    currentErrorBackoff = null;
+    backOffAndRetry(currentTimeoutBackoff, rte);
+  }
+
+  private void handleIOException(IOExceptionWithStatus ioe) throws IOException {
```
sduskis: Should currentTimeoutBackoff be reset here?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResumingStreamingResultScanner.java
```diff
@@ -94,27 +95,48 @@ public Row next() throws IOException {
...
     }
   }
 
+  private void handleScanTimeout(ScanTimeoutException rte) throws IOException {
+    logger.info("The client could not get a response in %d ms. Retrying the scan.",
+      retryOptions.getReadPartialRowTimeoutMillis());
+    if (currentTimeoutBackoff == null) {
+      // Create a backoff that retries immediately, up to the max number of times
+      currentTimeoutBackoff =
+          new TimeoutRetryBackOff(retryOptions.getMaxScanTimeoutRetries(), BackOff.ZERO_BACKOFF);
```
sduskis: do we need a BackOff.ZERO_BACKOFF / delegate?

garye: The alternative is to return 0 explicitly in the TimeoutRetryBackOff implementation. I thought this provided an easier path to doing something fancier (like exponential backoff) instead of immediate retries in the future. Happy to remove if you think the extra code isn't worth it.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResumingStreamingResultScanner.java
garye: If we reset currentTimeoutBackoff here then we don't have a mechanism to ever give up if we, for example, alternate between timeouts and retryable exceptions. By not resetting here we're saying that we will never accept more than a fixed number of timeouts for a given request which seems appropriate.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResumingStreamingResultScanner.java
sduskis: I think we should use more direct logic here showing that this is counter based rather than backoff/timespan based.  Since there's no need to sleep, might as well do a simple if statement to either reissueRequest() or throw a BigtableRetriesExhaustedException

garye:
Moved the timeout retry logic out of a BackOff and changed both counters in the scanner to Atomics. PTAL.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResumingStreamingResultScanner.java
```diff
@@ -93,52 +97,62 @@ public Row next() throws IOException {
         Row result = currentDelegate.next();
         if (result != null) {
           lastRowKey = result.getKey();
-          rowCount++;
-          // We've had at least one successful RPC, reset the backoff
-          currentBackoff = null;
+          rowCount.incrementAndGet();
+          // We've had at least one successful RPC, reset the backoff and retry counter
+          currentErrorBackoff = null;
+          timeoutRetryCount.set(0);
```
sduskis: timeoutRetryCount = null is a few nanoseconds faster for each call to .next().  I think that given performance considerations, we should also go back to long rowCount instead of AtomicLong.

garye:
PTAL

bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/scanner/ResumingStreamingResultScannerTest.java
```diff
@@ -261,4 +306,72 @@ private void doErrorsDoNotResume(Status status) throws IOException {
...
+
+   when(mockScannerFactory.createScanner(eq(originalRequest.build())))
+        .thenReturn(mockScanner);
+    ResultScanner<Row> afterError1Mock = mock(ResultScanner.class);
+    ResultScanner<Row> afterTimeoutMock = mock(ResultScanner.class);
+    ResultScanner<Row> afterError2Mock = mock(ResultScanner.class);
+
+    ReadRowsRequest.Builder expectedResumeRequest = originalRequest.build().toBuilder();
+    expectedResumeRequest.getRowRangeBuilder()
+        .setStartKey(ResumingStreamingResultScanner.nextRowKey(ByteString.copyFromUtf8("row1")));
```
sduskis: Can you please add a variable for ResumingStreamingResultScanner.nextRowKey(ByteString.copyFromUtf8("row1")) to break up the long command?

garye:
Addressed variable in test to break up long command

sduskis:
LGTM.  Please squash the commits.

PR #718, opened 2016-03-09
Adding shading rules in bigtable-hbase-dataflow.

sduskis:
This is in preparation for a new dataflow release which uses some of the
bigtable classes, but has its own shading rules.  This is needed for
both sets of shading to work nicely together.


garye:
LGTM

Issue #703, opened 2016-03-02
Fully Asynchronous ResultScanner


udoprog:
Hey,

Currently [`ResultScanner`](bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResultScanner.java) does not provide the means of subscribing for when a new batch is available. This forces users to potentially block, or perform awkward polling on on `ResultScanner#available()` if they want to build a fully asynchronous scanner.

To patch the current API one could introduce `ListenableFuture<Void> ResultScanner#fetchMoreRows()` as a complement. This would allow users to implement [something like this](https://github.com/spotify/heroic/blob/master/metric/datastax/src/main/java/com/spotify/heroic/metric/datastax/DatastaxBackend.java#L582) (example using Datastax).

As a more radical alternative the [extended Observer pattern](http://reactivex.io/documentation/observable.html) is a much better model of this behavior in my opinion.

sduskis:
gRPC has a concept of StreamObserver which we might be able to use: https://github.com/grpc/grpc-java/blob/master/stub/src/main/java/io/grpc/stub/StreamObserver.java

Would that work?  It's already built into the infrastructure.

udoprog:
Yeah, definitely!
In fact, that is the extended Observer pattern :).

sduskis:
We've been doing work towards this goal.  We still want to implement this feature.

udoprog:
Great to hear, thank you!

PR #733, opened 2016-03-13
Changing the unit of ResponseQueueReader from ReadRowResponse to Row.

sduskis:
This is both a performance enhancement and will help us with issue #703


sduskis:
I'm going to break this down into smaller PRs.

PR #736, opened 2016-03-14
Re-adding the ChannelPool ensure capacity feature.

sduskis:
Removing this logic caused a 300-400 microsecond drop in Get and Put performance.  I think we should add it back in and only remove this logic once we can find the reason for the 300-400 microsecond performance issue.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
@@ -16,15 +16,18 @@
 package com.google.cloud.bigtable.grpc.io;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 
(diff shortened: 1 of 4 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -122,13 +144,27 @@ protected void checkedStart(ClientCall.Listener<RespT> responseListener, Metadat
...
+  private void setChannels(List<ManagedChannel> newChannelList) {
+    channels.set(ImmutableList.copyOf(newChannelList));
+  }
+
   public int size() {
-    return channels.size();
+    return channels.get().size();
+  }
+
+  public int availbleSize() {
```
mgarolera: This is not used, remove?

sduskis: done.

mgarolera:
LGTM. I don't get how this is causing higher latencies. It moves more logic to the startup of the conenction and creates all the channels at the beginning instead of waiting for the fist request, but on normal operation it should be exactly the same or better because there is no channel creation on calls.

sduskis:
I don't get it either.  For now, we are going bak to the old way because of the performance implications.  We need to figure out why this change is required and fix the root cause, if we can.

PR #743, opened 2016-03-15
Refactor HeapSizeManager

garye:
- Rename to RpcThrottler
- Delegate heap size / rpc count limiting to single place


bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -137,13 +139,16 @@ protected AbstractBigtableConnection(Configuration conf, boolean managed, Execut
...
     this.session = new BigtableSession(options);
 
-    int defaultRpcCount = AsyncExecutor.MAX_INFLIGHT_RPCS_DEFAULT * options.getChannelCount();
-    int maxInflightRpcs = conf.getInt(MAX_INFLIGHT_RPCS_KEY, defaultRpcCount);
-    long maxMemory = conf.getLong(
-      BIGTABLE_BUFFERED_MUTATOR_MAX_MEMORY_KEY,
-      AsyncExecutor.ASYNC_MUTATOR_MAX_MEMORY_DEFAULT);
-
-    this.heapSizeManager = new HeapSizeManager(maxMemory, maxInflightRpcs);
+    synchronized (resourceLimitedLock) {
```
sduskis: Can you please put this in a synchronized static method rather than having a separate Lock?

bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -72,7 +73,7 @@ public void run() {
             int size = ACTIVE_BUFFERED_MUTATORS.size();
             new Logger(AbstractBigtableConnection.class).warn(
               "Shutdown is commencing and you have open %d buffered mutators."
-                  + "You need to close() or flush() them so that is not lost", size);
+                  + "You need to close() or awaitCompletion() them so that is not lost", size);
```
sduskis: This message should still be flush(), since it refers to the buffered mutator and not the RpcThrottler.

garye: Fixed

bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -163,7 +168,7 @@ public Table getTable(TableName tableName, ExecutorService pool) throws IOExcept
       pool = BigtableSessionSharedThreadPools.getInstance().getBatchThreadPool();
     }
     BatchExecutor batchExecutor = new BatchExecutor(
-         new AsyncExecutor(client, heapSizeManager),
+         new AsyncExecutor(client, new RpcThrottler(resourceLimiter)),
```
sduskis: Perhaps each individual call to .batch should have it's own RpcThrottler (or FlushManager?).  There could be multiple concurrent batches happening for a given BatchExecutor, and each batch should be handled by itself with separate flush management per batch.

garye: Let's keep this PR small and circle back on this when we have more time.

sduskis:
LGTM

PR #747, opened 2016-03-17
Always resolve API endpoints to IP addresses for now.

garye:
Remove after next gRPC release that has a bug fix related to
unreachable IPv6 addresses.


sduskis:
LGTM

PR #748, opened 2016-03-17
RpcThrottler waits for retries that don't count towards resource limits

garye:
BigtableBufferedMutator retries individual failures


sduskis:
LGTM

PR #751, opened 2016-03-21
Defer invoking isRetryable predicate

garye:
Doesn't move the needle on performance tests, but is defensive against more expensive predicates that could be introduced in the future.


sduskis:
LGTM

PR #754, opened 2016-03-21
Removing extraneous Executor from BigtableSession.

sduskis:
Also
- making the initial shared executor creation into an async process.
- Using the non-deprecated constructor for BigtableSession in TestBigtableSession.


garye:
LGTM

PR #755, opened 2016-03-21
Refactoring ResultQueueEntry methods to be distinct classes

sduskis:
This will hopefully improve scanning performance by reducing the number of checks required to get a result for the normal case of a response based ResultQueueEntry


garye:
LGTM

PR #758, opened 2016-03-23
Update no progress warning to include outstanding retry count

garye:


sduskis:
LGTM.

I think that we should print out this message periodically (30 seconds?).  Once doesn't seem to be enough.

garye:
It now logs every 30 seconds. Refactored and wrote a test to verify. PTAL.

sduskis:
LGTM after squashing

PR #759, opened 2016-03-23
Passing a ScheduledExecutorService instead of ExecutorService.

sduskis:
Currently, it's still used by RetryingRpcFunction as a plain ExecutorService.  It will be used to schedule changes in a future commit.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -75,6 +75,12 @@ public class BigtableSession implements Closeable {
   private static final Logger LOG = new Logger(BigtableSession.class);
   private static SslContextBuilder sslBuilder;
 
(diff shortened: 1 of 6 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -334,12 +336,12 @@ protected ManagedChannel createNettyChannel(final String host) throws IOExceptio
     BigtableSessionSharedThreadPools sharedPools = BigtableSessionSharedThreadPools.getInstance();
     return NettyChannelBuilder
         .forAddress(new InetSocketAddress(address, options.getPort()))
-        .maxMessageSize(256 * 1024 * 1024) // 256 MB, server has 256 MB limit.
+        .maxMessageSize(MAX_MESSAGE_SIZE) // 256 MB, server has 256 MB limit.
```
garye: Nit: move these comments up to the constant declaration

garye:
LGTM

PR #764, opened 2016-03-24
Updating TestBasicOps and TestCreateTable to execute in parallel.

sduskis:
This saves me about 30 seconds per integration test run.


bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestBasicOps.java
```diff
@@ -162,8 +167,25 @@ public void testPutAlmostTooBigValue() throws IOException {
...
+            testPutGetDeleteExists(10 << 20, true, true); // 10 MB
+            return null;
+          };
+        });
+      }
+    } finally {
+      es.shutdown();
+      while (!es.isTerminated()) {
+        try {
+          es.awaitTermination(1, TimeUnit.SECONDS);
```
garye: Why loop and await for 1 second at a time instead of a single awaitTermination()? Either way would be good to put an upper bound on amount of time to wait.

sduskis:
I put a 1 minute limit on the operations.  PTAL.

garye:
LGTM

PR #771, opened 2016-03-28
Moving persistence logic into BulkMutation.

sduskis:
The BulkMutation processing was duplicated, yet was inconsistent, in BatchExecutor and BigtableBufferedMutator.  Encapsulating all of the persistence logic into BulkMutation for better reuse.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
@@ -19,6 +19,7 @@ import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 
+import com.google.api.client.repackaged.com.google.common.annotations.VisibleForTesting;
 import com.google.bigtable.v1.MutateRowRequest;
 import com.google.bigtable.v1.MutateRowsRequest;
 import com.google.bigtable.v1.MutateRowsResponse;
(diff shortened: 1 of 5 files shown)


bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -307,24 +293,9 @@ private void offer(Mutation mutation) throws IOException {
...
-              || bulkMutation.getApproximateByteSize() >= options.getBulkMaxRequestSize()) {
-            mutateRowsAsync(bulkMutation);
-            bulkMutation = null;
-          }
-        }
+        MutateRowRequest request = adapt(mutation);
+        ListenableFuture<Empty> future = bulkMutation.add(request);
+        addExceptionCallback(future, mutation);
       } else {
         initializeAsyncMutators();
```
garye: nit: I don't think you need to do this until you decide to add to the asyncOperationsQueue

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
```diff
@@ -44,14 +45,133 @@
           .withDescription("Mutation does not have a status")
           .asRuntimeException();
 
-  private final List<SettableFuture<Empty>> futures = new ArrayList<>();
-  private final MutateRowsRequest.Builder builder;
-  
-  private long approximateByteSize = 0l;
+  @VisibleForTesting
+  class Batch {
```
garye: Can this be a static inner class?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
```diff
@@ -44,14 +45,133 @@
...
+        public void onFailure(Throwable t) {
+          for (SettableFuture<Empty> future : futures) {
+            future.setException(t);
+          }
+        }
+      };
+      Futures.addCallback(bulkFuture, callback);
+    }
+
+    private void mutateRowAsync() {
```
garye: This calls asyncExecutor#mutateRowsAsync, should it be called mutateRowsAsync?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
```diff
@@ -44,14 +45,133 @@
...
+      ListenableFuture<MutateRowsResponse> future = null;
+      try {
+        future = asyncExecutor.mutateRowsAsync(toRequest());
+      } catch (InterruptedException e) {
+        future = Futures.<MutateRowsResponse> immediateFailedFuture(e);
+      } finally {
+        if (future != null) {
```
garye: If it's possible that the future could be null here it might be better to raise an Exception rather than skip the callback.

sduskis:
All comments were addressed.  PTAL.

garye:
LGTM

Issue #616, opened 2015-12-29
FuzzyRowFilterAdapter is missing from FilterAdapter


ZubairNabi:
I've been trying to get `FuzzyRowFilter` to work, but out of the box I get the following exception:

```
Caused by: com.google.cloud.bigtable.hbase.adapters.filters.UnsupportedFilterException: Unsupported filters encountered: FilterSupportStatus{isSupported=false, reason='Don't know how to adapt Filter class 'class org.apache.hadoop.hbase.filter.FuzzyRowFilter''}
```

I can see that there's an adapter for it in the codebase but it's not in the filter map within `FileAdapter`. Putting it in the map causes the previous exception to obviously go away but then I get a dependency conflict:

```
Caused by: java.lang.NoSuchMethodError: org.apache.hadoop.hbase.protobuf.generated.HBaseProtos$BytesBytesPair.getFirst()Lcom/google/bigtable/repackaged/com/google/protobuf/ByteString;
	at com.google.cloud.bigtable.hbase.adapters.filters.FuzzyRowFilterAdapter.extractFuzzyRowFilterPairs(FuzzyRowFilterAdapter.java:90)
	at com.google.cloud.bigtable.hbase.adapters.filters.FuzzyRowFilterAdapter.adapt(FuzzyRowFilterAdapter.java:48)
	at com.google.cloud.bigtable.hbase.adapters.filters.FuzzyRowFilterAdapter.adapt(FuzzyRowFilterAdapter.java:39)
	at com.google.cloud.bigtable.hbase.adapters.filters.SingleFilterAdapter.adapt(SingleFilterAdapter.java:56)
	at com.google.cloud.bigtable.hbase.adapters.filters.FilterAdapter.adaptFilter(FilterAdapter.java:143)
	at com.google.cloud.bigtable.hbase.adapters.ScanAdapter.createUserFilter(ScanAdapter.java:116)
	at com.google.cloud.bigtable.hbase.adapters.ScanAdapter.buildFilter(ScanAdapter.java:77)
	at com.google.cloud.bigtable.hbase.adapters.ScanAdapter.adapt(ScanAdapter.java:94)
	at com.google.cloud.bigtable.hbase.adapters.HBaseRequestAdapter.adapt(HBaseRequestAdapter.java:70)
	at com.google.cloud.bigtable.hbase.BigtableTable.getScanner(BigtableTable.java:206)
```

[...]

PR #772, opened 2016-03-28
Fixing FuzzyRowFilterAdapter, but not hooking it up.

sduskis:
FuzzyRowFilterAdapter has a couple of problems:

1) It uses the hbase proto conversion to get the "fuzzyKeysDataList."  That conversion doesn't work when our shading is combined with non-shaded hbase.  Things just don't work.
2) The data and mask were switched in the adapter.  see https://github.com/apache/hbase/blob/master/hbase-client/src/main/java/org/apache/hadoop/hbase/filter/FuzzyRowFilter.java#L80 for more info.

This is the first step towards fixing #616.

### DIFF
--- a/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/FuzzyRowFilterAdapter.java
+++ b/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/FuzzyRowFilterAdapter.java
@@ -4,9 +4,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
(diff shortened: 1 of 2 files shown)


bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/FuzzyRowFilterAdapter.java
```diff
@@ -42,6 +40,18 @@
           .setCellsPerColumnLimitFilter(Integer.MAX_VALUE)
           .build();
 
+  private static Field FUZZY_KEY_DATA_FIELD;
+  private static Exception FUZZY_KEY_DATA_FIELD_EXCCEPTION;
```
garye: nit: EXCCEPTION

sduskis: Fixed.

garye:
LGTM

PR #773, opened 2016-03-28
Retrying bulkMutation failures in batch.

sduskis:
Retries are now done as a batch in BulkMutation.
Shared code from BatchExecutor and BigtableBufferedMutator moved to BulkMutation.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
@@ -15,14 +15,15 @@
  */
 package com.google.cloud.bigtable.grpc.async;
 
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.List;
-
-import com.google.api.client.repackaged.com.google.common.annotations.VisibleForTesting;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.api.client.util.BackOff;
 import com.google.bigtable.v1.MutateRowRequest;
 import com.google.bigtable.v1.MutateRowsRequest;
+import com.google.bigtable.v1.MutateRowsRequest.Entry;
 import com.google.bigtable.v1.MutateRowsResponse;
+import com.google.cloud.bigtable.config.Logger;
+import com.google.cloud.bigtable.config.RetryOptions;
+import com.google.cloud.bigtable.grpc.scanner.BigtableRetriesExhaustedException;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
@@ -31,7 +32,16 @@ import com.google.protobuf.Any;
 import com.google.protobuf.Empty;
 import com.google.rpc.Status;
 
+import io.grpc.Status.Code;
 import io.grpc.StatusRuntimeException;
+
+import java.io.IOException;
+import java.util.ArrayList;
+ import java.util.Iterator;
+import java.util.List;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * This class combines a collection of {@link MutateRowRequest}s into a single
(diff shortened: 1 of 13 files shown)


bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/async/TestRpcThrottler.java
```diff
@@ -99,11 +93,12 @@ public void run() {
       });
       // Give some time for the Runnable to be executed.
       Thread.sleep(10);
-      assertEquals(1, secondRequestRegisteredLatch.getCount());
+      assertFalse(secondRequestRegistered.get());
       resourceLimiter.markCanBeCompleted(id);
 
-      // Now wait for the second request to be registered
-      assertTrue(secondRequestRegisteredLatch.await(1, TimeUnit.MINUTES));
```
garye: Why did we go back to using the boolean/sleep instead of CountdownLatch?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
```diff
@@ -108,78 +151,171 @@ MutateRowsRequest toRequest(){
...
+        return;
+      }
+
+      long nextBackOff = BackOff.STOP;
+      try {
+        if(currentBackoff == null) {
+          currentBackoff = retryOptions.createBackoff();
+        }
+        nextBackOff = currentBackoff.nextBackOffMillis();
+      } catch (IOException e) {
```
garye: I don't think nextBackOffMillis will actually throw but a comment stating that we are ignoring on purpose would be helpful

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
```diff
@@ -108,78 +151,171 @@ MutateRowsRequest toRequest(){
...
+        } else {
+          this.requestManager = retryRequestManager;
+          LOG.info(
+            "Retrying failed call. Failure #%d, got #%d failures",
+            failedCount++,
+            requestManager.futures.size());
+          retryExecutorService.schedule(this, nextBackOff, TimeUnit.MILLISECONDS);
+        }
+      } catch (RuntimeException e) {
+        // TODO: THIS IS IMPORTANT
```
garye: Just noting this in case you wanted to handle before pushing

bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/async/TestRpcThrottler.java
```diff
@@ -241,52 +248,4 @@ public void run() {
       pool.shutdownNow();
     }
   }
-
-  @Test
-  public void testNoSuccessWarning() throws Exception {
```
garye: Merge error? I'd like to keep this test.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
```diff
@@ -108,78 +151,171 @@ MutateRowsRequest toRequest(){
...
+
+            @Override
+            public void onFailure(Throwable t) {
+              setFailure(t);
+            }
+          });
+    }
+
+
+    private synchronized void handleResult(MutateRowsResponse result) {
```
garye: This method is on the long side. Any clean way to factor some parts out to smaller methods?

sduskis: I refactored that method into smaller ones.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
```diff
@@ -108,78 +151,197 @@ MutateRowsRequest toRequest(){
...
+        extraStatusCount++;
+        LOG.error("Got %extra status: %s", statuses.next());
+      }
+      LOG.error("Got %d extra statuses", extraStatusCount);
+    }
+
+    private void doRetry(long nextBackOff, RequestManager retryRequestManager) {
+      if (retryRequestManager == null || retryRequestManager.futures.isEmpty()) {
+        this.currentRequestManager = null;
+        this.asyncExecutor.getRpcThrottler().onRetryCompletion(retryId);
```
garye: I might be missing it but where does retryId get set?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
garye: Thanks I think that's more readable!

sduskis:
Thanks for the catch with retryId.  I somehow removed that with my last PR.  I just submitted a fix for that issue.  I also added handling for a generic RuntimeException.  PTAL.

garye:
LGTM

PR #778, opened 2016-03-30
Turning on bulk writes in BufferedMutator and BatchExecutor by default.

sduskis:


sduskis:
I need to fix some failing tests first.

PR #780, opened 2016-03-30
Adding isRetryable functionality ot BigtableAsyncRpc.

sduskis:


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -75,7 +72,8 @@
 public class BigtableDataGrpcClient implements BigtableDataClient {
 
   private static final Logger LOG = new Logger(BigtableDataGrpcClient.class);
-  private static AtomicInteger LOG_COUNT = new AtomicInteger();
+
+  /// Retryable Predicates
```
garye: Nit: /// -> //

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -115,11 +113,16 @@ public boolean apply(CheckAndMutateRowRequest checkAndMutateRowRequest) {
         }
       };
 
-  private static final Predicate<SampleRowKeysRequest> IS_RETRYABLE_SAMPLE_ROW_KEY =
-      Predicates.<SampleRowKeysRequest> alwaysTrue();
+  private static final boolean allCellsHaveTimestamps(Iterable<Mutation> mutations) {
+    for (Mutation mut : mutations) {
+      if (mut.getSetCell().getTimestampMicros() == -1) {
```
garye: What about if it's 0? That seems to be the default value in the Cell proto and what it gets set to after a clear.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
sduskis: fixed.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
sduskis: I changed this.  I hope that this doesn't bite some of our users who may have not set timestamp.

garye:
LGTM

PR #782, opened 2016-03-30
Remove race between creating adding and completing retry futures

garye:
I was too aggressive with one of the deleted sleeps and the test became flaky. I like this better anyway.


sduskis:
LGTM

PR #785, opened 2016-03-31
Replacing FutureCallbacks with Call.Listener for RPC retries

sduskis:
This shaves off up to 40 microseconds and leaves the work to send out an RPC at less than 1 microsecond of processing time according to the PutMicrobenchmark.


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -271,28 +275,45 @@ public Row readModifyWriteRow(ReadModifyWriteRowRequest request) {
...
-      ReqT request, BigtableAsyncRpc<ReqT, RespT> rpc, CancellationToken cancellationToken) {
-    return new RetryingRpcFunction<>(
-            retryOptions, request, rpc, retryExecutorService, cancellationToken)
-        .callRpcWithRetry();
+  private <ReqT, RespT> List<RespT> performBlockingStreamingRpc(
+      ReqT request, BigtableAsyncRpc<ReqT, RespT> rpc) {
+    RetryingCollectingClientCallListener<ReqT, RespT> listener =
+        new RetryingCollectingClientCallListener<>(
+            retryOptions, request, rpc, retryExecutorService);
+    try {
```
garye: Can this try/catch block be factored out and shared between performBlockingUnaryRpc and performBlockingStreamingRpc?

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -271,28 +275,45 @@ public Row readModifyWriteRow(ReadModifyWriteRowRequest request) {
...
+      listener.cancel();
+      Thread.currentThread().interrupt();
+      throw Status.CANCELLED.withCause(e).asRuntimeException();
+    } catch (ExecutionException e) {
+      listener.cancel();
+      throw Status.fromThrowable(e).asRuntimeException();
+    }
   }
 
   // Scanner methods
```
garye: Up to you but streamRows doesn't seem to fit in that well since it has a bunch of logic instead of the more declarative style of the other methods here. Maybe something to encapsulate and move into a separate class?

sduskis: I'm going to punt on this for now.  We do need to do some refactoring of the scanner logic so that we can also have an observer type interface.  We'll do that another time.

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
sduskis: Good point.  I factored out the logic.  PTAL.

garye:
LGTM

PR #787, opened 2016-04-01
Moving User Agent header into NettyChannel construction.

sduskis:
Adding some methods in PutMicroBencmark to run against a real server.


garye:
LGTM

PR #792, opened 2016-04-05
Fixing a misspelled method name (retrowException)

### LINKED ISSUE #663: Misspelled method name
In `AbstractCloudBigtableTableDoFn`, the method name `retrowException` is misspelled (should be `rethrowException`).

I guess it's too late to fix this, but maybe we could at least support the correct spelling as well as the misspelling.

sduskis:
This fixes #663.


garye:
LGTM. Do we think anyone else might have extended AbstractCloudBigtableTableDoFn and would break because of this?

sduskis:
Good point.  We'll call this out in the release notes.

PR #793, opened 2016-04-05
Updating protobuf dependencies.

sduskis:
1) Removed all copies of the core protobuf files. The are no longer necessary and can cause problems with upgrades to newer protobuf jar versions.
2) Updated all googleapi related protobufs.
3) Updated the README -- removed the documented protoc3 requirement for the java generation, since it's not a requirement since we upgraded to use grpc generation.


garye:
LGTM

PR #796, opened 2016-04-12
Removing unused BIGTABLE_CHANNEL_TIMEOUT_MS_KEY option.

sduskis:
This was an option used a while back for custom code we had to refresh channels.  The retries we introdced allowed us to remove the channel refreshing, but we did not remove the BigtableOption that controlled it.


garye:
LGTM

PR #797, opened 2016-04-13
Introducing BulkOptions.

sduskis:
It encapsulates the options for bulk operations that were previously in BigtableOptions and AbstractBigtableConnection.


bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
```diff
@@ -205,6 +167,11 @@ public Builder setUsePlaintextNegotiation(boolean usePlaintextNegotiation) {
     }
 
     public BigtableOptions build() {
+      if (bulkOptions == null) {
+        int maxInflightRpcs =
```
garye: Is it possible in a non-HBase environment to have a non-null bulkOptions but to not have set maxInflightRpcs? If so we might not have a valid value to pass through to ResourceLimiter.

garye:
LGTM other than the one concern

sduskis:
I pushed a commit that addresses the concern.  PTAL

garye:
LGTM

PR #798, opened 2016-04-14
BigtableSession should manage the creation of BulkMutation.

sduskis:
Users who use the proto version of the API need to create BulkMutations.  This PR adds in a simple way of creating a BulkMutation.  It also moves existing ways of creating the BulkMutation and AsyncExecutor to use BigtableSession instead of creating it themselves.  This simplifies the amount of knowledge other classes need in constructing these objects.


garye:
LGTM

PR #800, opened 2016-04-15
Allowing a user to retry even if the client does not set the timestamp.

sduskis:
The concern about setting the client timestamp in hbase is safe from the standpoint of most situations.  A Column Family with a single version and gets/scans where version=1 will not have a problem with retries that put in two cells into a qualifier.  Allow a user to set an option that will ensure that a server-side timestamp is set.

This PR is in response to a user request.

### DIFF
--- a/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
+++ b/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
@@ -98,6 +98,16 @@ public class BigtableOptionsFactory {
   public static final String ENABLE_GRPC_RETRIES_KEY = "google.bigtable.grpc.retry.enable";
 
   /**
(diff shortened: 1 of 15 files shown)


bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
```diff
@@ -98,6 +98,15 @@
   public static final String ENABLE_GRPC_RETRIES_KEY = "google.bigtable.grpc.retry.enable";
 
   /**
+   * By default, the hbase client will set the timestamp on the client side before sending a Put if
+   * it's not set already. That's done to ensure that only a single cell is written for the put if retries occur.
+   * Retries aren't common, and writing multiple cells isn't a problem in many cases. Sometimes, seeing the server
```
garye: seeing -> setting

garye:
LGTM

PR #803, opened 2016-04-19
Updating google.auth.library.version depedency.

sduskis:
The 0.4.0 version has a fix for cloud shell and gae flex credentials.  We can remove our copy of CloudShellCredentials.java which has a necessary fix. We also get a fix for #750.


garye:
LGTM

PR #805, opened 2016-04-22
Changed method name to match proto method name to reduce confusion.

carterpage:
I think we should either change the underlying method, or change the method name here to match.


carterpage:
I signed it!

sduskis:
LGTM

PR #808, opened 2016-04-26
Adding modifyColumn to AbstractBigtableAdmin.

sduskis:
It should have been there before.  I'm not sure why it was missing.


garye:
LGTM

PR #809, opened 2016-04-26
Changing method signature to use byte[] for prefix

carterpage:


sduskis:
LGTM

sduskis:
We probably should add a test before releasing this out in the wild.

PR #812, opened 2016-04-28
Removing RenameTable from BigtableTableAdminClient.

sduskis:
RenameTable was never implemented on the server.  It shouldn't be there
in the client either.


garye:
LGTM

PR #824, opened 2016-05-09
remove tests for non-descending cells

garye:


sduskis:
LGTM.. as long as you ran ReadRowsAcceptanceTest and it passed.

garye:
I forgot that was pushed. I'll run the tests and address any issues.

garye:
Removed two validations that are no longer valid. PTAL.

PR #827, opened 2016-05-10
Adding generics to BigtableResultScannerFactory.

sduskis:
Removing V1 specific artifacts makes this class more useful in V2.


garye:
LGTM

PR #828, opened 2016-05-10
Renaming RowMergerV2 to be v2.RowMerger.

sduskis:
More v2 classes will be coming.  A v2 package is going to be easier to maintain.


garye:
LGTM

PR #830, opened 2016-05-10
move test json to new v2 package

garye:


garye:
Committing to fix build

PR #833, opened 2016-05-10
Adding some of the simpler v2 related scan and other data ops.

sduskis:
This should fix a bug in RequestCountManagingResponseQueueReader.java
This should fix a docs issue in v1 RowMerger.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/v2/RowMerger.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/v2/RowMerger.java
@@ -60,12 +60,39 @@ import java.util.TreeMap;
  * rm.onComplete();
  * </pre>
  * <p>
- * When a complete row is found, {@link StreamObserver#onNext(Row)} will be called.
+ * When a complete row is found, {@link StreamObserver#onNext(Object)} will be called.
  * {@link StreamObserver#onError(Throwable)} will be called for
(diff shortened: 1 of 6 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/v2/RowMerger.java
```diff
@@ -60,12 +60,39 @@
  * rm.onComplete();
  * </pre>
  * <p>
- * When a complete row is found, {@link StreamObserver#onNext(Row)} will be called.
+ * When a complete row is found, {@link StreamObserver#onNext(Object)} will be called.
```
garye: very minor, but should this be onNext(ReadRowsResponse)?

garye:
LGTM

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/v2/RowMerger.java
sduskis: RowMerger is an adapter.  It's a StreamObsever<ReadRowsResponse>, which calls a StreamObserver<Row>.  The intent of emitRow is to call StreamObserver<row>.onNext(Row).  I changed this to  {@link StreamObserver#onNext(Object)} since javadocs was giving a warning with the Generic parameter (Row).

PR #841, opened 2016-05-16
Integration tests have been taking longer.  

sduskis:
Fail after 10 minutes, not 5 minutes like we do now.

Hopefully both the bigtable and hbase cluster tests will fail less frequently, especially on developer machines.


garye:
LGTM

PR #848, opened 2016-05-17
Using Bigtable ReadRowsRequest instead of Scan for CloudBigtableIO.

sduskis:
This removes a use of HBase's ProtobufUtils which uses HBase protos under the cover, which requires old versions of protobuf to work.  Removing uses of ProtobufUtils will allow us to upgrade grpc and protobuf versions.

### DIFF
--- a/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableScanConfiguration.java
+++ b/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableScanConfiguration.java
@@ -15,17 +15,14 @@
  */
 package com.google.cloud.bigtable.dataflow;
 
-import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.bigtable.v1.ReadRowsRequest;
+import com.google.cloud.bigtable.grpc.BigtableClusterName;
+import com.google.cloud.bigtable.hbase.adapters.Adapters;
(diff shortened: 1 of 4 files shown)


bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableScanConfiguration.java
```diff
@@ -92,6 +90,17 @@ public Builder() {
      */
     public Builder withScan(Scan scan) {
       this.scan = scan;
+      this.request = null;
+      return this;
+    }
+
+    /**
+     * Specifies the {@link ReadRowsRequest} that will be used to filter the table.
+     * @param scan The {@link ReadRowsRequest} to add to the configuration.
```
garye: @param scan -> @param request.

bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableScanConfiguration.java
```diff
@@ -92,6 +90,17 @@ public Builder() {
...
+     * @return The {@link CloudBigtableScanConfiguration.Builder} for chaining convenience.
+     */
+    public Builder withRequest(ReadRowsRequest request) {
+      this.request = request;
```
garye: Should this.scan be set to null to mimic the behavior of withScan()?

sduskis:

garye:
LGTM

PR #850, opened 2016-05-17
Updating grpc, netty and protobuf versions.

sduskis:
We needed to do a bit of a hack with org.apache.hadoop.hbase.util.ByteStringer to get the bigtable-dataflow-import utilities create test files.


garye:
LGTM

sduskis:
I added tcnative v16 to the bigtable-hbase-integration-tests/pom.xml post review.  I needed to do that to make the tests pass.

PR #853, opened 2016-05-18
Adding v2 BulkMutation

sduskis:
- Updated BigtableDataClient to reflect the fact that MutateRowsResponse is a streaming response rather than unary.
- Copied AsyncExecutor to v2. Updated to v2 return types (for example MutateRowResponse in v2 vs. Empty in v1), and List<MutateRowsResponse> instead of a single response
- Copied BulkMutation to v2. The handleResult() method looks significantly differently with the streaming response.
- Copied the tests for AsyncExecutor and BulkMutation with minor changes.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/v2/BulkMutation.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/v2/BulkMutation.java
@@ -1,0 +1,418 @@
+/*
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.grpc.async.v2;
+
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.api.client.util.BackOff;
+import com.google.bigtable.v2.MutateRowRequest;
+import com.google.bigtable.v2.MutateRowResponse;
+import com.google.bigtable.v2.MutateRowsRequest;
+import com.google.bigtable.v2.MutateRowsResponse;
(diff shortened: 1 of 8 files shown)


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/v2/BulkMutation.java
```diff
@@ -0,0 +1,418 @@
...
+          bulkFuture,
+          new FutureCallback<List<MutateRowsResponse>>() {
+            @Override
+            public void onSuccess(List<MutateRowsResponse> result) {
+              handleResult(result);
+            }
+
+            @Override
+            public void onFailure(Throwable t) {
+              perfromFullRetry(new AtomicReference<Long>(), t);
```
garye: typo: perfrom -> perform

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/v2/BulkMutation.java
```diff
@@ -0,0 +1,418 @@
...
+        return;
+      }
+
+      List<MutateRowsResponse.Entry> entries = new ArrayList<>();
+      for (MutateRowsResponse response : result) {
+        entries.addAll(response.getEntriesList());
+      }
+
+      if (entries.isEmpty()) {
+        perfromFullRetry(backoffTime, new IllegalStateException("MutateRowResponse result or statuses"));
```
garye: MutateRowResponse -> MutateRowsResponse

bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/async/v2/TestAsyncExecutor.java
```diff
@@ -0,0 +1,213 @@
...
+        }
+      });
+      try {
+        future.get(50, TimeUnit.MILLISECONDS);
+        Assert.fail("The future.get() call should timeout.");
+      } catch(TimeoutException expected) {
+        // Expected Exception.
+      }
+      completeCall();
+      future.get(50, TimeUnit.MILLISECONDS);
```
garye: Could use a CountdownLatch or something that you can wait on if this turns out to be flaky

garye:
LGTM other than typos

bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/async/v2/TestAsyncExecutor.java
sduskis: We potentially need to evaluate a better alternative overall.  We spoke offline about this specific issue, and determined that CountdownLatch doesn't make sense in this scenario.

PR #855, opened 2016-05-18
Upgrading dataflow.

sduskis:
This required a a maven submodule refactoring.  There's some funky shading that happens in bigtable-hbase-dataflow/pom.xml which was done to fix a problem relating to bigtable-client-core's inclusion in the dataflow sdk.  The dataflow sdk had to do some funky shading because various projects they use have conflicting dependencies.  They added cloud-bigtable-core 0.2.3 in dataflow sdk 1.5.0+ and did shading causing conflicts with classes in the cloud-bigtable-hbase-* projects.  cloud-bigtable-hbase-dataflow's shading when combined with dataflow sdk shading causes some funkiness when running maven and using an IDE.  An intermediate submodule alleviates those issues.


garye:
LGTM

PR #859, opened 2016-05-20
Adding some infrastructure for liquid sharding.

sduskis:
Liquid sharding allows Dataflow to split off some work if a Reader is going too slow, and another worker has some capacity.  There are some open questions that need to be worked out with the Dataflow team, so the actual implementation doesn't exist yet.

see https://github.com/apache/incubator-beam/blob/master/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigtable/BigtableIO.java for more details on how Apache Beam does it.

### DIFF
--- a/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
+++ b/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
@@ -68,6 +68,9 @@ import com.google.cloud.dataflow.sdk.coders.Coder;
 import com.google.cloud.dataflow.sdk.coders.CoderRegistry;
 import com.google.cloud.dataflow.sdk.io.BoundedSource;
 import com.google.cloud.dataflow.sdk.io.BoundedSource.BoundedReader;
+import com.google.cloud.dataflow.sdk.io.range.ByteKey;
+import com.google.cloud.dataflow.sdk.io.range.ByteKeyRange;
+import com.google.cloud.dataflow.sdk.io.range.ByteKeyRangeTracker;
 import com.google.cloud.dataflow.sdk.options.PipelineOptions;
 import com.google.cloud.dataflow.sdk.transforms.Aggregator;
 import com.google.cloud.dataflow.sdk.transforms.DoFn;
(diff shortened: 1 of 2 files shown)


bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -252,6 +263,11 @@ public boolean isCompletionMarker(Result[] result) {
     public long getRowCount(Result[] result) {
       return result == null ? 0 : result.length;
     }
+
+    @Override
+    public ByteString getKey(Result[] result) {
+      return isCompletionMarker(result) ? null : ByteString.copyFrom(result[result.length - 1].getRow());
```
garye: Why does this get the key from last element of the array? A method comment in the interface would be helpful.

mbrukman: Is this code accessing the last element of the array without checking that the array is non-empty?

garye: I think isCompletionMarker() covers that case

mbrukman: `isCompletionMarker()` appears to check for `result == null` but I don't see the check for `result.length == 0` here or there.

sduskis: The CloudBigtableIO worked because of this code: https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/BigtableResultScannerAdapter.java#L39

We should still have a result.length == 0 check in isCompletionMarker().  I'll add it.

sduskis: hrm... it's already there for ResultArrayIterator:

    @Override
    public boolean isCompletionMarker(Result[] result) {
      return result == null || result.length == 0;
    }

sduskis:
@garye: I added a comment got getKey() (renamed to getLastKey()).  I should probably add more comments to getFractionConsumed() and splitAtFraction() in Reader.

garye:
LGTM

PR #860, opened 2016-05-20
Updating the v2 protos. Mostly, rename of the bulk delete protos

sduskis:
There were also some comments changes.

There's a subtle issue with the generated java files.  I had to revert to grpc 0.13.1, and the 3.0.0-beta-1 in order to get the integration tests to pass; the tests hang with grpc 0.14.0 and 3.0.0-beta-2 generated protos.  I'll try to use the 0.15.1-SNAPSHOT to see if the tests pass with that version.  In the meantime, we'll need to keep using generated files from previous grpc/protobuf version versions.


sduskis:
ping

garye:
LGTM

PR #866, opened 2016-05-26
Removing expensive logging from v2 RowMerger.

sduskis:


bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/v2/RowMerger.java
```diff
@@ -108,24 +108,20 @@ void handleLastScannedRowKey(ByteString lastScannedRowKey) {
...
-      void validateChunk(RowInProgress rowInProgess, ByteString previousKey, CellChunk newChunk) {
-        Preconditions.checkArgument(
-            rowInProgess == null, "A new row cannot have existing state: " + newChunk);
-        Preconditions.checkArgument(
-            newChunk.getRowStatusCase() != RowStatusCase.RESET_ROW,
-            "A new row cannot be reset: " + newChunk);
-        Preconditions.checkArgument(
-            !newChunk.getRowKey().isEmpty(), "A row key must be set: " + newChunk);
-        Preconditions.checkArgument(newChunk.hasFamilyName(), "A family must be set: " + newChunk);
+          void validateChunk(RowInProgress rowInProgess, ByteString previousKey,
```
garye: indented 2 extra spaces

garye:
LGTM

Makes sense not to print out the entire chunk (would the entire value get printed?) but there's a varargs version of checkArgument that uses a format string in case that solves your problem...

bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/v2/RowMerger.java
sduskis: Fixed the indentation.  It was related to eclipse auto formatting.

sduskis:
I updated RowMerger validation to use varags.  Thanks for the tip.

PR #874, opened 2016-06-02
Fixing the unit of the RpcThrottler warning timestamp.

sduskis:
300000 * 1000 nanoseconds == 300 microseconds.  Logging every 300 microseconds isn't helpful. I bumped up the logging to 30 seconds.

### DIFF
--- a/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/RpcThrottler.java
+++ b/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/RpcThrottler.java
@@ -41,7 +41,7 @@ public class RpcThrottler {
   // In awaitCompletion, wait up to this number of nanoseconds without any operations completing.  If
   // this amount of time goes by without any updates, awaitCompletion will log a warning.  Flush()
   // will still wait to complete.
-  private static final long INTERVAL_NO_SUCCESS_WARNING_NANOS = 300000 * 1000;
+  private static final long INTERVAL_NO_SUCCESS_WARNING_NANOS = TimeUnit.SECONDS.toNanos(30);
 
   private final ResourceLimiter resourceLimiter;
   private final NanoClock clock;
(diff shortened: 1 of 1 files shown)


garye:
LGTM

AngusDavis:
When the log line is emitted, the 'last successful' time is reset so this will always show that an op completed N seconds ago (where N is the INTERVAL_NO_SUCCESS_WARNING_NANOS).

sduskis:
resetNoSuccessWarningDeadline() does this at time A:
`noSuccessWarningDeadlineNanos = clock.nanoTime() + INTERVAL_NO_SUCCESS_WARNING_NANOS;
`

logNoSuccessWarning() does this at time B:

`long lastUpdateNanos = clock.nanoTime() - noSuccessWarningDeadlineNanos + INTERVAL_NO_SUCCESS_WARNING_NANOS;`

lastUpdateNanos = B - (A + INTERVAL) + INTERVAL =>
B - A - INTERVAL - INTERVAL =>
B - A

B - A should be greater or equals to INTERVAL.

AngusDavis:
Right - but in awaitCompletion:

        long now = clock.nanoTime();
        if (now >= noSuccessWarningDeadlineNanos) {
          logNoSuccessWarning(now);
          resetNoSuccessWarningDeadline();
          performedWarning = true;
        }

So subsequent calls to logNoSuccessWarning will always be INTERVAL +- 250 ms (DEFAULT_FINISH_WAIT_MILLIS).

sduskis:
Ah.  I'll fix that.

sduskis:
I think we actually found that printing every INTERVAL (or so) is a good idea, and that's what's happening.  Do you see a problem with that?

AngusDavis:
Nope, I like that. It's just that "time since last log line" and "time since last RPC completion" are being used interchangeably. The log message, to me, reads "The last time an RPC completed was N seconds ago", but that is not correct.

Re-reading the log message, I do realize that it could also read "We haven't seen any operations complete in the last 30 seconds when we printed that last message", but this means that the end-user needs to try to correlate log lines to hunt down when that particular throttler started emitting log entries.

PR #880, opened 2016-06-15
Moving classes in v2 packages to their parent package.

sduskis:
There was a minor warning fix in BulkMutation as well.


garye:
LGTM

PR #882, opened 2016-06-19
ResumingStreamingResultScanner should always close calls.

sduskis:
ResumingStreamingResultScanner doesn't close calls when it throws an exception.  Close the call for both types of exceptions.


garye:
LGTM

PR #885, opened 2016-06-23
Fixing some BulkMutation issues.

sduskis:
There was a misconfiguration of the default table admin endpoints, so I fixed that as well.


garye:
LGTM

PR #888, opened 2016-06-26
Updating v2 protos.

sduskis:
See https://github.com/googleapis/googleapis/tree/master/google/bigtable for the latest version of the protos. Most of the changes are in the comments.  There was one change of name to parent that clarifies cases where name is the name of the parent.


garye:
LGTM

PR #900, opened 2016-06-29
Updating grpc, io.netty, protobuf and tcnative versions.

sduskis:
This version of grpc works a lot better for us in terms of scan performance.

Since protobuf was updated, I regenerated all of the java classes.  It turns out that we didn't regenerate the services before we released 0.9.0, so the comments from the service protos got updated in the generated java files.


garye:
LGTM

PR #901, opened 2016-06-30
Fix documentation of data proto

rgruener:
This was bothering me and pretty confusing at first glance. Could be better to switch the order as well so that closed comes before open as in every other range key.


rgruener:
pretty sure spotify signed the cla

sduskis:
Thanks for the catch!  We have a bit of a workflow to update protos.  I'll start the process.

Note to other Mergers: please don't merge this yet.

sduskis:
I fixed this in another PR.  Closing this one.

PR #903, opened 2016-07-01
Adding RetryOptions.toBuilder()

### LINKED ISSUE #899: Add toBuilder() to RetryOptions and BulkOptions
There is no easy way to change a RetryOptions into a RetryOptions.Builder. Same for BulkOptions.

sduskis:
This should fix #899.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/RetryOptions.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/RetryOptions.java
@@ -92,6 +92,22 @@ public class RetryOptions implements Serializable {
     private Set<Status.Code> statusToRetryOn = new HashSet<>(DEFAULT_ENABLE_GRPC_RETRIES_SET);
     private boolean allowRetriesWithoutTimestamp;
 
+    public Builder() {
+    }
+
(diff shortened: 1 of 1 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/RetryOptions.java
```diff
@@ -92,6 +92,22 @@
     private Set<Status.Code> statusToRetryOn = new HashSet<>(DEFAULT_ENABLE_GRPC_RETRIES_SET);
     private boolean allowRetriesWithoutTimestamp;
 
+    public Builder() {
+    }
+
+    public Builder(RetryOptions options) {
+      enableRetries = options.retriesEnabled;
```
garye: minor, but the constructor in BulkOptions uses this.<field> for everything. Would be good to make consistent.

garye:
LGTM other than minor comment

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/RetryOptions.java
sduskis: Good point. Fixed.

PR #905, opened 2016-07-06
Improving getTable() performance.

sduskis:
AbstractBigtableConnection.getTable() takes 1-2 microseconds.  Caching or deferring instantiation of objects reduces that to 100-300 nanoseconds.


garye:
LGTM

PR #906, opened 2016-07-06
Removing the generics in ReasponseQueueReader.

sduskis:
The generics in ReasponseQueueReader were needed for the v1 / v2 transition.  Removing them will help set the stage to create a more performat system that takes advantage of the v2 semantics.


garye:
LGTM

PR #908, opened 2016-07-11
Pushing out CallOptions settings to BigtableDataGrpcClient.

sduskis:
There ought to be user control over timeouts. This change updates the
plumbing, but stops short of changing public interfaces that allow that
control.  The public interface changes will happen in a future PR.


garye:
LGTM

PR #909, opened 2016-07-11
Updating use of grpc cancel methods to new signature.

sduskis:
ClientCall.cancel() was deprecated in favor of ClentCall.cancel(message,
throwable).


garye:
LGTM

PR #914, opened 2016-07-13
Adding configuration based CallOptions generator.

sduskis:
We use the term Options to convey that an object is a configuration object (BigtableOptions, RetryOptions and etc).  That naming was a bit awkward when used as CallOptions configuration; CallOptionsOptions just doesn't sound right, so I used CallOptionsConfig.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
@@ -73,9 +73,9 @@ public class BigtableOptions implements Serializable {
 
     private RetryOptions retryOptions = new RetryOptions.Builder().build();
     private CallOptionsConfig callOptionsConfig = new CallOptionsConfig.Builder().build();
-    // The default credentials get credential from well known locations, such as the GCE
-    // metdata service or gcloud configuration in other environments. A user can also override
-    // the default behavior with P12 or JSon configuration.
+    // CredentialOptions.defaultCredentials() gets credentials from well known locations, such as
+    // the GCE metdata service or gcloud configuration in other environments. A user can also
+    // override the default behavior with P12 or JSon configuration.
     private CredentialOptions credentialOptions = CredentialOptions.defaultCredentials();
 
     public Builder() {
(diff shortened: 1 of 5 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
```diff
@@ -66,18 +66,18 @@ private static int getDefaultDataChannelCount() {
...
-    // Performance tuning options.
-    private RetryOptions retryOptions = new RetryOptions.Builder().build();
     private int dataChannelCount = BIGTABLE_DATA_CHANNEL_COUNT_DEFAULT;
 
     private BulkOptions bulkOptions;
     private boolean usePlaintextNegotiation = false;
 
+    private RetryOptions retryOptions = new RetryOptions.Builder().build();
+    private CallOptionsConfig callOptionsConfig = new CallOptionsConfig.Builder().build();
+    // The default credentials get credential from well known locations, such as the GCE
```
garye: "get credential" -> "get credentials"

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
```diff
@@ -0,0 +1,106 @@
...
+ */
+package com.google.cloud.bigtable.config;
+
+import java.io.Serializable;
+
+import com.google.common.base.MoreObjects;
+import com.google.common.base.Preconditions;
+
+/**
+ * Experimental options to turn on timeout options and .
```
garye: Incomplete comment

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/CallOptionsFactory.java
```diff
@@ -47,4 +53,27 @@
       return CallOptions.DEFAULT;
     }
   }
+
+  /**
+   * Creates a new {@link CallOptions} based on a {@link CallOptionsConfig}. This factory assumes
+   * {@link CallOptionsConfig#isUseTimeout()} is true.
```
garye: Why does it assume isUseTimeout == true? It seems to handle both cases in create().

garye:
LGTM other than minor comments

PR #918, opened 2016-07-17
PutAdapter performance improvement.

sduskis:
Creating a proto with protoBuilder.addXXXBuilder() is slower than protoBuilder.addAllXXX(List<XXX>).  This makes me sad.  Also, cell.getXXXLength() methods are not simple getters; they do non trivial work, so caching the value saves some CPU time as well.

Our PutAdapterPerf test has a single Put with 100 small Cells in the same column family.  Before the two changes, that took 18-45 microseconds on my machine in various runs on my laptop; the results were varied depending on the CPU load, whether a parallel run happened right before.

After the change, that reliably took 9-14 microseconds, with more results toward the 9-11 us.

### DIFF
--- a/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/PutAdapter.java
+++ b/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/PutAdapter.java
@@ -19,7 +19,6 @@ import com.google.bigtable.v2.MutateRowRequest;
 import com.google.bigtable.v2.Mutation;
 import com.google.bigtable.v2.Mutation.MutationCase;
 import com.google.bigtable.v2.Mutation.SetCell;
-import com.google.bigtable.v2.Mutation.SetCell.Builder;
 import com.google.cloud.bigtable.hbase.BigtableConstants;
 import com.google.cloud.bigtable.hbase.adapters.read.RowCell;
 import com.google.protobuf.BigtableZeroCopyByteStringUtil;
@@ -30,6 +29,7 @@ import org.apache.hadoop.hbase.HConstants;
 import org.apache.hadoop.hbase.client.Put;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map.Entry;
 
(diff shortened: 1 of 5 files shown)


bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/PutAdapter.java
```diff
@@ -57,56 +57,55 @@ public PutAdapter(int maxKeyValueSize, boolean setClientTimestamp) {
...
+        int qualifierLength = cell.getQualifierLength();
+        int valueLength = cell.getValueLength();
         // Since we are not using the interface involving KeyValues, we reconstruct how big they would be.
         // 20 bytes for metadata plus the length of all the elements.
-        int keyValueSize = (20 +
-            cell.getRowLength() +
-            cell.getFamilyLength() +
-            cell.getQualifierLength() +
-            cell.getValueLength());
+        int keyValueSize = (20 + rowLength + familySize + qualifierLength + valueLength);
```
rameshdharan: Where does the magic 20 come from? Could we hoist it into `BigtableConstants`?

Also, would it make sense to put this calculation into a private static getKeyValueSize method that takes an operation and cell?

rameshdharan:
Could you add some before/after timing information to the commit and maybe a one-liner about the source of the perf improvement? It took me a little while staring at this to grok the difference and so something like "use `Mutation.Builder.addAllMutations` to batch all the mutations for the row request with a single Builder" in the commit would have been helpful IMO.

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/PutAdapter.java
sduskis: I'm not quite sure where it came from.  I'll need to ping @AngusDavis for that.

sduskis:
I updated the comment.  PTAL.

rameshdharan:
Consider this an LGTM. I'd still like to see the constant defined somewhere but that was not introduced in this change so no need to hold it up. 

Meanwhile, the new comment looks fine to me. However, the actual commit still has the old comment, so would be nice to amend it there as well.

sduskis:
I updated the commit message.  I also added #920.

PR #921, opened 2016-07-18
Add versions for maven javadoc and source plugins

rameshdharan:
Quashes some spammy warnings by specifying versions to use for the
javadoc and source plugins.


sduskis:
LGTM

PR #926, opened 2016-07-21
Switch <offlineLinks> -> <links> for javadoc

rameshdharan:
We don't have a particular need to use the <offlineLink>/-linkOffline
capability and it doesn't work as currently written so let's just use
regular <links> for now and we can revisit this later if we need to.

Testing Done:
- ran the javadoc:javadoc build, verified the link errors went away


sduskis:
LGTM

PR #927, opened 2016-07-22
More javadoc fixes

rameshdharan:
- Exclude v1 API 
- Add fixes generated by javadoc:fix


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/Logger.java
```diff
@@ -20,68 +20,140 @@
...
+   */
   public Logger(Class<?> logClass) {
     this.log = LogFactory.getLog(logClass);
   }
 
+  /**
+   * <p>trace.</p>
+   *
+   * @param message a {@link java.lang.String} object.
+   * @param args a {@link java.lang.Object} object.
```
sduskis: The tool doesn't mention that this is a varag.  Can you please update the various method with varags to clarify?

rameshdharan: Nice catch, will do.

sduskis:
LGTM

PR #929, opened 2016-07-27
Scans where start == end should be the same as a Get

### LINKED ISSUE #928: Scans where start == end should be the same as a Get
Externally-reported issue:

> When I submit a scan that has the same start and end row then i get this exception: 
> 
> ```
> INVALID_ARGUMENT: Error in field 'row_ranges' : Error in element #0 : start_key
> must be less than end_key 
> ```
>
> This will now require me to duplicate a bunch of code so that I can use a `Get` when there is only one row I am selecting.
>
> Can you treat those scans as gets the way that HBase does?

sduskis:
This should fix #928.

### DIFF
--- a/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ScanAdapter.java
+++ b/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ScanAdapter.java
@@ -93,14 +93,14 @@ public class ScanAdapter implements ReadOperationAdapter<Scan> {
     RowFilter filter = buildFilter(scan, readHooks);
 
     RowSet.Builder rowSetBuilder = RowSet.newBuilder();
+    ByteString startRow = ByteString.copyFrom(scan.getStartRow());
     if (scan.isGetScan()) {
-      rowSetBuilder.addRowKeys(ByteString.copyFrom(scan.getStartRow()));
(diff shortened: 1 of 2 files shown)


bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ScanAdapter.java
```diff
@@ -90,16 +90,17 @@ public RowFilter buildFilter(Scan scan, ReadHooks hooks) {
...
-              RowRange.newBuilder()
-                  .setStartKeyClosed(ByteString.copyFrom(scan.getStartRow()))
-                  .setEndKeyOpen(ByteString.copyFrom(scan.getStopRow()))));
+    RowSet.Builder rowSetBuilder = RowSet.newBuilder();
+    if (scan.isGetScan()) {
+      rowSetBuilder.addRowKeys(ByteString.copyFrom(scan.getStartRow()));
+    } else {
+      rowSetBuilder.addRowRanges(
+        RowRange.newBuilder().setStartKeyClosed(ByteString.copyFrom(scan.getStartRow()))
+            .setEndKeyOpen(ByteString.copyFrom(scan.getStopRow())));
```
rameshdharan: Indentation here seems a bit inconsistent. Personally I'm fine with the style the way it was (with leading periods) but in both cases the indentation seems off (4 spaces instead of 2)?

rameshdharan:
LGTM (except for the very minor indentation issue).

PR #930, opened 2016-07-28
Support for connecting to the gcloud bigtable emulator

garye:
The BIGTABLE_EMULATOR_HOST environment variable can be set to the host:port of the emulator.

Note that, if set, the emulator host value will override any host, port or plaintext negotiation settings that the user overrode in other ways. The log message should make it clear that this is happening.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
@@ -32,6 +32,11 @@ public class BigtableOptions implements Serializable {
 
   private static final long serialVersionUID = 1L;
 
+  // If set to a host:port address, this environment variable will configure the client to connect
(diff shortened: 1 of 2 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
```diff
@@ -166,6 +171,37 @@ public Builder setCallOptionsConfig(CallOptionsConfig callOptionsConfig) {
...
+     */
+    private void applyEmulatorEnvironment() {
+      // Look for a host:port for the emulator.
+      String emulatorHost = System.getenv(BIGTABLE_EMULATOR_HOST_ENV_VAR);
+      if (emulatorHost == null) {
+        return;
+      }
+
+      String[] hostPort = emulatorHost.split(":");
+      if (hostPort.length != 2) {
```
sduskis: Preconditions?

garye:
PTAL

sduskis:
setTestEnv() is unfortunate, but I see it's necessary.

LGTM.

PR #931, opened 2016-07-29
Replacing channel expand pool with pre-created pool.

sduskis:
I've been meaning to remove the expand pool for a while.  I ran the client load test, and there was no discernible difference between a pre-created pool and one that gets created on demand.  I'm going to run the latency test before assigning a reviewer.


sduskis:
This slowed down the latency test by .3 ms. closing.

Issue #505, opened 2015-10-09
Investigate a mechanism for performance insight into CBT client


sduskis:
Dropwizard (http://metrics.dropwizard.io/) is a good candidate.  Same thing for https://cloud.google.com/monitoring/

sduskis:
We can call this done.  This was implemented and deployed.

PR #933, opened 2016-08-02
Initial work for Dropwizard integration.

sduskis:
- Adds a wrapper for Timer and Counter
- Adds RPC and Operation level timers, and retry counters.

This should start towards the goals of #505.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/metrics/BigtableClientMetrics.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/metrics/BigtableClientMetrics.java
@@ -16,20 +16,43 @@
 package com.google.cloud.bigtable.metrics;
 
 import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.Reporter;
 import com.codahale.metrics.Slf4jReporter;
 import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * Singleton Container for a MetricRegistry.
- *
+ * Singleton Container for a {@link MetricRegistry}. The default behavior is to return
+ * implementations that do nothing. Exporting of metrics can be turned on by either adding TRACE
+ * level logging for this class, which will write out all metrics to a log file, or via a call to
+ * {@link BigtableClientMetrics#enable} followed by a programmatic configuration of reporters as per
+ * the instructions on <a href="http://metrics.dropwizard.io/3.1.0/getting-started/">the Dropwizards
+ * Metrics Getting Started docs</a>.
  * @author sduskis
- *
  */
 public final class BigtableClientMetrics {
 
   private static MetricRegistry CLIENT_STATS;
+
+  // Simplistic initialization via slf4j
+  static {
+    // This adds a simple mechanism of enabling statistics via slf4j configuration.
+    // More complex configuration is available programmatically.
+    Logger logger = LoggerFactory.getLogger(BigtableClientMetrics.class);
+    if (logger.isTraceEnabled()) {
+      enable();
+      final Slf4jReporter reporter =
+          Slf4jReporter.forRegistry(getClientStats())
+              .outputTo(logger)
+              .convertRatesTo(TimeUnit.SECONDS)
+              .convertDurationsTo(TimeUnit.MILLISECONDS)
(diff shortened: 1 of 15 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/metrics/BigtableClientMetrics.java
```diff
@@ -0,0 +1,112 @@
...
+  private static MetricRegistry CLIENT_STATS;
+
+  private static final Counter NULL_COUNTER = new Counter() {
+    @Override
+    public void inc() {
+    }
+  };
+
+  private static final Timer NULL_TIMER = new Timer() {
+
```
garye: Kill the newline

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/metrics/BigtableClientMetrics.java
```diff
@@ -0,0 +1,112 @@
...
+          Slf4jReporter.forRegistry(CLIENT_STATS)
+              .outputTo(logger)
+              .convertRatesTo(TimeUnit.SECONDS)
+              .convertDurationsTo(TimeUnit.MILLISECONDS)
+              .build();
+      reporter.start(1, TimeUnit.MINUTES);
+    }
+  }
+
+  public static void enable() {
```
garye: Might as well synchronize, unless this is being called somewhere safe that I missed.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/metrics/Counter.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.metrics;
+
+/**
+ * Wrapper around a Dropwizard Metrics {@link com.codahale.metrics.Counter}.
```
garye: No need to mention the Dropwizard implementation in the interface comments

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/metrics/BigtableClientMetrics.java
```diff
@@ -0,0 +1,112 @@
...
+    Logger logger = LoggerFactory.getLogger(BigtableClientMetrics.class);
+    if (logger.isTraceEnabled()) {
+      CLIENT_STATS = new MetricRegistry();
+      final Slf4jReporter reporter =
+          Slf4jReporter.forRegistry(CLIENT_STATS)
+              .outputTo(logger)
+              .convertRatesTo(TimeUnit.SECONDS)
+              .convertDurationsTo(TimeUnit.MILLISECONDS)
+              .build();
+      reporter.start(1, TimeUnit.MINUTES);
```
garye: Maybe make this configurable?

sduskis:
I updated the comments for Timer, Counter and BigtableClientMetrics.

In terms of configurability, I was thinking of programmatic configuration.  The Metrics is a singleton thing, but most of the configuration we have is Connection level.   Users can customize reporters programmatically or get the out of the box simple implementation.  Do you have a better alternative?

garye:
I think programmatic configurability makes sense. It seemed like this default reporter could be potentially be more or less verbose than someone wanted and I wanted to make sure that anyone who wanted tracing didn't also have to stick with the once-per-minute log dump.

garye:
LGTM - I think I missed that just enabling that certain logger for tracing did all of this. That seems ok for now.

Perhaps could also add some pre-canned reporters that are accessible via static methods on BigtableClientMetrics as a middle ground between unconfigurable tracing and doing whatever you want with the metric API. Like newLoggingReporter(TimeUnit reportingInterval), newConsoleReporter(TimeUnit reportingInterval), etc. 

Something to think about for next revision, though.

sduskis:
Thanks for the review.  I like those suggestions.  Lets discuss the options in more detail.

PR #934, opened 2016-08-02
Upgrading grpc version and regenerating java protos.

sduskis:
Updating the ZeroCopyByteString objects.  Please take a look at those
Please checkout the top pom.xml and BigtableSession changes as well.

There's a whole bunch of generated code that doesn't need to be reviewed too carefully.


garye:
LGTM

PR #935, opened 2016-08-02
Updating the top level project to cloud-bigtable-client-parent.

sduskis:
I'd like to use cloud-bigtable-client for the new API, so renaming the top level to include the '-parent' extension which is more maven-y.


garye:
LGTM. I have a similar change on a throwaway branch.

PR #936, opened 2016-08-02
Adding active channel count and per channel latency metrics.

sduskis:


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -45,11 +49,38 @@
   /** Constant <code>LOG</code> */
   protected static final Logger LOG = new Logger(ChannelPool.class);
 
+  private static final AtomicInteger ChannelIdGenerator = new AtomicInteger();
```
rameshdharan: Random aside - I find it surprising that there's no existing unique identifier on the channel. 

What about just using the ManagedChannel hashCode()?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -45,11 +49,38 @@
...
+
+  /**
+   * Best effort counter of active channels. There may be some cases where channel termination
+   * counting may not accurately be decremented.
+   */
+  protected static Counter ACTIVE_CHANNEL_COUNTER;
+
+  protected synchronized static Counter getActiveChannelCounter() {
+    if (ACTIVE_CHANNEL_COUNTER == null) {
```
rameshdharan: Curious, why not just initialize in the ChannelPool constructor rather than lazily? 

Is it because you want to avoid the initialization in the case where the metrics stuff is turned off?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -66,17 +97,22 @@
...
     }
   }
 
+  private ChannelAndMetrics createChannelAndMetrics() throws IOException {
+    return new ChannelAndMetrics(factory.create());
+  }
+
   /**
    * Makes sure that the number of channels is at least as big as the specified capacity.  This
    * method is only synchornized when the pool has to be expanded.
```
rameshdharan: Not part of this change but mind fixing the spelling mistake in the comment? s/synchornized/synchronized/

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -92,29 +128,18 @@ public void ensureChannelCount(int capacity) throws IOException {
     if (channels.get().size() < capacity) {
       synchronized (this) {
         if (channels.get().size() < capacity) {
-          List<ManagedChannel> newChannelList = new ArrayList<>(channels.get());
+          Counter counter = getActiveChannelCounter();
+          List<ChannelAndMetrics> newChannelList = new ArrayList<>(channels.get());
           while(newChannelList.size() < capacity) {
-            newChannelList.add(factory.create());
+            newChannelList.add(createChannelAndMetrics());
+            counter.inc();
```
rameshdharan: Maybe just getActiveChannelCounter().inc()?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -134,15 +159,51 @@ public String authority() {
   public <ReqT, RespT> ClientCall<ReqT, RespT> newCall(
       MethodDescriptor<ReqT, RespT> methodDescriptor, CallOptions callOptions) {
     Preconditions.checkState(!shutdown, "Cannot perform operations on a closed connection");
-    ClientCall<ReqT, RespT> delegate = getNextChannel().newCall(methodDescriptor, callOptions);
+    int currentRequestNum = requestCount.getAndIncrement();
+    ImmutableList<ChannelAndMetrics> channelsList = channels.get();
+    int index = Math.abs(currentRequestNum % channelsList.size());
```
rameshdharan: Why nuke the getNextChannel() logic here? NBD but personally I think it is a bit more readable to keep the modulus/round-robin logic in a separate function. 

Separately it makes me a *bit* nervous that there's nothing inherently ensuring that the channel we choose here is actually valid AFAICT? I guess the channelsList only ever grows, so it's not like we'll get an invalid index, but don't we want to ensure that the channel we select isn't already shutdown (and wouldn't doing that correctly require more synchronization)?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
sduskis: There's a classloading vs. user configuration timing issue.  A user should have the opportunity to configure metrics gathering even if the class loader loads the ChannelPool.  If a user turns on metrics, these metrics should not be lost.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
sduskis: Is there a reason to incur the cost of the synchronized method again?  Do you think getActiveChannelCounter().inc() looks nicer?  maybe rename counter to activeChannelCounter?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
sduskis: I'll add this as a comment.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -206,12 +272,18 @@ public ManagedChannel shutdownNow() {
   @Override
   public boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException {
     long endTimeNanos = System.nanoTime() + unit.toNanos(timeout);
-    for (ManagedChannel channel : channels.get()) {
+    Counter counter = getActiveChannelCounter();
```
rameshdharan: NB: the logic here is slightly strange, although I think it's probably fine. We potentially spend our entire time budget awaiting for termination on the first non-terminated channel we find. Of course the implementation of awaitTermination doesn't actually do anything either other than just sleeping for the entire allotted time, so again I guess none of this matters, but it's still weird to me. 

It would read more natural IMO if we just first checked for any non-terminated channels, then if we find any, slept for some increment, checking all of them again, but I imagine this is not worth worrying about.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -45,11 +49,38 @@
...
+    }
+    return ACTIVE_CHANNEL_COUNTER;
+  }
+
   public interface ChannelFactory {
     ManagedChannel create() throws IOException;
   }
 
-  private final AtomicReference<ImmutableList<ManagedChannel>> channels = new AtomicReference<>();
+  private static class ChannelAndMetrics {
```
rameshdharan: FWIW I think InstrumentedChannel is a better name than ChannelAndMetrics.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
sduskis: On second thought, I'm going to move the inc()/dec() responsibility to ChannelAndMetrics.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -196,8 +257,13 @@ public boolean isTerminated() {
...
   @Override
   public ManagedChannel shutdownNow() {
-    for (ManagedChannel channel : channels.get()) {
-      channel.shutdownNow();
+    Counter counter = getActiveChannelCounter();
+    for (ChannelAndMetrics channelWrapper : channels.get()) {
+      ManagedChannel channel = channelWrapper.channel;
+      if (!channel.isTerminated()) {
+        channel.shutdownNow();
+        counter.dec();
```
rameshdharan: getActiveChannelCounter().dec()?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
rameshdharan: ^ that sounds like a good idea.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
sduskis: will do

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
sduskis: hashCode() is interesting.  I think that it's useful to see an incrementing number.  It will help group and order channels.

rameshdharan: Yeah, I agree, hashCode is likely to be too unwieldy / will look like a large annoying random number in log output.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
rameshdharan: On a related note, instead of doing modulo+indexing on the list, you could use a cycling deque+re-enque pattern with an ArrayDeque, or the Guava cycling iterator: 

https://github.com/google/guava/blob/165ef0d34adffb8a5ac84b859e9d58cd68412193/guava/src/com/google/common/collect/Iterables.java#L400

sduskis:
I updated comments, fixed accounting logic for closing channels, and reinstituted getNextChannel().  PTAL

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -57,26 +58,55 @@
...
     }
     return ACTIVE_CHANNEL_COUNTER;
   }
 
-  public interface ChannelFactory {
-    ManagedChannel create() throws IOException;
-  }
-
+  /**
+   * Contains a {@link ManagedChannel} and metrics for the channel
```
rameshdharan: Sorry for not asking this earlier, but did you consider subclassing ManagedChannel so that this is a real wrapper?

If you did that you could roll in the channelId concept, and fold in the markInactive semantics to the shutdown/shutdownNow overrides.

rameshdharan:
Overall LGTM, though I think it's worth pondering the possibility of making ChannelAndMetrics (a.k.a. InstrumentedChannel :) a subclass of ManagedChannel (is-a rather than has-a). I don't feel too strongly in favor of it so feel free to ignore.

sduskis:
InstrumentedChannel makes sense.  I'm going to Merge this so that we can see if there any impacts in tonights performance tests.  I'll mull over is-a.

I'll rename ChannelAndMetrics to InstrumentedChannel before I do that.

sduskis:
InstrumentedChannel now implements ManagedChannel.  It looks a lot better.  PTAL.

rameshdharan:
Awesome, yes this looks great now. Nicely done!

PR #938, opened 2016-08-02
Fix a couple new javadoc warnings

rameshdharan:


sduskis:
LGTM.  Thanks for the catch!

PR #942, opened 2016-08-05
Fix javadoc:aggregate-jar target

rameshdharan:
This change makes the javadoc:aggregate-jar target work at the
top-level.

Doing this correclty is complicated because javadoc:aggregate is
confused by the repackaged/shadedPattern renaming of the bigtable client
in the dataflow submodules.

After a lot of experimentation, banging my head against the wall, and
discussions with Jeff I decided the best way to handle javadoc at
the top-level was simply to exclude the dataflow projects; we already
have a way to handle and publish those docs that we aren't going to
change, so we simply skip the com.google.cloud.bigtable.dataflow*
packages when doing a top-level javadoc:aggregate[-jar] for the main
project.

I also made a change to ensure that you can turn back on the doclint
behavior by passing -Ddoclint on the commandline if you want to inspect
it for any javadoc target, though it's still not particularly useful
given all the noise from grpc-java (and note that they have it disabled
in their build).

I also made some tweaks to the way we express configuration so the
javadoc stuff will apply for all profiles, and the aggregate-jar target
will be used during the package phase.

Finally I threw in a couple of fixes for new warnings that I noticed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,27 +75,9 @@ You can run the following command for Java 7:
         -Dgoogle.bigtable.project.id=[your cloud project id] \
         -Dgoogle.bigtable.instance.id=[your cloud bigtable instance id] \
 
-## Deploying
-**TODO**
-
-### How to setup the deployment environment
-**TODO**
-
-
-### How to deploy
-**TODO**
-
-
-## Troubleshooting & useful tools
-**TODO**
-
-### Examples of common tasks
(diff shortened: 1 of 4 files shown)


sduskis:
LGTM.

I have a follow up request for another PR.  Let's please get rid of the v1 proto files and the excludePackageNames.

PR #944, opened 2016-08-05
Removing v1 protos

sduskis:
This removes 117 java files that no longer need to be compiled every build or add bulk to our jars.
While regenerating the protos, it turned out that v2 grpc Service objects needed to be regenerated
Upgraded one test class that was using v1 classes to use v2.


garye:
LGTM

I looked at the upgraded test class but the rest isn't really reviewable.

PR #945, opened 2016-08-05
Upgrading plugins.

sduskis:
Basically ran "mvn versions:display-plugin-updates" to get the latest versions.
Also removed some explicit version settings in child pom.xml and moved them to depend on the top pom.xml's versions.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/pom.xml
+++ b/bigtable-client-core-parent/bigtable-client-core/pom.xml
@@ -133,7 +133,6 @@ limitations under the License.
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-release-plugin</artifactId>
-                <version>2.5.1</version>
             </plugin>
         </plugins>
     </build>
--- a/bigtable-dataflow-parent/bigtable-dataflow-import/pom.xml
+++ b/bigtable-dataflow-parent/bigtable-dataflow-import/pom.xml
@@ -124,7 +124,6 @@ limitations under the License.
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-eclipse-plugin</artifactId>
-                <version>2.9</version>
                 <configuration>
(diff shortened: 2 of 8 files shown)


bigtable-client-core-parent/bigtable-client-core/pom.xml
```diff
@@ -133,7 +133,6 @@ limitations under the License.
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-release-plugin</artifactId>
-                <version>2.5.1</version>
```
rameshdharan: How come the version was removed in this case? I guess that makes it default to 'latest' but still makes me a bit nervous, maybe we can pin it more explicitly?

bigtable-dataflow-parent/bigtable-dataflow-import/pom.xml
```diff
@@ -124,7 +124,6 @@ limitations under the License.
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-eclipse-plugin</artifactId>
-                <version>2.9</version>
```
rameshdharan: Same question here, how come no version for maven-eclipse-plugin?

rameshdharan:
LGTM except curious why the versions were removed for two of the plugins.

bigtable-dataflow-parent/bigtable-dataflow-import/pom.xml
sduskis: It was removed from the child, and now the version is managed in the top pom.xml for all usages of maven-eclipse-plugin

bigtable-client-core-parent/bigtable-client-core/pom.xml
sduskis: This isn't latest... it depends on the top pom's version management.

sduskis:
The top pom has dependencyManagement and pluginManagement which drive the version management for child poms.

rameshdharan:
Got it, thanks for clarifying.

PR #948, opened 2016-08-07
Adding more metrics

sduskis:


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/StreamingBigtableResultScanner.java
```diff
@@ -48,7 +53,13 @@ public StreamingBigtableResultScanner(ResponseQueueReader responseQueueReader,
   /** {@inheritDoc} */
   @Override
   public Row next() throws IOException {
-    return responseQueueReader.getNextMergedRow();
+    resultsMeter.mark();
```
garye: Isn't this counting calls to next() instead of Rows returned?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/metrics/BigtableClientMetrics.java
```diff
@@ -164,6 +174,31 @@ public void close() {
     }
   }
 
+  /**
+   * Creates a named {@link Meter}.
+   *
+   * @param name
+   * @return a Dropwizard Metrics {@link com.codahale.metrics.Meter} or {@link BigtableClientMetrics#NULL_METER}.
```
garye: Don't need a comment that leaks implementation details

sduskis: +1 on removing the leakiness.  Currently, all 3 types of metrics have comments leak implementation details.  I plan to a MetricsFactory interface, with Null and Dropwizard implementations.  Is it ok to fix all of the comments with the next PR?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/StreamingBigtableResultScanner.java
sduskis: Does it look better now?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/metrics/BigtableClientMetrics.java
garye: Works for me

garye:
LGTM

PR #949, opened 2016-08-07
Moving scan retry logic into a separate class.

sduskis:
This will allow us to use the retry functionality in other scenarios, specifically observer based responses as described in #703.


garye:
LGTM

PR #955, opened 2016-08-09
Adding Logging levels to BigtableClientMetrics.

sduskis:


garye:
Does this implementation prevent us from changing metric log levels at runtime? I think we want to be able to go from INFO -> DEBUG without restarting, for example.

sduskis:
I'm going to punt on runtime changes for now.  It would be great if we could add it in the near future.  Can this PR proceed without the runtime change flexibility?

sduskis:
I don't think anything in this PR prevents that kind of future enhancement.

rameshdharan:
To build on Gary's point, I wonder is it necessary to have the *collection* based on the level - is there some way we could get away with doing the filtering in the output instead? That is to say, always collect everything but filter the log output according to whatever level you've selected?

garye:
My concern is the implementation detail that a NULL metrics registry could get handed out depending on the log level at the time that `getMetricRegistry` was called, for example. I suppose this can be revisited in a future PR as long as it wouldn't require any changes to the interface. As long as you don't think that it would then LGTM and we'll tackle it later. Ramesh's suggestion might be a way forward when we get to it.

rameshdharan:
Yeah so looks like there's not really a way to do what I was asking in the current Dropwizard release AFAICT. 

Basically I was hoping you could get away with a single registry and have the reporters do filtering. I believe that's what https://github.com/dropwizard/metrics/issues/178 is asking for.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BigtableAsyncRpc.java
```diff
@@ -43,14 +44,14 @@
...
+      MetricRegistry debugRegistry = getMetricRegistry(MetricLevel.Debug);
+      MetricRegistry infoRegistry = getMetricRegistry(MetricLevel.Info);
       return new RpcMetrics(
-          registry.createTimer(fullMethodName + ".operation.latency"),
-          registry.createTimer(fullMethodName + ".rpc.latency"),
-          registry.createCounter(fullMethodName + ".retry.count"),
-          registry.createCounter(fullMethodName + ".failure.count"),
-          registry.createCounter(fullMethodName + ".retries.exhausted.count")
-          );
+          infoRegistry.createTimer(fullMethodName + ".operation.latency"),
```
rameshdharan: Suggested this out-of-band to @sduskis but one nice way to handle this instead would be to have our own createTimer, createCounter methods that take the metric string name and return the appropriate registry. 

For example we could append ".loglevel" to the end of the strings and then look them up in the slf4j config.

sduskis:
As per a conversation with Ramesh, I did some quick testing, and each Dropwizard Timer adds ~500 nanos.  There's going to be some metrics that are useful in very few cases, and we shouldn't cause a runtime impact for all users for those few cases.

At this point, all metrics have to be turned on before creating a Connection.  I think we can support  runtime changes in the future with another level of indirection in a future PR.

PR #956, opened 2016-08-09
Fix javadoc build (broken link)

rameshdharan:
Not sure why but the at-links sometimes need full resolution or they
produce Javadoc warnings. This is hopefully the last thing preventing
the jenkins javadoc:aggregate job from going green. Let's find out.


sduskis:
I've seen that before with nested classes.  I'm not sure why that happens.

PR #959, opened 2016-08-10
Refactoring BigtableAsyncRpc

sduskis:
The goal here is to move the read rows functionality into the same
listener model that all of the other RPCs use.  That will help with
metrics gathering and the "observer" read rows functionality described
in #703.


garye:
LGTM

Issue #958, opened 2016-08-10
Cannot use bigtable-hbase and bigtable-dataflow together


vc13819:
If the two jars are used together, the BigtableOpitons is one is shaded, but not in another. Cannot use them together.

sduskis:
This was fixed in one of the previous releases of the cloud bigtable client.  Which version are you on now?

sduskis:
I'm closing this issue, since I don't think this is an actionable request on our end.  Please upgrade to the latest client.  If that does not work, please reopen this issue.

vc13819:
I am using 0.9.1. This is the latest release. Reopening the issue.

sduskis:
Can you please help me reproduce this problem?  Code snippets and pom.xml are welcome.

sduskis:
FYI, if you want to use bigtable-hbase, see https://github.com/sduskis/cloud-bigtable-examples/tree/master/java/dataflow-connector-examples for the way we do it.  If you'd like, feel free to reach out to me in the email on my profile with more detailed information, or you can add the information here.

sduskis:
maybe this helps... There are two versions of BigtableOptions and BigtableSession.  Use the ones prefixed by the package name: com.google.bigtable.repackaged.com.google.cloud

vc13819:
The following code when compiled with both bigtable-hbase-1.1 and
bigtable-dataflow in

its dependencies. The version I am using is 0.9.1

Note that the usage of Options is done deep into a jar (not in the code I

have written). I do not even have an import for BigtableOptions in my code.

public static void main(String... args) {
  CloudBigtableOptions options =
      PipelineOptionsFactory.fromArgs(args).withValidation().as(CloudBigtableOptions.class);
  options.setBigtableInstanceId(System.getProperty("bigtable-instance-id"));
  options.setBigtableProjectId(System.getProperty("bigtable-project-id"));
  options.setBigtableTableId(System.getProperty("bigtable-name"));
  options.setStagingLocation("gs://" +
      System.getProperty("storage-bucket-id") + "/staging/dataflow");
  Scan scan = new Scan();
  scan.addFamily("p".getBytes());
  scan.setCacheBlocks(false);
  CloudBigtableScanConfiguration config =
CloudBigtableScanConfiguration.fromCBTOptions(options)
      .toBuilder().withScan(scan).build();
  Pipeline p = Pipeline.create(options);
  p
[...]

PR #968, opened 2016-08-12
Fixing hbase/Dataflow shading.

sduskis:
A user who has bigtable-hbase-1.* in the classpath should be able to use bigtable-hbase-dataflow.  Currently, as see in #958, the shading we added so that BigtableIO and CloudBigtableIO can coexist doesn't quite work for bigtable-hbase.


garye:
LGTM

PR #974, opened 2016-08-16
Removing Dynamic rebalancing needs to be carefully tested.

sduskis:
The current implementation causes duplicate values while reading.


garye:
I think it would be cleaner to just delete the code

sduskis:
Deleted the code.  PTAL.

garye:
LGTM

PR #977, opened 2016-08-17
Adding pom.xml exclusions for shaded artifacts.

sduskis:
This helps the eclipse plugin understand what dependencies not to include for shaded projects.


rameshdharan:
LGTM. 

I imagine all of this could still be simplified further in the future if we revisit the fundamental split between bigtable-client-core, bigtable-hbase, and dataflow. Future...

PR #986, opened 2016-08-23
Remove attach-javadocs from package target

rameshdharan:
Causing problems for the release process, so let's disable it for now.


sduskis:
LGTM

PR #989, opened 2016-08-23
Restore excluded package names in javadoc default-cli

rameshdharan:
Removing the exclusions for dataflow breaks the javadoc:aggregate target, which we
still want to keep working when it is invoked directly.

With this change, we add the additional dataflow exclusions for the
javadoc plugin under the default-cli execution id (used when the plugin
is invoked directly), but leave them out in the attach-javadocs
execution (used as part of the package phase / release process).


sduskis:
LGTM

PR #990, opened 2016-08-29
Fixing an issue in the HBaseMutationCoder.

sduskis:
PutAdapter is used by HBaseMutation coder and PutAdapter has a bit of a hack to set the timestamp on the client side.  That breaks Dataflow's requirement of having a consistent serialization of objects.  We've been telling users to set client side timestamps when they hit this obscure problem.  This PR fixes the serialization to not add the client side timestamp for the Dataflow serialization, thereby fixing the underlying problem.


garye:
LGTM

PR #991, opened 2016-08-30
Removing the streamingBatchSize option, since it's no longer useful.

sduskis:


rameshdharan:
Code change LGTM. 

What did this option used to do / why is it no longer useful?

sduskis:
This option was very useful back in the V1 days when we read many small responses, one per row.  This option would allow us to configure how many requests we'd ask back from the server.  For small rows, asking gRPC to queue up many responses improved performance.  The bottleneck was the latency of streaming responses between the client and the proxy. These days, we get a large response with many rows, and the old bottleneck is gone.  Reading n responses at a time doesn't change performance in V2, and requires more complicated code.

PR #992, opened 2016-08-30
Use null credentials with bigtable emulator

rgruener:
When using the bigtable emulator, you should not need to define credentials since the emulator does not require them. Currently in a completely new terminal instance when running the emulator the following exception is thrown:
```
Caused by: java.io.IOException: The Application Default Credentials are not available.
They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information
```

This should allow somebody to experiment on top of the emulator without having to worry about credentials.


rgruener:
Hmm, definitely used my @spotify address for the commit. It seems to not find an sla for spotify either.

mbrukman:
@rgruener — sorry about that, looks like you've done everything correctly and it's a glitch on our end; I'll look into it.

garye:
Thanks, I don't think anyone had ever tried this without default credentials set up in some way. Will merge when CLA issue is sorted out.

rgruener:
I signed it!

PR #993, opened 2016-09-01
Update all files to have 2016 copyright notices.

sduskis:
Les, can you please confirm that this is the right thing to do?


lesv:
I agree w/ Misha.  PTAL

sduskis:
The policy should be to copyright as of the file creation.  Noted.  Closing the PR without merging.

PR #1000, opened 2016-09-07
Use offline links for Javadoc

hegemonic:
This PR configures both of the Javadoc builds (the aggregate build and the Dataflow build) to use offline links to external packages. It also sets an appropriate window title for the aggregate build.


rameshdharan:
LGTM! Thanks.

Issue #1001, opened 2016-09-07
Cannot use ColumnRangeFilter with null bounds


Naveg:
Using a minColumn or maxColumn of null causes a NullPointerException [here](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/6713653f90bf2a2b981f07349ae2d1d89541dfb4/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/ColumnRangeFilterAdapter.java#L50)

See [here](https://hbase.apache.org/apidocs/org/apache/hadoop/hbase/filter/ColumnRangeFilter.html) for a description of the intended use of the filter.

sduskis:
Thanks for the catch.  This is a quick fix.  I'm working on testing the fix now, and hopefully we'll have a cloud-bigtable-client 0.9.3-SNAPSHOT with this change sometime tomorrow.

PR #1002, opened 2016-09-07
Fix column range filter

sduskis:
This fixes issue #1001.

I added defensive null checks, unit tests and integration tests.


rameshdharan:
LGTM! Glad to see the thorough new tests.

PR #1003, opened 2016-09-08
Spell out Google Compute Engine, and update some links

hegemonic:
This PR makes the following changes to the Javadoc comments:

+ Spells out "Google Compute Engine" rather than using the unapproved acronym "GCE."
+ Refers to Compute Engine "instances" rather than "VMs."
+ Updates outdated links.
+ Fixes typos.


sduskis:
LGTM

PR #1006, opened 2016-09-11
Remove unused log parameter.

gamorris:


sduskis:
The parameter is used... It's an exception; we log the stackTrace() of that exception.

gamorris:
Ugh. Of course. I'm an idiot... :-) totally misread that...

sduskis:
I saw where you were coming from.  The use isn't immediately obvious.  Usually, one does not print a stack trace on info logging.

PR #1017, opened 2016-09-20
Remove cancel

sduskis:
We call cancel gratuitously.  If we already get a response status, it means the client call is complete.  There is no need to call cancel again.  It's unclear if this is causing any problems, although there's a possibility that there's a performance hit for doing another round trip.


sduskis:
Travis is slow today.  Merging.

PR #1018, opened 2016-09-20
Adding v1 proto package.

sduskis:
The next step is to add a v1/v2 compatibility project for Dataflow purposes.


bigtable-client-core-parent/bigtable-v1-protos/README.md
```diff
@@ -0,0 +1,76 @@
+# Bigtable Protos
```
rameshdharan: Perhaps bigtable-legacy-protos might be a better name for this? Or alternatively we could rename bigtable-protos to bigtable-v2-protos? I think either would be slightly less confusing for new people in the codebase.

bigtable-client-core-parent/bigtable-v1-protos/README.md
```diff
@@ -0,0 +1,76 @@
...
+find . -name '*.java' | grep -v bigtable | xargs rm
+```
+
+# Generating the Python protos
+
+The gRPC Python generation process requires protoc 3
+
+Protoc can be installed via the instructions at:
+https://github.com/grpc/grpc-java#build-protobuf.
+Installing it isn't always trivial;
```
rameshdharan: This comment makes me sad. Not a huge deal, but I kind of feel like we should just take it out, if we aren't actually going to explain the issues in detail or link to anything else. Seems like the advice should be contained in the upstream #build-protobuf link?

sduskis: The comment is out of date.  I'll fix the v1/legacy README for this commit.  I'll update the core proto doc in a separate commit.

bigtable-client-core-parent/bigtable-v1-protos/README.md
sduskis: I think that bigtable-legacy-protos is better.  We have users who directly use the bigtable-protos maven artifacts, so I'd rather not change the name.

sduskis:
Can you please re-review the README.md?  I changed it quite a bit.

bigtable-client-core-parent/bigtable-legacy-protos/README.md
```diff
@@ -0,0 +1,41 @@
+# Bigtable Protos
+
+## Purpose
+This project houses the raw .proto files that are used by bigtable-legacy-compatibility, which is primarily geared towards v1 protobuff objects used by Dataflow.  This project generates a jar with java artifacts derived from the proto files in this project.  These proto files can also be used to generate other clients.
```
rameshdharan: s/protobuff/protobuf

rameshdharan: You might also want to use .proto everywhere instead of proto, but I don't feel strongly about it.

rameshdharan: Also maybe change Dataflow to Google Cloud Dataflow?

bigtable-client-core-parent/bigtable-legacy-protos/README.md
```diff
@@ -0,0 +1,41 @@
...
+This project relies on artifacts from https://github.com/grpc/grpc-java. You can see more about gRPC at https://github.com/grpc/grpc-common.
+
+## Bigtable proto files
+
+Primarily, Dataflow uses Mutation, Row and RowFilter as the top level objects that users would manipulate.  There are some objects that are used by those top level objects, as well as some objects that are not used by Dataflow, but are still included for the simplicity of managing this project.
+
+Here are the files directly associated with Bigtable functionality 
+* src/main/proto/google/bigtable/v1/bigtable_data.proto
+
+Copied Protos
```
rameshdharan: Change this to ### Copied Protos ### instead of the dash-underlining?

bigtable-client-core-parent/bigtable-legacy-protos/README.md
```diff
@@ -0,0 +1,41 @@
...
+--------------
+
+There were some proto files copied into this project from other projects used to have a full set of dependencies while generating Java artifacts from the bigtable v1 protos.  The generated artifacts are duplicate of objects in the bigtable-proto project, so they are removed after the are generated.
```
rameshdharan: Replace "used to have" with "in order to have"?

"After the are generated" -> "After they are generated"

sduskis:
I addressed the README comments.  PTAL.

bigtable-client-core-parent/bigtable-legacy-protos/README.md
sduskis: done

bigtable-client-core-parent/bigtable-legacy-protos/README.md
sduskis: done

sduskis:
Good catches. I updated the README.  I'm not sure why the UI isn't showing them.

sduskis:
I'm going to merge this.  If we need to change docs, we can do that a bit later.

PR #1021, opened 2016-09-22
Fixing some read/scan related issues.

sduskis:
- Removing close from the getNext() method.
- Adding some synchronization around closing a scan.


sduskis:
I spoke to Gary offline.  This shouldn't have any issues.  We'll revisit this PR if necessary.

PR #1022, opened 2016-09-22
Adding a bigtable-legacy-compatibility project

sduskis:
This will be used by Google Cloud Dataflow to convert between v1 and v2 protos.


sduskis:
@garye: I was thinking of either throwing an UnsupportedException right now, or leaving this as is; and do another PR another time.  Alternatively just do more testing in this PR.  I can go either way.  The second approach will just take a bit longer before there's something in the head branch.

garye:
Ok, LGTM. Might be best to throw UnsupportedException if we aren't sure it works but we can circle back on that if you want.

sduskis:
I added more tests.  PTAL.

sduskis:
The Travis build failed, but it looks like it happened after the maven build succeeded.  Something strange with Travis post processing.

Issue #1019, opened 2016-09-20
AbstractBigtableAdmin doesn't throw expected exceptions


heldtogether:
Calling `createTable()` on any instance of `org.apache.hadoop.hbase.client.Admin` should throw various exceptions depending on why the table can't be created.

According to the docs in the interface:

```
@throws IllegalArgumentException if the table name is reserved
@throws org.apache.hadoop.hbase.MasterNotRunningException if master is not running
@throws org.apache.hadoop.hbase.TableExistsException if table already exists (If concurrent
    threads, the table may have been created between test-for-existence and attempt-at-creation).
@throws IOException
```

**Expected behavior:**
If table exists, throw `TableExistsException`.

**Observed behavior:**
If table exists, `IOException` is thrown.

Here is an example of the issue:

```java
    try {
        connection = BigtableConfiguration.connect("projectId", "instanceId");
        Admin admin = connection.getAdmin();

        HTableDescriptor descriptor = new HTableDescriptor(TableName.valueOf("tableName"));
        descriptor.addFamily(new HColumnDescriptor("cf"));
        
        admin.createTable(descriptor);
    } catch (TableExistsException e) {
        // This Exception is never thrown!
    } catch (IOException e) {
        // This Exception is always thrown!
    }
```

I can see the `IOException` is being thrown at https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java#L372 but can't trace it back to the original exception so can't submit a pull request.

sduskis:
The Cloud Bigtable API uses gRPC infrastructure to send errors.  We currently don't have an easy way to send information about the specifics of the issue to ensure the expected behavior.

In this case, we'd need to send back a well defined HTTP error code, as well as protobuf describing the specifics of the problem (error_code = "FOUND", object_type="table", name = "table_name") or (error_code = "NOT_FOUND", object_type="instanceId", name = "instanceId") and etc.

We do not have a mechanism to send that type of information through our infrastructure at the beginning of the Cloud Bigtable project, so we decided to hold off from doing anything fancy and provide an error reason String.  The String can change over time.

In the long term, we need to fix this, but we don't have a timeline on this.

heldtogether:
Can you suggest the best way of handling the exceptions, given the error reason string can change over time?

[...]

PR #1025, opened 2016-09-27
Throw `TableExistsException` 

heldtogether:
The `TableExistsException` would be really useful when calling `createTable()`. See issue #1019 for a use case. It looks like we can use the `StatusRuntimeException` to distinguish between the table already exists and all other errors.

### DIFF
--- a/bigtable-hbase-parent/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestCreateTable.java
+++ b/bigtable-hbase-parent/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestCreateTable.java
@@ -351,7 +351,6 @@ public class TestCreateTable extends AbstractTest {
   }
 
   @Test
-  @Category(KnownGap.class)
   public void testAlreadyExists() throws IOException {
     thrown.expect(TableExistsException.class);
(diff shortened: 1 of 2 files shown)


heldtogether:
I signed it!

sduskis:
Thanks for the contribution!

I pushed a change that tests for this condition.  Would you be able to rebase, and remove the `@Category(KnownGap.class)` annotation from the new test in TestCreateTable, then check the fix with:
```
mvn clean integration-test \
   -PbigtableIntegrationTest \
   -Dgoogle.bigtable.project.id=[your cloud project id] \
   -Dgoogle.bigtable.instance.id=[your cloud bigtable instance id]
```

Greatly appreciated!

heldtogether:
I think I'm on the right lines but the exception being caught is `com.google.bigtable.repackaged.io.grpc.StatusRuntimeException` rather than `io.grpc.StatusRuntimeException`. 

I'm not familiar with the way the plugin shades packages. If I try importing `com.google.bigtable.repackaged.io.grpc.StatusRuntimeException` it doesn't compile. Can you point me in the right direction to use this repackaged exception?

sduskis:
Here's the secret sauce:

>       if (Status.fromThrowable(throwable).getCode() == Status.Code.ALREADY_EXISTS) {
>         throw new TableExistsException(desc.getTableName());
>       }

PR #1028, opened 2016-09-28
Readding zero copy which significantly improves reads.

sduskis:


bigtable-hbase-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/RowMergerPerf.java
```diff
@@ -21,26 +23,37 @@
 public class RowMergerPerf {
 
   public static void main(String[] args) {
-    for (int i = 0; i < 10; i++) {
-      rowMergerPerf(createResponses());
+    for(int i =0; i<5; i++) {
```
rameshdharan: Minor: tweak the spacing here?

```i = 0; i < 5; i++```

File: bigtable-hbase-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/RowMergerPerf.java [hunk 2]
```diff
@@ -21,26 +23,37 @@
...
-        .setValue(ByteString.copyFrom(RandomStringUtils.randomAlphanumeric(10000).getBytes()))
-        .setTimestampMicros(0L)
-        .setCommitRow(true)
-        .build();
+    int size = 100_000 / cellCount;
+    for (int i = 0; i < cellCount; i++) {
+      CellChunk contentChunk =
+          CellChunk.newBuilder()
+              .setFamilyName(StringValue.newBuilder().setValue("Family" + i))
+              .setQualifier(BytesValue.newBuilder().setValue(ByteString.copyFromUtf8("Qaulifier")))
```
rameshdharan: s/Qualifier/Qualifier

bigtable-hbase-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/RowMergerPerf.java
```diff
@@ -21,26 +23,37 @@
...
 
-    CellChunk contentChunk = CellChunk.newBuilder()
-        .setFamilyName(StringValue.newBuilder().setValue("Family1"))
-        .setQualifier(BytesValue.newBuilder().setValue(ByteString.copyFromUtf8("Qaulifier")))
-        .setRowKey(ByteString.copyFrom(Bytes.toBytes("rowkey-0")))
-        .setValue(ByteString.copyFrom(RandomStringUtils.randomAlphanumeric(10000).getBytes()))
-        .setTimestampMicros(0L)
-        .setCommitRow(true)
-        .build();
+    int size = 100_000 / cellCount;
```
rameshdharan: *Maybe* make sure that cellCount > 0 and 100_000 % cellCount == 0?

bigtable-dataflow-parent/bigtable-dataflow-import/src/test/java/com/google/protobuf/HBaseZeroCopyByteString.java
```diff
@@ -28,6 +28,28 @@
  * @since 0.96.1
  */
 public final class HBaseZeroCopyByteString  {
+
```
rameshdharan: How is this file used? I can't actually find any references to HBaseZeroCopyByteString anywhere...

bigtable-hbase-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/RowMergerPerf.java
sduskis: done

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/protobuf/BigtableZeroCopyByteStringUtil.java
```diff
@@ -12,6 +12,29 @@
  * @version $Id: $Id
  */
 public final class BigtableZeroCopyByteStringUtil {
+
```
rameshdharan: So, AFAICT this code is doing the same thing:

https://github.com/tsuna/asynchbase/commit/b55d9092efa7b13de7022cd02aa2a465971056ca

In fact, this is sort of how HBaseBigtableZeroCopyUtil used to work. Can you explain why that approach doesn't work here? Presumably it breaks in protobuf 3.0 but it wasn't immediately obvious to me how/why?

bigtable-dataflow-parent/bigtable-dataflow-import/src/test/java/com/google/protobuf/HBaseZeroCopyByteString.java
sduskis: This is a bit of a nuanced thing.  It's a class from the HBase API that doesn't work well with newer versions of protubf.  I can explain offline

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/protobuf/BigtableZeroCopyByteStringUtil.java
sduskis: It does indeed break in protubuf 3.0.0; LiteralByteString is now a private class, and this code doesn't work anymore.

FYI [here's](https://github.com/apache/hbase/blob/master/hbase-protocol/src/main/java/com/google/protobuf/HBaseZeroCopyByteString.java#L67) the grandaddy of both our copy and the asynchbase code.

bigtable-hbase-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/RowMergerPerf.java
sduskis: done

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/protobuf/BigtableZeroCopyByteStringUtil.java
rameshdharan: Ok, tracing this further looks like we can get rid of our .wrap() method here and use UnsafeByteOperations. 

But the ByteString class is still missing a way to go the other direction (i.e. "unwrap") AFAICT.

bigtable-hbase-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/RowMergerPerf.java
sduskis: I added a Precondition for cellCount > 0.  I added a comment that 100_000 % cellCount !=0 is fine.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/protobuf/BigtableZeroCopyByteStringUtil.java
sduskis: UnsafeByteOperations.unsafeWrap(ByteBuffer.wrap(byte[])) would end up using BoundedByteString which isn't ideal.  LiteralByteString is better when you're using the whole byte[] rather than a segment of it.

PR #1030, opened 2016-09-29
Adding timeouts for gets and scans.

sduskis:


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
```diff
@@ -38,48 +38,93 @@
...
-  private final int timeoutMs;
+  private final int shortRpcTimeoutMs;
+  private final int longRpcTimeoutMs;
 
   /**
    * <p>Constructor for CallOptionsConfig.</p>
    *
    * @param useTimeout a boolean.
-   * @param timeoutMs a int.
+   * @param unaryRpcTimeoutMs a int.
```
garye: nit: "an int"

PR #1031, opened 2016-10-01
RFC: cleaner zero copy without reflection

rameshdharan:
Not to be merged as is, but FYI I spoke with the protobuf guys and they clarified how to do the zero-copy optimization without needing the janky reflection hack. 

I wasn't quite sure how to convert everything over to this new way; the shaded thing really confused me and so this diff is a mess, but the interesting bit is in ZeroCopyByteStringUtil.. Note the use of UnsafeOperations.unsafeWrap and UnsafeOperations.writeTo. Neither of these incur any additional copies, or introduce BoundedByteBuffer objects or anything like that. 

I verified with the RowMerger benchmark that the numbers look identical (attached); I do think this is a better approach than mucking with the visibility so assuming I clean it up I'd like to merge some form of it. 

[new-results.txt](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/files/504034/new-results.txt)

[old-results.txt](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/files/504035/old-results.txt)


sduskis:
Closing this in favor of #1035.

PR #1032, opened 2016-10-03
Introduce backwards compatibility options for v1

garye:
Zone and cluster can be specified instead of instance, in which case a lookup
will be done to find the corresponding instance.

Note that this is NOT recommended and every effort should be made
to migrate directly to instance-based APIs.


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterName.java
```diff
@@ -0,0 +1,29 @@
+package com.google.cloud.bigtable.grpc;
```
sduskis: Copyright please.

garye:
Added copyrights and modified BigtableClusterName. PTAL.

PR #1038, opened 2016-10-06
InstanceId and cluster/zone id configs can coexist.

sduskis:


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -307,11 +307,13 @@ private BigtableOptions resolveLegacyOptions(BigtableOptions options) throws IOE
...
           );
-      return options.toBuilder()
-          .setZoneId(null)
-          .setClusterId(null)
-          .setInstanceId(instanceId)
-          .build();
+      if (options.getInstanceId() != null) {
+        Preconditions.checkArgument(
+            options.getInstanceId().equals(instanceId),
+            "Supplied instanceId: '%s', zoneId: '%s' and clusterId: '%s'. They do not match."
```
garye: Missing the variables for replacement?

PR #1042, opened 2016-10-10
Some refactoring of BulkMutation.

sduskis:
- Keeping track of all active batches.  This will enable more elaborate retries in the future for hung batches.
- Making Batch a nested class instead of a static nested class.  Batch needs access to most of BulkMutation's objects anyway.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
@@ -87,9 +87,10 @@ public class BulkMutation {
   static class RequestManager {
     private final List<SettableFuture<MutateRowResponse>> futures = new ArrayList<>();
     private final MutateRowsRequest.Builder builder;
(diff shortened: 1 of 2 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
```diff
@@ -338,10 +328,10 @@ public synchronized void run() {
       ListenableFuture<List<MutateRowsResponse>> future = null;
       try {
         if (retryId == null) {
-          retryId = Long.valueOf(this.asyncExecutor.getRpcThrottler().registerRetry());
+          retryId = Long.valueOf(asyncExecutor.getRpcThrottler().registerRetry());
         }
-        MutateRowsRequest request = currentRequestManager.build();
-        future = asyncExecutor.mutateRowsAsync(request);
+        activeBatches.put(currentBatch.id, currentBatch);
```
garye: I don't think it's strictly necessary given the current code but I think it would be best to implement hashCode/equals for Batch since it's being used as a map key. Otherwise LGTM.

Issue #1027, opened 2016-09-28
Add a new readRows API for a SimpleRow


sduskis:
Currently, the Row object is hierarchical Row -> Family -> Column -> Cell.  The creation of that structure is more expensive than what we need for HBase Result objects.  We can improve scan and Get performance by creating a simple Row -> List(Tuple(Family, Column, Cell)).

sduskis:
This is done.  FlatRow is the new SImpleRow.

PR #1045, opened 2016-10-15
Adding a new SimpleRow class

tyagihas:
Adding new data structure, SimpleRow, for the issue [https://github.com/GoogleCloudPlatform/cloud-bigtable-client/issues/1027](url)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,22 @@
 [![Maven][maven-shield]][maven-link]
 [![Stack Overflow][stackoverflow-shield]][stackoverflow-link]
 
-Bigger than a data warehouse, fast enough for real-time access, and less expensive than running virtual machines. The world-renowned database that powers Google is now available to you worldwide.
+[Google Cloud Bigtable](https://cloud.google.com/bigtable/) is Google's NoSQL
+Big Data database service. It's the same database that powers many core Google
+services, including Search, Analytics, Maps, and Gmail.
 
-## Overview
+Bigtable is designed to handle massive workloads at consistent low latency and
+high throughput, so it's a great choice for both operational and analytical
+applications, including IoT, user analytics, and financial data analysis.
 
-[Google Cloud Bigtable](https://cloud.google.com/bigtable/) offers you a fast, fully managed, almost infinitely scalable NoSQL database service that's ideal for web, mobile, and IoT applications requiring terabytes to petabytes of data. Unlike comparable market offerings, Cloud Bigtable doesn't require you to sacrifice speed, scale, or cost efficiency when your applications grow. The Bigtable service that Cloud Bigtable relies upon has been battle-tested at Google for more than 10 years—it's the database driving major applications such as Search, Analytics, Maps and Gmail.
+Bigtable provisions and scales to hundreds of petabytes automatically, and can
+smoothly handle millions of operations per second. Changes to the deployment
+configuration are immediate, so there is no downtime during reconfiguration.
+
+Bigtable [integrates easily][integrations] with popular Big Data tools like
+Hadoop, as well as Google Cloud Platform products like Cloud Dataflow and
+Dataproc. Plus, Bigtable supports the open-source, industry-standard HBase API,
+which makes it easy for development teams to get started.
 
 ## Project setup, installation, and configuration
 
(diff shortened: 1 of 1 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+  private final ArrayDeque<SimpleColumn> columns;
+	
+  public SimpleRow() {
+    columns = new ArrayDeque<SimpleColumn>();
+  }
+	
+  public void addCell(String family, ByteString qualifier, Cell cell) {
+    SimpleColumn column = new SimpleColumn();
+    column.family = family.getBytes();
```
garye: I don't think you need to turn this into bytes and then back into a String. Can it be left a String and compared against previous families using .equals() instead of the byte array comparison?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
```diff
@@ -0,0 +1,68 @@
...
+
+import com.google.bigtable.v2.Cell;
+import com.google.bigtable.v2.Column;
+import com.google.bigtable.v2.Family;
+import com.google.bigtable.v2.Row;
+import com.google.protobuf.ByteString;
+
+import java.util.ArrayDeque;
+
+public class SimpleRow {
```
garye: A class-level comment would be helpful

tyagihas: Will add a class-level comment in the next commit.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
```diff
@@ -0,0 +1,68 @@
...
+  public class SimpleColumn {
+    public byte[] family;
```
tyagihas: I agree. I will make it as private and immutable for this time.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
tyagihas: I chose this approach because the first merge process was faster than using String. However, this is probably an edge case so I will change it to String.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
```diff
@@ -0,0 +1,68 @@
...
+      if (!compare(prevKey, column.family)) {
+        familyBuilder = rowBuilder.addFamiliesBuilder().setName(new String(column.family));
+      }
+      familyBuilder.addColumns(column.builder);
+      prevKey = column.family;
+    }
+    return rowBuilder;
+  }
+  
+  private boolean compare(byte[] arr1, byte[] arr2) {
```
tyagihas: Will remove this method.

sduskis:
At this point, all we should have is SimpleRow that is just a SimpleRow.  It should not be referenced yet from RowMerger.  There could be a separate converter between SimpleRow and Row, but that conversion should not be in SimpleRow.

SimpleRow should have an ImmutableList<SimpleColumn>, and probably a Builder that creates the list.  SimpleColumn should not have a Column.Builder, but rather a value, a timestamp and labels.

The performance improvement will come later, when we can use SimpleRow as a first class citizen instead of a Row.  Creating a Row costs 500 nanos per column, due to the ordering of RowMerger.FamilyBuilderManager.  SimpleRow will not be ordered, hence the performance boost we need.  The server returns sorted values anyway, and the HBase adapter does another round of ordering.  The ordering we do for Row is superflous most of the time, so the goal is to remove it.

PR #1049, opened 2016-10-21
Moving user agent calculation into bigtable-client-core.

sduskis:
We always want to include cloud bigtable client version and jdk version in the user agent.  The BigtableOptions user agent is an additional piece of information rather than a substitute, so treat it that way.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVersionInfo.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVersionInfo.java
@@ -14,7 +14,7 @@ public class BigtableVersionInfo {
   public static final String CORE_UESR_AGENT = "bigtable-" + CLIENT_VERSION +",jdk-" + JDK_VERSION;
 
   /**
-   * Gets user agent from bigtable-hbase.properties. Returns a default dev user agent with current
+   * Gets user agent from bigtable-version.properties. Returns a default dev user agent with current
    * timestamp if not found.
    */
   private static String getVersion() {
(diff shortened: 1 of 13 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVersionInfo.java
```diff
@@ -0,0 +1,53 @@
...
+
+  private static final Logger LOG = new Logger(BigtableVersionInfo.class);
+
+  public static final String CLIENT_VERSION = getVersion();
+  public static final String JDK_VERSION = getJavaVersion();
+
+  public static final String CORE_UESR_AGENT = "bigtable-" + CLIENT_VERSION +",jdk-" + JDK_VERSION;
+
+  /**
+   * Gets user agent from bigtable-hbase.properties. Returns a default dev user agent with current
```
rameshdharan: s/bigtable-hbase.properties/bigtable-version.properties

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVersionInfo.java
```diff
@@ -0,0 +1,53 @@
...
+   * timestamp if not found.
+   */
+  private static String getVersion() {
+    final String defaultVersion = "dev-" + System.currentTimeMillis();
```
rameshdharan: We can do it in a follow-up change but I'd strongly recommend using the ident gitattributes facility so we always embed the git commit hash in the user agent: 

https://git-scm.com/docs/gitattributes

That should be fairly straightforward to do.

PR #1050, opened 2016-10-21
Using a pooled connection for CloudBigtableIO reading.

sduskis:
This should improve retries for reading, by virture of reading on a different connection.


bigtable-dataflow-parent/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -844,7 +845,6 @@ protected void setScanner(ResultScanner<Row> scanner) {
     @Override
     public void close() throws IOException {
       scanner.close();
-      session.close();
```
rameshdharan: I assume these just get cleaned up when the whole pool is shut down / AbstractBigtableConnection is cleaned up?

PR #1051, opened 2016-10-21
Resetting the httpTransport and credentials in the interceptor.

sduskis:
If there's an IOException, it could very well be due to the http transport and/or the cached credentials.  In that case, get a new transport and new credentials before getting a new OAuth2 token.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
@@ -19,6 +19,7 @@ import io.grpc.Metadata;
 import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
 import java.io.IOException;
+import java.security.GeneralSecurityException;
 import java.util.Date;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -32,6 +33,8 @@ import com.google.api.client.util.Preconditions;
 import com.google.api.client.util.Sleeper;
 import com.google.auth.oauth2.AccessToken;
 import com.google.auth.oauth2.OAuth2Credentials;
+import com.google.cloud.bigtable.config.CredentialFactory;
+import com.google.cloud.bigtable.config.CredentialOptions;
 import com.google.cloud.bigtable.config.Logger;
 import com.google.cloud.bigtable.config.RetryOptions;
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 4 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -276,21 +284,23 @@ private HeaderCacheElement getCachedHeader() throws IOException {
...
-      return false;
-    }
-    HeaderCacheElement cacheElement = refreshCredentialsWithRetry();
-    synchronized (isRefreshing) {
-      headerCache.set(cacheElement);
-      isRefreshing.set(false);
-      isRefreshing.notifyAll();
+    try {
+      HeaderCacheElement cacheElement = refreshCredentialsWithRetry();
+      synchronized (isRefreshing) {
```
rameshdharan: Does this actually need to be inside the synchronized block? it looks like headerCache.set is atomic anyway...?

sduskis: The synchronization is there to attempt to stop the thundering herd problem.  We don't need multiple threads creating parallel calls to get an oauth token.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -381,4 +393,17 @@ protected RetryState getRetryState(BackOff backoff) throws IOException {
...
   }
+
+  /**
+   * Call {@link CredentialFactory#clearHttpTransport()} and retrieve a new {@link OAuth2Credentials}.
+   */
+  @VisibleForTesting
+  void refreshCredentials() {
+    CredentialFactory.clearHttpTransport();
+    try {
+      this.credentials = (OAuth2Credentials) CredentialFactory.getCredentials(credentialOptions);
```
sduskis: We made a check at an earlier point, but it definitely doesn't hurt to do it again here.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
rameshdharan: Sorry, yeah I understand the intent, what I mean is that I don't think the headerCache call specifically needs to be inside the block. You've already done the refreshCredentialsWithRetry outside the lock so this is just populating an atomic reference. 

A different way of asking this is actually, shouldn't refreshCredentialsWithRetry be *inside* the synchronized block? Otherwise you're still kicking off a bunch of requests?

sduskis: there's a return statement inside the synchronized block.

rameshdharan: Ah, ok I see.

So the point only one of these queued requests (coming through doRefresh via asyncRefresh) will actually make it through the first synchronized block and *actually* issue the request. 

But so moving back a step, in the case where the header is stale, we're still likely creating/enqueueing a big burst of doRefresh calls via asyncRefresh(), most of which are going to just run through and not actually do anything. 

Why not just change the implementation of asyncRefresh so that it runs the first part (the check for whether a refresh should be enqueued) synchronously, and then just use the executor to enqueue the actual real I/O work i.e. refreshCredentialsWithRetry(). 

The simple and benignly racy way is to just use isRefreshing as a hint, i.e. just check if isRefreshing.get() is true before the call to executor.execute. That would cut down on most of the executor calls. Another way to accomplish this would be to introduce another CacheState -- `Refreshing` -- and skip multiple calls to asyncRefresh() that way.

PR #1052, opened 2016-10-21
Adding SimpleRow class

tyagihas:


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
```diff
@@ -0,0 +1,79 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.protobuf.ByteString;
+
+/**
+ * <p>This class stores values from {@link com.google.bigtable.v2.ReadRowsResponse} objects and represents a single row.</p>
+ *
+ * @author tyagihas
+ * @version $Id: $Id
+ */
+public class SimpleRow {
```
sduskis: I think that SimpleRow ought to be immutable, since it is intended to be used as the result of a Get or Scan.  I was thinking that we could have a SimpleRow.Builder similar to the way we do [BigtableConfiguration](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConfiguration.java).

What are your thoughts on this?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
```diff
@@ -0,0 +1,79 @@
...
+  public final class SimpleColumn {
+	private final String family;
+	private final ByteString qualifier;
+	private final long timestamp;
+	private final ByteString value;
+	private final List<String> labels;
+
+    private SimpleColumn(String family, ByteString qualifier, long timestamp, ByteString value, List<String> labels) {
```
sduskis: I like that this is an immutable class.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
```diff
@@ -0,0 +1,79 @@
...
+    public ByteString getValue() {
+      return value;
+    }
+
+    public List<String> getLabels() {
+      return labels;
+    }
+  }
+
+  private final ImmutableList.Builder<SimpleColumn> builder;
```
sduskis: This should probably be an ImmutableList here.  We shouldn't have a builder in the SimpleRow class.  The ImmutableList.Builder could be in a SimpleRow.Builder, but probably not in SimpleRow.  The list of SimpleColumn could be used multiple times.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/SimpleRowConverter.java
```diff
@@ -0,0 +1,53 @@
...
+import com.google.common.collect.ImmutableList;
+
+/**
+ * <p>This class converts an instance of {@link com.google.cloud.bigtable.grpc.SimpleRow} to {@link com.google.bigtable.v2.Row}.</p>
+ *
+ * @author tyagihas
+ * @version $Id: $Id
+ */
+public class SimpleRowConverter {	
+  public Row.Builder buildRow(SimpleRow row, Row.Builder rowBuilder) {
```
sduskis: Is there a reason we need Row.Builder here?  Could we have something like:

`public RowbuildRow(SimpleRow simpleRow)`

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/SimpleRowConverter.java
```diff
@@ -0,0 +1,53 @@
...
+    String prevKey = "";
+    Family.Builder familyBuilder = null;
+    ImmutableList<SimpleColumn> list = row.getList();
```
sduskis: To have a Row, we'd probably need a sorted list here.  We'd need to sort the SimpleRows by family and column.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
tyagihas: It sounds good to me. I'll reflect it to the next commit.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/SimpleRowConverter.java
tyagihas: Agree. Will remove the second parameter, "rowBuilder" and create a new instance within the method.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
tyagihas: I assume that the situation where the list is used multiple times is within a single ReadRowsResponse. Is that correct?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/SimpleRowConverter.java
tyagihas: Agree. Will add ordering logic.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
```diff
@@ -0,0 +1,117 @@
...
+    private Builder() {
+      listBuilder = new ImmutableList.Builder<SimpleColumn>();
+    }
+
+    public SimpleRow createRow() {
+      return new SimpleRow();
+    }
+  }
+
+  private ImmutableList<SimpleColumn> columns = null;
```
sduskis: There should also be a row key here, I think.

tyagihas: Agree. Just added it to the latest commit.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
```diff
@@ -0,0 +1,124 @@
...
+
+    public static Builder getDefaultInstance() {
+      return DEFAULT_INSTANCE;
+    }
+
+    private Builder() {
+      listBuilder = new ImmutableList.Builder<SimpleColumn>();
+    }
+
+    public SimpleRow createRow(ByteString rowKey) {
```
sduskis: In our other classes, we have separate setters for all variables.  It would be nice to follow that same style here.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
```diff
@@ -0,0 +1,124 @@
...
+      int result = left.family.compareTo(right.family);
+      if (result == 0) {
+        result = left.qualifier.toStringUtf8().compareTo(right.qualifier.toStringUtf8());
+      }
+      return result;
+    }
+  }
+
+  public static final class Builder {
+	private static Builder DEFAULT_INSTANCE = new Builder();
```
sduskis: A static variable seems like it could have many bad side effects.  We should either have a public Builder constructor or a newBuilder() method.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
```diff
@@ -0,0 +1,124 @@
...
+	private final ImmutableList.Builder<SimpleColumn> listBuilder;
+
+    public static Builder getDefaultInstance() {
+      return DEFAULT_INSTANCE;
+    }
+
+    private Builder() {
+      listBuilder = new ImmutableList.Builder<SimpleColumn>();
```
sduskis: How can we populate this listBuilder?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    public SimpleRow createRow(ByteString rowKey) {
+      return new SimpleRow(rowKey);
+    }
+  }
+
+  private final ByteString rowKey;
+  private ImmutableList<SimpleColumn> columns = null;
```
sduskis: Why isn't this also final and not null like rowKey?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
```diff
@@ -0,0 +1,124 @@
...
+
+  private SimpleRow(ByteString rowKey) {
+	  this.rowKey = rowKey;
+  }
+  
+  public ByteString getRowKey() {
+  	return rowKey;
+  }
+	
+  public void addCell(String family, ByteString qualifier, long timestamp, ByteString value,
```
sduskis: This should be in Builder :)

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
```diff
@@ -0,0 +1,124 @@
...
+      List<String> labels) {
+    Builder.getDefaultInstance().listBuilder.add(
+      new SimpleColumn(family, qualifier, timestamp, value, labels));
+  }
+  
+  public ImmutableList<SimpleColumn> getList() {
```
sduskis: This should just be a simple method that returns the simpleColumn immutable List.  Also, there is a preference by Dataflow to use standard java objects like List and Set rather than guava classes.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
tyagihas: Yeah, it shouldn't be here.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
tyagihas: Move addCell to Builder class.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
tyagihas: Changed it to List interface.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/SimpleRow.java
tyagihas: Added a setter method and removed a rowKey parameter from this method to conform with the existing style.

sduskis:
It's looking good!  Thanks for putting up with the review.

PR #1056, opened 2016-10-25
RetryOptions: remove Guava from public API

dhalperi:
Guava makes breaking changes quite often, exposing it to users is considered harmful.


dhalperi:
NB: I made this patch in github editor, so not tested yet.

sduskis:
I tested this change.  I approve.  I'll merge.

dhalperi:
Thanks!

PR #1061, opened 2016-11-01
ManyThreadDriver - base code for a long running test.

garye:


bigtable-hbase-parent/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/ManyThreadDriver.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
+
+public class ManyThreadDriver {
+
+  private static void runTest(String projectId, String instanceId, final String tableName) throws Exception {
+    Configuration configuration = new Configuration();
+    configuration.set("hbase.client.connection.impl", "com.google.cloud.bigtable.hbase1_0.BigtableConnection");
```
sduskis: We should refactor the configuration (later?).  All of this stuff is is bigtable-test.xml, and new Configuration pulls this in by default.

bigtable-hbase-parent/bigtable-hbase-integration-tests/pom.xml
```diff
@@ -232,6 +232,16 @@ limitations under the License.
 
     <dependencies>
         <dependency>
+            <groupId>${project.groupId}</groupId>
```
sduskis: These shouldn't be here.  it would be ok to use the bigtableIntegrationTest profile for running ManyThreadDriver.  This would have the impact of having the bigtable dependencies when we run the integration test with -P hbaseLocalMiniClusterTest which may cause problems running the HBase tests.

bigtable-hbase-parent/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/ManyThreadDriver.java
```diff
@@ -0,0 +1,94 @@
+package com.google.cloud.bigtable.hbase;
```
garye: Added.

bigtable-hbase-parent/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/ManyThreadDriver.java
garye: Ok, to-be-refactored later.

garye:
PTAL

PR #1064, opened 2016-11-02
Expand channel pool in getStreamFuture

juruen:
It seems that expandPoolIfNecessary is only called from getUnaryFuture.
The result is that if only bulk mutations are used, the pool doesn't get
expanded and we end up with just one TCP connection which hurts performance.

I'm not sure if this is intended or not. But by changing this we get a great performance boost.


sduskis:
I'm going to have to change this expand pool thing to front load the creation of the pool.

PR #1066, opened 2016-11-02
Setting the number of connections in ChannelPool's constructor.

sduskis:
ensureChannelCount was flaky.  We should just create the whole pool to begin with.

This will have an impact on the way we do testing, but we should change our tests rather than have a complicated client.


sduskis:
The old method helped with our YCSB testing.  We should change the YCSB testing instead :)

PR #1071, opened 2016-11-06
Enhancing SimpleRow, SimpleRowConverter

sduskis:
- SimpleRow now has equals and toString for testing purposes
- SimpleRow is now Serializable
- SimpleRowConverter converters Row -> SimpleRow now.
- Added TestSimpleRowConverter


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/SimpleRow.java
```diff
@@ -0,0 +1,248 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.grpc.scanner;
+
+import java.io.Serializable;
+import java.util.Collections;
+import java.util.List;
+
+import com.google.appengine.repackaged.com.google.common.base.Objects;
```
rameshdharan: Hah, what's the story with this repackaging? Does grpc actually have a com.google.appengine dependency? and/or where is that coming in from?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/SimpleRow.java
```diff
@@ -0,0 +1,248 @@
...
+ * @author tyagihas
+ * @version $Id: $Id
+ */
+public class SimpleRow implements Serializable {
+
+  private static final long serialVersionUID = 1L;
+
+  public final static class SimpleCell implements Serializable {
+
+    private static final long serialVersionUID = 1L;
```
rameshdharan: Curious about this.. Do we need "real" serialization here/do we actually depend on the class being serialized/deserialized correctly (by Dataflow or something)?

Seems like unless we both need it to be serialized and we have some kind of testing for that, we could just rely on the default serialization and/or use @SuppressWarnings("serial"). 

Asking more for my own edification; I don't have particularly strong feelings about this.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/SimpleRow.java
sduskis: shoot.  I didn't mean for that to happen.  That happens with automated eclipse imports.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/SimpleRow.java
sduskis: You're correct about this class will eventually be used by Dataflow for serialization.

rameshdharan: Ok, makes sense. Only concern I have is whether we (or Dataflow) has a test to make sure that we do/don't maintain serialization compatibility for whatever boundary is crossed there, e.g. if we rework this class/change/remove/add fields or whatever. Maybe it doesn't matter because the versions will always match?

sduskis: The serialization is for sending data across machines that are guaranteed to always with the same verisions of files classpath.

PR #1075, opened 2016-11-07
Using FlatRow in RowMerger as a performance boost

sduskis:
It's more performant to use FlatRow than the current structure of RowMerger.  It's not quite 2X faster, but it's close.

I had to comment out the timestamp and label ordering, since the integration tests fail.  I think we'll need to change the timestamp ordering in our json file to be reverse timestamp ordering.


sduskis:
As per an offline discussion, the json file is the acceptance test file for ReadRowsResponse conversion.

PR #1077, opened 2016-11-07
Creating new methods that export FlatRow instead of Row.

sduskis:
There was quite a bit of fallout from this change.  If you need a smaller set of changes, let's figure out how to split this up.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
@@ -418,28 +418,28 @@ public class BigtableDataGrpcClient implements BigtableDataClient {
     // TODO(sduskis): Figure out a way to perform operation level metrics with the
     // AbstractBigtableResultScanner implementations.
     final ResultScanner<FlatRow> delegate = readFlatRows(request);
-    return new ResultScanner<Row>(){
-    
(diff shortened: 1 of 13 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -394,11 +417,47 @@ private Metadata createMetadata(String tableName) {
...
+    
+      @Override
+      public void close() throws IOException {
+        delegate.close();
+      }
+    
+      @Override
+      public Row[] next(int count) throws IOException {
+        FlatRow[] flatRows = delegate.next(count);
+        Row rows [] = new Row[flatRows.length];
```
garye: kill the space before the brackets

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResponseQueueReader.java
```diff
@@ -69,17 +68,17 @@ public ResponseQueueReader(int readPartialRowTimeoutMillis, int capacityCap) {
   }
 
   /**
-   * Get the next complete Row object from the response queue.
+   * Get the next complete FlatRow object from the response queue.
```
garye: Could add javadoc links for consistency with other comments.

sduskis:
I fixed the javadoc and spacing issues.  PTAL.

PR #1080, opened 2016-11-08
Update read rows acceptance test with reverse-ordered timestamps

garye:
This is consistent with actual server behavior.


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRow.java
```diff
@@ -164,20 +165,6 @@ public int compare(Cell left, Cell right) {
       if (result != 0) {
         return result;
       }
-//      // reverse timestamp ordering
```
sduskis: Why wasn't this logic reenabled now that the tests changed?

PR #1082, opened 2016-11-08
Adding FlatRowAdapter and a test.

sduskis:


bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/FlatRowAdapter.java
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   * Convert a {@link org.apache.hadoop.hbase.client.Result} to a {@link FlatRow}.
+   *
+   * @param result a {@link org.apache.hadoop.hbase.client.Result} object.
+   * @return a {@link FlatRow} object.
+   */
+  public FlatRow adaptToRow(Result result) {
+    FlatRow.Builder rowBuilder = FlatRow.newBuilder();
+
+    // Result.getRow() is derived from its cells.  If the cells are empty, the row will be null.
```
garye: I don't quite follow this. If there is no Row in the hbase result we don't set a row key in the FlatRow builder. That doesn't seem valid...

sduskis: Updated.  PTAL

PR #1097, opened 2016-11-15
Renaming MutationAdapter to HBaseMutationAdapter.

sduskis:
I'd like to introduce a new interface for Put/Delete/RowMutations called MutationAdapter. I'm proposing that we rename the current MutationAdapter to HBaseMutation adapter.


sduskis:
BigtableBufferedMutator transforms Put/Delete/RowMutations into a MutateRowRequest using PutAdapter/DeleteAdapter/RowMutationsAdapter and passes the MutateRowRequest to BulkMutation.  BulkMutation takes the MutateRowRequest and converts it to a MutateRowsRequest.Entry.  It's expensive (but I don't know how much).  We should do a single conversion in 
PutAdapter/DeleteAdapter/RowMutationsAdapter to MutateRowsRequest.Entry in addition to a MutateRowRequest.

The new adapter will be an abstract parent class for PutAdapter, DeleteAdapter and RowMutationsAdapter which will allow creation of both a MutateRowRequest and a MutateRowsRequest.Entry in the abstract parent class, and defer to the subclass for extraction of the rowKey, and conversion of Hbase Cells to Cloud Bigtable Mutations.

PR #1100, opened 2016-11-16
RowMerger performance improvements.

sduskis:
When I ran RowMerger with this change plus #1099, the average cell took roughly 30 nanoseconds when compared to the current 70.

At this performance level, creating fewer objects helps, and so does replacing a chunk iterator with a getChunkCount()/getChunk(i).

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RowMerger.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RowMerger.java
@@ -19,7 +19,6 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Objects;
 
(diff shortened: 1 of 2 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RowMerger.java
```diff
@@ -216,138 +213,99 @@ abstract void validateChunk(RowInProgress rowInProgess, ByteString previousKey,
...
-      return currentId.rowKey;
+    private ByteString getRowKey() {
+      return flatRowBuilder == null ? null : flatRowBuilder.getRowKey();
+    }
+
+    private boolean hasRowKey() {
+      return flatRowBuilder != null;
+    }
+
+    private void reset() {
```
garye: I've seen reset() methods get out of sync with internal state and cause hard-to-diagnose problems. Is it worth the tradeoff? If so then maybe a comment along with the member variables that mentions keeping reset() up to date.

sduskis:
I removed the reset method.  It didn't actually help performance in the RowMergerPerf case.

PR #1104, opened 2016-11-18
Minor improvement to ChannelPoll performance.

sduskis:
Replacing AtomicBooleans with boolean and synchronized methods improved performance by 100 nanoseconds.

The `wrap(final ClientCall.Listener<RespT> delegate ...)` was moved to make that happen.


sduskis:
This change doesn't work consistently across environments.  Closing.

PR #1112, opened 2016-11-25
Adding adapters for HBase Mutations -> MutateRowsRequest.Entry.

sduskis:
BigtableBufferedMutor and BatchExecutor should convert from a Put/Delete directly to a MutateRowsRequest.Entry rather than first converting to a MutateRowRequest and then to a MutateRowsRequest.Entry. This PR adds the ability to convert to a MutateRowsRequest.Entry. A future PR will use the new ability from BufferedMutator and BatchExecutor.

### DIFF
--- a/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/PutAdapter.java
+++ b/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/PutAdapter.java
@@ -17,12 +17,12 @@ package com.google.cloud.bigtable.hbase.adapters;
 
 import com.google.api.client.util.Clock;
 import com.google.bigtable.v2.MutateRowRequest;
+import com.google.bigtable.v2.MutateRowsRequest;
 import com.google.bigtable.v2.Mutation;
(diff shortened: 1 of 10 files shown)


bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/PutAdapter.java
```diff
@@ -31,16 +31,17 @@
...
+ * Adapt an HBase {@link Put} Operation into a Bigtable {@link MutateRowRequest.Builder} or
+ * {@link MutateRowsRequest.Entry}.
  * @author sduskis
  * @version $Id: $Id
  */
-public class PutAdapter implements OperationAdapter<Put, MutateRowRequest.Builder> {
+public class PutAdapter extends MutationAdapter<Put> {
   private final int maxKeyValueSize;
   private final boolean setClientTimestamp;
   public Clock clock = Clock.SYSTEM;
```
garye: Not from this change, but does this have to be public? I imagine it's used for testing but is package protected sufficient?

sduskis: I'll make it package protected.

PR #1116, opened 2016-11-30
Performance increase for scanning.

sduskis:
Results from Bigtable are already ordered, although not ideally for HBase.  The order we get:
- family
- some internally ordered mechanism for qualifier
- reverse timestamp within the qualifier

Currently, we don't assume any ordering when we convert from the results we get from Cloud Bigtable.  We currently dedup using a TreeMap orderd by family/qualifier/timstamp via an HBase Comparator.  That's an expensive operation, or at least more expensive than it could be.  With the the assumptions of the server returning the order above, we can dedup within a single qualifer by simply checking against the previously returned response to see if the key matches, and if so, omit the value.  We can sort qualifier/timestamp within a family rather than sorting all cells across all families.

These changes made a 2-3X difference in the performance of the FlatRowAdapter as measured by RowMergerPerf.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRow.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRow.java
@@ -29,7 +29,7 @@ import com.google.protobuf.ByteString;
 
 /**
  * <p>
- * This class stores represents a single row. It's a flattened version of the adata of aa
+ * This class stores represents a single row. It's a flattened version of the data of a
  * {@link Row}
  * </p>
  * @author tyagihas
@@ -128,16 +128,17 @@ public class FlatRow implements Serializable {
         return false;
       }
       Cell other = (Cell) obj;
-      return hasEqualKeys(other) &&
+      return equalFamilyQualifierAndTimestamp(other) &&
           Objects.equal(value, other.value) &&
           Objects.equal(labels, other.labels);
     }
 
     /**
      * @param other
-     * @return true if the family, qualifier and timestamps are the same
+     * @return true if the family, qualifier and timestamps are the same in this Cell as it is in
+     *         the other Cell.
      */
-    public boolean hasEqualKeys(Cell other) {
+    public boolean equalFamilyQualifierAndTimestamp(Cell other) {
       return other != null &&
           Objects.equal(family, other.family) &&
           Objects.equal(qualifier, other.qualifier) &&
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRow.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRow.java
@@ -128,11 +128,20 @@ public class FlatRow implements Serializable {
         return false;
       }
       Cell other = (Cell) obj;
-      return Objects.equal(family, other.family) &&
-          Objects.equal(qualifier, other.qualifier) &&
-          timestamp == other.timestamp &&
+      return hasEqualKeys(other) &&
           Objects.equal(value, other.value) &&
           Objects.equal(labels, other.labels);
+    }
+
+    /**
+     * @param other
+     * @return true if the family, qualifier and timestamps are the same
+     */
+    public boolean hasEqualKeys(Cell other) {
+      return other != null &&
+          Objects.equal(family, other.family) &&
+          Objects.equal(qualifier, other.qualifier) &&
+          timestamp == other.timestamp;
     }
 
     @Override
--- a/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/FlatRowAdapter.java
+++ b/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/FlatRowAdapter.java
@@ -15,17 +15,22 @@
  */
 package com.google.cloud.bigtable.hbase.adapters.read;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.TreeSet;
-
-import org.apache.hadoop.hbase.KeyValue;
-import org.apache.hadoop.hbase.client.Result;
-import org.apache.hadoop.hbase.util.Bytes;
+import com.google.bigtable.v2.ReadRowsRequest;
+import com.google.bigtable.v2.RowFilter.Interleave;
 import com.google.cloud.bigtable.grpc.scanner.FlatRow;
 import com.google.cloud.bigtable.hbase.BigtableConstants;
 import com.google.cloud.bigtable.hbase.adapters.ResponseAdapter;
 import com.google.cloud.bigtable.util.ByteStringer;
+import com.google.common.base.Objects;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import org.apache.hadoop.hbase.Cell;
+import org.apache.hadoop.hbase.client.Result;
+import org.apache.hadoop.hbase.filter.FilterList;
+import org.apache.hadoop.hbase.filter.WhileMatchFilter;
+import org.apache.hadoop.hbase.util.Bytes;
 
 /**
  * Adapt between a {@link FlatRow} and an hbase client {@link org.apache.hadoop.hbase.client.Result}.
@@ -39,6 +44,19 @@ public class FlatRowAdapter implements ResponseAdapter<FlatRow, Result> {
   static final long TIME_CONVERSION_UNIT = BigtableConstants.BIGTABLE_TIMEUNIT.convert(1,
     BigtableConstants.HBASE_TIMEUNIT);
 
+  static final Comparator<RowCell> QUALIFIER_COMPARATOR =
+      new Comparator<RowCell>() {
+        @Override
+        public int compare(RowCell left, RowCell right) {
+          int result = Bytes.compareTo(left.getQualifierArray(), right.getQualifierArray());
+          if (result != 0) {
+            return result;
+          }
+          result = Long.compare(right.getTimestamp(), left.getTimestamp());
+          return result;
+        }
+      };
+
   /**
    * {@inheritDoc}
    * 
@@ -49,8 +67,8 @@ public class FlatRowAdapter implements ResponseAdapter<FlatRow, Result> {
     if (flatRow == null) {
       return Result.EMPTY_RESULT;
     }
-
-    return convert(flatRow, new TreeSet<org.apache.hadoop.hbase.Cell>(KeyValue.COMPARATOR));
+    List<Cell> hbaseCells = addCells(flatRow, false);
+    return Result.create(hbaseCells.toArray(new Cell[hbaseCells.size()]));
   }
 
   /**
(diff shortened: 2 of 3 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRow.java
```diff
@@ -128,13 +128,22 @@ public boolean equals(Object obj) {
         return false;
       }
       Cell other = (Cell) obj;
-      return Objects.equal(family, other.family) &&
-          Objects.equal(qualifier, other.qualifier) &&
-          timestamp == other.timestamp &&
+      return hasEqualKeys(other) &&
```
kevinsi4508: Can you update the class doc to: 

This class stores represents a single row. It's a flattened version of the data of a
  * {@link Row}

Can you confirm that the change in this file is a no-op?

You introduced a concept of the key in this class. Could you mention this in the class doc or at least at the place where it is used.

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/FlatRowAdapter.java
```diff
@@ -60,20 +78,42 @@ public Result adaptResponsePresortedCells(FlatRow flatRow) {
...
-    return convert(flatRow, new ArrayList<org.apache.hadoop.hbase.Cell>());
+    List<Cell> hbaseCells = addCells(flatRow, false);
+    return Result.create(hbaseCells.toArray(new Cell[hbaseCells.size()]));
   }
 
-  private Result convert(FlatRow flatRow, Collection<org.apache.hadoop.hbase.Cell> hbaseCells) {
+  /**
+   * Converts all of the {@link FlatRow#getCells()} into HBase {@link RowCell}s.
+   * <p>
+   * All @link FlatRow.Cell}s with labels are discarded, since those are used only for
```
kevinsi4508: All {@link FlateRow.Cell}s

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRow.java
sduskis: I updated hasEqualKeys to equalFamilyQualifierAndTimestamp.

I updated the comment.

The logic new method was an IDE refactoring of the three tests and added `other != null` which will not change the functionality of the original code in `equals`.

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/FlatRowAdapter.java
```diff
@@ -39,6 +44,19 @@
   static final long TIME_CONVERSION_UNIT = BigtableConstants.BIGTABLE_TIMEUNIT.convert(1,
     BigtableConstants.HBASE_TIMEUNIT);
 
+  static final Comparator<RowCell> QUALIFIER_COMPARATOR =
```
sduskis: ack

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/FlatRowAdapter.java
```diff
@@ -60,20 +78,42 @@ public Result adaptResponsePresortedCells(FlatRow flatRow) {
...
-    for (FlatRow.Cell cell : flatRow.getCells()) {
+    List<FlatRow.Cell> cells = flatRow.getCells();
+    List<Cell> hbaseCells = new ArrayList<>(cells.size());
+    List<RowCell> familyCells = new ArrayList<>(cells.size());
+
+    for (FlatRow.Cell cell : cells) {
       // Cells with labels are for internal use, do not return them.
       // TODO(kevinsi4508): Filter out targeted {@link WhileMatchFilter} labels.
-      if (!cell.getLabels().isEmpty()) {
+      if (!cell.getLabels().isEmpty() || cell.hasEqualKeys(previousCell)) {
```
kevinsi4508: Add comment: Perform deduplication by skipping cells with matching key (family, qualifier, timestamp).

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/FlatRowAdapter.java
```diff
@@ -82,23 +122,38 @@ private Result convert(FlatRow flatRow, Collection<org.apache.hadoop.hbase.Cell>
...
-
-      hbaseCells.add(keyValue);
+      RowCell rowCell =
+          new RowCell(
+              rowKey,
+              family,
+              ByteStringer.extract(cell.getQualifier()),
+              hbaseTimestamp,
+              ByteStringer.extract(cell.getValue()));
+      if (!sort ) {
```
kevinsi4508: nit:  if (!sort)

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/FlatRowAdapter.java
```diff
@@ -82,23 +122,38 @@ private Result convert(FlatRow flatRow, Collection<org.apache.hadoop.hbase.Cell>
...
+        hbaseCells.add(rowCell);
+      } else {
+        if (previousFamily != null && family != previousFamily) {
```
kevinsi4508: !Arrays.equals(previousFamily, family)

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/FlatRowAdapter.java
```diff
@@ -82,23 +122,38 @@ private Result convert(FlatRow flatRow, Collection<org.apache.hadoop.hbase.Cell>
...
       previousCell = cell;
       previousFamily = family;
     }
+    addAll(hbaseCells, familyCells);
+    return hbaseCells;
+  }
 
-    return Result.create(hbaseCells.toArray(new org.apache.hadoop.hbase.Cell[hbaseCells.size()]));
+  void addAll(List<Cell> hbaseCells, List<RowCell> familyCells) {
+    if (!familyCells.isEmpty()) {
```
kevinsi4508: nit: remove "if (!familyCells.isEmpty())"

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/FlatRowAdapter.java
```diff
@@ -60,20 +78,42 @@ public Result adaptResponsePresortedCells(FlatRow flatRow) {
...
+   * more than once, if there are {@link Interleave}s in the {@link ReadRowsRequest#getFilter()},
+   * but the duplicates will appear one after the other. An {@link Interleave}s can originate from a
+   * {@link FilterList} with a {@link FilterList.Operator#MUST_PASS_ONE} operator or a
+   * {@link WhileMatchFilter}.
+   * <p>
+   * @param flatRow
+   * @param sort
+   * @return
+   */
+  private List<Cell> addCells(FlatRow flatRow, boolean sort) {
```
kevinsi4508: nit: convertCells()? addCells() sounds like adding cells to an input parameter (like the one we had with TreeSet).

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/FlatRowAdapter.java
sduskis: I'm actually going to add `familyCells != null && !familyCells.isEmpty())`.  Wouldn't that have a performance benefit for the non-sorted case?  I'm not going to create familyCells for the non-sorted case to begin with...

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/FlatRowAdapter.java
```diff
@@ -82,23 +122,38 @@ private Result convert(FlatRow flatRow, Collection<org.apache.hadoop.hbase.Cell>
...
+      }
       previousCell = cell;
       previousFamily = family;
     }
+    addAll(hbaseCells, familyCells);
+    return hbaseCells;
+  }
 
-    return Result.create(hbaseCells.toArray(new org.apache.hadoop.hbase.Cell[hbaseCells.size()]));
+  void addAll(List<Cell> hbaseCells, List<RowCell> familyCells) {
```
kevinsi4508: nit: Rearrange parameter ordering?  void addAll(List<RowCell> familyCells, List<Cell> hbaseCells, )
Usually, we order the input as "from" and then "to". "Add cells from 'familyCells' to 'hbaseCells'".

bigtable-hbase-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/RowMergerPerf.java
```diff
@@ -100,7 +100,7 @@ private static void rowMergerPerf(ReadRowsResponse response, int cellCountPerRow
       // The adapter is slower than the RowMerger, so decrease the number of rows by a factor of 3
       // so that the test finishes faster. This will be enough of a sample to get an idea about the
       // adapter's performance
-      long rowCount = CUMULATIVE_CELL_COUNT / cellCountPerRow / 4;
+      long rowCount = CUMULATIVE_CELL_COUNT / cellCountPerRow;
```
kevinsi4508: Sorry! why removing " / 4" here?

sduskis: The new change is more performance sensitive to families with a larger number of qualifiers.  The sorting is of values within the qualifiers.  The more qualifiers, the more `(n * log n)` takes.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRow.java
kevinsi4508: There are some typos in the class doc: "This class stores represents a single row. It's a flattened version of the data of a".

sduskis:
The comments were addressed.  PTAL.

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/FlatRowAdapter.java
kevinsi4508: It is a nit. Not checking it is for readability. It is up to you. Performance difference is minimum as it is done once with empty list. familyCells is always initialized, no?

kevinsi4508:
LGTM

PR #1118, opened 2016-12-01
Fixing the Dataflow import proess.

sduskis:
There were a few things that needed to be fixed given Dataflow updates and performance for large imports.
- Shuffle the splits so that splits are more likely to hit different cloud bigtable nodes.
- Use a floor for the smallest meaningful split.  Dataflow passes in a value of 1 byte.  We should use a minimum of 100 MB or so.
- Implement a new Dataflow method for dynamic rebalancing (or notifying of the lack of dynamic rebalancing, in this case): getSplitPointsRemaining()

### DIFF
--- a/bigtable-dataflow-parent/bigtable-dataflow-import/src/main/java/com/google/cloud/bigtable/dataflowimport/HadoopFileSource.java
+++ b/bigtable-dataflow-parent/bigtable-dataflow-import/src/main/java/com/google/cloud/bigtable/dataflowimport/HadoopFileSource.java
@@ -107,7 +107,7 @@ public class HadoopFileSource<K, V> extends BoundedSource<KV<K, V>> {
   private static final long serialVersionUID = 0L;
   private static final Logger SOURCE_LOG = LoggerFactory.getLogger(HadoopFileSource.class);
 
-  private static final long ONE_HUNDRED_MB = 100 * (1 << 20);
+  private static final long MIN_BUNDLE_SIZE_BYTES = 100 * (1 << 20);
 
   // Work-around to suppress confusing warning and stack traces by gcs-connector.
   // See setIsRemoteFileFromLaunchSite() for more information. This variable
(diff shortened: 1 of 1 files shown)


bigtable-dataflow-parent/bigtable-dataflow-import/src/main/java/com/google/cloud/bigtable/dataflowimport/HadoopFileSource.java
```diff
@@ -367,11 +390,13 @@ public long getEstimatedSizeBytes(PipelineOptions options) {
       for (FileStatus st : listStatus(createFormat(job), job)) {
         size += st.getLen();
       }
+      return size;
     } catch (IOException | NoSuchMethodException | InvocationTargetException
         | IllegalAccessException | InstantiationException e) {
       // ignore, and return 0
+      SOURCE_LOG.warn("Got exception", e);
```
kevinsi4508: shouldn't this be error()?

bigtable-dataflow-parent/bigtable-dataflow-import/src/main/java/com/google/cloud/bigtable/dataflowimport/HadoopFileSource.java
```diff
@@ -278,7 +284,24 @@ public void validate() {
   public List<? extends BoundedSource<KV<K, V>>> splitIntoBundles(long desiredBundleSizeBytes,
       PipelineOptions options) throws Exception {
     if (serializableSplit == null) {
-      return Lists.transform(computeSplits(desiredBundleSizeBytes),
+      // Dataflow can send a really small number desiredBundleSizeBytes, as low as 1. 100 MB chunks
+      // seem to be a good floor.
+      desiredBundleSizeBytes = Math.max(desiredBundleSizeBytes, ONE_HUNDRED_MB);
```
kevinsi4508: nit: try not to mutate input parameter. Up to you though.

bigtable-dataflow-parent/bigtable-dataflow-import/src/main/java/com/google/cloud/bigtable/dataflowimport/HadoopFileSource.java
```diff
@@ -102,6 +105,9 @@
  */
 public class HadoopFileSource<K, V> extends BoundedSource<KV<K, V>> {
   private static final long serialVersionUID = 0L;
+  private static final Logger SOURCE_LOG = LoggerFactory.getLogger(HadoopFileSource.class);
+
+  private static final long ONE_HUNDRED_MB = 100 * (1 << 20);
```
kevinsi4508: Name it by how it is used, not the value is has. Perhaps MIN_BUNDLE_SIZE_BYTES? or MIN_BUNDLE.

kevinsi4508:
some nits. LGTM

PR #1121, opened 2016-12-02
Using a new endpoint for dataflow operations.

sduskis:
We want all of our users to use batch-bigtable.googleapis.com for their dataflow work.  That will isolate users with good latency requirements from batch users who need good throughput.

I tested reading and writing 1,000,000 rows through Dataflow with this change, both of which worked as expected.

### DIFF
--- a/bigtable-dataflow-parent/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
+++ b/bigtable-dataflow-parent/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
@@ -253,6 +253,9 @@ public class CloudBigtableConfiguration implements Serializable {
   public Configuration toHBaseConfig() {
     Configuration config = new Configuration(false);
 
+    // Dataflow should use a different endpoint for data operations than online traffic.
+    config.set(BigtableOptionsFactory.BIGTABLE_HOST_KEY, "batch-bigtable.googleapis.com");
+
     // This setting can potentially decrease performance for large scale writes. However, this
     // setting prevents problems that occur when streaming Sources, such as PubSub, are used.
     // To override this behavior, call:


bigtable-dataflow-parent/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
```diff
@@ -253,6 +253,9 @@ public BigtableOptions toBigtableOptions() throws IOException {
   public Configuration toHBaseConfig() {
     Configuration config = new Configuration(false);
 
+    // Dataflow should use a different endpoint for data operations than online traffic.
+    config.set(BigtableOptionsFactory.BIGTABLE_HOST_KEY, "batch-bigtable.googleapis.com");
```
kevinsi4508: Should this be configurable like "<google.bigtable.endpoint.host>bigtable.googleapis.com</google.bigtable.endpoint.host>"

kevinsi4508:
I am fine with hardcoding it. LGTM

bigtable-dataflow-parent/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java
sduskis: Dataflow works with programmatic configuration, so we have to write code rather than XML.

The point of this PR is to use a different endpoint for google.bigtable.endpoint.host.  batch-bigtable.googleapis.com will be for Dataflow and other batch processes, and bigtable.googleapis.com will be for user's online operations (i.e. traffic from live serving systems).

sduskis: FWIW, users can still override the default and user other endpoints.

PR #1123, opened 2016-12-06
FlatRowAdapter should sort families, not column qualifiers

sduskis:
FlatRow used to be completely ordered on the client side.  The server actually returns sorted rows, albeit with a different ordering that we need.  I mistakenly sorted by column qualifier before; we actually need to sort by family and not column qualifer.


bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/FlatRowAdapter.java
```diff
@@ -88,79 +78,87 @@ private Result adaptResponse(FlatRow flatRow, final boolean sort) {
...
     byte[] rowKey = ByteStringer.extract(flatRow.getRowKey());
 
     FlatRow.Cell previousCell = null;
     byte[] previousFamily = null;
 
-    List<FlatRow.Cell> cells = flatRow.getCells();
-    List<Cell> hbaseCells = new ArrayList<>(cells.size());
-    List<RowCell> familyCells = sort ? new ArrayList<RowCell>(cells.size()) : null;
+    Map<String, List<RowCell>> familyMap = new TreeMap<>();
+    List<RowCell> currentFamilyRowCells = new ArrayList<>(flatRow.getCells().size());
```
sduskis: Thanks for the review.

I tested just a `familyMap.get(familyString);` instead of a cached variable.  That adds about 20-30 nanoseconds per cell, or roughly a 50% increase in processing time per cell.  Given the volume of cells we need to process, even that relatively minor cost per cell can add up.  Due to performance and CPU overhead being super important, I personally feel that a readability issue, as long as it's relatively minor, is ok in order to achieve faster performance.

What were you hoping to achieve with the suggestion?

sduskis: Your point about readability is well taken.  There's too much going on in that method.  Let me see what I can do to improve readability while maintaining performance.

sduskis:
I moved the sorting logic into its own class so that it's easier to see the forest for the trees.  PTAL.

sduskis:
@kevinsi4508, I removed the method you commented on previously (FlatRowAdapter.adaptResponsePresortedCells).  PTAL

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/FlatRowAdapter.java
```diff
@@ -88,79 +54,151 @@ private Result adaptResponse(FlatRow flatRow, final boolean sort) {
...
-    List<RowCell> familyCells = sort ? new ArrayList<RowCell>(cells.size()) : null;
-
-    for (FlatRow.Cell cell : cells) {
-      // Cells with labels are for internal use, do not return them.
-      // TODO(kevinsi4508): Filter out targeted {@link WhileMatchFilter} labels.
-      //
-      // Perform deduplication by skipping cells with matching key (family, qualifier, timestamp).
-      if (!cell.getLabels().isEmpty() || cell.equalFamilyQualifierAndTimestamp(previousCell)) {
-        continue;
+  private static class CellSorter {
```
kevinsi4508: nit: @NotThreadSafe

kevinsi4508:
LGTM

PR #1124, opened 2016-12-06
BigtableClusterUtilities should have no-arg methods.

sduskis:
BigtableClusterUtilities allows users to programmatically control the size of their cluster.  It is generally used with a user configuring a project and an instance.  A user can also set '-' as the instance id for all projects, which is useful when we know the cluster id but not the instance id.

Currently, one project/instance combination which only have a single cluster and zone is likely to be our most common scenario.  As such, users should not have to specify clusterId and zoneId.  In those scenarios, the code should look like:

`BigtableClusterUtilities.forInstance(projectId, instanceId).setClusterSize(30);`

It currently looks like:

`BigtableClusterUtilities.forInstance(projectId, instanceId).setClusterSize(clusterId, zone, 30);`

Add new methods for setSize, incrementSize, getClusterSize that don't have clusterId and zone.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterName.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterName.java
@@ -27,11 +27,15 @@ import java.util.regex.Pattern;
 public class BigtableClusterName {
   // Use a very loose pattern so we don't validate more strictly than the server.
   private static final Pattern PATTERN =
-      Pattern.compile("projects/.*/instances/(.*)/clusters/(.*)");
+      Pattern.compile("projects/[^/]+/instances/([^/]+)/clusters/([^/]+)");
 
   private final String clusterName;
   private final String instanceId;
   private final String clusterId;
+
+  public static BigtableClusterName parse(String clusterName) {
+    return new BigtableClusterName(clusterName);
+  }
 
   public BigtableClusterName(String clusterName) {
     this.clusterName = clusterName;
@@ -41,6 +45,9 @@ public class BigtableClusterName {
     this.clusterId = matcher.group(2);
   }
 
+  /**
+   * Returns the fully qualified cluster name same thing as {@link #getClusterName()}.
+   */
   @Override
   public String toString() {
     return clusterName;
@@ -48,22 +55,31 @@ public class BigtableClusterName {
 
   /**
    * @return The id of the instance that contains this cluster. It's the second group in the Cluster
-   *         name: projects/(.*)/instances/(.*)/clusters/(.*)
+   *         name: "projects/(.+)/instances/(.+)/clusters/(.+)".
    */
   public String getInstanceId() {
     return instanceId;
   }
 
   /**
-   * @return The id of this cluster. It's the third group in the Cluster name:
-   *         projects/(.*)/instances/(.*)/clusters/(.*)
+   * @return The id of this cluster. It will look like the following
+   *         projects/(.+)/instances/(.+)/clusters/(.+).
    */
   public String getClusterName() {
+    return clusterName;
+  }
+
+  /**
+   * @return The id of this cluster. It's the third group in the Cluster name:
+   *         "projects/(.+)/instances/(.+)/clusters/(.+)".
+   */
+  public String getClusterId() {
     return clusterId;
   }
 
   /**
    * Create a fully qualified snapshot name based on the the clusterName and the snapshotId.
+   * Snapshot name will look like: "projects/(.+)/instances/(.+)/clusters/(.+)/snapshots/(.+)"
    * @param snapshotId The id of the snapshot
    * @return A fully qualified snapshot name that contains the fully qualified cluster name as the
    *         parent and the snapshot name as the child.
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterName.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterName.java
@@ -27,16 +27,18 @@ import java.util.regex.Pattern;
 public class BigtableClusterName {
   // Use a very loose pattern so we don't validate more strictly than the server.
   private static final Pattern PATTERN =
-      Pattern.compile("projects/.*/instances/(.*)/clusters/.*");
+      Pattern.compile("projects/.*/instances/(.*)/clusters/(.*)");
 
   private final String clusterName;
   private final String instanceId;
+  private final String clusterId;
 
   public BigtableClusterName(String clusterName) {
     this.clusterName = clusterName;
     Matcher matcher = PATTERN.matcher(clusterName);
     Preconditions.checkArgument(matcher.matches(), "Malformed cluster name");
     this.instanceId = matcher.group(1);
+    this.clusterId = matcher.group(2);
   }
 
   @Override
@@ -45,9 +47,27 @@ public class BigtableClusterName {
   }
 
   /**
-   * @return The id of the instance that contains this cluster.
+   * @return The id of the instance that contains this cluster. It's the second entry in the Cluster
+   *         name: projects/(.*)/instances/(.*)/clusters/(.*)
    */
   public String getInstanceId() {
     return instanceId;
+  }
+
+  /**
+   * @return The id of this cluster. It's the thir entry in the Cluster name:
+   *         projects/(.*)/instances/(.*)/clusters/(.*)
+   */
+  public String getClusterName() {
+    return clusterId;
+  }
+
+  /**
+   * @param snapshotId The id of the snapshot
+   * @return A fully qualified snapshot name that contains the fully qualified cluster name as the
+   *         parent and the snapshot name as the child.
+   */
+  public String toSnapshotName(String snapshotId) {
+    return clusterName + "/snapshots/" + snapshotId;
   }
 }
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterName.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterName.java
@@ -55,15 +55,15 @@ public class BigtableClusterName {
 
   /**
    * @return The id of the instance that contains this cluster. It's the second group in the Cluster
-   *         name: "projects/(.+)/instances/(.+)/clusters/(.+)".
+   *         name: "projects/{projectId}/instances/{instanceId}/clusters/{clusterId}".
    */
   public String getInstanceId() {
     return instanceId;
   }
 
   /**
-   * @return The id of this cluster. It will look like the following
-   *         projects/(.+)/instances/(.+)/clusters/(.+).
+   * @return The name of this cluster. It will look like the following
+   *         "projects/{projectId}/instances/{instanceId}/clusters/{clusterId}".
    */
   public String getClusterName() {
     return clusterName;
@@ -71,7 +71,7 @@ public class BigtableClusterName {
 
   /**
    * @return The id of this cluster. It's the third group in the Cluster name:
-   *         "projects/(.+)/instances/(.+)/clusters/(.+)".
+   *         "projects/{projectId}/instances/{instanceId}/clusters/{clusterId}".
    */
   public String getClusterId() {
     return clusterId;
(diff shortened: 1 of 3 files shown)


sduskis:
FYI, snapshot integration needs this change so that we can easily git (projectId, instanceId) -> (clusterId, locationId)

kevinsi4508:
Read through the description. I am not sure why we need 'zone' when updating a cluster.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterName.java
```diff
@@ -45,9 +47,28 @@ public String toString() {
   }
 
   /**
-   * @return The id of the instance that contains this cluster.
+   * @return The id of the instance that contains this cluster. It's the second group in the Cluster
```
kevinsi4508: This comment is not correct, unless you change the PATTERN to have 3 groups.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterName.java
```diff
@@ -45,9 +47,28 @@ public String toString() {
...
    */
   public String getInstanceId() {
     return instanceId;
   }
+
+  /**
+   * @return The id of this cluster. It's the third group in the Cluster name:
+   *         projects/(.*)/instances/(.*)/clusters/(.*)
+   */
+  public String getClusterName() {
```
kevinsi4508: getClusterId()

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterUtilities.java
```diff
@@ -73,6 +70,66 @@ public static BigtableClusterUtilities forAllInstances(String projectId)
...
+      } catch (Exception e) {
+        logger.warn("Error closing BigtableClusterUtilities: ", e);
+      }
+    }
+  }
+
+  /**
+   * @return The cluster associated with the given project and instance
+   */
+  public static Cluster lookupCluster(String projectId, String instanceId)
```
kevinsi4508: Should lookupSingleCluster throw exception if there are more than one cluster.
Otherwise, this method doesn't make sense.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterUtilities.java
```diff
@@ -97,17 +154,26 @@ public BigtableClusterUtilities(final BigtableOptions options)
...
+   */
+  public int getClusterSize() {
+    return getSingleCluster().getServeNodes();
+  }
+
+  /**
    * Gets a {@link ListClustersResponse} that contains all of the clusters for the
    * projectId/instanceId configuration.
-   * @return the current state of the instance
+   * @return the current state of the instance or the project if the instance is "-"
```
kevinsi4508: "@return the current state of instance"? Please clarify, I thought this method returns a cluster list.

kevinsi4508: Maybe something like: @Return all clusters in the instance if the instance ID is provided, all clusters in project otherwise.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterUtilities.java
```diff
@@ -118,32 +184,73 @@ public synchronized ListClustersResponse getClusters() {
...
     if (currentSize == newSize) {
-      logger.info("Cluster %s already has %d nodes.", clusterId, newSize);
+      logger.info("Cluster %s already has %d nodes.", getClusterId(cluster), newSize);
     } else {
       updateClusterSize(cluster.getName(), newSize);
     }
   }
 
+  public Cluster getSingleCluster() {
+    Preconditions.checkState(getClusters().getClustersCount() != 0,
```
kevinsi4508: Delete this precondition check, the second one is what we need.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterUtilities.java
```diff
@@ -118,32 +184,73 @@ public synchronized ListClustersResponse getClusters() {
...
    * @param clusterId
    * @param zoneId
    * @param incrementCount a positive or negative number to add to the current node count
    * @return the new size of the cluster.
-   * @throws InterruptedException
+   * @throws InterruptedException if the cluster is in the middle of updating, and an interrupt was
+   *           received
    */
-  public synchronized int incrementClusterSize(String clusterId, String zoneId, int incrementCount)
+  public int incrementClusterSize(String clusterId, String zoneId, int incrementCount)
```
kevinsi4508: I am not sure why we even have this method. Wouldn't setClusterSize() is sufficient.

kevinsi4508: "increment" usually means increasing size.  Perhaps: AdjustClusterSize(... int adjustment/delta) ?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterUtilities.java
```diff
@@ -154,6 +261,10 @@ public synchronized int incrementClusterSize(String clusterId, String zoneId, in
     return newSize;
   }
 
+  private String getClusterId(Cluster cluster) {
```
kevinsi4508: nit: static

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterUtilities.java
```diff
@@ -217,7 +328,7 @@ public synchronized int getClusterNodeCount(String clusterId, String zoneId) {
    * @return the {@link Cluster} if it was set. If the cluster is not found, throw a {@link
    *     NullPointerException}.
    */
-  public synchronized Cluster getCluster(String clusterId, String zoneId) {
+  public Cluster getCluster(String clusterId, String zoneId) {
     Cluster response = null;
     for (Cluster cluster : getClusters().getClustersList()) {
```
sduskis: We allow users to set only the projectId.  This method allows us to use v1 configuration with a v2 client.  Let's speak offline about this.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterUtilities.java
sduskis: lookupCluster() depends on getSingleCluster().  getSingleCluster() throws an exception.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterUtilities.java
sduskis: I think you're right.  It's better to have fewer methods here.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterUtilities.java
kevinsi4508: I think my point is that with v2 client, getClusters() could return all clusters within a project (belong to different instances).

The following is valid and "cluster1" cluster ID will match both of the clusters.
Instance1:cluster1
instance2:cluster1

sduskis: Do we allow a single project in a single zone to have duplicate clusterids?  Last time I asked, I thought the answer was no.  This is not allowed:

> Instance1:cluster1 (where cluster is in zone1)
> instance2:cluster1 (where cluster is in zone1)

This is:

> Instance1:cluster1 (where cluster is in zone1)
> instance2:cluster1 (where cluster is in zone2)

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterName.java
sduskis: The pattern has 3 groups.

kevinsi4508: Hm? the following has two groups, no?

"projects/[^/]+/instances/([^/]+)/clusters/([^/]+)"
!=
"projects/(.+)/instances/(.+)/clusters/(.+)"

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterUtilities.java
```diff
@@ -73,6 +70,66 @@ public static BigtableClusterUtilities forAllInstances(String projectId)
   }
 
   /**
+   * @return The instance id associated with the given project, zone and cluster.
+   */
```
kevinsi4508: Can we add a comment here saying that we expect instance and cluster to have one-to-one relationship? I know this for HBase right now, but it doesn't hurt to clarify.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterUtilities.java
```diff
@@ -73,6 +70,66 @@ public static BigtableClusterUtilities forAllInstances(String projectId)
...
+    throws IOException {
+    BigtableClusterUtilities utils;
+    try {
+      utils = BigtableClusterUtilities.forAllInstances(projectId);
+    } catch (GeneralSecurityException e) {
+      throw new RuntimeException("Could not initialize BigtableClusterUtilities", e);
+    }
+
+    try {
+      Cluster cluster = utils.getCluster(clusterId, zoneId);
```
kevinsi4508: Maybe add comment like this:
Since we don't have instance ID, we first list all clusters in the project and try to match by cluster by cluster ID. There must be one and only one matching cluster.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterUtilities.java
```diff
@@ -73,6 +70,66 @@ public static BigtableClusterUtilities forAllInstances(String projectId)
   }
 
   /**
+   * @return The instance id associated with the given project, zone and cluster.
+   */
+  public static String lookupInstanceId(String projectId, String clusterId, String zoneId)
+    throws IOException {
+    BigtableClusterUtilities utils;
+    try {
+      utils = BigtableClusterUtilities.forAllInstances(projectId);
```
kevinsi4508: nit:
Does it make more sense if we only allow BigtableClusterUtilities to be instantiated when either an instance ID or a cluster ID is provided? With that, we don't need to pass in any cluster ID or Instance ID in those non-static public methods.

example: no getClusterSize(String clusterId, String zoneId)  and getClusterSize() 
we will only have getClusterSize()

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterUtilities.java
sduskis: done

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterUtilities.java
kevinsi4508: Zone is not needed, can be refactored later.

kevinsi4508:
Minors things most you have addressed. The code looks good to me, some tests for BigtableClusterUtilities.java would be nice.
LGTM

PR #1125, opened 2016-12-06
Consistently applying Thread.currentThread().interrupt().

sduskis:
All InterruptedException handling should include Thread.currentThread().interrupt() as per:
http://stackoverflow.com/questions/4906799/why-invoke-thread-currentthread-interrupt-when-catch-any-interruptexception

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
@@ -389,6 +389,7 @@ public class BigtableDataGrpcClient implements BigtableDataClient {
       listener.start();
       return listener.getCompletionFuture().get();
     } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
       listener.cancel();
       throw Status.CANCELLED.withCause(e).asRuntimeException();
     } catch (ExecutionException e) {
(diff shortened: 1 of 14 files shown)


igorbernstein2:
Should this be added to the wider `catch(Exception|Throwable)` cases as well?

* https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java#L248

* https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java#L374

* https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java#L371

* https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RowMerger.java#L374

sduskis:
Great feedback! 

- BatchExecutor.java#L248 needs it.  Added it.
- BigtableBufferedMutator.java#L374 seems to not need it.  It uses a slight different path; the BigtableBufferedMutator.offer() method is the one that has the method that's prone to interruption.
- BulkMutation.java#L371 needs it.  Added it. 
- RowMerger.java#L374 does not need it.  There isn't any waiting involved.

PTAL

igorbernstein2:
LGTM

PR #1126, opened 2016-12-06
Optimize the mapreduce Import job.

igorbernstein2:
* Randomize the splits to spread load across all tablet servers
* Increase the min splitsize to give each mapper enough work to do

### DIFF
--- a/bigtable-hbase-parent/bigtable-hbase-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/ShuffledSequenceFileInputFormat.java
+++ b/bigtable-hbase-parent/bigtable-hbase-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/ShuffledSequenceFileInputFormat.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


bigtable-hbase-parent/bigtable-hbase-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/ShuffledSequenceFileInputFormat.java
```diff
@@ -0,0 +1,23 @@
+package com.google.cloud.bigtable.mapreduce;
```
sduskis: Please add the standard copyright header we have in most of our classes.  It should be set to 2016 for new files.

mbrukman: FYI, you can use [`autogen`](https://github.com/mbrukman/autogen) to add copyright headers to files, and yes, it's a shameless plug. :-)

It automatically uses the current year, of course.

igorbernstein2: updated

PR #1130, opened 2016-12-08
Adding a basic snapshot tests.

sduskis:

PR #1132, opened 2016-12-08
Adding support for Snapshots.

sduskis:
This is the core of the integration.  We'll be adding tests in #1130, and then making sure that this PR works.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
@@ -18,6 +18,7 @@ package com.google.cloud.bigtable.config;
 import java.io.Serializable;
 import java.util.Objects;
 
+import com.google.cloud.bigtable.grpc.BigtableClusterName;
 import com.google.cloud.bigtable.grpc.BigtableInstanceName;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
@@ -271,6 +272,7 @@ public class BigtableOptions implements Serializable {
   private final boolean usePlaintextNegotiation;
 
   private final BigtableInstanceName instanceName;
+  private final BigtableClusterName clusterName;
 
   private final BulkOptions bulkOptions;
   private final CallOptionsConfig callOptionsConfig;
@@ -290,6 +292,7 @@ public class BigtableOptions implements Serializable {
       zoneId = null;
       dataChannelCount = 1;
       instanceName = null;
+      clusterName = null;
       usePlaintextNegotiation = false;
 
       bulkOptions = null;
@@ -337,8 +340,14 @@ public class BigtableOptions implements Serializable {
     if (!Strings.isNullOrEmpty(projectId)
         && !Strings.isNullOrEmpty(instanceId)) {
       this.instanceName = new BigtableInstanceName(projectId, instanceId);
+      if (!Strings.isNullOrEmpty(clusterId)) {
+        this.clusterName = instanceName.toClusterName(clusterId);
+      } else {
+        this.clusterName = null;
+      }
     } else {
       this.instanceName = null;
+      this.clusterName = null;
     }
 
     LOG.debug("Connection Configuration: projectId: %s, instanceId: %s, data host %s, "
@@ -471,6 +480,10 @@ public class BigtableOptions implements Serializable {
    */
   public BigtableInstanceName getInstanceName() {
     return instanceName;
+  }
+
+  public BigtableClusterName getClusterName() {
+    return clusterName;
   }
 
   /**
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
@@ -18,7 +18,6 @@ package com.google.cloud.bigtable.config;
 import java.io.Serializable;
 import java.util.Objects;
 
-import com.google.cloud.bigtable.grpc.BigtableClusterName;
 import com.google.cloud.bigtable.grpc.BigtableInstanceName;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
@@ -272,7 +271,6 @@ public class BigtableOptions implements Serializable {
   private final boolean usePlaintextNegotiation;
 
   private final BigtableInstanceName instanceName;
-  private final BigtableClusterName clusterName;
 
   private final BulkOptions bulkOptions;
   private final CallOptionsConfig callOptionsConfig;
@@ -292,7 +290,6 @@ public class BigtableOptions implements Serializable {
       zoneId = null;
       dataChannelCount = 1;
       instanceName = null;
-      clusterName = null;
       usePlaintextNegotiation = false;
 
       bulkOptions = null;
@@ -340,14 +337,8 @@ public class BigtableOptions implements Serializable {
     if (!Strings.isNullOrEmpty(projectId)
         && !Strings.isNullOrEmpty(instanceId)) {
       this.instanceName = new BigtableInstanceName(projectId, instanceId);
-      if (!Strings.isNullOrEmpty(clusterId)) {
-        this.clusterName = instanceName.toClusterName(clusterId);
-      } else {
-        this.clusterName = null;
-      }
     } else {
       this.instanceName = null;
-      this.clusterName = null;
     }
 
     LOG.debug("Connection Configuration: projectId: %s, instanceId: %s, data host %s, "
@@ -480,10 +471,6 @@ public class BigtableOptions implements Serializable {
    */
   public BigtableInstanceName getInstanceName() {
     return instanceName;
-  }
-
-  public BigtableClusterName getClusterName() {
-    return clusterName;
   }
 
   /**
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -21,7 +21,6 @@ import java.io.IOException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.UnknownHostException;
-import java.nio.charset.Charset;
 import java.security.GeneralSecurityException;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -36,11 +35,7 @@ import javax.annotation.Nullable;
 import javax.net.ssl.SSLException;
 
 import com.google.api.client.util.Strings;
-import com.google.bigtable.admin.v2.Cluster;
 import com.google.bigtable.admin.v2.ListClustersResponse;
-import com.google.bigtable.v2.ColumnRange;
-import com.google.bigtable.v2.RowFilter;
-import com.google.bigtable.v2.RowFilter.Chain;
 import com.google.cloud.bigtable.config.BigtableOptions;
 import com.google.cloud.bigtable.config.BigtableVersionInfo;
 import com.google.cloud.bigtable.config.CredentialOptions;
@@ -244,6 +239,12 @@ public class BigtableSession implements Closeable {
       .synchronizedList(new ArrayList<ManagedChannel>());
   private final ImmutableList<HeaderInterceptor> headerInterceptors;
 
+  /**
+   * This cluster name is either configured via BigtableOptions' clusterId, or via a lookup of the
+   * clusterID based on BigtableOptions projectId and instanceId.  See {@link BigtableClusterUtilities}
+   */
+  private BigtableClusterName clusterName;
+
 
   /**
    * Use {@link com.google.cloud.bigtable.grpc.BigtableSession#BigtableSession(BigtableOptions)} instead;
@@ -386,10 +387,12 @@ public class BigtableSession implements Closeable {
 
 
   /**
-   * Return options with legacy input options, if any, resolved into currently supported options.
+   * Snapshot operations need various aspects of a {@link BigtableClusterName}. This method gets a
+   * clusterId from either a lookup (projectId and instanceId translate to a single clusterId when
+   * an instance has only one cluster) or from {@link BigtableOptions#getClusterId()}.
    */
   public synchronized BigtableClusterName getClusterName() throws IOException {
-    if (options.getClusterName() == null) {
+    if (this.clusterName == null) {
       try (BigtableClusterUtilities util =
           BigtableClusterUtilities.forInstance(options.getProjectId(), options.getInstanceId())) {
         ListClustersResponse clusters = util.getClusters();
@@ -397,16 +400,12 @@ public class BigtableSession implements Closeable {
           String.format(
             "Project '%s' / Instance '%s' has %d clusters. There must be exactly 1 for this operation to work.",
             options.getProjectId(), options.getInstanceId(), clusters.getClustersCount()));
-        Cluster cluster = clusters.getClusters(0);
-        options = options.toBuilder()
-            .setClusterId(new BigtableClusterName(cluster.getName()).getClusterId())
-            .setZoneId(BigtableClusterUtilities.getZoneId(cluster))
-            .build();
+        clusterName = new BigtableClusterName(clusters.getClusters(0).getName());
       } catch (GeneralSecurityException e) {
         throw new IOException("Could not get cluster Id.", e);
       }
     }
-    return options.getClusterName();
+    return clusterName;
   }
 
   /**
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -35,6 +35,7 @@ import javax.annotation.Nullable;
 import javax.net.ssl.SSLException;
 
 import com.google.api.client.util.Strings;
+import com.google.bigtable.admin.v2.ListClustersResponse;
 import com.google.cloud.bigtable.config.BigtableOptions;
 import com.google.cloud.bigtable.config.BigtableVersionInfo;
 import com.google.cloud.bigtable.config.CredentialOptions;
@@ -232,7 +233,7 @@ public class BigtableSession implements Closeable {
   private BigtableTableAdminClient tableAdminClient;
   private BigtableInstanceGrpcClient instanceAdminClient;
 
-  private final BigtableOptions options;
+  private  BigtableOptions options;
   private final List<ManagedChannel> managedChannels = Collections
       .synchronizedList(new ArrayList<ManagedChannel>());
   private final ImmutableList<HeaderInterceptor> headerInterceptors;
@@ -375,6 +376,28 @@ public class BigtableSession implements Closeable {
       return options.toBuilder().setInstanceId(instanceId).build();
     }
     return options;
+  }
+
+
+  /**
+   * Return options with legacy input options, if any, resolved into currently supported options.
+   */
+  public synchronized BigtableClusterName getClusterName() throws IOException {
+    if (options.getClusterName() == null) {
+      try (BigtableClusterUtilities util =
+          BigtableClusterUtilities.forInstance(options.getProjectId(), options.getInstanceId())) {
+        final ListClustersResponse clusters = util.getClusters();
+        Preconditions.checkState(clusters.getClustersCount() == 0,
+          "Project '%s' / Instance '%s' has %d clusters. There must be exactly 1 for this operation to work.",
+          options.getProjectId(), options.getInstanceId(), clusters.getClustersCount());
+        String clusterName = util.getSingleCluster().getName();
+        String clusterId = BigtableClusterName.parse(clusterName).getClusterId();
+        options = options.toBuilder().setClusterId(clusterId).build();
+      } catch (GeneralSecurityException e) {
+        throw new IOException("Could not get cluster Id.", e);
+      }
+    }
+    return options.getClusterName();
   }
 
   /**
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.UnknownHostException;
+import java.nio.charset.Charset;
 import java.security.GeneralSecurityException;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -35,7 +36,11 @@ import javax.annotation.Nullable;
 import javax.net.ssl.SSLException;
 
 import com.google.api.client.util.Strings;
+import com.google.bigtable.admin.v2.Cluster;
 import com.google.bigtable.admin.v2.ListClustersResponse;
+import com.google.bigtable.v2.ColumnRange;
+import com.google.bigtable.v2.RowFilter;
+import com.google.bigtable.v2.RowFilter.Chain;
 import com.google.cloud.bigtable.config.BigtableOptions;
 import com.google.cloud.bigtable.config.BigtableVersionInfo;
 import com.google.cloud.bigtable.config.CredentialOptions;
@@ -57,6 +62,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableList.Builder;
+import com.google.protobuf.ByteString;
 
 import io.grpc.ManagedChannel;
 import io.grpc.netty.GrpcSslContexts;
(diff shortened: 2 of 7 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
```diff
@@ -337,8 +340,14 @@ private BigtableOptions(
...
         && !Strings.isNullOrEmpty(instanceId)) {
       this.instanceName = new BigtableInstanceName(projectId, instanceId);
+      if (!Strings.isNullOrEmpty(clusterId)) {
+        this.clusterName = instanceName.toClusterName(clusterId);
+      } else {
+        this.clusterName = null;
+      }
     } else {
       this.instanceName = null;
+      this.clusterName = null;
```
kevinsi4508: nit: explicitly set/default them to null first

this.instanceName = null;
this.clusterName = null;

if (...)
  this.instanceName = ... 
  if (...)
    this.clusterName = ...

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -330,6 +331,28 @@ private BigtableOptions resolveLegacyOptions(BigtableOptions options) throws IOE
     return options;
   }
 
+
+  /**
+   * Return options with legacy input options, if any, resolved into currently supported options.
+   */
+  public synchronized BigtableClusterName getClusterName() throws IOException {
+    if (options.getClusterName() == null) {
```
kevinsi4508: This is a public method; can you give me an example how it should be used?

I thought what need to done in the constructor is:
if (instance ID provided) {
  cluster ID = get single cluster ID from instance ID.
  check cluster ID is the same as input cluster ID if provided, otherwise set cluster ID in options.
} else {
  instance ID = get single instance ID by cluster ID.
  set instance ID in options.
}

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -330,6 +331,28 @@ private BigtableOptions resolveLegacyOptions(BigtableOptions options) throws IOE
...
+      try (BigtableClusterUtilities util =
+          BigtableClusterUtilities.forInstance(options.getProjectId(), options.getInstanceId())) {
+        final ListClustersResponse clusters = util.getClusters();
+        Preconditions.checkState(clusters.getClustersCount() == 0,
+          "Project '%s' / Instance '%s' has %d clusters. There must be exactly 1 for this operation to work.",
+          options.getProjectId(), options.getInstanceId(), clusters.getClustersCount());
+        String clusterName = util.getSingleCluster().getName();
```
kevinsi4508: I thought getSingleCluster() does all the checking already. Why do we need to getClusters() first?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSnapshotName.java
```diff
@@ -0,0 +1,90 @@
...
+    this.snapshotName = snapshotName;
+    Matcher matcher = PATTERN.matcher(snapshotName);
+    Preconditions.checkArgument(matcher.matches(), "Malformed snapshot name");
+    this.instanceId = matcher.group(1);
+    this.clusterId = matcher.group(2);
+    this.snapshotId = matcher.group(3);
+  }
+
+  /**
+   * Returns the fully qualified Snapshot name same thing as {@link #getClusterName()}.
```
kevinsi4508: same things as getSnapshotName

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSnapshotName.java
```diff
@@ -0,0 +1,90 @@
...
+   * @return The id of the instance that contains this cluster. It's the second group in the Snapshot name
+   *         name: "projects/{projectId}/instances/{instanceId}/clusters/{clusterId}/snapshots/{@snapshotId}".
+   */
+  public String getInstanceId() {
+    return instanceId;
+  }
+
+  /**
+   * @return The id of this cluster. It's the third group in the Snapshot name:
+   *         "projects/{projectId}/instances/{instanceId}/clusters/{clusterId}/snapshots/{@snapshotId}".
```
kevinsi4508: why "@snapshotI", should be just "snapshotId"?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSnapshotName.java
```diff
@@ -0,0 +1,90 @@
...
+   * Returns the fully qualified Snapshot name same thing as {@link #getClusterName()}.
+   */
+  @Override
+  public String toString() {
+    return snapshotName;
+  }
+
+  /**
+   * @return The id of the instance that contains this cluster. It's the second group in the Snapshot name
+   *         name: "projects/{projectId}/instances/{instanceId}/clusters/{clusterId}/snapshots/{@snapshotId}".
```
kevinsi4508: remove the extra "name".

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSnapshotName.java
```diff
@@ -0,0 +1,90 @@
...
+ * This class encapsulates a Snapshot name of the form
+ * projects/(projectId)/instances/(instanceId)/clusters/(clusterId)
+ */
+public class BigtableSnapshotName {
+  // Use a very loose pattern so we don't validate more strictly than the server.
+  private static final Pattern PATTERN =
+      Pattern.compile("projects/[^/]+/instances/([^/]+)/clusters/([^/]+)/snapshots/([^/]+)");
+
+
+  public static BigtableSnapshotName parse(String snapshotName) {
```
kevinsi4508: Do we need a static factory method where the public constructor is doing the same thing?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableTableName.java
```diff
@@ -24,10 +28,40 @@
...
 public class BigtableTableName {
+  // Use a very loose pattern so we don't validate more strictly than the server.
+  private static final Pattern PATTERN =
+      Pattern.compile("projects/[^/]+/instances/([^/]+)/tables/([^/]+)");
+
   private final String tableName;
+  private final String instanceId;
+  private final String tableId;
+
+  public static BigtableTableName parse(String name) {
```
kevinsi4508: delete?

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -819,93 +836,172 @@ public boolean isTableAvailable(TableName tableName, byte[][] splitKeys) throws
   @Override
   public void snapshot(String snapshotName, TableName tableName)
```
kevinsi4508: Why snaphostName is a string and tableName is TableName type?

kevinsi4508: Never mind, this is HBase

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -819,93 +836,172 @@ public boolean isTableAvailable(TableName tableName, byte[][] splitKeys) throws
...
+      waitForOperation(operation.getName());
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new IOException("Create snapshot operation was interrupted", e);
+    }
+  }
+
+  private com.google.longrunning.Operation snapshotTable(String snapshotName, TableName tableName)
+      throws IOException {
+    BigtableClusterName clusterName = this.connection.getSession().getClusterName();
```
kevinsi4508: Why are we getting BigtableClusterName and BigtableInstanceName from two different places? Are they V1 and V2 names? If both are V2 name we can get InstanceName from ClusterName.

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -819,93 +836,172 @@ public boolean isTableAvailable(TableName tableName, byte[][] splitKeys) throws
...
+    com.google.longrunning.Operation operation = snapshotTable(snapshotName, tableName);
+    try {
+      waitForOperation(operation.getName());
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new IOException("Create snapshot operation was interrupted", e);
+    }
+  }
+
+  private com.google.longrunning.Operation snapshotTable(String snapshotName, TableName tableName)
```
kevinsi4508: private Operation

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -819,93 +836,172 @@ public boolean isTableAvailable(TableName tableName, byte[][] splitKeys) throws
...
+    BigtableTableName btTableName = instanceName.toTableName(tableName.getNameAsString());
+    return bigtableTableAdminClient.snapshotTable(SnapshotTableRequest.newBuilder()
+        .setCluster(clusterName.getClusterName())
+        .setSnapshotId(snapshotName)
+        .setName(btTableName.toString())
+        .build());
+  }
+
+  /**
+   * Waits for an operation like cluster resizing to complete.
```
kevinsi4508: This is copied form cluster resizing. Please put it in a common place to be reused and also update the code/comment to be generic operation.

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -819,93 +836,172 @@ public boolean isTableAvailable(TableName tableName, byte[][] splitKeys) throws
...
   /** {@inheritDoc} */
   @Override
   public void cloneSnapshot(String snapshotName, TableName tableName)
       throws IOException, TableExistsException, RestoreSnapshotException {
-    throw new UnsupportedOperationException("cloneSnapshot");  // TODO
+    BigtableInstanceName instanceName = options.getInstanceName();
+    BigtableClusterName clusterName = this.connection.getSession().getClusterName();
+    String btSnapshotName = clusterName.toSnapshotName(snapshotName);
+
+    bigtableTableAdminClient.createTableFromSnapshot(CreateTableFromSnapshotRequest.newBuilder()
```
kevinsi4508: Same as snapshot(), this is a LRO, we need to poll.

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -819,93 +836,172 @@ public boolean isTableAvailable(TableName tableName, byte[][] splitKeys) throws
...
   /** {@inheritDoc} */
   @Override
   public MasterProtos.SnapshotResponse takeSnapshotAsync(HBaseProtos.SnapshotDescription snapshot)
       throws IOException, SnapshotCreationException {
-    throw new UnsupportedOperationException("takeSnapshotAsync");  // TODO
+    snapshotTable(snapshot.getName(), TableName.valueOf(snapshot.getTable()));
+    LOG.warn("isSnapshotFinished() is not currently supported by BigtableAdmin.\n"
+        + "You may poll for existence of the snapshot with listSnapshots(snpashotName)");
+    return MasterProtos.SnapshotResponse.newBuilder()
+        .setExpectedTimeout(TimeUnit.MINUTES.toMillis(5)).build();
```
kevinsi4508: Just curious, why setting expected timeout to 5 minutes in the response?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
sduskis: The value are final.  Java only lets you set the value once.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
sduskis: We should defer lookup of the cluster until a snapshot operation is performed.  Looking up the cluster is an extra call to an admin method, which we only want to do if it's absolutely necessary.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
sduskis: doh.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSnapshotName.java
sduskis: hm.  I cut and pasted from BigtableClusterName.  You're right.  we don't need parse.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableTableName.java
sduskis: yup

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -819,93 +836,172 @@ public boolean isTableAvailable(TableName tableName, byte[][] splitKeys) throws
   @Override
   public void snapshot(String snapshotName, TableName tableName)
       throws IOException, SnapshotCreationException, IllegalArgumentException {
-    throw new UnsupportedOperationException("snapshot");  // TODO
+    com.google.longrunning.Operation operation = snapshotTable(snapshotName, tableName);
```
sduskis: oh... right.

sduskis: currentOperationState.getDone() isn't enough.  You still need to check for the resultCase.  Your point about checkout the operation that's returned is well taken.  No need to look it up again in the quick cases.

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
sduskis: :)

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
sduskis: It's a good point about shared code, but I think I'd like to defer the refactoring to another PR.

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -819,93 +836,172 @@ public boolean isTableAvailable(TableName tableName, byte[][] splitKeys) throws
...
+
+  private com.google.longrunning.Operation snapshotTable(String snapshotName, TableName tableName)
+      throws IOException {
+    BigtableClusterName clusterName = this.connection.getSession().getClusterName();
+    BigtableInstanceName instanceName = options.getInstanceName();
+    BigtableTableName btTableName = instanceName.toTableName(tableName.getNameAsString());
+    return bigtableTableAdminClient.snapshotTable(SnapshotTableRequest.newBuilder()
+        .setCluster(clusterName.getClusterName())
+        .setSnapshotId(snapshotName)
+        .setName(btTableName.toString())
```
sduskis: Let's discuss this offline.  Let's stick with the default if we use HBase.

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
sduskis: It was a guess... Is there a better guess?  I need to put in something.

sduskis:
All comments were addressed, and integration tests are passing against our test environment.  PTAL.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
kevinsi4508: ah... final variables, didn't see that. anyway, it is a nit.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
kevinsi4508: That is where I can't see it clearly; how it is used. Can you give an example? I understand that most of time, we don't need cluster ID; it is only needed when a snapshot operation is called. (* please put this in the comment *)

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -232,7 +239,7 @@ private synchronized static void initializeResourceLimiter(BigtableOptions optio
   private BigtableTableAdminClient tableAdminClient;
   private BigtableInstanceGrpcClient instanceAdminClient;
 
-  private final BigtableOptions options;
+  private  BigtableOptions options;
```
kevinsi4508: nit: remove an extra whitespace.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -377,6 +384,31 @@ private BigtableOptions resolveLegacyOptions(BigtableOptions options) throws IOE
...
+    if (options.getClusterName() == null) {
+      try (BigtableClusterUtilities util =
+          BigtableClusterUtilities.forInstance(options.getProjectId(), options.getInstanceId())) {
+        ListClustersResponse clusters = util.getClusters();
+        Preconditions.checkState(clusters.getClustersCount() == 1,
+          String.format(
+            "Project '%s' / Instance '%s' has %d clusters. There must be exactly 1 for this operation to work.",
+            options.getProjectId(), options.getInstanceId(), clusters.getClustersCount()));
+        Cluster cluster = clusters.getClusters(0);
+        options = options.toBuilder()
```
kevinsi4508: The reason you want to mutate "options" here so that the next time when this is called, we don't have to lookup for cluster again. I don't know if it is worth it. I would rather just 1) do this in constructor or 2) not mutating "options" and return cluster ID here.

Up to you.

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -819,93 +832,174 @@ public boolean isTableAvailable(TableName tableName, byte[][] splitKeys) throws
...
-    throw new UnsupportedOperationException("snapshot");  // TODO
+    waitForOperation(snapshotTable(snapshotName, tableName));
+  }
+
+  private Operation snapshotTable(String snapshotName, TableName tableName)
+      throws IOException {
+    return bigtableTableAdminClient.snapshotTable(SnapshotTableRequest.newBuilder()
+        .setCluster(getClusterName().toString())
+        .setSnapshotId(snapshotName)
+        .setName(options.getInstanceName().toTableNameStr(tableName.getNameAsString()).toString())
```
kevinsi4508: nit: Already a string, don't need to call toString(), correct?

bigtable-hbase-parent/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestSnapshots.java
```diff
@@ -44,37 +44,51 @@
...
-   public void delete() throws IOException {
-     Admin admin = getConnection().getAdmin();
-     admin.disableTable(tableName);
-     admin.deleteTable(tableName);
-     admin.disableTable(clonedTableName);
-     admin.deleteTable(clonedTableName);
-     admin.close();
-   }
+  @After
+  public void delete() throws IOException {
```
kevinsi4508: nit: cleanup()? we already have a delete method: "delete(Admin admin, TableName tableName) "

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -819,93 +832,174 @@ public boolean isTableAvailable(TableName tableName, byte[][] splitKeys) throws
...
+              "System returned invalid response for Operation check: " + currentOperationState);
+        }
+      }
+      try {
+        Thread.sleep(200);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        throw new IOException("Wating for operation was interrupted.", e);
+      }
+      currentOperationState =
```
kevinsi4508: We should either do a info log to inform users that we are still waiting for the operation to complete or we should timeout at some point.

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -819,93 +832,174 @@ public boolean isTableAvailable(TableName tableName, byte[][] splitKeys) throws
   @Override
   public void snapshot(String snapshotName, TableName tableName)
       throws IOException, SnapshotCreationException, IllegalArgumentException {
-    throw new UnsupportedOperationException("snapshot");  // TODO
+    waitForOperation(snapshotTable(snapshotName, tableName));
+  }
+
+  private Operation snapshotTable(String snapshotName, TableName tableName)
+      throws IOException {
+    return bigtableTableAdminClient.snapshotTable(SnapshotTableRequest.newBuilder()
```
kevinsi4508: Sorry! I don't know if you have addressed my previous comment here. Do we really want default to Max TTL? This makes me think of having a better default on server side.

kevinsi4508:
Some more minor things. We are most there. Thanks!

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
sduskis: Good point.  Do you have a feel about the right amount of time for timeouts and the right frequency of logging?

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
sduskis: HBase snapshots are not cleaned up.  The current behavior would be consistent with HBase.  HBase should use our server-side default, what ever that is.

sduskis:
All comments were addressed.  PTAL.

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
kevinsi4508: Could you document it here? Thanks!

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
kevinsi4508: We can log every 5 seconds.

kevinsi4508:
LGTM

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
sduskis: What should I document as the default?  Should we say it's 1 year?

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
sduskis: 5 seconds per logging seems reasonable.  How long should we wait before we timeout?

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
kevinsi4508: We should mention here that we are not setting TTL and it is whatever the server TTL.

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
kevinsi4508: Since we are logging something every 5 seconds, we can have longer timeout, like 30 minutes.

sduskis: 30 minutes is way too much to wait.  I was thinking a maximum of 2.  How long would our largest current table take to snapshot?

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
sduskis: Is there a way for users to determine what the server TTL is?

kevinsi4508: Yes, API Doc mentions the default TTL.

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
kevinsi4508: Largest tables could take a few minutes.

kevinsi4508:
LGTM

PR #1134, opened 2016-12-09
Row merger does sorting and deduping

sduskis:
RowMerger is the right place to sort and dedup cells.  Move the sorting logic from FlatRowAdapter to RowMerger.


igorbernstein2:
Might be a good idea to add a test to verify that RowMerger will sort the column families correctly, but other wise LGTM

sduskis:
I added an integration test.  I was thinking the same thing.

PR #1136, opened 2016-12-10
Dataproc job packaging

igorbernstein2:
Repackage the dataproc mapreduce job in a single shaded jar.

### DIFF
--- a/bigtable-hbase-parent/bigtable-hbase-mapreduce/pom.xml
+++ b/bigtable-hbase-parent/bigtable-hbase-mapreduce/pom.xml
@@ -56,6 +56,9 @@ limitations under the License.
             <artifactId>bigtable-hbase-1.2</artifactId>
             <version>${project.version}</version>
 
+            <!-- Exclude all of hadoop's provided deps. To keep the list up to date, run
+            `mvn dependency:tree` and exclude all of hadoop-client's transitive dependencies that
+            are not in the provided scope -->
             <exclusions>
                 <exclusion>
                     <groupId>io.netty</groupId>
@@ -141,7 +144,6 @@ limitations under the License.
             <groupId>org.apache.hbase</groupId>
             <artifactId>hbase-server</artifactId>
             <version>${hbase.version}</version>
-
             <exclusions>
                 <exclusion>
                     <groupId>*</groupId>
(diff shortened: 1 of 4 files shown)


bigtable-hbase-parent/bigtable-hbase-mapreduce/pom.xml
```diff
@@ -33,54 +33,125 @@ limitations under the License.
...
-            <artifactId>bigtable-hbase-1.1</artifactId>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>bigtable-hbase-1.2</artifactId>
             <version>${project.version}</version>
+
             <exclusions>
                 <exclusion>
-                    <groupId>${project.groupId}</groupId>
-                    <artifactId>bigtable-hbase-shaded</artifactId>
+                    <groupId>io.netty</groupId>
```
sduskis: Are these required outside of Dataproc, if so, then the exclusion list should probably be a set ofshade plugin artifactSet/exclusions

sduskis: Or even better, an explicit list of inclusions.

igorbernstein2: I should've added a comment explaining this. Its a not a question if these deps are used, but that shading them would create a conflict with hadoop's classpath. Since hadoop (< 3.0) doesn't isolate its classpath, mapreduce jobs live in hadoop's classpath. I'm only trying to de-duplicate the deps on the classpath.

To make this situation somewhat manageable, hadoop publishes hadoop-client pom.xml which lists all of the deps as provided. Unfortunately maven doesn't have a way to force conflict resolution in favor of the provided scope and mixes the deps between hadoop-client's provided deps and hbase-client's compile deps.

The easiest way I found to discover the overlap is to list hadoop-client first and exclude all hadoop-client deps that are marked as compile (since hadoop-client is listed first, it will keep its deps and if maven picks hbase-client's version it will inherit the compile scope).  Using this approach, we will be able upgrade the hbase dep in the future. Switching it to a shade inclusion model, while less verbose, will make it harder to upgrade hbase.

igorbernstein2:
I added a comment explaining the exclusions. This should be ready to be merged

Issue #1107, opened 2016-11-22
BulkMutation / BufferedMutator needs an auto flush timer


sduskis:
The go implementation of bulk mutation has a configurable timer that will auto-flush after a period of inactivity.  Java should have that option as well.

PR #1138, opened 2016-12-13
Add option to autoflush the bulk mutations

igorbernstein2:
When enabled, this will ensure that rows don't linger in bulk mutations indefinitely


sduskis:
LGTM other than equals()/toString().

Noting that this fixes #1107.

igorbernstein2:
updated

PR #1142, opened 2016-12-15
Removing an unnecessary generics argument.

sduskis:
For reference, this generic argument was added when we were upgrading between V1 and V2.  There were 2 versions of ReadRowsRequest, one for each version, and we wanted to support both.  At this point, we only need to specify the v2 ReadRowsRequest.


igorbernstein2:
LGTM

PR #1143, opened 2016-12-15
Refactoring ReadRowsRequestRetryHandler to only manage ReadRowsRequest

sduskis:
- renaming ReadRowsRequestRetryHandler to ReadRowsRequestManager.
- ReadRowsRequestManager does not do backoffs and retries.  It only manages ReadRowsRequests for retries
- Retries and backoffs are performed by ResumingStreamingResultScanner
- Tests for ResumingStreamingResultScanner were broken up into tests to for ReadRowsRequest management in TestReadRowsRequestManager and tests for retry behavior in TestResumingStreamingResultScanner
- This is a step towards performing retries via AbstractRetryingRpcListener

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ReadRowsRequestManager.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ReadRowsRequestManager.java
@@ -17,47 +17,27 @@ package com.google.cloud.bigtable.grpc.scanner;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
-import com.google.api.client.util.BackOff;
-import com.google.api.client.util.Sleeper;
 import com.google.bigtable.v2.ReadRowsRequest;
 import com.google.bigtable.v2.RowRange;
 import com.google.bigtable.v2.RowRange.EndKeyCase;
 import com.google.bigtable.v2.RowRange.StartKeyCase;
 import com.google.bigtable.v2.RowSet;
-import com.google.cloud.bigtable.config.Logger;
-import com.google.cloud.bigtable.config.RetryOptions;
-import com.google.cloud.bigtable.grpc.async.BigtableAsyncRpc.RpcMetrics;
-import com.google.cloud.bigtable.grpc.io.IOExceptionWithStatus;
 import com.google.protobuf.ByteString;
 
-import io.grpc.Status;
-
-import java.io.IOException;
 import java.nio.ByteBuffer;
-import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * Keeps track of Rows returned from a readRows RPC for information relevant to resuming the RPC
- * after temporary problems like DEADLINE_EXCEEDED.
+ * after temporary problems.
  *
  * @author sduskis
  * @version $Id: $Id
  */
-public class ReadRowsRequestRetryHandler {
+public class ReadRowsRequestManager {
 
   // Member variables from the constructor.
-  private final RetryOptions retryOptions;
   private final ReadRowsRequest originalRequest;
-  private final RpcMetrics rpcMetrics;
-  private final Logger logger;
 
-  // Internal member variables.
-
-  // The number of times we've retried after a timeout
-  private AtomicInteger timeoutRetryCount = null;
-
-  private BackOff currentErrorBackoff;
-  private Sleeper sleeper = Sleeper.DEFAULT;
   // The number of rows read so far.
   private long rowCount = 0;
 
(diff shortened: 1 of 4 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ReadRowsRequestManager.java
```diff
@@ -67,77 +47,18 @@
...
    * </p>
-   * @param retryOptions a {@link RetryOptions} object.
    * @param originalRequest a {@link ReadRowsRequest} object.
-   * @param rpcMetrics a {@link com.google.cloud.bigtable.grpc.async.BigtableAsyncRpc.RpcMetrics} object to keep track of retries and
-   *          failures.
-   * @param logger a {@link Logger} to log info messages about the state of retries.
    */
-  public ReadRowsRequestRetryHandler(RetryOptions retryOptions, ReadRowsRequest originalRequest,
-      RpcMetrics rpcMetrics, Logger logger) {
+  public ReadRowsRequestManager(ReadRowsRequest originalRequest) {
```
igorbernstein2: Might be a good idea to make this class package protected to signify that its not part of the public api

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResumingStreamingResultScanner.java
```diff
@@ -87,25 +103,82 @@ public FlatRow next() throws IOException {
       try {
         FlatRow result = currentDelegate.next();
         if (result != null) {
-          retryHandler.update(result);
+          requestManager.updateLastFoundKey(result.getRowKey());
+          // We've had at least one successful RPC, reset the backoff and retry counter
+          currentErrorBackoff = null;
+          timeoutRetryCount = null;
```
igorbernstein2: Might be less error prone to initialize the AtomicInteger in the constructor & set it to 0 here

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResumingStreamingResultScanner.java
```diff
@@ -87,25 +103,82 @@ public FlatRow next() throws IOException {
...
 
+
+  private ReadRowsRequest handleScanTimeout(ScanTimeoutException rte) throws IOException {
+    logger.info("The client could not get a response in %d ms. Retrying the scan.",
+      retryOptions.getReadPartialRowTimeoutMillis());
+
+    // Reset the error backoff in case we encountered this timeout after an error.
+    // Otherwise, we will likely have already exceeded the max elapsed time for the backoff
+    // and won't retry after the next error.
+    currentErrorBackoff = null;
```
igorbernstein2: similar to timeoutRetryCount, might be less error prone to construct once & reset when necessary

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResumingStreamingResultScanner.java
```diff
@@ -46,6 +51,14 @@
   private Context operationContext;
   private Context rpcContext;
 
+  private RetryOptions retryOptions;
+
+  // The number of times we've retried after a timeout
+  private AtomicInteger timeoutRetryCount = null;
```
igorbernstein2: This feels a bit weird, it doesn't seem like this class is generally thread safe, why use AtomicInteger?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ReadRowsRequestManager.java
```diff
@@ -17,47 +17,27 @@
...
 /**
  * Keeps track of Rows returned from a readRows RPC for information relevant to resuming the RPC
- * after temporary problems like DEADLINE_EXCEEDED.
+ * after temporary problems.
  *
  * @author sduskis
  * @version $Id: $Id
  */
-public class ReadRowsRequestRetryHandler {
+public class ReadRowsRequestManager {
```
igorbernstein2: Might be a good idea to make this class package protected to signify that its not part of the public api

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ReadRowsRequestManager.java
```diff
@@ -67,77 +47,18 @@
...
-      rpcMetrics.markRetriesExhasted();
-      throw new BigtableRetriesExhaustedException("Exhausted streaming retries.", ioe);
-    }
-
-    sleep(nextBackOffMillis);
-    return updateRequest();
-  }
-
-  private ReadRowsRequest updateRequest() {
+  public ReadRowsRequest getUpdatedRequest() {
```
igorbernstein2: build would be more appropriate

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ReadRowsRequestManager.java
```diff
@@ -67,77 +47,18 @@
...
-    rowCount++;
-    // We've had at least one successful RPC, reset the backoff and retry counter
-    currentErrorBackoff = null;
-    timeoutRetryCount = null;
   }
 
-  void updateLastFoundKey(ByteString key) {
+  public void updateLastFoundKey(ByteString key) {
     this.lastFoundKey = key;
+    rowCount++;
```
igorbernstein2: Would it be make sense to add a sanity check that the key is >= than the previous lastFoundKey?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResumingStreamingResultScanner.java
sduskis: There's a non trivial (relatively speaking) cost to building the Backoff object, which IIRC, promoted this us to set the AtomicInteger and Backoff object at failure time rather than incur the creation cost for the vast majority of successful operations.

I just tested AtomicInteger creation, but the cost is less than 1 nanosecond per create.  a new Backoff object is 75 nanoseconds. 

I think you're right about thread safety being a potential concern.  Do you think synchronized handleScanTimeout and handleIOException is enough?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResumingStreamingResultScanner.java
sduskis: This was the way we did things, which I refactored for this PR, but didn't look at too carefully.  This code was updated over time by different people, and I didn't look at it too carefully.  I'm glad you're reviewing it.

I think synchronizing the handle exception methods is a better approach... an int would work fine in that case.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ReadRowsRequestManager.java
sduskis: SGTM.    Added #1144 to do a sweep of our code for this issue.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResumingStreamingResultScanner.java
igorbernstein2: Please correct me if I'm wrong, but doesn't ReadRowsRequestManager get created once per scan attempt whose response would contain many rows? If so, isn't the 75 ns negligible with respect to network latency and protobuf parsing

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResumingStreamingResultScanner.java
igorbernstein2: I'm not certain, but it doesn't look like this class is accessed from multiple threads. If that's true, then I think we should just mark this class as @NotThreadSafe

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResumingStreamingResultScanner.java
sduskis: I dug into this a bit more, and there seems to be another reason for a null timer.  There's a timer in the Backoff object that starts when the ExponentialBackOff is built.  (The code is [here](https://github.com/google/google-http-java-client/blob/dev/google-http-client/src/main/java/com/google/api/client/util/ExponentialBackOff.java#L171)).

Maybe calling reset on the backoff object for every scan result was the costly part?  I don't completely remember the details, but I do recall settling on this.

I think we'd need to do a bit more analysis and performance testing before we reinstate the creation of the backoff outside of exception handling.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ReadRowsRequestManager.java
sduskis: FYI, #1146

sduskis:
I'm merging as per an offline comment from @igorbernstein2.

PR #1149, opened 2016-12-16
Fixing some eclipse issues for bigtable-hbase-mapreduce.

sduskis:
The shading causes a bit of havoc with Eclipse.  This is how we've fixed issues in the past.

### DIFF
--- a/bigtable-hbase-parent/bigtable-hbase-mapreduce/pom.xml
+++ b/bigtable-hbase-parent/bigtable-hbase-mapreduce/pom.xml
@@ -137,6 +137,10 @@ limitations under the License.
                     <groupId>org.slf4j</groupId>
                     <artifactId>slf4j-api</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>${project.groupId}</groupId>
+                    <artifactId>bigtable-hbase-shaded</artifactId>
+                </exclusion>
             </exclusions>
         </dependency>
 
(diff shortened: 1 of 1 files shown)


bigtable-hbase-parent/bigtable-hbase-mapreduce/pom.xml
```diff
@@ -137,6 +137,10 @@ limitations under the License.
                     <groupId>org.slf4j</groupId>
                     <artifactId>slf4j-api</artifactId>
                 </exclusion>
+                <exclusion>
```
igorbernstein2: Would you mind adding a comment that this is a eclipse specific workaround?

igorbernstein2:
Otherwise LGTM

bigtable-hbase-parent/bigtable-hbase-mapreduce/pom.xml
igorbernstein2: Hmm, that's a bit weird...maven's shade plugin should already be excluding the shaded deps from the resulting pom for bigtable-hbase-1.2. I was guessing that eclipse wasn't parsing the transformed pom.xml and thats why the exclusion was necessary.

Without this change, running `mvn dependency:tree` in bigtable-hbase-mapreduce I get the following:

> [INFO] +- com.google.cloud.bigtable:bigtable-hbase-1.2:jar:0.9.5-SNAPSHOT:compile
> [INFO] |  +- org.apache.hbase:hbase-client:jar:1.2.4:compile
> [INFO] |  |  +- org.apache.hbase:hbase-annotations:jar:1.2.4:compile
> [INFO] |  |  +- org.apache.hbase:hbase-protocol:jar:1.2.4:compile
> [INFO] |  |  \- com.yammer.metrics:metrics-core:jar:2.2.0:compile
> [INFO] |  \- org.apache.hbase:hbase-common:jar:1.2.4:compile
> [INFO] |     \- com.github.stephenc.findbugs:findbugs-annotations:jar:1.3.9-1:compile

sduskis: interesting.  I guess we'll have to add the comment everywhere we do this.  I guess it really is only eclipse.

PR #1155, opened 2016-12-21
Hooking up ResumingStreamingResultScanner with ReadRowsRetryListener.

sduskis:


igorbernstein2:
I think there is a bug here.

It seems like ReadRowsRetryListener#setException should only be called once all of the retries have been exhausted and the Listener is as good as dead. So resetting the RowMerger there doesn't affect anything. However, the RowMerger must be reset when we are about to retry due to timeouts & errors.  So I think the the rowMerger reset must happen in run() and handleTimeout should also delegate to run to retry

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -450,42 +441,33 @@ public int available() {
...
-        readRowsAsync.newCall(callOptions);
-
-    ResponseQueueReader reader = new ResponseQueueReader(
-        retryOptions.getReadPartialRowTimeoutMillis(), retryOptions.getStreamingBufferSize());
-
-    final StreamObserverAdapter<ReadRowsResponse> listener =
-        new StreamObserverAdapter<>(readRowsCall, new RowMerger(reader));
+    ResponseQueueReader reader = new ResponseQueueReader(retryOptions.getStreamingBufferSize());
+    final ReadRowsRetryListener listener = createReadRowsRetryListener(request, reader);
+    return new ResumingStreamingResultScanner(reader, new ScanHandler() {
```
igorbernstein2: Might be cleaner to have the listener implement ScanHandler. I know that you had it that way in the previous iteration, sorry to flip flop on this.

sduskis:
Re: RowMerger reset in run()

[Agreed](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ReadRowsRetryListener.java#L96).  That's how I ended up coding the functionality.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
sduskis: Done.  PTAL.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResumingStreamingResultScanner.java
```diff
@@ -38,176 +32,74 @@
...
-      } catch (IOExceptionWithStatus ioe) {
-        closeRpcContext();
-        closeCurrentDelegate();
-        ReadRowsRequest newRequest = handleIOException(ioe);
-        currentDelegate = scannerFactory.createScanner(newRequest);
-        this.rpcContext = rpcMetrics.timeRpc();
+        scanHandler.handleTimeout(rte);
+      } catch (Throwable e) {
+        scanHandler.cancel();
+        throw new BigtableRetriesExhaustedException("Exhausted streaming retries.", e);
```
igorbernstein2: Might be good to add a separate catch clause for `BigtableRetriesExhaustedException` to avoid double wrapping it

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResumingStreamingResultScanner.java
```diff
@@ -38,176 +32,74 @@
...
-      operationContext = null;
-    }
-  }
-
-  private void closeRpcContext() {
-    if (rpcContext != null) {
-      rpcContext.close();
-      rpcContext = null;
-    }
+    scanHandler.cancel();
```
igorbernstein2: I don't think `synchronized` does anything here

PR #1162, opened 2016-12-27
Add observable API for fetching rows

udoprog:
Hey,

This is a suggestion for a solution to #703.

You don't have to merge this PR, but I'd like to use it as an opportunity to present the kind of API that I'd like to see for #703.

Some notes:
* `ReadRowsStreamObservableRetryListener` is almost a straight up copy of `ReadRowsRetryListener`. A few things could be shared here.
* `StreamObservable` probably should not live in the `scanner` package.
* Cancellation of an in-flight request is not supported, but could be accomplished by returning a future from `void observe()` which can be handled, all though it would not be possible to use `io.grpc.stub.StreamObserver` in that case (all though I believe that you should probably wrap that API regardless).

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataClient.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataClient.java
@@ -26,11 +26,13 @@ import com.google.bigtable.v2.MutateRowsResponse;
 import com.google.bigtable.v2.ReadModifyWriteRowRequest;
 import com.google.bigtable.v2.ReadModifyWriteRowResponse;
 import com.google.bigtable.v2.ReadRowsRequest;
+import com.google.bigtable.v2.ReadRowsResponse;
 import com.google.bigtable.v2.Row;
 import com.google.bigtable.v2.SampleRowKeysRequest;
 import com.google.bigtable.v2.SampleRowKeysResponse;
 import com.google.cloud.bigtable.grpc.scanner.FlatRow;
 import com.google.cloud.bigtable.grpc.scanner.ResultScanner;
+import com.google.cloud.bigtable.grpc.scanner.StreamObservable;
 import com.google.common.collect.ImmutableList;
 import com.google.common.util.concurrent.ListenableFuture;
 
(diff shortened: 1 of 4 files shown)


sduskis:
Why do you want to use `ReadRowsResponse` as the observed object rather than `FlatRow`?  `ReadRowsResponse` is meant as an object for wire-protocol efficiency rather than something that's generally usable for applications.

udoprog:
We can implement a simplified decoding procedure that _could_ also give better performance, because:
* We generally filter to a single row and column family.
* We know exactly what the upper bound size is for a given cell.
* We can do clever things because of our use-case, like speculatively aggregate incoming data (that can be dropped on a reset) to reduce total live memory util at any given time.

But the main reason was that it was the simplest to implement (that would allow for experimenting with the above), and would finally allow us to move forward on #703. I agree that the chunks API is not very ergonomic for general use.

sduskis:
FWIW, #703 was the reason we built `ReadRowsRetryListener`.  I am actively working on this issue.

FlatRow should perform well (in 0.9.5) for most cases.  We improve the performance of the `RowMerger` algorithm quite a bit in the up coming release.  I would really rather not expose ReadRowsResponse unless all other options were exhausted; most users wouldn't know what to do with it.  I am mulling over other alternatives other than FlatRow that will allow for a streaming experience within a Row.

For now, I'd still like to build an observable interface on FlatRow for the next release in early January.  We can build a more robust streaming cell experience if needed after that.  Does that sound reasonable?

udoprog:
Hey,

Right now there is no usable tier 1 client unless you want to generate grpc stubs yourself. But then you lose out on some of the behavior implemented in this client (retries, connection pooling). From our perspective we need to at least have immediate access to a stream of `ReadRowsResponse` for a comparison. If this client doesn't provide it we'll have to build something dedicated specifically to our tests. Exposing a low level interface somehow would be beneficial for us in this regard.
Whatever experience we garner related to this might help inform future interface design?

But, the biggest beef right now is not having asynchronous result scanning. If that would come in the form of `StreamObservable<FlatRow>` it would be a big leap forward compared with today.

On a side note: how and when would you encounter resets? This is probably something we'd like to instrument since it seems like it would directly effect latency (i.e. plot a comparison between latency over resets/s?). Are there any schema designs/write/read patterns one can employ to avoid this?

sduskis:
+1 on having an async observer.  We're working on it.  The last tricky part is handling the scenario where there could be a mysterious dropped connection, and we need to restart the request.  One way to do it is to require that a deadline is set.  Another is to have a timer that periodically checks that there was a recent streamed response... we'll deal with it.

In terms of resets, those happen within large Rows.  If a row is larger than some size (which is configurable on the server side), it could span multiple `ReadRowResponse`s.  A large cell could also be chopped up into multiple chunks.  If that row changed between `ReadRowResponse` transmissions, a reset will be issued.

The row has to be large enough to be chunked and the row has to change via a Mutation in order to get a reset.

sduskis:
Why do we need `StreamObservable`?  Can't we just pass in a `StreamObserver` in a method in `BigtableDataClient`?

udoprog:
@sduskis Yeah that should work. Then the method could immediately return a handle that allows for cancellation.

sduskis:
@udoprog, can you please see [here](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataClient.java#L160) to confirm that the changes make sense in your context?

sduskis:
Closing this.  If there's more we need to do, can you please open an issue or another PR, please?

udoprog:
@sduskis Sorry, missed the first poke. It looks good!

PR #1163, opened 2016-12-28
Fixing ReadRowsListener major bug

sduskis:
Make sure to set lastResponseMs when `run()` is called.  If it's not set, requests that are a bit slow will continuously restart.


sduskis:
FYI @igorbernstein2

PR #1166, opened 2016-12-29
Fix two usability issues with the HBase client

pmkc:
This addresses two issue I had with HBase Client:

* BigtableConfiguration is currently very hard to use with existing Hadoop configurations and clients that call ConnectionFactory.createConnection (which in my opinion should be all of them). I tried to preserve backwards compatibility and convenient one liners, while being more inline with HBaseConfiguration.

* The timeouts in BigtableOptionsFactory seem to match poorly with the HBase settings loaded. In the case of the short timeout, the default HBase value (120s) is outside of the required range and breaks the client. I simply use the Bigtable specific keys.

If you would like this split out or tracked as issues, I'm happy to close and refile. I just wanted to leave this before the weekend. No rush to review this either.

### DIFF
--- a/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConfiguration.java
+++ b/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConfiguration.java
@@ -52,7 +52,7 @@ public class BigtableConfiguration {
    * @return the default bigtable {@link Connection} implementation class found in the classpath.
    */
   public static Class<? extends Connection> getConnectionClass() {
-    Preconditions.checkState(CONNECTION_CLASS != null
+    Preconditions.checkState(CONNECTION_CLASS != null,
         "Could not find an appropriate BigtableConnection class");
     return CONNECTION_CLASS;
   }
--- a/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConfiguration.java
+++ b/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConfiguration.java
@@ -17,6 +17,7 @@ package com.google.cloud.bigtable.hbase;
 
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.client.Connection;
+import org.apache.hadoop.hbase.client.HConnection;
 
 import com.google.api.client.repackaged.com.google.common.base.Preconditions;
 
@@ -34,7 +35,7 @@ public class BigtableConfiguration {
   };
 
   private static final Class<? extends Connection> CONNECTION_CLASS = chooseConnectionClass();
-  
+
   @SuppressWarnings("unchecked")
   private static Class<? extends Connection> chooseConnectionClass() {
     for (String className : CONNECTION_CLASS_NAMES) {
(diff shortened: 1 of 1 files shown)


bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConfiguration.java
```diff
@@ -47,42 +49,87 @@
...
-    return connect(configure(projectId, instanceId));
+    return connect(create(projectId, instanceId));
   }
 
   /**
    * <p>connect.</p>
    *
-   * @param conf a {@link org.apache.hadoop.conf.Configuration} object.
+   * Use {@link org.apache.hadoop.hbase.client.ConnectionFactory#createConnection(Configuration)} if
+   * you want predictable code.
```
sduskis: I think that this comment needs to be updated.  #createConnection is a fine alternative, but it needs HConnection.HBASE_CLIENT_CONNECTION_IMPL.  The intent here is to bypass that, and have a simple way to create a BigtableConnection object.

It's unclear to me what "if you want predictable code" means in this context.

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConfiguration.java
```diff
@@ -47,42 +49,87 @@
     return null;
   }
 
+  private static Configuration setConnectionClass(Configuration config) {
```
sduskis: I'm not a fan of this approach.  This seems to be a setup for ConnectionFactory, which is understandable.  I'd rather have a getConnectionClass() which returns CONNECTION_CLASS.  ConnectionFactory can use the value, but I don't think we should directly mix ConnectionFactory with BigtableConfiguration.

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConfiguration.java
```diff
@@ -47,42 +49,87 @@
...
+    config.set(HConnection.HBASE_CLIENT_CONNECTION_IMPL, CONNECTION_CLASS.getCanonicalName());
+    return config;
+  }
+
   /**
-   * <p>configure.</p>
+   * <p>Create a configuration with HBase resources loaded and Connection Class set.</p>
+   */
+  public static Configuration create() {
+    return setConnectionClass(HBaseConfiguration.create());
```
sduskis: I'm not happy with HBaseConfiguration.create().  It caused us some significant pain in some cases due to the way it looks up XML files in the classpath.

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
```diff
@@ -392,16 +392,10 @@ private static void setClientCallOptions(Configuration configuration,
 
     clientCallOptionsBuilder
         .setUseTimeout(configuration.getBoolean(BIGTABLE_USE_TIMEOUTS_KEY, USE_TIMEOUT_DEFAULT));
-    if (configuration.get(BIGTABLE_RPC_TIMEOUT_MS_KEY) != null) {
```
sduskis: Can you please separate out the BigtableOptionsFactory.java changes into a different PR?

pmkc:
As discussed offline, I left the default behaviors of Configuration.Create(false), but added a new method which takes an existing Configuration.

PTAL.

sduskis:
Travis says that there's a compilation error:

[ERROR] /home/travis/build/GoogleCloudPlatform/cloud-bigtable-client/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConfiguration.java:[55,54] ')' expected
[ERROR] /home/travis/build/GoogleCloudPlatform/cloud-bigtable-client/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConfiguration.java:[56,65] illegal start of expression

Please fix them :)

pmkc:
Oops. Fixed.

PR #1170, opened 2017-01-02
Simplifying commit/reset information retrieval in RowMerger.

sduskis:
There was a duplicate check in RowMerger to confirm that the RowStatus is the right type.  That check happens in the java proto object already.


igorbernstein2:
LGTM

PR #1171, opened 2017-01-02
Removing last scanned row key handling in RowMerger.

sduskis:
Currently, `RowMerger`'s last scanned row handling involves always throwing an exception, which is the wrong behavior.  At this point, `ReadRowsRetryListener` has the right behavior.  It never made sense to throw exceptions, and there doesn't seem to be a good reason to have any reference to las scanned row in `RowMerger` anymore.


igorbernstein2:
LGTM

PR #1172, opened 2017-01-05
Fixing issue in gets that causes CANCELs.

sduskis:
Closing a scan before the OK is processed causes CANCELED, but improves latency.  CANCELs show up on our error graphs, and are scary.  It's better to have higher latency than lots of CANCELs.


igorbernstein2:
LGTM

PR #1174, opened 2017-01-05
Dataflow export

igorbernstein2:
Working on adding tests now


bigtable-dataflow-parent/bigtable-dataflow-import/src/main/java/com/google/cloud/bigtable/dataflowimport/HBaseExportJob.java
```diff
@@ -0,0 +1,303 @@
...
+  public static Pipeline buildPipeline(HBaseExportOptions options) throws IOException {
+    if(options.getDiskSizeGb() == 0) {
+      options.setDiskSizeGb(20);
+    }
+    if(options.getBigtableProjectId() == null) {
+      options.setBigtableProjectId(options.getProject());
+    }
+
+    // Configure Bigtable scan parameters
+    final Scan scan = new Scan()
```
igorbernstein2: To avoid overlap, how about `buildPipeline(DataflowOptions, CloudBigtableScanConfiguration, String destination)`?

bigtable-dataflow-parent/bigtable-dataflow-import/src/main/java/com/google/cloud/bigtable/dataflowimport/HBaseExportJob.java
```diff
@@ -0,0 +1,303 @@
...
+
+      final List<? extends BoundedSource<T>> boundedSources = source
+          .splitIntoBundles(DESIRED_SHARD_SIZE, c.getPipelineOptions());
+
+      LOG.info("Split into {} shards", boundedSources.size());
+
+      // randomize order to make sure that scans are distributed across tablets
+      Collections.shuffle(boundedSources);
+
+      // emit each shard with a random uuid, note that uuids are used instead of counters to prevent misleading the
```
igorbernstein2: My main concern about using ordered keys is that the next step is a GroupByKey, which I assume will sort the elements causing hotspotting.

bigtable-dataflow-parent/bigtable-dataflow-import/src/main/java/com/google/cloud/bigtable/dataflowimport/HBaseExportJob.java
```diff
@@ -0,0 +1,303 @@
...
+      );
+      hadoopConfig.set("fs.gs.project.id", c.getPipelineOptions().as(DataflowPipelineOptions.class).getProject());
+      for (Map.Entry<String, String> entry : HBaseImportIO.CONST_FILE_READER_PROPERTIES.entrySet()) {
+        hadoopConfig.set(entry.getKey(), entry.getValue());
+      }
+
+      super.startBundle(c);
+    }
+
+    @Override
```
sduskis: The overall flow of this method is difficult to follow.  I think the primary culprit is the way timing is done.  Let's discuss how we can improve that.

bigtable-dataflow-parent/bigtable-dataflow-import/src/main/java/com/google/cloud/bigtable/dataflowimport/HBaseExportJob.java
```diff
@@ -0,0 +1,303 @@
...
+      };
+
+      final long start = System.nanoTime();
+      long localStart = System.nanoTime();
+
+      long reportedTotalDuration = 0;
+
+      try (BoundedSource.BoundedReader<Result> reader = source.createReader(c.getPipelineOptions())) {
+        boolean hasMore = reader.start();
+        readDuration.addValue(System.nanoTime() - localStart);
```
sduskis: `System.nanoTime() - localStart` is repeated often.  Does it make sense to encapsulate that?

bigtable-dataflow-parent/bigtable-dataflow-import/src/main/java/com/google/cloud/bigtable/dataflowimport/HBaseExportJob.java
sduskis: sgtm

bigtable-dataflow-parent/bigtable-dataflow-import/src/main/java/com/google/cloud/bigtable/dataflowimport/HBaseExportJob.java
sduskis: Then let's not have ordered keys. Can we separate keys from file name?

igorbernstein2: The keys allow for the bundles keep their filenames stable. If a shard fails to write to gcs, we want it to retry using the same name. Keeping the filename as a key makes that very easy.

sduskis: We do want stable file names and random keys.  A first time (or maybe even later) user wants to see some measure of verification that the process works correctly.  We can add another String or integer to the value that would allow us to have a consistent file name that's different from the key.

igorbernstein2: ok

bigtable-dataflow-parent/bigtable-dataflow-import/src/main/java/com/google/cloud/bigtable/dataflowimport/HBaseExportJob.java
```diff
@@ -162,193 +159,130 @@ public static Pipeline buildPipeline(DataflowPipelineOptions pipelineOptions,
...
-      Collections.shuffle(boundedSources);
+      Collections.shuffle(shards);
 
-      // emit each shard with a random uuid, note that uuids are used instead of counters to prevent misleading the
-      // user into thinking that the filenames are ordered in any way
-      for (BoundedSource<T> sourceBundle : boundedSources) {
-        c.output(KV.of(UUID.randomUUID().toString(), sourceBundle));
+      i = 0;
+      for (ExportShard<T> shard : shards) {
+        c.output(KV.of(i++, shard));
```
sduskis: Is there a benefit of a number vs. a random UUID here?  Just curious.

bigtable-dataflow-parent/bigtable-dataflow-import/src/main/java/com/google/cloud/bigtable/dataflowimport/HBaseExportJob.java
```diff
@@ -162,193 +159,130 @@ public static Pipeline buildPipeline(DataflowPipelineOptions pipelineOptions,
...
-      writeDuration = new BatchCounter(createAggregator("writeDuration(secs)", new SumLongFn()), 1_000);
-      totalDuration = new BatchCounter(createAggregator("totalDuration(secs)", new SumLongFn()), 1_000);
-      bytesWritten = new BatchCounter(createAggregator("bytesWritten(MB)", new SumLongFn()), 1024 * 1024);
+      bytesWritten = createAggregator("bytesWritten", new SumLongFn());
     }
 
     @Override
-    public void populateDisplayData(DisplayData.Builder builder) {
-      super.populateDisplayData(builder);
+    public void processElement(ProcessContext c) throws Exception {
```
sduskis: This is much clearer.  Thanks for the change.

bigtable-dataflow-parent/bigtable-dataflow-import/src/main/java/com/google/cloud/bigtable/dataflowimport/HBaseExportOptions.java
```diff
@@ -0,0 +1,59 @@
...
+
+import com.google.cloud.bigtable.dataflow.CloudBigtableOptions;
+import com.google.cloud.dataflow.sdk.options.Default;
+import com.google.cloud.dataflow.sdk.options.Description;
+
+/**
+ * An extension of {@link com.google.cloud.bigtable.dataflow.CloudBigtableOptions} that contains additional configuration
+ * for exporting Bigtable tables into GCS as sequence files.
+ *
+ * @author sduskis
```
sduskis: @author igorbernstein

bigtable-dataflow-parent/bigtable-dataflow-import/src/main/java/com/google/cloud/bigtable/dataflowimport/HBaseExportJob.java
```diff
@@ -0,0 +1,288 @@
...
+
+      LOG.info("Starting new shard");
+
+      final String projectId = c.getPipelineOptions().as(DataflowPipelineOptions.class).getProject();
+      final String finalFilePath = basePath + "/" + shard.filename;
+
+      try (BoundedSource.BoundedReader<Result> reader = shard.source.createReader(c.getPipelineOptions())) {
+        boolean hasMore = reader.start();
+
+        if (!hasMore) {
```
sduskis: `if (!reader.start()) { return }`

sduskis:
LGTM

igorbernstein2:
@sduskis this is ready for final review PTAL

PR #1179, opened 2017-01-13
Fix calculation of HadoopFileSource shard size.

igorbernstein2:


bigtable-dataflow-parent/bigtable-dataflow-import/src/main/java/com/google/cloud/bigtable/dataflowimport/HadoopFileSource.java
```diff
@@ -384,19 +384,25 @@ public long getEstimatedSizeBytes(PipelineOptions options) {
     if (isRemoteFileFromLaunchSite) {
       return 0;
     }
+
     long size = 0;
```
sduskis: should `long size ...` be moved to right above `Job job = ...`?

PR #1182, opened 2017-01-24
Fix typo in UnknownHostException

Temikus:
Yes, tiny silly change, I know. It's just been driving me nuts since I see it in logs quite a bit :)


sduskis:
I fixed this with #1183.  Thanks for the find!  I'll release a new 0.9.6-SNAPSHOT with this change.

PR #1195, opened 2017-02-03
Fix batch api behavior to support partial results

igorbernstein2:
This should match hbase behavior

### DIFF
--- a/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
+++ b/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
@@ -290,35 +290,7 @@ public class BatchExecutor {
     if (results == null) {
       results = new Object[actions.size()];
     }
-    Preconditions.checkArgument(results.length == actions.size(),
-        "Result array must have same dimensions as actions list.");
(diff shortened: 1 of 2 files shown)


bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
```diff
@@ -340,8 +317,10 @@ public void batch(List<? extends Row> actions, @Nullable Object[] results)
    */
   public Result[] batch(List<? extends Row> actions) throws IOException {
     try {
-      Result[] results = new Result[actions.size()];
-      batch(actions, results);
+      Object[] resultsOrErrors = new Object[actions.size()];
+      batch(actions, resultsOrErrors);
+      Result[] results = new Result[resultsOrErrors.length];
+      System.arraycopy(resultsOrErrors, 0, results, 0, results.length);
```
sduskis: Will this work if there are Throwables in the resultsOrErrors array?

igorbernstein2: resultsOrErrors can't contain exceptions at this point. batch would've thrown an exception if any action failed

PR #1197, opened 2017-02-05
Allowing override of http transport in credentials factory.

sduskis:
Some users want to set a proxy.  This will allow users to do so.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CredentialFactory.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CredentialFactory.java
@@ -79,16 +79,35 @@ public class CredentialFactory {
   // HTTP transport used for created credentials to perform token-refresh handshakes with remote
   // credential servers. Initialized lazily to move the possibility of throwing
   // GeneralSecurityException to the time a caller actually tries to get a credential.
-  private static HttpTransportFactory httpTransportFactory = new HttpTransportFactory() {
-    @Override
-    public HttpTransport create() {
-      try {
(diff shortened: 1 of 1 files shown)


igorbernstein2:
This feels a bit weird, shouldn't this be configured via BigtableOptions?

sduskis:
We have some caching of credentials.  Allowing this to be set via CredentialsOptions would require us to remove that cache.  This is a stop gap solution.

The best answer here might be JWT credentials, which makes it so that we don't have to do remote calls at all.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CredentialFactory.java
```diff
@@ -79,16 +79,35 @@
...
   // credential servers. Initialized lazily to move the possibility of throwing
   // GeneralSecurityException to the time a caller actually tries to get a credential.
-  private static HttpTransportFactory httpTransportFactory = new HttpTransportFactory() {
-    @Override
-    public HttpTransport create() {
-      try {
-        return GoogleNetHttpTransport.newTrustedTransport();
-      } catch (Exception e) {
-        throw new RuntimeException("Could not create a GoogleNetHttpTransport: " + e.getMessage(), e);
+  private static HttpTransportFactory httpTransportFactory;
```
igorbernstein2: This needs to be volatile as well:

http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

sduskis: I removed the synchronized check as per our offline conversation.

igorbernstein2:
LGTM

PR #1199, opened 2017-02-06
Minor fix to FlatRowAdapter to check for empty flatrows.

sduskis:


igorbernstein2:
I think we should add a comment that this shouldn't happen under normal circumstances, but otherwise LGTM

PR #1200, opened 2017-02-06
BulkRead should execute all filter requests in parallel

igorbernstein2:


bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
```diff
@@ -15,11 +15,13 @@
  */
 package com.google.cloud.bigtable.hbase;
 
+import com.google.cloud.bigtable.grpc.BigtableSessionSharedThreadPools;
```
sduskis: are these imports superfluous?

igorbernstein2: oops, thanks for catching that

PR #1201, opened 2017-02-06
Batch timer is now Info level instead of debug

sduskis:
This will allow batches to be part of the logging that's enabled via
SLF4J.


igorbernstein2:
LGTM

PR #1209, opened 2017-02-09
Catching Throwable instead of IOException for BulkRead.

sduskis:
This will fix some flakiness we've seen with our TestBatch.testBatchDoesntHang() test.


igorbernstein2:
LGTM

PR #1212, opened 2017-02-10
Refactor batch executor

sduskis:
- replacing `Future<List<FlatRow>>` with `Future<FlatRow>` in `BulkRead`
- Adding a transformer that will convert FlatRows into something else in `BulkRead`.  This will improve performance of situations where there are multiple Gets of the same key.


igorbernstein2:
I like the clean up of Future<List<FlatRow>>. But I'm not sure about the transformer...I feel like it mixes concerns in BulkRead. How often do overlapping keys occur in batches?

sduskis:
I'll back out the transformer.   Overlapping keys are possible if a user puts in two equal Gets in the List<Get>.  I guess it's ok for us to do slightly more work if a user does that.

sduskis:
Reverted the transformer change.  PTAL

igorbernstein2:
LGTM

PR #1213, opened 2017-02-10
Adding ByteStringComparator to compare ByteStrings

sduskis:
This is functionality that will be useful for other cases where we need to compare ByteStrings, including splitting BulkRead batches.


igorbernstein2:
LGTM

PR #1217, opened 2017-02-10
Breaking up large batches of gets into smaller batches.

### LINKED ISSUE #1207: Improve performance of batch gets with smaller parallel batches
Right now, batch gets via `Table.get(List<Get> gets)` (a.k.a. batch gets) does a single scan, reads the results, and adds them to the result array one at a time.

Quick Note:

- the Cloud Bigtable ReadRows API allows for an arbitrary set of row ranges and/or row keys.  Responses will be returned in lexicographical order.
- the HBase batch get API is somewhat expensive in terms of client side CPU.  Each Get is transformed to a ReadRowsRequest.  The RowFilter, which contains things like column families and columns to read, is extracted.  

New suggested approach:

1. Convert the list of gets into a list of RowFilter -> rowKeys
2. Given some specified batch size in `BigtableOptions` (which will have a default of 100 or so?), create batches of keys and create 1 ReadRowsRequest per batch.  The keys should be sorted lexicographically for better server-side performance.
3. Run scans  in parallel per request. As results come in, fill in the result Object[] in locations corresponding to the original Get.

sduskis:
More parallel reads should help batch get performance.  Sorting the row keys in lexicographical order will improve the chances of getting efficient lookups of multiple keys against a single tablet.

This should fix #1207

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -450,7 +450,7 @@ public class BigtableSession implements Closeable {
    * @return a {@link com.google.cloud.bigtable.grpc.async.BulkRead} object.
    */
   public BulkRead createBulkRead(BigtableTableName tableName) {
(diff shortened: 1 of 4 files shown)


igorbernstein2:
LGTM

PR #1229, opened 2017-02-24
Fix ordering of ScanAdapter filter chaining

sduskis:
Max versions implies get the top N remaining cells.  Order matters with chaining, and the order was previously broken.  In most cases where a Column Family is configured with a single latest cell this doesn't matter.  In cases where there are multiple cells, and the user wants an earlier cell, the ordering of the chain of filters has to change from

1) max versions filter
2) timestamp filter
3) other filters

to

1) timestamp filter
2) other filters
3) max version filter


igorbernstein2:
LGTM, but I think it could use a test to demonstrate the subtlety.

sduskis:
The test was there from a previous commit.  It's the testGetScannerBeforeTimestamp() on which I removed the @Category(KnownIssue.class) annotation; it failed before, and now it passes.

Issue #1227, opened 2017-02-23
HBaseResultCoder fails to serialize KeyValue instances


gsgalloway:
`HBaseResultCoder` (in bigtable-hbase-dataflow) works well for `Result` instances that use the `RowKey` implementation of hbase's `Cell` interface, but when given a `Result` with hbase's `KeyValue` instead of Bigtable's `RowKey` it corrupts the output `Result`.

I couldn't find where exactly the breakdown occurs, but it seems related to the fact that a call to `RowCell.toString()` causes an ArrayIndexOutOfBoundsException.

A simple test for `RowCell.toString()`:
```java
import com.google.bigtable.repackaged.com.google.cloud.hbase.adapters.read.RowCell;
import org.apache.hadoop.hbase.Cell;
import org.apache.hadoop.hbase.CellUtil;
import org.apache.hadoop.hbase.KeyValue;

@Test
public void testStringifyRowCell() throws Exception {
    Cell keyValue = CellUtil.createCell("key".getBytes(), "family".getBytes(), "value".getBytes());
    Cell rowCell = new RowCell("key".getBytes(), "family".getBytes(), "qualifier".getBytes(), System.currentTimeMillis(), "value".getBytes());

    keyValue.toString();
    rowCell.toString();  // ArrayIndexOutOfBoundsException
}
```
Stack trace from test failure:
```
java.lang.ArrayIndexOutOfBoundsException: 27495

	at org.apache.hadoop.hbase.KeyValue.keyToString(KeyValue.java:1231)
	at org.apache.hadoop.hbase.KeyValue.keyToString(KeyValue.java:1190)
	at com.google.bigtable.repackaged.com.google.cloud.hbase.adapters.read.RowCell.toString(RowCell.java:234)
```

Another test illustrating `HBaseCoder`'s incompatibility with `KeyValue`

```java
import org.apache.hadoop.hbase.Cell;
import org.apache.hadoop.hbase.CellComparator;
import org.apache.hadoop.hbase.CellUtil;
import org.apache.hadoop.hbase.KeyValue;
import org.apache.hadoop.hbase.client.Result;
import com.google.cloud.bigtable.dataflow.coders.HBaseResultCoder;
import java.io.ByteArrayInputStream;
import java.io.ByteArrayOutputStream;
import static org.junit.Assert.assertTrue;


@Test
public void testHBaseResultCoderWithKeyValue() throws Exception {
    // Given
    // -----
[...]

PR #1235, opened 2017-02-28
 Fix FlatRowAdapter to properly handle hbase Result.

### LINKED ISSUE #1231: Bigtable HBase Exports are actually not compatible with HBase
Exports from Bigtable are no longer compatible with HBase.  The culprit is [`RowCell`](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/RowCell.java) which is a performance optimization for Cloud Bigtable scans.  It replaces HBase's [`KeyValue`](https://github.com/apache/hbase/blob/master/hbase-common/src/main/java/org/apache/hadoop/hbase/KeyValue.java) which is a great wire protocol, but caused performance bottlenecks for general Cloud Bigtable scans.

The solution is to translate `RowCell`s to `KeyValue` cells in the export process.  Bonus: have an option in the export process to allow `RowCell` for more performant disaster recovery backups.

igorbernstein2:
This should also fix #1231 & #1227

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/util/ByteStringer.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/util/ByteStringer.java
@@ -54,6 +54,15 @@ public class ByteStringer {
     return USE_ZEROCOPYBYTESTRING? ZeroCopyByteStringUtil.wrap(array): ByteString.copyFrom(array);
   }
 
+  /**
+   * Wraps a byte array in a {@link ByteString} without copying it.
+   */
+  public static ByteString wrap(final byte[] array, int offset, int length) {
+    return USE_ZEROCOPYBYTESTRING?
+        ZeroCopyByteStringUtil.wrap(array, offset, length)
+        : ByteString.copyFrom(array, offset, length);
+  }
+
   public static byte[] extract(ByteString buf) {
     return USE_ZEROCOPYBYTESTRING ? ZeroCopyByteStringUtil.get(buf) : buf
         .toByteArray();
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/util/ByteStringer.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/util/ByteStringer.java
@@ -58,9 +58,12 @@ public class ByteStringer {
    * Wraps a byte array in a {@link ByteString} without copying it.
    */
   public static ByteString wrap(final byte[] array, int offset, int length) {
-    return USE_ZEROCOPYBYTESTRING?
-        ZeroCopyByteStringUtil.wrap(array, offset, length)
-        : ByteString.copyFrom(array, offset, length);
+    if (USE_ZEROCOPYBYTESTRING && offset == 0 && length == array.length) {
+      return ZeroCopyByteStringUtil.wrap(array);
+    }
+    else {
+      return ByteString.copyFrom(array, offset, length);
+    }
   }
 
   public static byte[] extract(ByteString buf) {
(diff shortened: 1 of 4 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/util/ByteStringer.java
```diff
@@ -54,6 +54,15 @@ public static ByteString wrap(final byte[] array) {
     return USE_ZEROCOPYBYTESTRING? ZeroCopyByteStringUtil.wrap(array): ByteString.copyFrom(array);
   }
 
+  /**
+   * Wraps a byte array in a {@link ByteString} without copying it.
+   */
+  public static ByteString wrap(final byte[] array, int offset, int length) {
+    return USE_ZEROCOPYBYTESTRING?
```
igorbernstein2: I think I'm missing something here...
When adapting hbase's Result, isn't it always the case that offset !=0? So this should prevent 3 array copies per cell?

bigtable-hbase-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/adapters/read/TestFlatRowAdapter.java
```diff
@@ -15,21 +15,28 @@
...
 import org.apache.hadoop.hbase.client.Result;
 import org.apache.hadoop.hbase.util.Bytes;
+import org.junit.Assert;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 
 import com.google.cloud.bigtable.grpc.scanner.FlatRow;
 import com.google.protobuf.ByteString;
+import org.omg.CORBA.TypeCode;
```
igorbernstein2: oops

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/FlatRowAdapter.java
```diff
@@ -75,7 +76,8 @@ private static RowCell toRowCell(byte[] rowKey, FlatRow.Cell cell, byte[] family
         // Bigtable timestamp has more granularity than HBase one. It is possible that Bigtable
         // cells are deduped unintentionally here. On the other hand, if we don't dedup them,
         // HBase will treat them as duplicates.
-        cell.getTimestamp() / TIME_CONVERSION_UNIT,
+        cell.getTimestamp() == HConstants.LATEST_TIMESTAMP ?
```
igorbernstein2: done: #1236

igorbernstein2:
rebased to include the timestamp change

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/util/ByteStringer.java
sduskis: When adapting a Result RowCell, `(offset == 0 && length == array.length)` will be true.

The 3 byte[] copies come into play for Puts that are sent via gRPC RPCs.  This case isn't an RPCE, but rather Dataflow's serialization process.  in the normal Dataflow serialization process, this will occur:

1. A scan happens, producing a Result generated from FlatRows.
2. Dataflow may serialize the Result 
3. if 2 occurs, deserialization occurs.

2 and 3 require byte array copies, regardless of whether the `Result` has `RowCells` or `KeyValue`s.  If the `Result` has `RowCell`s, there's no need to copy the `byte[]` as part of the serialization process.  If the `Result` has `KeyValue`s, then we the optimal solution would probably look different, and I don't think that case is something we should tackle in the Dataflow code, since it's not going to happen in production anyway.

igorbernstein2:
Updated, PTAL.

PR #1237, opened 2017-02-28
Optimize prefix filter

igorbernstein2:

PR #1241, opened 2017-03-07
move row prefix calculation into a utility class

igorbernstein2:
This part of #1237, I think the prefix range calculation functionality better fits with the PrefixRowFilter. I don't think its general enough to deserve its own utility class, but I don't feel strongly about it.


bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/PrefixFilterAdapter.java
```diff
@@ -53,4 +52,46 @@ public FilterSupportStatus isFilterSupported(
       FilterAdapterContext context, PrefixFilter filter) {
     return FilterSupportStatus.SUPPORTED;
   }
+
```
sduskis: Is there a reason to put this method here vs. BigtableExtendedScan?  This seems like a utility method, and perhaps should be in com.google.cloud.bigtable.util.

igorbernstein2: Just thought that it made sense to keep prefix range calculations in PrefixFilterAdapter. But I don't have strong feelings...I moved it to its own utility

Issue #1244, opened 2017-03-10
Concurrent modification exceptions using CloudBigtableIO


ankurcha:
While long running (streaming) dataflow job on our production dataset I saw some exceptions in the worker logs (about 200).

```java
Exception: java.util.ConcurrentModificationException
java.util.HashMap$HashIterator.nextNode(HashMap.java:1437)
java.util.HashMap$ValueIterator.next(HashMap.java:1466)
com.google.bigtable.repackaged.com.google.cloud.grpc.async.RpcThrottler.awaitCompletion(RpcThrottler.java:177)
com.google.bigtable.repackaged.com.google.cloud.grpc.async.AsyncExecutor.flush(AsyncExecutor.java:330)
com.google.bigtable.repackaged.com.google.cloud.hbase.BigtableBufferedMutator.flush(BigtableBufferedMutator.java:236)
com.google.bigtable.repackaged.com.google.cloud.hbase.BigtableBufferedMutator.close(BigtableBufferedMutator.java:213)
org.apache.hadoop.hbase.client.AbstractBigtableConnection$2.close(AbstractBigtableConnection.java:181)
com.google.cloud.bigtable.dataflow.CloudBigtableIO$CloudBigtableSingleTableBufferedWriteFn.finishBundle(CloudBigtableIO.java:989)
```

It seems like the `RpcThrottler#awaitCompletion` is being called for multiple bundles at the same time from multiple worker threads. I saw there is a `@GuardedBy("lock")` annotation at the declaration site but looks like that's not catching the condition here.

* Connector version `0.9.5.1`

sduskis:
Ugh.  I'll fix this for 0.9.6.  I'll add a lock around the iteration.

sduskis:
wait a second... this shouldn't happen.  There should only be one call to CloudBigtableSingleTableBufferedWriteFn.finishBundle().  How did you set up that scenario?

sduskis:
I should probably make CloudBigtableSingleTableBufferedWriteFn.finishBundle() final, but I'm really confused about why I need to do that.

garye:
I think `RpcThrottler#awaitCompletion` can handle multiple concurrent calls. But that retry code at line 177 can end up calling into `onRetryCompletion` and removing from `outstandingRetries` which would cause the exception.

4d55397500:
I'm also experiencing this problem with 0.9.5. When will 0.9.6 be released? Until then which stable release is suggested?

PR #1246, opened 2017-03-10
Fixing #1244 by adding some synchronization in RpcThrottler and CloudBigtableIO

sduskis:


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/RpcThrottler.java
```diff
@@ -172,13 +172,18 @@ public void awaitCompletion() throws InterruptedException {
...
-            retryHandler.performRetryIfStale();
-          }
-          if (isFlushed()) {
-            break;
+          lock.lock();
+          try {
+            // There are unusual cases where an RPC could be completed, but we don't clean up
+            // the state and the locks.  Try to clean up if there is a timeout.
+            for (RetryHandler retryHandler : outstandingRetries.values()) {
+              retryHandler.performRetryIfStale();
```
garye: I don't think this PR addresses the problem. There's an extra lock here but the lock was already acquired on line 168 and the lock is reentrant so I don't think this does anything. As best I can tell the problem is the call stack that can happen here (all while holding that lock):
`RetryHandler.performRetryIfStale`
`BulkMutation.Batch#setRetryComplete`
`RpcThrottler#onRetryCompletion`

At this point `outstandingRetries.remove(id)` is called, which will cause a `ConcurrentModificationException` on the next iteration on this line.

If that's all true, then we should make a local copy of `oustandingRetries.values()` (because it's backed by the map: https://docs.oracle.com/javase/7/docs/api/java/util/Map.html#values()) and iterate over that.

bigtable-dataflow-parent/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableIO.java
```diff
@@ -983,10 +983,11 @@ public void processElement(ProcessContext context) throws Exception {
      * Closes the {@link BufferedMutator} and {@link Connection}.
      */
     @Override
-    public void finishBundle(Context context) throws Exception {
+    public synchronized void finishBundle(Context context) throws Exception {
```
garye: I don't think is necessary

sduskis: It also doesn't hurt, so can we keep this?

PR #1253, opened 2017-03-21
Refactoring ChannelPool

sduskis:
InstrumentedChannel is now decoupled from header processing in ChannelPool.  This will allow a future PR to cache data API InstrumentedChannels for reuse across BigtableSession / AbstractBigtableConnection instances, which is a useful feature for Dataflow performance.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingRpcListener.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingRpcListener.java
@@ -166,9 +166,10 @@ public abstract class AbstractRetryingRpcListener<RequestT, ResponseT, ResultT>
     }
 
     // Perform Retry
-    String channelId = trailers != null ? trailers.get(ChannelPool.CHANNEL_ID_KEY) : "";
-    LOG.info("Retrying failed call. Failure #%d, got: %s on channel %s",
-        status.getCause(), failedCount, status, channelId);
+    String channelId =
+        trailers != null ? trailers.get(ChannelPool.InstrumentedChannel.CHANNEL_ID_KEY) : "";
+    LOG.info("Retrying failed call. Failure #%d, got: %s on channel %s", status.getCause(),
+      failedCount, status, channelId);
 
     rpc.getRpcMetrics().markRetry();
     retryExecutorService.schedule(this, nextBackOff, TimeUnit.MILLISECONDS);
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
@@ -16,6 +16,7 @@
 package com.google.cloud.bigtable.grpc.io;
 
 import java.io.IOException;
+import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -53,20 +54,22 @@ public class ChannelPool extends ManagedChannel {
 
   private static final AtomicInteger ChannelIdGenerator = new AtomicInteger();
 
+  public static ImmutableList<InstrumentedChannel>
+      toInstrummentedChannels(List<ManagedChannel> channels) {
+    Preconditions.checkArgument(channels != null && !channels.isEmpty(),
+      "Channels have to be non-null and not empty.");
+    ImmutableList.Builder<InstrumentedChannel> response = ImmutableList.builder();
+    for (ManagedChannel channel : channels) {
+      response.add(new InstrumentedChannel(channel));
+    }
+    return response.build();
+  }
+
   protected static Stats STATS;
 
   /** Constant <code>CHANNEL_ID_KEY</code> */
-  public static final Key<String> CHANNEL_ID_KEY = Key.of("bigtable-channel-id", Metadata.ASCII_STRING_MARSHALLER);
-
-  /**
-   * A factory for creating ManagedChannels to be used in a {@link ChannelPool}.
-   *
-   * @author sduskis
-   *
-   */
-  public interface ChannelFactory {
-    ManagedChannel create() throws IOException;
-  }
+  public static final Key<String> CHANNEL_ID_KEY =
+      Key.of("bigtable-channel-id", Metadata.ASCII_STRING_MARSHALLER);
 
   private static class Stats {
     /**
@@ -99,7 +102,7 @@ public class ChannelPool extends ManagedChannel {
    * @author sduskis
    *
    */
-  private class InstrumentedChannel extends ManagedChannel {
+  public static class InstrumentedChannel extends ManagedChannel {
     private final ManagedChannel delegate;
     // a uniquely named timer for this channel's latency
     private final Timer timer;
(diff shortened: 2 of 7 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -53,20 +54,22 @@
 
   private static final AtomicInteger ChannelIdGenerator = new AtomicInteger();
 
+  public static ImmutableList<InstrumentedChannel>
```
igorbernstein2: This seems like an internal implementation detail of of ChannelPool, might better to keep it private

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -53,20 +54,22 @@
...
+      toInstrummentedChannels(List<ManagedChannel> channels) {
+    Preconditions.checkArgument(channels != null && !channels.isEmpty(),
+      "Channels have to be non-null and not empty.");
+    ImmutableList.Builder<InstrumentedChannel> response = ImmutableList.builder();
+    for (ManagedChannel channel : channels) {
+      response.add(new InstrumentedChannel(channel));
+    }
+    return response.build();
```
igorbernstein2: nit: response is a bit misleading since it doesn't do any io, might be clearer to name it wrappedChannels or results

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -99,7 +102,7 @@ protected static synchronized Stats getStats() {
    * @author sduskis
    *
    */
-  private class InstrumentedChannel extends ManagedChannel {
+  public static class InstrumentedChannel extends ManagedChannel {
```
igorbernstein2: Can we keep InstrumentedChannel private to ChannelPool? or at least package private?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -239,28 +239,35 @@ public String authority() {
...
+  /**
+   * <p>
+   * Constructor for ChannelPool.
+   * </p>
+   * @param headerInterceptors a {@link java.util.List} object.
+   * @param channels a {@link ImmutableList} of {@link InstrumentedChannel}s.
+   * @throws java.io.IOException if any.
+   */
+  public ChannelPool(List<HeaderInterceptor> headerInterceptors,
+      ImmutableList<InstrumentedChannel> channels) throws IOException {
```
igorbernstein2: Similar to above: might be better to keep this as a private implementation detail

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -295,10 +302,20 @@ public String authority() {
...
+        getNextChannel().newCall(methodDescriptor, callOptions)) {
+      @Override
+      protected void checkedStart(ClientCall.Listener<RespT> listener, Metadata headers)
+          throws Exception {
+        for (HeaderInterceptor interceptor : headerInterceptors) {
+          interceptor.updateHeaders(headers);
+        }
+        delegate().start(listener, headers);
+      }
+    };
```
igorbernstein2: Might be clearer to promote this to a static inner class

sduskis:
I addressed the comments, and did a bit more refactoring in `ChannelPool`.  I replaced uses of `InstrumentedChannel` with `ManagedChannel`.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -51,55 +51,61 @@
   /** Constant <code>LOG</code> */
   protected static final Logger LOG = new Logger(ChannelPool.class);
 
-  private static final AtomicInteger ChannelIdGenerator = new AtomicInteger();
-
-  protected static Stats STATS;
-
-  /** Constant <code>CHANNEL_ID_KEY</code> */
-  public static final Key<String> CHANNEL_ID_KEY = Key.of("bigtable-channel-id", Metadata.ASCII_STRING_MARSHALLER);
```
igorbernstein2: Not sure if this would conflict with the follow up refactors, but if we keep CHANNEL_ID_KEY on the ChannelPool then we can make InstrumentedChannel private to the ChannelPool.

sduskis: I'd rather go in the other direction and move `InstrumentedChannel` into its own file.  `ChannelPool` and `InstrumentedChannel` are orthogonal concerns.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingRpcListener.java
```diff
@@ -166,9 +166,10 @@ public void onClose(Status status, Metadata trailers) {
     }
 
     // Perform Retry
-    String channelId = trailers != null ? trailers.get(ChannelPool.CHANNEL_ID_KEY) : "";
-    LOG.info("Retrying failed call. Failure #%d, got: %s on channel %s",
-        status.getCause(), failedCount, status, channelId);
+    String channelId =
+        trailers != null ? trailers.get(ChannelPool.InstrumentedChannel.CHANNEL_ID_KEY) : "";
```
igorbernstein2: Might be a good idea to hide the key extraction behind a function on the channel pool. Something like:

ChannelPool.extractIdentifier(trailers)

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
sduskis: This is a pretty small class.  We don't have a separate class for the `InstrumentedChannel`'s `CheckedForwardingClientCall`.  I don't have a strong feeling about this either way.

sduskis:
I should break this up into smaller PRs.  There's a lot going on here.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
igorbernstein2: Does InstrumentedChannel make sense outside of a ChannelPool? I thought it mainly existed to track the number of connections that are open and their identifiers.

sduskis: You're right `InstrumentedChannel` belongs in `ChannelPool`.  I think that the headers are the problem here, and I was trying to solve for that.  I'll move out the header processing.

PR #1254, opened 2017-03-21
Remove paths to HConstants

sduskis:
Some confluence of events in Dataflow caused a deadlock.  There were some strange stack traces that pointed towards the root cause being the class loading of HConstants.  Scans refer to HConstants.  KeyValues refer to HConstants.  Puts and Adapters -> KeyOnlyFilterAdapter both refer to HContants.

This PR removes a couple of paths to HConstants in hope that this removes the class loading issue.

### DIFF
--- a/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/KeyOnlyFilterAdapter.java
+++ b/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/KeyOnlyFilterAdapter.java
@@ -17,8 +17,9 @@ package com.google.cloud.bigtable.hbase.adapters.filters;
 
 import com.google.bigtable.v2.RowFilter;
 import com.google.bigtable.v2.RowFilter.Chain;
+import com.google.cloud.bigtable.hbase.adapters.read.RowCell;
+
 import org.apache.hadoop.hbase.Cell;
-import org.apache.hadoop.hbase.KeyValue;
 import org.apache.hadoop.hbase.filter.KeyOnlyFilter;
 import org.apache.hadoop.hbase.util.Bytes;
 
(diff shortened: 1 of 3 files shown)


sduskis:
As per our offline discussion, I added a static loading of HConstants.class.  PTAL.

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/KeyOnlyFilterAdapter.java
```diff
@@ -32,7 +33,7 @@
  */
 public class KeyOnlyFilterAdapter extends TypedFilterAdapterBase<KeyOnlyFilter> {
   /** Constant <code>TEST_CELL</code> */
-  protected static final Cell TEST_CELL = new KeyValue(
+  protected static final Cell TEST_CELL = new RowCell(
```
igorbernstein2: This is a bit weird...I don't think we should be sending our own subclass as a probe into hbase's impl of transformCell

bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -73,6 +74,11 @@ public void run() {
       }
     };
     Runtime.getRuntime().addShutdownHook(new Thread(shutDownRunnable));
+
+    // Force the loading of HConstants.class, which shares a bi-directional reference with KeyValue.
+    // This bi-drectional relationship causes problems when KeyVale and HConstants are class loaded
+    // in different threads.  This forces a clean class loading of both classes.
+    HConstants.class.getName();
```
igorbernstein2: I think we should go a step further & make the class loading more deterministic by adding Adapters here instead... No reason to have threads fight over classloading locks

sduskis: The problem is a bidirectional relationship between `KeyValue` and `HConstant`.  Is there a similar bidirectional relationship with `Adapters`?  What does loading `Adapters` achieve?

igorbernstein2: I can't think of another bidirectional dependency cycle. But I think that the static initializer in Adapters has a fairly large graph of dependencies that we will have to be loaded at some point. By eagerly loading Adapters in the user's main thread we are drastically lowering our chances of another deadlock due to multiple threads contending for classloading locks.

sduskis:
I addressed the comments.  I reverted `KeyOnlyFilterAdapter` and now class load the graph via `Adapter.class.getName()`

igorbernstein2:
LGTM

PR #1261, opened 2017-03-22
Add MultiRowRangeFilter support 

igorbernstein2:
Depends on #1260 for the hbase upgrade.

### DIFF
--- a/bigtable-hbase-parent/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestFilters.java
+++ b/bigtable-hbase-parent/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestFilters.java
@@ -1648,6 +1648,43 @@ public class TestFilters extends AbstractTest {
   }
 
   @Test
+  @Category(KnownGap.class)
+  public void testMultiRangeFilterOrList() throws IOException {
+    String prefix = "testMultiRangeFilterOrList";
+    int rowCount = 10;
+    byte[][] rowKeys = dataHelper.randomData(prefix, rowCount);
+    Arrays.sort(rowKeys, Bytes.BYTES_COMPARATOR);
+    List<Put> puts = new ArrayList<>();
+    for (byte[] rowKey : rowKeys) {
+      puts.add(
+          new Put(rowKey)
+              .addColumn(COLUMN_FAMILY, Bytes.toBytes("q1"), Bytes.toBytes("val1")));
+    }
+    Table table = getTable();
+    table.put(puts);
+
+    MultiRowRangeFilter rangeFilter = new MultiRowRangeFilter(Arrays.asList(
+        // rows 1 & 2
+        new RowRange(rowKeys[1], true, rowKeys[3], false)
+    ));
+
+    PrefixFilter prefixFilter = new PrefixFilter(rowKeys[8]);
+
+    FilterList filterList = new FilterList(rangeFilter, prefixFilter);
+
+    Scan scan = new Scan().addFamily(COLUMN_FAMILY).setFilter(filterList);
+    ResultScanner scanner = table.getScanner(scan);
+    Result[] results = scanner.next(rowCount + 2);
+    Assert.assertEquals(3, results.length);
+
+    // first range: rows 1 & 2
+    Assert.assertArrayEquals(rowKeys[1], results[0].getRow());
+    Assert.assertArrayEquals(rowKeys[2], results[1].getRow());
+    // second range: rows 9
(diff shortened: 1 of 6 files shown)


bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/MultiRowRangeFilterAdapter.java
```diff
@@ -0,0 +1,67 @@
+package com.google.cloud.bigtable.hbase.adapters.filters;
```
sduskis: we need the standard apache license header on all files.

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/MultiRowRangeFilterAdapter.java
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Adapter for {@link MultiRowRangeFilter}, it converts the filter into an index scan hint
+ */
+public class MultiRowRangeFilterAdapter extends TypedFilterAdapterBase<MultiRowRangeFilter>  {
+  @Override
+  public RowFilter adapt(FilterAdapterContext context, MultiRowRangeFilter filter)
+      throws IOException {
+
+    return RowFilter.newBuilder()
```
sduskis: What happen in cases where scan hints won't be invoked, like in a must pass one filter list?

sduskis: We somehow need to error out in that case...

bigtable-hbase-parent/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestFilters.java
```diff
@@ -1610,6 +1612,40 @@ public void testPrefixFilter() throws IOException {
       Assert.assertArrayEquals(rowKeys[i+1], boundedResults[i].getRow());
     }
   }
+  @Test
+  public void testMultiRangeFilter() throws IOException {
```
sduskis: What happens when the tests are set to HBase 1.0?

sduskis: Where's the shim?

sduskis:
Overall, I'm concerned about this filter.  There are some corner cases that need to be handled, and there doesn't seem to be a good way to handle them now.  There are some other solutions that solve the same problem that this PR solves, specifically `BigtableExtendedScan`, which better aligns with the Cloud Bigtable model than this implementation, and doesn't have the same corner cases.

I think we should continue working on this, but not include this as part of the 0.9.6 release.

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/MultiRowRangeFilterAdapter.java
igorbernstein2: done

bigtable-hbase-parent/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestFilters.java
igorbernstein2: Oops, added the shim.

sduskis:
I would feel more comfortable with this if this required explicit user action to enable `MultiRowRangeFilterAdapter.java`. 

I would approve this PR if the integration test were marked `KnownGap`, with a comment, and that there were instructions on `MultiRowRangeFilterAdapter` on how to enable it.

In a separate PR, let's make `FilterAdapter.addFilterAdapter()` public so that users can actually enable MultiRowRangeFilterAdapter.

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/MultiRowRangeFilterAdapter.java
igorbernstein2: Good call, I added the limitation to isFilterSupported

PR #1262, opened 2017-03-22
Adding a cache for the data ChannelPool.

sduskis:
This is useful for cases where connections are opened frequently, such and Dataflow scenarios.  Invoking the new cache for CloudBigtableIO, and removing CloudBigtableConnectionPool.


sduskis:
I updated the config to use BigtableOptions instead of a static variable for configuration.  PTAL

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -308,6 +309,20 @@ public BigtableSession(BigtableOptions opts) throws IOException {
     initializeResourceLimiter(options);
   }
 
+  private ChannelPool getDataChannelPool() throws IOException {
+    String host = options.getDataHost();
+    int channelCount = options.getChannelCount();
+    if (options.useCachedChannel()) {
+      synchronized (BigtableSession.class) {
+        if (cachedDataChannelPool == null) {
```
igorbernstein2: Might be a good idea to check if the cachedDataChannelPool was constructed with the same host/channelCount

sduskis: There isn't a quick way to do that (channels don't know their host name), and this is an experimental feature.  I'll add a TODO.

Issue #1274, opened 2017-04-05
Inclusion of google well known proto types makes it impossible to use other services


ankurcha:
The version 0.9.6 bigtable-client package includes all the `com.google.type` classes but remaps the `com.google.protobuf.Descriptors.FileDescriptor` to `repackaged` package. This makes it impossible to use for example datastore in the same project as it may encounter the repackaged version of the class before the "real" version from `grpc-google-common`. 
Causing an error:

```
Caused by: java.lang.NoSuchMethodError: com.google.type.LatLngProto.getDescriptor()Lcom/google/protobuf/Descriptors$FileDescriptor;
	at com.google.datastore.v1.EntityProto.<clinit>(EntityProto.java:107)
	at com.google.datastore.v1.Entity$PropertiesDefaultEntryHolder.<clinit>(Entity.java:160)
	at com.google.datastore.v1.Entity.<init>(Entity.java:67)
	at com.google.datastore.v1.Entity.<init>(Entity.java:16)
	at com.google.datastore.v1.Entity$1.parsePartialFrom(Entity.java:1014)
	at com.google.datastore.v1.Entity$1.parsePartialFrom(Entity.java:1009)
	at com.google.protobuf.CodedInputStream$StreamDecoder.readMessage(CodedInputStream.java:2315)
	at com.google.datastore.v1.EntityResult.<init>(EntityResult.java:56)
	at com.google.datastore.v1.EntityResult.<init>(EntityResult.java:13)
	at com.google.datastore.v1.EntityResult$1.parsePartialFrom(EntityResult.java:922)
	at com.google.datastore.v1.EntityResult$1.parsePartialFrom(EntityResult.java:917)
	at com.google.protobuf.CodedInputStream$StreamDecoder.readMessage(CodedInputStream.java:2315)
	at com.google.datastore.v1.QueryResultBatch.<init>(QueryResultBatch.java:68)
	at com.google.datastore.v1.QueryResultBatch.<init>(QueryResultBatch.java:13)
	at com.google.datastore.v1.QueryResultBatch$1.parsePartialFrom(QueryResultBatch.java:1521)
	at com.google.datastore.v1.QueryResultBatch$1.parsePartialFrom(QueryResultBatch.java:1516)
	at com.google.protobuf.CodedInputStream$StreamDecoder.readMessage(CodedInputStream.java:2315)
[...]

PR #1275, opened 2017-04-05
Relocating common protos in the hbase shading process.

sduskis:
This should be a short term fix for #1274
This PR relocates the following packages:

- com/google/type/
- com/google/rpc/
- com/google/logging/
- com/google/cloud/audit/
- com/google/longrunning/


igorbernstein2:
I think com/google/api should be added as well.
LGTM

sduskis:
Thankfully, com/google/api was there already: https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/bigtable-client-0.9.6/bigtable-hbase-parent/bigtable-hbase-shaded/pom.xml#L135

PR #1280, opened 2017-04-10
next round of refactoring authenticator

igorbernstein2:
* Wrap background refresh in a Future
* Remove canRefresh

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
@@ -246,14 +246,16 @@ public class RefreshingOAuth2CredentialsInterceptor implements ClientInterceptor
         this.futureToken = executor.submit(new Callable<HeaderCacheElement>() {
           @Override
           public HeaderCacheElement call() throws Exception {
-            HeaderCacheElement newToken = refreshCredentialsWithRetry();
-
-            synchronized (RefreshingOAuth2CredentialsInterceptor.this) {
+            try {
+              HeaderCacheElement newToken = refreshCredentialsWithRetry();
               headerCache.set(newToken);
-              futureToken = null;
-              isRefreshing.set(false);
+              return newToken;
+            } finally {
+              synchronized (isRefreshing) {
+                futureToken = null;
+                isRefreshing.set(false);
+              }
             }
-            return newToken;
           }
         });
       }
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
@@ -287,7 +287,6 @@ public class RefreshingOAuth2CredentialsInterceptor implements ClientInterceptor
       try {
         return readerFuture.get(250, TimeUnit.MILLISECONDS);
       } catch (ExecutionException|TimeoutException e) {
-        // Should never happen
         throw new IOException(e.getCause());
       } catch (InterruptedException e) {
         Thread.currentThread().interrupt();
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
@@ -25,6 +25,10 @@ import com.google.cloud.bigtable.config.RetryOptions;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Throwables;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListeningExecutorService;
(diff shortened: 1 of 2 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -220,53 +230,90 @@ protected void checkedStart(Listener<RespT> responseListener, Metadata headers)
...
-      executor.execute(new Runnable() {
-        @Override
-        public void run() {
-          doRefresh();
-        }
-      });
+      SettableFuture<HeaderCacheElement> f = SettableFuture.create();
+      try {
+        f.set(syncRefresh());
+      } catch (Exception e) {
```
sduskis: Throwable instead of Exception

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -220,53 +230,90 @@ protected void checkedStart(Listener<RespT> responseListener, Metadata headers)
...
-  }
 
-  private boolean canRefresh() {
-    return !isRefreshing.get() && getCacheState(this.headerCache.get()) != CacheState.Good;
+    synchronized (this.isRefreshing) {
+      if (!this.isRefreshing.compareAndSet(false, true)) {
+        this.futureToken = executor.submit(new Callable<HeaderCacheElement>() {
+          @Override
+          public HeaderCacheElement call() throws Exception {
+            HeaderCacheElement newToken = refreshCredentialsWithRetry();
```
sduskis: Do we need a try-catch here to unset isRefreshing?

igorbernstein2: I don't think so, refreshCredentialsWithRetry catches all exceptions.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
igorbernstein2: I'm sure that would be good. I don't think we want to proxy Errors to user code.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
sduskis: we currently have this in `doRefresh()`:

 ```
   } finally {
      synchronized (isRefreshing) {
        isRefreshing.set(false);
        isRefreshing.notifyAll();
      }
    }
```
Can you please add that to the Callable if there's no harm in doing so?

File: bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java (hunk 1 above)
Comment (sduskis) [inline bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java]: ok.  I buy that.

File: bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java [hunk 3]
```diff
@@ -220,53 +230,89 @@ protected void checkedStart(Listener<RespT> responseListener, Metadata headers)
...
+    ListenableFuture<HeaderCacheElement> readerFuture = null;
+    SettableFuture<HeaderCacheElement> writerFuture = null;
+
+    // Check if another thread is already refreshing and set the appropriate future
+    synchronized (this.isRefreshing) {
+      if (this.isRefreshing.get()) {
+        readerFuture = this.futureToken;
+      }
+      else {
+        writerFuture = SettableFuture.create();
```
sduskis: Is there a reason for sync and async not to share logic around creating Futures?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
igorbernstein2: I added the finally block. But I think notifyAll is out of place here

sduskis: I get it now :).  The `readerFuture.get(250, TimeUnit.MILLISECONDS); ` will deliver the "notification"

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
igorbernstein2: I couldnt figure out a way to share the logic without causing sync stale refreshes from blocking reads. I'm planning on proposing eliminating the sync  path altogether and just wait on the async future for expired tokens in a follow up CL.

PR #1283, opened 2017-04-10
Reduce deadlocking scenarios between RpcThrottler and BulkMutation

sduskis:
- RpcThrottler.onRetryCompletion() is non-locking if the id is not in outstandingRetries.
- Reduced the type of calls from BulkMutation to RpcThrottler that can potentially cause deadlocking between the two classes.


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/RpcThrottler.java
```diff
@@ -258,6 +258,9 @@ void onRpcCompletion(long id) {
    * @param id a long.
    */
   public void onRetryCompletion(long id) {
+    if (!outstandingRetries.containsValue(id)) {
+      return;
+    }
```
igorbernstein2: I don't think this is threadsafe

sduskis:
I moved the check to see if the value was in the collection into the section guarded by the lock.

PR #1284, opened 2017-04-10
Unlock when fixing staleness in RpcThrottler.

sduskis:


sduskis:
This should work to fix some of the problems in #1289

PR #1285, opened 2017-04-11
WIP: refactor authentication

igorbernstein2:
Still a work in progress, just requesting feedback.  The major change here is that authentication retries are now handled by the general retry mechanism. This was a necessary change to be able to handle tokens that were revoked before their expiration. 

* bubble retries to RetryingListener
* run all token refreshes in a background thread (even appengine)
* cimplify OAuth interceptor by using futures & avoiding throwing exceptions
* handle tokens that were revoked serverside

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/RetryOptions.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/RetryOptions.java
@@ -54,7 +54,8 @@ public class RetryOptions implements Serializable {
       Status.Code.DEADLINE_EXCEEDED,
       Status.Code.INTERNAL,
       Status.Code.UNAVAILABLE,
-      Status.Code.ABORTED);
+      Status.Code.ABORTED,
+      Code.UNAUTHENTICATED);
 
   /** We can timeout when reading large cells with a low value here. With a 10MB
    * cell limit, 60 seconds allows our connection to drop to ~170kbyte/s. A 10 second
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/CredentialInterceptorCache.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/CredentialInterceptorCache.java
@@ -110,11 +110,8 @@ public class CredentialInterceptorCache {
     // The RefreshingOAuth2CredentialsInterceptor uses the credentials to get a security token that
     // will live for a short time.  That token is added on all calls by the gRPC interceptor to
     // allow users to access secure resources.
-    //
-    // Perform that token lookup asynchronously. This permits other work to be done in
-    // parallel. The RefreshingOAuth2CredentialsInterceptor has internal locking that assures that
-    // the oauth2 token is loaded before the interceptor proceeds with any calls.
-    oauth2Interceptor.asyncRefresh();
+    oauth2Interceptor.syncRefresh();
+
     if (isDefaultCredentials) {
       defaultCredentialInterceptor = oauth2Interceptor;
     }
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/CredentialInterceptorCache.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/CredentialInterceptorCache.java
@@ -104,8 +104,7 @@ public class CredentialInterceptorCache {
             credentials.getClass().getName()));
 
     RefreshingOAuth2CredentialsInterceptor oauth2Interceptor =
-        new RefreshingOAuth2CredentialsInterceptor(executor, (OAuth2Credentials) credentials,
-            retryOptions);
+        new RefreshingOAuth2CredentialsInterceptor(executor, (OAuth2Credentials) credentials);
 
     // The RefreshingOAuth2CredentialsInterceptor uses the credentials to get a security token that
     // will live for a short time.  That token is added on all calls by the gRPC interceptor to
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
@@ -33,14 +33,12 @@ import io.grpc.CallOptions;
 import io.grpc.Channel;
 import io.grpc.ClientCall;
 import io.grpc.ClientInterceptor;
-import io.grpc.ClientInterceptors.CheckedForwardingClientCall;
+import io.grpc.ForwardingClientCall.SimpleForwardingClientCall;
 import io.grpc.Metadata;
 import io.grpc.MethodDescriptor;
 import io.grpc.Status;
-import io.grpc.StatusException;
 import io.grpc.StatusRuntimeException;
 import java.io.IOException;
-import java.util.Date;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
@@ -48,7 +46,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
-import javax.annotation.Nullable;
 
 /**
  * Client interceptor that authenticates all calls by binding header data provided by a credential.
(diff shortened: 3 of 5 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/RetryOptions.java
```diff
@@ -54,7 +54,8 @@
       Status.Code.DEADLINE_EXCEEDED,
       Status.Code.INTERNAL,
       Status.Code.UNAVAILABLE,
-      Status.Code.ABORTED);
+      Status.Code.ABORTED,
```
sduskis: These changes seem like a useful feature outside of any client auth interceptor changes.  Can you please make a PR for just this change + tests?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/CredentialInterceptorCache.java
```diff
@@ -104,17 +104,13 @@ public synchronized ClientInterceptor getCredentialsInterceptor(
...
 
     // The RefreshingOAuth2CredentialsInterceptor uses the credentials to get a security token that
     // will live for a short time.  That token is added on all calls by the gRPC interceptor to
     // allow users to access secure resources.
-    //
-    // Perform that token lookup asynchronously. This permits other work to be done in
-    // parallel. The RefreshingOAuth2CredentialsInterceptor has internal locking that assures that
-    // the oauth2 token is loaded before the interceptor proceeds with any calls.
-    oauth2Interceptor.asyncRefresh();
+    oauth2Interceptor.syncRefresh();
```
sduskis: The async refresh was there as a performance boost for the creation of a BigtableSession.  Why do we need to remove this?  What are the performance impacts of making this a synchronous call?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/RetryOptions.java
igorbernstein2: I'm a bit concerned about having 2 levels of retries. Won't that cause excessive retries? Auth will retry x times and the retry listener will retry y times ending up with x * y times?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/CredentialInterceptorCache.java
igorbernstein2: I think this already done in its own thread on startup and I'm concerned about it executing outside of the request thread in appengine

sduskis: it's not done on startup.  App engine always needs to do this synchronously, but that doesn't mean that we should reduce performance elsewhere for the 80%+ case, does it?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/RetryOptions.java
sduskis: There are two types of retries.  CBT RPC retries and auth retries.  Those should be orthogonal concerns.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -76,339 +76,296 @@
...
+  private HeaderCacheElement getHeader() throws ExecutionException, InterruptedException {
+    final Future<HeaderCacheElement> deferredResult;
+
+    synchronized (lock) {
+      CacheState state = headerCache.getCacheState();
+
+      if (state == CacheState.Good || state == CacheState.Exception) {
+        return headerCache;
+        // AppEngine doesn't allow threads to outlive the request, so disable background refresh
+      } else if (!isAppEngine && state == CacheState.Stale) {
```
sduskis: I think that `isAppEngine && state == CacheState.Stale` should do a synchRefresh().

sduskis: Currently, `isAppEngine && stale` will be  Status.UNAUTHENTICATED, which seems wrong.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -76,339 +76,296 @@
...
-      } catch (Exception e) {
-        logger.warn("Got an unexpected exception while trying to refresh google credentials.", e);
-        return new HeaderCacheElement(new IOException("Could not read headers", e));
+        // Update the token only if the new token is good or the old token is bad
+        CacheState newState = newToken.getCacheState();
+        boolean newTokenOk = newState == CacheState.Good || newState == CacheState.Stale;
+        CacheState oldCacheState = headerCache.getCacheState();
+        boolean oldTokenOk = oldCacheState == CacheState.Good || oldCacheState == CacheState.Stale;
+
+        if (newTokenOk || !oldTokenOk) {
```
sduskis: Shouldn't a newer token that's good always over-write an older token?  Wouldn't it have a later expiration?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -76,339 +76,296 @@
...
-            throw new IOException(e);
-          }
-        }
+        return new HeaderCacheElement(
+            Status.UNAUTHENTICATED
+                .withCause(new IllegalStateException("Could not process state: " + state))
+        );
       }
     }
+    return deferredResult.get();
```
sduskis: This should timeout after some amount of time, probably 250 ms. like the sync refresh.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -76,339 +76,296 @@
...
-      }
-      isRefreshing.set(true);
-    }
-    try {
-      HeaderCacheElement cacheElement = refreshCredentialsWithRetry();
-      synchronized (isRefreshing) {
-        headerCache.set(cacheElement);
+  ListenableFuture<HeaderCacheElement> asyncRefresh() {
+    synchronized (lock) {
+      if (refreshing) {
```
sduskis: Can we get rid of `refreshing`, and use `futureToken != null`?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -76,339 +76,296 @@
...
+              .withDescription("Authentication rate limit has been exceeded, failing fast")
+      );
+    }
+
     try {
-      nextBackOffMillis = backoff.nextBackOffMillis();
-    } catch (IOException e) {
-      // Should never happen since we use an ExponentialBackoff
-      throw Throwables.propagate(e);
+      LOG.info("Refreshing the OAuth token");
```
sduskis: this should be a `debug`

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
igorbernstein: That is a mistake, but I think it should be Status.GOOD....no reason to cut the expiration short for app engine

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
igorbernstein: I'd prefer to keep it explicit

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
igorbernstein: A new good or stale token will always overwrite the old token. This is mainly to prevent a new pre-emptive bad token from expiring a stale old token.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/RetryOptions.java
igorbernstein: I think keeping retries in the retry listener makes this code a lot easier to reason about, especially with the addition of serverside token revocations. I don't think having 2 separate sets of retry loops adds a lot of value.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -76,339 +67,329 @@
...
-   * <p>syncRefresh.</p>
-   *
-   * @throws java.io.IOException if any.
+   * Get the http credential header we need from a new oauth2 AccessToken.
    */
-  public void syncRefresh() throws IOException {
-    synchronized (isRefreshing) {
-      if (!isRefreshing.get()) {
-        doRefresh();
+  private HeaderCacheElement getHeader()
```
sduskis: Why do we need to incur the lock penalty when CacheState == CacheState.Good?  Can we optimize for that situation?

igorbernstein2:
closing in favor of the broken out PRs #1290 #1291 #1292

PR #1290, opened 2017-04-13
Auth refactor 1: minor

igorbernstein2:
This part 1 of the #1285. This should have no behavioral changes.

* Only store real expiration time in HeaderCacheElement & compute derived
* use explicit lock object & refresh flag instead of AtomicBoolean
* rearrange method order to make it easier to read the class top down
* update visibility
* reformat code

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
@@ -194,11 +194,8 @@ public class RefreshingOAuth2CredentialsInterceptor implements ClientInterceptor
     this.retryOptions = Preconditions.checkNotNull(retryOptions);
     this.logger = Preconditions.checkNotNull(logger);
 
-    // com.google.auth.oauth2.AppEngineCredentials is package private, so we don't have direct
-    // visibility to it. This is a way to detect that this application runs on App Engine so that
-    // we can always get credentials synchronously, which must be done for AppEngineCredentials
-    // which relies on a ThreadLocal.
-    this.isAppEngine = credentials.getClass().getName().contains("AppEngineCredentials");
+    // From MoreExecutors
+    this.isAppEngine = System.getProperty("com.google.appengine.runtime.environment") != null;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -190,14 +179,13 @@ public RefreshingOAuth2CredentialsInterceptor(ExecutorService scheduler,
     this.retryOptions = Preconditions.checkNotNull(retryOptions);
     this.logger = Preconditions.checkNotNull(logger);
 
-    // com.google.auth.oauth2.AppEngineCredentials is package private, so we don't have direct
-    // visibility to it. This is a way to detect that this application runs on App Engine so that
-    // we can always get credentials synchronously, which must be done for AppEngineCredentials
-    // which relies on a ThreadLocal.
-    this.isAppEngine = credentials.getClass().getName().contains("AppEngineCredentials");
+    // From MoreExecutors
+    this.isAppEngine = System.getProperty("com.google.appengine.runtime.environment") != null;
```
sduskis: Could this be a static field?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -285,12 +225,58 @@ String getHeader()  {
...
+      executor.execute(new Runnable() {
+        @Override
+        public void run() {
+          doRefresh();
+        }
+      });
+    }
+  }
+
+  private boolean canRefresh() {
```
sduskis: There should be a comment here about why there is a lack of synchronization.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
igorbernstein2: yep, updated

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
igorbernstein2: oops, it does need synchronization

Issue #1223, opened 2017-02-17
Problem doing a read with 0.9.5.1 version on App Engine.


vc13819:
I am unable to do reads when using 0.9.5.1 version. The code works fine with 0.9.4 though.

Attached is the [code](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/files/783879/backend.zip) which reproduces the problem. For most part, this is adapted from the gae-flex example. A connection to the bigtable is created using a servletcontext weblistener. The connection is then used in a servlet to do a scan.
The error I get is that is
The root cause appears to be
"Caused by: java.lang.IllegalStateException: Request threads can only be created within the context of a running request".

The entire stacktrace.

22:39:07.288
com.yawasa.server.ReadHandler doPost: java.lang.RuntimeException: com.google.cloud.bigtable.grpc.scanner.BigtableRetriesExhaustedException: Exhausted streaming retries.
	at org.apache.hadoop.hbase.client.AbstractClientScanner$1.hasNext(AbstractClientScanner.java:97)
	at com.yawasa.server.ReadHandler.doPost(ReadHandler.java:31)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:707)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)
	at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:821)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1685)
	at com.google.apphosting.utils.servlet.ParseBlobUploadFilter.doFilter(ParseBlobUploadFilter.java:125)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1668)
	at com.google.apphosting.runtime.jetty9.SaveSessionFilter.doFilter(SaveSessionFilter.java:51)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1668)
	at com.google.apphosting.utils.servlet.TransactionCleanupFilter.doFilter(TransactionCleanupFilter.java:50)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1668)
	at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:581)
[...]

PR #1291, opened 2017-04-14
Auth refactor 2: futures

igorbernstein2:
This PR depends on #1290 and includes it. See here for actual differences:
https://github.com/igorbernstein2/cloud-bigtable-client/compare/auth-refactor-minor...igorbernstein2:auth-refactor-futures?expand=0

This PR uses Futures to manage background auth refreshes & uses grpc statuses instead of exceptions to communicate failure

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
@@ -36,8 +36,10 @@ import io.grpc.Status;
 import io.grpc.StatusException;
 import io.grpc.StatusRuntimeException;
 import java.io.IOException;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.Future;
 import javax.annotation.concurrent.GuardedBy;
 
 /**
@@ -153,11 +155,14 @@ public class RefreshingOAuth2CredentialsInterceptor implements ClientInterceptor
 
   @VisibleForTesting
   @GuardedBy("lock")
-  final AtomicReference<HeaderCacheElement> headerCache = new AtomicReference<>();
+  HeaderCacheElement headerCache = null;
 
   @VisibleForTesting
   @GuardedBy("lock")
   boolean isRefreshing = false;
+
+  @GuardedBy("lock")
+  private Future<HeaderCacheElement> futureToken = null;
 
 
   /**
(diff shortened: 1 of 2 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -190,150 +192,139 @@ public RefreshingOAuth2CredentialsInterceptor(ExecutorService scheduler,
...
+
+    synchronized (lock) {
+      CacheState state = headerCache.getCacheState();
+
+      // AppEngine doesn't allow threads to outlive the request, so disable background refresh
+      if (isAppEngine && state == CacheState.Stale) {
+        state = CacheState.Good;
+      }
+
+      if (state == CacheState.Good || state == CacheState.Exception) {
```
sduskis: We need to try to get a new header when `state == CacheState.Exception`.  How would that happen normally?  How would it happen on AppEngine?

igorbernstein2: Added a background refresh

igorbernstein2:
Rebased, this is ready for review

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -187,155 +202,139 @@ public RefreshingOAuth2CredentialsInterceptor(ExecutorService scheduler,
...
-          break;
+          return headerCache;
         case Expired:
-          syncRefresh();
-          headerCache = getCachedHeader();
-          break;
         case Exception:
-          asyncRefresh();
-          throw asUnauthenticatedException(headerCache.exception);
+          // defer the future resolution (asyncRefresh will spin up a thread that will try to acquire the lock)
```
sduskis: What should AppEngine do for Exception cases?  Should it behave like Expired?

sduskis:
I think that we should drop direct support for AppEngine due to #1223, and add a TODO.  There are some other issues that need to be worked out first before we can even confirm that our client auth refresh works well.  

That should help us (or at least me) be able to better reason about the changes we need to make to the client auth interceptor.

Once we get through the core use case, we can come back and figure out how to make AppEngine work

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -187,155 +202,139 @@ public RefreshingOAuth2CredentialsInterceptor(ExecutorService scheduler,
...
-        updateToken(cacheElement);
-      }
-    } finally {
-      synchronized (lock) {
+
+      try {
+        this.futureToken = executor.submit(new Callable<HeaderCacheElement>() {
+          @Override
+          public HeaderCacheElement call() throws Exception {
+            HeaderCacheElement newToken = refreshCredentialsWithRetry();
```
sduskis: refreshCredentialsWithRetry() can throw a RuntimeException.  Please add a try-finally clause to set `isRefreshing = false`.  (see `getRetryState()`)

igorbernstein2: Thats no longer the case, refreshCredentialsWithRetry() catches all of its Exceptions now & converts them to HeaderCacheElements

sduskis: Can you please fix the handling of getRetryState()'s `throw Throwables.propagate(e);` case?  I'd feel more comfortable with that change.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -187,155 +199,135 @@ public RefreshingOAuth2CredentialsInterceptor(ExecutorService scheduler,
...
-          + "New token state: {}", newState);
-      return headerCache.get();
+    synchronized (lock) {
+      // Update the token only if the new token is good or the old token is bad
+      CacheState newState = newToken.getCacheState();
+      boolean newTokenOk = newState == CacheState.Good || newState == CacheState.Stale;
+      CacheState oldCacheState = headerCache.getCacheState();
+      boolean oldTokenOk = oldCacheState == CacheState.Good || oldCacheState == CacheState.Stale;
+
+      if (newTokenOk || !oldTokenOk) {
```
sduskis: Why do we use the old token if it's stale?

igorbernstein2: Unless I made a mistake, we would only use the old stale token if the new token is an error, which allows us to continue making requests.

sduskis: I keep on getting tripped up on this.  Sorry about that.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
igorbernstein2: Ahh, I see what you mean now.

PR #1293, opened 2017-04-14
Reorganizing tests relating to netty recycler.

sduskis:


igorbernstein2:
LGTM. Out of curiosity, what issue was the recycler causing us?

PR #1294, opened 2017-04-14
Removing ALPN logging

sduskis:
The ALPN logging was needed when we were going through a transition from alpn_boot -> tcnative / OpenSSL -> tcnative / BoringSSL.  Tcnative is our prerferred solution.  This logging is unnecessary and unnecessarily digs into the internals of grpc.


igorbernstein2:
LGTM

PR #1295, opened 2017-04-14
Use the ManagedChannelBuilder public API.

sduskis:
Ideally, we shouldn't care about NettyChannelBuilder vs. OkHttpChannelBuilder or any other specific implementation detail.  All we should care about is using ManagedChannelBuilder.  For now, use the ManagedChannelBuilder's public methods rather than resorting to NettyChannelBuilder's custom extensions.  We needed those extensions back in the alpn_boot days, but don't really need them anymore.  We can't use ManagedChannelBuilder.forAddress() yet, because it doesn't seem to work in shaded environments.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -55,14 +55,10 @@ import com.google.common.collect.ImmutableList.Builder;
 
 import io.grpc.ClientInterceptor;
 import io.grpc.ManagedChannel;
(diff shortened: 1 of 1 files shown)


igorbernstein2:
What was the shading issue?

sduskis:
Here's the exception I get when I use ManagedChannelBuilder:

> Caused by: java.lang.reflect.InvocationTargetException
> Caused by: com.google.bigtable.repackaged.io.grpc.ManagedChannelProvider$ProviderNotFoundException: No functional channel service provider found. Try adding a dependency on the grpc-okhttp or grpc-netty artifact
> 

[Here](https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/ManagedChannelProvider.java#L51) is the relevant class.

PR #1299, opened 2017-04-16
Readding sslContext creation to BigtableSession.

sduskis:
Standalone uses of BigtableSession were somehow failing.  Readding creation of the sslContext.


sduskis:
Merging without review to fix some failing tests.

PR #1302, opened 2017-04-24
Adding the ability to set a json string for credentials.

### LINKED ISSUE #1206: How to authenticate and get a Connection instance, from credentials in a JSON string?
I am working on accessing Bigtable from Cloud Foundry, using a new version of the [GCP service broker](https://github.com/GoogleCloudPlatform/gcp-service-broker).  When an application is bound to an instance of this service,  credentials are passed in via an environment variable, and the service account JSON credentials are accessible here.  When using PubSub, we can get a PubSub client like this:

```
PubSub pubSub = PubSubOptions.newBuilder().setProjectId(projectId).setCredentials(cred).build().getService();
```

where `cred` comes from an InputStream from our JSON string:

```
Credentials cred = ServiceAccountCredentials.fromStream(in);
```

This is handy in such an environment, and we have been able to find similar, although not identical, approaches in the following Google Cloud Platform APIs: Vision, BigQuery, and Storage.  So far, I don't see how we can do this using the Bigtable Java client API, and am just seeking advice on this.

sduskis:
This should fix #1206.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CredentialOptions.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CredentialOptions.java
@@ -15,8 +15,10 @@
  */
 package com.google.cloud.bigtable.config;
 
+import java.io.ByteArrayInputStream;
 import java.io.InputStream;
 import java.io.Serializable;
+import java.nio.charset.StandardCharsets;
 import java.util.Objects;
 
 import com.google.auth.Credentials;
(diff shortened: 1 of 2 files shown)


mahmoudhanafy:
@sduskis How can I get this feature ?
I tried to use version `0.9.6.2` but it still not there.

sduskis:
@mahmoudhanafy You need to use 0.9.7-SNAPSHOT.   See [here](https://github.com/GoogleCloudPlatform/cloud-bigtable-examples/blob/master/java/hello-world/pom.xml#L41) for some tips on using the snapshot repository.

PR #1307, opened 2017-04-24
Refactoring AbstractRetryingRpcListener

sduskis:
- It no longer implements Runnable.
- It now provideds methods to get a result synchronously or asynchronously.
- Move rpc.start() into a synchronzied block.


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingRpcListener.java
```diff
@@ -174,7 +175,16 @@ public void onClose(Status status, Metadata trailers) {
...
+    retryExecutorService.schedule(getRunnable(), nextBackOff, TimeUnit.MILLISECONDS);
+  }
+
+  protected Runnable getRunnable() {
+    return new Runnable() {
+      @Override
+      public void run() {
+        AbstractRetryingRpcListener.this.run();
+      }
+    };
```
igorbernstein2: If we are supporting java 8 only, then this can be simplified to 
```java
retryExecutorService.schedule(this::run, nextBackOff, TimeUnit.MILLISECONDS);
```

PR #1314, opened 2017-05-01
Moving ScanIterator to batch/common.

sduskis:
This is part of an effort to move all of the Cloud Bigtable operations into batch/common.  This refactor is part of an effort to create a Beam connector.

### DIFF
--- a/bigtable-dataflow-parent/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/batch/common/ScanIterator.java
+++ b/bigtable-dataflow-parent/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/batch/common/ScanIterator.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.batch.common;
+
+import java.io.IOException;
+import java.io.Serializable;
+
+import org.apache.hadoop.hbase.client.Result;
+
+import com.google.bigtable.repackaged.com.google.cloud.grpc.scanner.FlatRow;
+import com.google.bigtable.repackaged.com.google.cloud.grpc.scanner.ResultScanner;
+
+/**
+ * Performs a {@link ResultScanner#next()} or {@link ResultScanner#next(int)}.  It also checks if
+ * the ResultOutputType marks the last value in the {@link ResultScanner}.
+ *
+ * @param <ResultOutputType> is either a {@link Result} or {@link Result}[];
+ */
+public interface ScanIterator<ResultOutputType> extends Serializable {
+  /**
+   * Get the next unit of work.
+   * @param resultScanner The {@link ResultScanner} on which to operate.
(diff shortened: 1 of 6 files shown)


bigtable-dataflow-parent/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/batch/common/SplitTracker.java
```diff
@@ -0,0 +1,26 @@
...
+
+import com.google.bigtable.repackaged.com.google.protobuf.ByteString;
+
+/**
+ * Determines whether or not a split occurred before a rowKey.
+ *
+ */
+public interface SplitTracker {
+  public boolean isAfterSplit(ByteString rowKey);
+}
```
igorbernstein2: Will we have multiple implementations of this? If not, I'm not sure the abstraction is worth it

bigtable-dataflow-parent/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/batch/common/ScanIteratorFactory.java
```diff
@@ -0,0 +1,109 @@
...
+        }
+        results.add(FLAT_ROW_ADAPTER.adaptResponse(row));
+      }
+      return results.toArray(new Result[results.size()]);
+    }
+
+
+    @Override
+    public boolean isCompletionMarker(Result[] result) {
+      return result == null || result.length == 0;
```
igorbernstein2: When will result be null? (same for below)

bigtable-dataflow-parent/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/batch/common/SplitTracker.java
sduskis: There will be 2 implementations: one for beam, one for Dataflow.

bigtable-dataflow-parent/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/batch/common/ScanIteratorFactory.java
```diff
@@ -0,0 +1,109 @@
...
+      return result == null ? 0 : result.length;
+    }
+  }
+
+  public static ScanIterator<Result> getSingleResultIterator() {
+    return RESULT_ADVANCER;
+  }
+
+  public static ScanIterator<Result[]> getMultiResultIterator(int resultCount) {
+    return new ResultArrayIterator(resultCount);
```
sduskis: We'd have to discuss how to do this with a Source / Reader combination.  They work differently than plain DoFns.

bigtable-dataflow-parent/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/batch/common/ScanIteratorFactory.java
sduskis: scanner.next() can be null if there are no results.

bigtable-dataflow-parent/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/batch/common/ScanIterator.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @param <ResultOutputType> is either a {@link Result} or {@link Result}[];
+ */
+public interface ScanIterator<ResultOutputType> extends Serializable {
+  /**
+   * Get the next unit of work.
+   * @param resultScanner The {@link ResultScanner} on which to operate.
+   * @param splitTracker The {@link SplitTracker} that determines if a split occurred.
+   */
+  ResultOutputType next(ResultScanner<FlatRow> resultScanner, SplitTracker splitTracker)
```
sduskis: The ScanIterator is "configuration" that's created in the vm that creates the pipeline.  The ResultScanner and SplitTracker is created per bundle.

PR #1320, opened 2017-05-15
Fixes to BigtableClusterUtilities

sduskis:
- fixed a problem with zoneId extraction
- returning Cluster from setClusterSize()
This is intended to be used in a Java autoscaler example


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableClusterUtilities.java
```diff
@@ -124,12 +125,14 @@ public static Cluster lookupCluster(String projectId, String instanceId)
...
-    String key = "/locations/";
-    int index = name.lastIndexOf(key) + key.length() + 1;
-    return name.substring(index);
+    return getZoneId(cluster.getLocation());
   }
 
+  @VisibleForTesting
+  static String getZoneId(String name) {
+    final String prefix = "/locations/";
+    return name.substring(name.lastIndexOf(prefix) + prefix.length());
```
igorbernstein2: nit: Might be better to extract this to BigtableLocationName?

PR #1322, opened 2017-05-17
Do not shade instrumentation and io.grpc.Context

bogdandrutu:
io.grpc.Context is a global state (per-thread context propagation) that cannot be shaded in order to allow context propagation from the application.

com.google.instrumentation:instrumentation-api uses reflection to load the implementation and shading this package will create a different library that cannot load the implementation. One of the biggest problem is the interaction with the io.grpc.Context, you cannot get the same Context.Key (even if the same string is used) so if this library is shaded then interaction between libraries via the io.grpc.Context will be broken.

### DIFF
--- a/bigtable-hbase-parent/bigtable-hbase-shaded/pom.xml
+++ b/bigtable-hbase-parent/bigtable-hbase-shaded/pom.xml
@@ -116,7 +116,6 @@ limitations under the License.
                                     <include>com.google.auth:*</include>
                                     <include>com.google.guava:guava</include>
                                     <include>com.google.http-client:*</include>
-                                    <include>com.google.instrumentation:instrumentation-api</include>
                                     <include>com.google.oauth-client:*</include>
                                     <include>com.google.protobuf:*</include>
                                     <include>com.twitter:hpack</include>
@@ -125,6 +124,9 @@ limitations under the License.
                                     <include>io.dropwizard.metrics:metrics-core</include>
 
                                 </includes>
+                                <excludes>
+                                    <exclude>io.grpc:grpc-context</exclude>>
+                                </excludes>
                             </artifactSet>
                             <relocations>
                                 <relocation>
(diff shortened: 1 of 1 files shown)


bogdandrutu:
cc/ @ejona86 @zhangkun83 from gRPC

sduskis:
We have to support a variety of situations that require shading, including Dataflow v1, which has a dependency on grpc 0.13.0.  We would like to support tracing even in Dataflow v1, if possible.

I do understand that shading is a major source of problems.  For now, we can't submit this PR as is.

ejona86:
> com.google.instrumentation:instrumentation-api uses reflection to load the implementation and shading this package will create a different library that cannot load the implementation.

You would need a shaded implementation as well, so that it will match the expected name.

bogdandrutu:
@ejona86  you are correct but the problem is that we need multiple libraries (bigtable and the main application for example) to communicate to each other in the same binary. If bigtable library shades the implementation then the application must use the shaded API in order to pass a Span to bigtable otherwise is not going to work.

Same problem you have with grpc.Context if I set a deadline (or any other property) in my application bigtable library will not be able to read it because of this problem.

ejona86:
Shading is incompatible with those use cases. You have to either resolve the dependency problems to remove the need to shade, shade the other library instead of this one, or give up on the features.

The instrumentation library is one of the most important things to shade, since it is the least API stable. io.grpc.Context could be fine to not shade, except that we only split Context into a separate artifact in 1.0.1.

sduskis:
Given @ejona86 comments, it sounds like we should not merge this PR. @bogdandrutu are you ok with that?

sduskis:
We have plans to change around our shading logic over the next couple of weeks.  In most cases that matter, users will not use shaded Cloud Bigtable artifacts.  The only exceptions will be Dataflow v1 and Hadoop.  If users ask for propagation of Spans for those environments, we'll figure something out later.

PR #1327, opened 2017-05-18
Refactor ResoureLimiterStats

sduskis:
- All touch-points of resource limiter stats relate to `BulkMutation`, so all uses were move to BulkMutation.  I renamed `ResourceLimiterStats` to `BulkMutationsStats` to clarify the relationship
- There was a bug in `BulkMutationsStats` relating to throttler stats that is now fixed.
- Adding simple test
- Adding a stats snapshot/reset method, which will be useful for the throttling logic.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
@@ -385,10 +385,13 @@ public class BulkMutation {
           retryId = Long.valueOf(
             asyncExecutor.getOperationAccountant().registerComplexOperation(createRetryHandler()));
         }
+        MutateRowsRequest request = currentRequestManager.build();
         long start = clock.nanoTime();
-        mutateRowsFuture = asyncExecutor.mutateRowsAsync(currentRequestManager.build());
+        long operationId = asyncExecutor.getOperationAccountant()
+            .registerOperationWithHeapSize(request.getSerializedSize());
         long now = clock.nanoTime();
         BulkMutationsStats.getInstance().markThrottling(now - start);
+        mutateRowsFuture = asyncExecutor.mutateRowsAsync(request, operationId);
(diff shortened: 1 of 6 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
```diff
@@ -385,8 +385,11 @@ public synchronized void run() {
           retryId = Long.valueOf(
             asyncExecutor.getOperationAccountant().registerComplexOperation(createRetryHandler()));
         }
+        long start = clock.nanoTime();
         mutateRowsFuture = asyncExecutor.mutateRowsAsync(currentRequestManager.build());
-        currentRequestManager.lastRpcSentTimeNanos = clock.nanoTime();
+        long now = clock.nanoTime();
+        BulkMutationsStats.getInstance().markThrottling(now - start);
```
igorbernstein2: This looks a bit weird. Are you intending on mutateRowsAsync blocking to achieve throttling?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutationsStats.java
```diff
@@ -25,14 +25,75 @@
...
 
   private Timer mutationTimer;
   private Meter mutationMeter;
   private Timer throttlingTimer;
 
+  public synchronized Snapshot snapshotAndReset() {
+    Snapshot snap = getSnapshot();
+    reset();
+    return snap;
+  }
```
igorbernstein2: Can you add a comment as to what the purpose of this method is?

sduskis:
I added comments, and added the call that can throttle directly into BulkMutation.  PTAL

sduskis:
Point taken.  I removed the method.  I'll build the throttler without the reset.

sduskis:
I updated the Stats to not be nullable anymore, as per our offline conversation.  I also remove the SnapshotStats, and just exposed the metrics directly.

PTAL.

PR #1343, opened 2017-05-23
Adding a BulkMutation throttler.

sduskis:
Adding a throttler and throttling options.


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BulkOptions.java
```diff
@@ -50,13 +50,25 @@
...
+   */
+  public static final int BIGTABLE_BULK_THROTTLE_TARGET_MS_DEFAULT = 40;
+
+  /**
    * The maximum amount of time a row will be buffered for. By default 0: indefinitely.
    */
   public static long BIGTABLE_BULK_AUTOFLUSH_MS_DEFAULT = 0;
 
   // Default rpc count per channel.
   /** Constant <code>BIGTABLE_MAX_INFLIGHT_RPCS_PER_CHANNEL_DEFAULT=50</code> */
```
garye: Comment is out of sync

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
```diff
@@ -64,7 +64,7 @@ private static int getDefaultDataChannelCount() {
     // 10 Channels seemed to work well on a 4 CPU machine, and this seems to scale well for higher
     // CPU machines. Use no more than 250 Channels by default.
```
garye: Comment out of sync.

Why the change to 5?

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
```diff
@@ -216,6 +216,19 @@
   public static final String BIGTABLE_BUFFERED_MUTATOR_MAX_MEMORY_KEY =
       "google.bigtable.buffered.mutator.max.memory";
 
+  /**
+   * Turn on a feature that will reduce the likelihood of BufferedMutator overloading a Cloud
+   * Bigtable cluster.
+   */
+  public static final String BIGTABLE_BUFFERED_MUTATOR_ENABLE_THROTTLING =
+      "google.bigtable.buffered.mutator.enable.throttling";
```
garye: nit - both options should maybe have the same prefix: google.bigtable.buffered.mutator.throttling

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/ResourceLimiter.java
```diff
@@ -89,6 +91,22 @@ public int getMaxInFlightRpcs() {
   }
 
   /**
+   * <p>Getter for the field <code>currentInFlightRpcs</code>.</p>
+   *
+   * @return The current number of allowed in-flight RPCs
+   */
+  public int getCurrentInFlightRpcs() {
```
garye: The name of the method doesn't seem to go with the comment. Is it allowed RPCs or currently executing RPCs?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutationThrottling.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.concurrent.TimeUnit;
+
+import com.google.cloud.bigtable.grpc.BigtableSessionSharedThreadPools;
+
+public class BulkMutationThrottling {
+
+  private static boolean isThrottling = false;
+
+  public static synchronized void throttle(final ResourceLimiter resourceLimiter,
```
garye: There are a lot of heuristics and magic here, would be nice if there was some explanation as to why things are done the way they are (eventually - maybe once this is no longer an experimental feature).

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutationThrottling.java
```diff
@@ -0,0 +1,65 @@
...
+      return;
+    }
+    final long highTargetNanos =
+        TimeUnit.MILLISECONDS.toNanos((long) (bulkMutationRpcTargetMs * 1.2));
+    final long lowTargetNanos =
+        TimeUnit.MILLISECONDS.toNanos((long) (bulkMutationRpcTargetMs * 0.8));
+    // Change at 5% increments
+    final int throttlingChangeStep = resourceLimiter.getMaxInFlightRpcs() / 20;
+    int current = resourceLimiter.getCurrentInFlightRpcs();
+    resourceLimiter.setCurrentInFlightRpcs(current / 2);
```
garye: Why does enabling throttling reduce the # of outstanding RPCs in half right away, even before latency measurements are examined?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
sduskis: Our internal performance tests showed that we should be adding fewer requests on the same channel.  I doubled the channels, and significantly reduced the outstanding RPCs.  I think that this deserves a separate PR, so I'll undo these changes.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutationThrottling.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+package com.google.cloud.bigtable.grpc.async;
+
+import java.util.concurrent.TimeUnit;
+
+import com.google.cloud.bigtable.grpc.BigtableSessionSharedThreadPools;
+
+public class BulkMutationThrottling {
+
+  private static boolean isThrottling = false;
```
igorbernstein2: What happens if a user creates 2 sessions?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutationThrottling.java
igorbernstein2: I don't love this being a static method, I feel like its something that the session should own

sduskis:
@igorbernstein2 I moved the throttling logic to `ResourceLimiter` so that there isn't a separate static method.  That said, `ResourceLimiter` is accessed as a static variable on `BigtableSession`.  In dataflow, `BigtableSession`s are ephemeral.  Each bundle gets its own `BigtableSession`.  There are a set of static variables (`cachedDataChannelPool` and `resourceLimiter` in `BigtableSession`, a couple of executors in grpc, and `BigtableSessionSharedThreadPools`) that allow for performance and VM level controls, like the throttling mechanism.

sduskis:
I added quite a few comments, and moved the throttling logic to ResourceLimiter.  PTAL.

igorbernstein2:
Looks good, one last request: can you add a dropwizard gauge for currentInFlightMaxRpcs...it should make it easier for users to tune & debug performance

sduskis:
We discussed metrics offline.  The scope of the work is large enough to require another PR.

sduskis:
@garye, I'm going to merge this.  If we need to clarify docs, let's please discuss this off line.

PR #1349, opened 2017-05-30
Check And Mutate semantics should be the inverse of ValueFilter

### LINKED ISSUE #1348: Possible incorrect behavior in checkAndPut
From https://stackoverflow.com/questions/44174094/checkandput-always-succeeds

There are only tests for equality comparators in checkAnd*.
Add tests for "<" or ">" (non commutative comparators) and verify that their behavior is as defined in the documentation:
"GREATER operator means expected value > existing <=> add the put."

sduskis:
It's not clear why that is, but CheckAnd* CompareOps should be reversed in order for Bigtable to Comply with HBase.  This should fix #1348

### DIFF
--- a/bigtable-hbase-parent/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestCheckAndMutate.java
+++ b/bigtable-hbase-parent/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestCheckAndMutate.java
@@ -4,9 +4,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


igorbernstein2:
LGTM

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableTable.java
```diff
@@ -701,6 +702,32 @@ protected boolean wasMutationApplied(
...
+   * For some reason, the ordering of CheckAndMutate operations is the inverse order of normal
+   * {@link ValueFilter} operations.
+   *
+   * @param compareOp
+   * @return the inverse of compareOp
+   */
+  private static CompareOp reverseCompareOp(CompareOp compareOp) {
+    switch (compareOp) {
+    case EQUAL:
+    case NOT_EQUAL:
```
rameshdharan: Shouldn't these be inverted also?

```
case EQUAL: 
   return CompareOp.NOT_EQUAL;
case NOT_EQUAL: 
   return CompareOp.EQUAL;
```

Doesn't look like this commit has tests for those cases but I believe the original poster encountered unexpected behavior even with strict equality?

bigtable-hbase-parent/bigtable-hbase-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestCheckAndMutate.java
```diff
@@ -285,4 +288,52 @@ public void testCheckAndMutate() throws IOException {
         row.containsColumn(SharedTestEnvRule.COLUMN_FAMILY, qualPut));
     table.close();
   }
+
+  @Test
+  public void testCompareOps() throws IOException {
```
rameshdharan: I think we should add tests for CompareOp.EQUAL and CompareOp.NOT_EQUAL as well.

PR #1359, opened 2017-06-05
Refactoring ResourceLimiter throttling.

sduskis:
Moving the increase / decrease logic to their own methods.  These methods will be reused in a subsequent enhancement.


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/ResourceLimiter.java
```diff
@@ -240,34 +240,45 @@ public synchronized void throttle(final int bulkMutationRpcTargetMs) {
...
-          if (newValue != current) {
-            setCurrentInFlightMaxRpcs(newValue);
-            LOG.debug("Latency is at %d ms.  Increasing paralellelism from %d to %d.",
-              TimeUnit.NANOSECONDS.toMillis(meanLatencyNanos), current, newValue);
-          }
+          increaseParallelism(minimumRpcCount, meanLatencyNanos, throttlingChangeStep / 2);
+        }
+      }
+
+      protected void reduceParallelism(final int minimumRpcCount, long meanLatencyNanos,
```
igorbernstein2: do you plan on subclassing? if not I think private would be more appropriate

PR #1374, opened 2017-06-07
Staleness handling is now scheduled in BulkMutation.

sduskis:
Removing the notion of complex operation in favor of a `Batch` scheduling its own staleness cleanup.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
@@ -29,7 +29,6 @@ import com.google.cloud.bigtable.config.Logger;
 import com.google.cloud.bigtable.config.RetryOptions;
 import com.google.cloud.bigtable.grpc.BigtableDataClient;
 import com.google.cloud.bigtable.grpc.BigtableTableName;
-import com.google.cloud.bigtable.grpc.async.OperationAccountant.ComplexOperationStalenessHandler;
 import com.google.cloud.bigtable.grpc.scanner.BigtableRetriesExhaustedException;
 import com.google.cloud.bigtable.metrics.BigtableClientMetrics;
 import com.google.cloud.bigtable.metrics.Meter;
@@ -76,6 +75,7 @@ public class BulkMutation {
 
   public static final long MAX_RPC_WAIT_TIME_NANOS = TimeUnit.MINUTES.toNanos(7);
   private final AtomicLong batchIdGenerator = new AtomicLong();
+  public ScheduledFuture<?> stalenessFuture;
 
   private static StatusRuntimeException toException(Status status) {
     io.grpc.Status grpcStatus = io.grpc.Status
@@ -138,8 +138,11 @@ public class BulkMutation {
     }
 
     public boolean isStale() {
-      return lastRpcSentTimeNanos != null
-          && lastRpcSentTimeNanos <= (clock.nanoTime() - MAX_RPC_WAIT_TIME_NANOS);
+      return lastRpcSentTimeNanos != null && calculateTimeUntilStaleNanos() <= 0;
+    }
+
+    public long calculateTimeUntilStaleNanos() {
+      return lastRpcSentTimeNanos + MAX_RPC_WAIT_TIME_NANOS - clock.nanoTime();
     }
 
     public boolean wasSent() {
@@ -149,7 +152,6 @@ public class BulkMutation {
 
   @VisibleForTesting
   class Batch implements Runnable {
-    
     private final Meter mutationMeter =
         BigtableClientMetrics.meter(MetricLevel.Info, "bulk-mutator.mutations.added");
     private final Meter mutationRetryMeter =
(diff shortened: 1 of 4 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
```diff
@@ -416,28 +417,35 @@ public synchronized void run() {
...
                 io.grpc.Status.UNKNOWN.withDescription("Stale requests.").asRuntimeException());
+            } else {
+              setupStalenessChecker();
             }
           }
         }
       };
+      if (currentRequestManager.lastRpcSentTimeNanos != null) {
+        long delay = currentRequestManager.calculateTimeUntilStaleNanos();
+        stalenessFuture = retryExecutorService.schedule(runnable, delay, TimeUnit.NANOSECONDS);
```
Orbz: Should stalenessFuture be contained in class Batch instead of BulkMutation? It looks like one batch calling setRetryComplete will cancel the latest batch's future, not the batch that created it.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/OperationAccountant.java
```diff
@@ -169,8 +124,8 @@ private void logNoSuccessWarning(long now) {
     long lastUpdated = TimeUnit.NANOSECONDS.toSeconds(lastUpdateNanos);
     LOG.warn(
       "No operations completed within the last %d seconds. "
-          + "There are still %d simple operations and %d complex operations in progress.",
-      lastUpdated, operations.size(), complexOperations.size());
+          + "There are still %d simple operations in progress.",
```
Orbz: No need to call them simple anymore if there's only one type of operation.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
```diff
@@ -76,6 +75,7 @@
 
   public static final long MAX_RPC_WAIT_TIME_NANOS = TimeUnit.MINUTES.toNanos(7);
   private final AtomicLong batchIdGenerator = new AtomicLong();
+  public ScheduledFuture<?> stalenessFuture;
```
garye: Does this need to be public?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
```diff
@@ -416,28 +417,35 @@ public synchronized void run() {
...
     }
 
-    /**
-     * A {@link ComplexOperationStalenessHandler} is a method of last resort in cases where somehow the various
-     * listeners failed to get a response, or had some kind of strange issue.
-     */
-    private ComplexOperationStalenessHandler createRetryHandler() {
-      return new ComplexOperationStalenessHandler() {
+    private void setupStalenessChecker() {
+      if (currentRequestManager == null){
```
garye: We also check for empty() below. Is that necessary here?  If so does it help to have this check be a method with a self-documenting name?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/OperationAccountant.java
```diff
@@ -134,23 +106,6 @@ public void awaitCompletion() throws InterruptedException {
 
         long now = clock.nanoTime();
         if (now >= noSuccessCheckDeadlineNanos) {
-          // There are unusual cases where an RPC could be completed, but we don't clean up
```
garye: Love all the code that is getting deleted!

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/OperationAccountant.java
garye: Can probably just refer to them as "operations"

bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/async/TestBulkMutationAwaitCompletion.java
```diff
@@ -137,10 +145,38 @@ public void run() {
...
     originalOperationAccountantLog = OperationAccountant.LOG;
     BulkMutation.LOG = mockLogger;
     OperationAccountant.LOG = mockLogger;
+    Runnable doTimout = new Runnable() {
+      @Override
+      public void run() {
+        try {
+          performTimeout();
+        } catch (InterruptedException e) {
+          // TODO Auto-generated catch block
```
garye: kill

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
sduskis: Ugh.  Good catch.

sduskis:
All of the comments were addressed.  PTAL.

garye:
LGTM

PR #1379, opened 2017-06-08
OperationAccountant now uses ListenableFuture as its interface.

sduskis:
This is an intermediate step towards separating the variety of concerns found in BulkMutation.


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/OperationAccountant.java
```diff
@@ -58,6 +61,7 @@
 
   private long noSuccessCheckDeadlineNanos;
   private int noSuccessWarningCount;
+  private int idGenerator = 0;
```
garye: nit: Seems weird to call an int a "generator". Also why an int guarded by a lock vs an atomic like above for the batch generator?

sduskis:
I'm now using an AtomicLong.  PTAL.

PR #1392, opened 2017-06-09
exclude unused api-common from protos

igorbernstein2:
This dep is pulled in as a transitive dep of grpc-google-common-protos. But we don't need any of the functionality that it provides. We only need the generated stubs


sduskis:
LGTM

PR #1393, opened 2017-06-09
Adding a new RetryingMutateRowsOperation to perform retries.

sduskis:
BulkMutation shouldn't have to do any retrying.  It already does too much.  The new RetryingMutateRowsOperation will perform the nuances of the retries.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
@@ -211,11 +211,6 @@ public class BulkMutation {
     @VisibleForTesting
     synchronized void handleResult(List<MutateRowsResponse> results) {
       mutateRowsFuture = null;
-      if (results.size() != 1) {
-        setFailure(io.grpc.Status.INTERNAL.withDescription("Expected only a single response.")
-            .asRuntimeException());
-        return;
-      }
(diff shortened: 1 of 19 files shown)


sduskis:
There are a lot of changes in this PR, but I think that it's good to go as is.  There are some things that I can pull out to a different PR, if you'd prefer.

igorbernstein2:
Looking now

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
```diff
@@ -223,136 +211,79 @@ protected void markCompletion() {
     @VisibleForTesting
     synchronized void handleResult(List<MutateRowsResponse> results) {
       mutateRowsFuture = null;
-      AtomicReference<Long> backoffTime = new AtomicReference<>();
+      if (results.size() != 1) {
+        setFailure(io.grpc.Status.INTERNAL.withDescription("Expected only a single response.")
+            .asRuntimeException());
+        return;
+      }
```
igorbernstein2: Would you mind adding a TODO to add support for multiple MutateRowsResponses? The server doesn't currently send them, but in the future it will support sub-batching of the Mutations

sduskis: We support it via the retrying framework.  I'll add a comment here.

Issue #1382, opened 2017-06-08
Upgrade to grpc 1.4.0


sduskis:
This is more tricky than previous upgrades.  The shading logic doesn't work well if tcnative is a separate artifact.  We should consider having an explicit dependency on tcnative, and including tcnative in our hbase shaded jar.

sduskis:
We upgraded to 1.4.0, but did not upgrade netty and tcnative.

PR #1395, opened 2017-06-09
Upgrading grpc to 1.4.0

sduskis:


igorbernstein2:
Don't forget to bump tcnative to 2.0.1-Final

sduskis:
Ugh.  We have a not backwards compatible change on our hands.  Older versions of grpc are not compatible with  tcnative to 2.0.1-Final.

igorbernstein2:
Maybe we should hold off upgrading grpc until https://github.com/grpc/grpc-java/pull/2485 lands?

sduskis:
This should fix #1382, if and when this is complete.

It turns out that the grpc upgrade can use older versions of netty, so I'll remove the netty upgrade and keep grpc and protobuf upgrades.

PR #1406, opened 2017-06-12
shade proto-google-common-protos as well

igorbernstein2:
grpc-google-common-protos no longer includes the protos, need to shade proto-google-common-protos


sduskis:
Do we need a relocation as well?

igorbernstein2:
Nope, they are under com.google.api, these classes used to live in grpc-google-common-protos, but were moved out into their own jar sometime between 0.1.6 & 0.1.9

sduskis:
LGTM

PR #1408, opened 2017-06-12
ResourceLimiter is now used from a ClientInterceptor.

sduskis:
AsyncExecutor and BulkMutator shouldn't have to know about ResourceLimiter directly.


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/ThrottlingClientInterceptor.java
```diff
@@ -0,0 +1,90 @@
...
+    final AtomicBoolean interrupted = new AtomicBoolean(false);
+    return new SimpleForwardingClientCall<ReqT, RespT>(next.newCall(method, callOptions)) {
+      @Override
+      public void start(Listener<RespT> delegate, Metadata headers) {
+        if (interrupted.get()) {
+          if (id.get() != -1) {
+            resourceLimiter.markCanBeCompleted(id.get());
+          }
+          delegate.onClose(Status.INTERNAL.withDescription("Operation was interrupted"),
+            new Metadata());
```
igorbernstein2: start must be called before any other method, so this half of the if statement will never be true

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/ThrottlingClientInterceptor.java
```diff
@@ -0,0 +1,90 @@
...
+              };
+          delegate().start(listener, headers);
+        }
+      }
+
+      @Override
+      public void sendMessage(ReqT message) {
+        try {
+          id.set(resourceLimiter.registerOperationWithHeapSize(((MessageLite)message).getSerializedSize()));
+          super.sendMessage(message);
```
igorbernstein2: It seems like this interceptor doesn't support client streaming calls (calling sendMessage multiple times). It would be good to add a precondition that prevents that from happening. Also would be good to handle the case of no sendMessage being called between start & close

igorbernstein2: Also, would be good to only enclose registerOperationWithHeapSize in try/catch for InterruptedException.

bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/async/TestAsyncExecutor.java
```diff
@@ -59,11 +62,13 @@
   private SettableFuture future;
 
   private AsyncExecutor underTest;
+  private ResourceLimiter resourceLimiter;
```
igorbernstein2: Maybe we should mock the resourceLimiter?

sduskis:
Most of the comments were addressed.  PTAL

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/ThrottlingClientInterceptor.java
```diff
@@ -0,0 +1,108 @@
...
+          delegate().request(numMessages);
+        }
+      }
+
+      @Override
+      public void sendMessage(ReqT message) {
+        Preconditions.checkState(delegateCall == null);
+        Preconditions.checkState(delegateListener != null);
+        Preconditions.checkState(headers != null);
+        Preconditions.checkState(id == null);
```
igorbernstein2: Please add a message to the id not being null. Something like "This interceptor only supports unary calls"

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -198,6 +199,9 @@ private synchronized static void initializeResourceLimiter(BigtableOptions optio
    */
   private BigtableClusterName clusterName;
 
+  private ThrottlingClientInterceptor throttlingClientInterceptor =
+      new ThrottlingClientInterceptor();
+
```
igorbernstein2: This doesn't appear to be used in the channel creation. Also would be possible to construct this in the constructor, passing it the resource limiter and avoiding the enable() bit?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/ThrottlingClientInterceptor.java
```diff
@@ -0,0 +1,108 @@
...
+        this.headers = headers;
+      }
+
+      @Override
+      public void request(int numMessages) {
+        if (id == null) {
+          requested += numMessages;
+        } else {
+          delegate().request(numMessages);
+        }
```
sduskis: All of the methods should be called in the same thread.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
sduskis: The interceptor is now added to the array of interceptors in BigtableSession's constructor.

sduskis:
I addressed the comments.  PTAL.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -230,16 +232,19 @@ public BigtableSession(BigtableOptions opts) throws IOException {
...
-        headerInterceptorBuilder.add(headerInterceptor);
+      if (credentialsInterceptor != null) {
+        clientInterceptorsList.add(credentialsInterceptor);
       }
     } catch (GeneralSecurityException e) {
       throw new IOException("Could not initialize credentials.", e);
     }
 
-    headerInterceptors = headerInterceptorBuilder.build().toArray(new ClientInterceptor[0]);
+    clientInterceptorsList.add(throttlingClientInterceptor);
```
igorbernstein2: I think we should throttle before fetching the oauth token. If the token is about to expire and we throttle the request then when we unblock the request, the token can be expired.

sduskis: Good point.

PR #1411, opened 2017-06-16
Adding uninitialized state to RefreshingOAuth2CredentialsInterceptor.

sduskis:


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -197,7 +197,7 @@ public RefreshingOAuth2CredentialsInterceptor(ExecutorService scheduler, OAuth2C
       public void start(Listener<RespT> responseListener, Metadata headers) {
         HeaderCacheElement headerCache = getHeaderSafe();
 
-        if (!headerCache.status.isOk()) {
+        if (headerCache.getCacheState() != CacheState.Good) {
```
igorbernstein2: I don't think this is necessary. As discussed, status is never null & I'd prefer to keep this logic in HeaderCacheElement

PR #1423, opened 2017-06-21
Updating BulkMutation.

sduskis:
- Minor refactoring in `BigtableDataGrpcClient` to consolidate logic in mutateRows methods
- Moving logic to onOK for RetryingMutateRowsOperation with invalid responses.
- When a User cancels a request, cancel the GrpcFuture in AbstractRetryingOperation


sduskis:
This is ready for review.  PTAL.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
```diff
@@ -304,11 +299,16 @@ private void setFailure(Throwable t) {
     private synchronized void setComplete() {
       cancelIfNotDone(mutateRowsFuture);
       cancelIfNotDone(stalenessFuture);
+      mutateRowsFuture = null;
+      for(Future<?> future: futures) {
+        if (!future.isDone()) {
+          future.cancel(true);
```
igorbernstein2: I think it would make things easier to debug if instead of cancelling the future, we set a custom exception instead. ie future.setException(new MissingBulkResponseEntry(entry))

PR #1440, opened 2017-07-02
Using fast checking for exists()

sduskis:
KeyOnlyFilter is a filter that returns an empty value for the first qualifier in a row.  This functionality was created to make exist() fast, but the client never took advantage of it.


igorbernstein2:
Looks good, but please add an integration test that ensures that this doesn't have any weird interactions some of the more exotic filters like pagination

sduskis:
I added a whileMatchFilter test.   PTAL.

PR #1442, opened 2017-07-03
Fixing a bug relating to RetryingReadRowsOperation.

sduskis:
completionFuture is null in RetryingReadRowsOperation, so add some defensive checks in AbstractRetryingOperation onError() for canceled operations.


igorbernstein2:
why is the completion future null for RetryingReadRowsOperation?

sduskis:
GrpcFuture is now always set.  PTAL.

Issue #1456, opened 2017-07-12
Failed to perform operation 'get'


jsw:
Since upgrading to `com.google.cloud.bigtable:bigtable-hbase-1.x:1.0.0-pre1` I am occasionally seeing the following error. I've redacted some confidential elements of the stack trace. Any idea what may be causing it?

```
2017-07-12 18:35:06,260 [ERROR] XXX:67 Table xxx: Other exception thrown: Failed to perform operation. Operation='get', projectId='xxx', tableName='xxx', rowKey='xxx'
java.io.IOException: Failed to perform operation. Operation='get', projectId='xxx', tableName='xxx', rowKey='1406519542728802'
	at com.google.cloud.bigtable.hbase.BigtableTable.logAndCreateIOException(BigtableTable.java:515)
	at com.google.cloud.bigtable.hbase.BigtableTable.get(BigtableTable.java:266)
        ... frames omitted
Caused by: java.util.concurrent.CancellationException: Task was cancelled.
	at com.google.common.util.concurrent.AbstractFuture.cancellationExceptionWithCause(AbstractFuture.java:1123)
	at com.google.common.util.concurrent.AbstractFuture.getDoneValue(AbstractFuture.java:501)
	at com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:404)
	at com.google.common.util.concurrent.AbstractFuture$TrustedFuture.get(AbstractFuture.java:86)
	at com.google.cloud.bigtable.hbase.BigtableTable.get(BigtableTable.java:253)
	... 38 common frames omitted
```

jsw:
@sduskis Any ideas on what could be causing the cancellation, or what I can do to help troubleshoot this? Unfortunately, I don't have an easy repro. It must be a race condition or timeout issue or similar. The issue must be on the client and not the server, given we just started seeing this issue. Our access pattern is to have several worker threads continually scanning for and reading individual rows from Bigtable. Unfortunately, we will have to rollback to our previous known good release, which depends on `bigtable-hbase-1.2:0.9.4`, and pins us to using an alpha release of the google pubsub library so that the netty dependencies align. Let me know what additional info I can help provide.

sduskis:
[...]

PR #1467, opened 2017-07-19
Touching up AbstractRetryingOperation cancels.

sduskis:
- Clarifying GrpcFuture.interruptTask() with comments.
- Adding a check for isDone() in interruptTask().
- Using completionFuture.cancel(false) instead of (true) in onError(), so that GrpcFuture.interruptTask() is not called.
- Adding AbstractRetryingOperation.getExhaustedRetriesException() so that it can be used for RetryingReadRowsOperation scan timeouts.

This might address #1456

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
@@ -18,6 +18,7 @@ package com.google.cloud.bigtable.grpc.async;
 import io.grpc.Status.Code;
 import java.io.IOException;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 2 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
```diff
@@ -141,7 +144,10 @@ protected void onError(Status status, Metadata trailers) {
     Code code = status.getCode();
     // CANCELLED
     if (code == Status.Code.CANCELLED) {
-      completionFuture.cancel(true);
+      if (!completionFuture.isDone()) {
```
igorbernstein2: I'm not sure if this is necessary, cancel is a no-op if isDone() is true

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
```diff
@@ -141,7 +144,10 @@ protected void onError(Status status, Metadata trailers) {
...
+        // cancel(true) would trigger GrpcFuture.interuptTask().  We don't want that to happen.
+        completionFuture.cancel(false);
```
igorbernstein2: Would you mind updating the comment as to why we don't want to interrupt the task? If I understand correctly its because task interruption should only come from the user, and since this is coming from the grpc/server the task is already interrupted?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
```diff
@@ -165,20 +171,20 @@ protected void onError(Status status, Metadata trailers) {
...
       LOG.info("Retrying failed call. Failure #%d, got: %s on channel %s",
           status.getCause(), failedCount, status, channelId);
+      performRetry(nextBackOff);
     }
+  }
 
-    performRetry(nextBackOff);
+  protected BigtableRetriesExhaustedException getExhaustedRetriesException(Status status) {
+    rpc.getRpcMetrics().markRetriesExhasted();
+    operationTimerContext.close();
```
igorbernstein2: nit: Its a bit surprising to have these side effects in this method. I can't think of anything better

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
sduskis: I'm going to revert to cancel(true).  I was worried about this invocation of cancel() cascading to another RPC cancel.  `GrpcFuture.cancel(true)` calls `GrpcFuture.interruptTask()` which calls `AbstractRetryingOperation.cancel()` which does an RPC cancel.  However, it turns out that in this case `AbstractRetryingOperation.cancel()` should be a no-op.  I'll ensure that it's a no-op by extending the synchronized block in `onOK()`

Issue #1470, opened 2017-07-25
Thread leak in BigTable client


yaojuncn:
BigTable client: com.google.cloud.bigtable:bigtable-hbase-1.x-hadoop:1.0.0-pre1
Runtime: dataproc 1.1  (Apache Spark 2.0.2), 

Problem:
after running the spark jobs for about 8 hours, it get out of memory 

the heap dump showed that there are too many grpc threads created, i think there is a thread leak. 

there are 229 grpc threads created, each contain about 10MB, which is about 2.3GB in total.

sduskis:
We were discussing this issue internally.  Here was my response:

We had a similar problem using Dataflow.  The problem turned out to be that we were opening too many Connections in a single VM, and each connection needed some threads from the thread pool.  We initially solved that problem by caching the HBase Connection object in a transient member variable, so that there would be only a single Connection per JVM.  We later added a more elegant solution that cached the underlying gRPC Channel objects.  I would recommend the following option settings that can be set either in an hbase-site.xml, or a hadoop Configuration object:

`"google.bigtable.use.cached.data.channel.pool" = "true"`   // This enables caching of the underlying gRPC channels
`"google.bigtable.endpoint.host" = "batch-bigtable.googleapis.com"`  // This directs operations to a set of production servers dedicated to batch jobs, which helps overall system health.

// These options help with settings meant for batch operations rather than "online" operations:
`"google.bigtable.grpc.retry.initial.elapsed.backoff.ms" = "5000"`      // Wait 5 seconds before retrying operations.  This helps reduce the likelihood of overloading the server-side 
`"google.bigtable.grpc.retry.max.elapsed.backoff.ms" = "300,000"`  // Keep on retrying for 5 minutes rather than 1 minute.

[Here]( https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-dataflow-parent/bigtable-hbase-dataflow/src/main/java/com/google/cloud/bigtable/dataflow/CloudBigtableConfiguration.java#L165)'s where we set those values for Dataflow.

yaojuncn:
@sduskis   
thanks for the quick response,  so trying to understand the case,  

1. when you say "we were opening too many Connections in a single VM, and each connection needed some threads from the thread pool",  I assume you are referring to the internal implementation in the library, right? 

2. Also I am a little bit confused by 
[...]

PR #1471, opened 2017-07-28
Ensuring that scans don't block grpc threads.

sduskis:
When scan queues are full, the client blocks. That could cause thread leaks, like those in issue #1470. This change will elminiate blocking on the grpc tread, and instead relies on client-side flow control to cache up to 3 responses, and then only request more results when all of the rows returned by a single ReadRowsResponse are removed from the queue.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
@@ -403,9 +403,10 @@ public class BigtableDataGrpcClient implements BigtableDataClient {
   public ResultScanner<FlatRow> readFlatRows(ReadRowsRequest request) {
     // Delegate all resumable operations to the scanner. It will request a non-resumable scanner
     // during operation.
-    ResponseQueueReader reader = new ResponseQueueReader(retryOptions.getStreamingBufferSize());
-    final RetryingReadRowsOperation listener = createReadRowsRetryListener(request, reader);
-    return new ResumingStreamingResultScanner(reader, listener);
+    ResponseQueueReader reader = new ResponseQueueReader();
+    ScanHandler scanHandler = createReadRowsRetryListener(request, reader);
+    reader.setScanHandler(scanHandler);
+    return new ResumingStreamingResultScanner(reader, scanHandler);
   }
 
   /** {@inheritDoc} */
@@ -425,6 +426,8 @@ public class BigtableDataGrpcClient implements BigtableDataClient {
           getCallOptions(readRowsAsync.getMethodDescriptor(), request),
           retryExecutorService,
           createMetadata(request.getTableName()));
+
+    // Start the operation.
     listener.getAsyncResult();
     return listener;
   }
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
@@ -403,32 +403,43 @@ public class BigtableDataGrpcClient implements BigtableDataClient {
   public ResultScanner<FlatRow> readFlatRows(ReadRowsRequest request) {
     // Delegate all resumable operations to the scanner. It will request a non-resumable scanner
     // during operation.
-    ResponseQueueReader reader = new ResponseQueueReader();
-    ScanHandler scanHandler = createReadRowsRetryListener(request, reader);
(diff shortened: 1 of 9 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -403,9 +403,10 @@ public int available() {
   public ResultScanner<FlatRow> readFlatRows(ReadRowsRequest request) {
     // Delegate all resumable operations to the scanner. It will request a non-resumable scanner
     // during operation.
-    ResponseQueueReader reader = new ResponseQueueReader(retryOptions.getStreamingBufferSize());
-    final RetryingReadRowsOperation listener = createReadRowsRetryListener(request, reader);
-    return new ResumingStreamingResultScanner(reader, listener);
+    ResponseQueueReader reader = new ResponseQueueReader();
+    ScanHandler scanHandler = createReadRowsRetryListener(request, reader);
+    reader.setScanHandler(scanHandler);
```
igorbernstein2: To avoid having both the scanner & reader calling out to the ScanHandler(RetryingReadRowsOperation) can the scanner just call the reader and the reader delegates to the ScanHandler?

igorbernstein2: Maybe consider just merging the Reader & Scanner? Its a bit hard to tell what the distinction between them is.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RetryingReadRowsOperation.java
```diff
@@ -83,22 +83,31 @@ public void run() {
...
+    }
   }
 
   /** {@inheritDoc} */
   @Override
   public void onMessage(ReadRowsResponse message) {
-    // Pre-fetch the next request
-    call.request(1);
     resetStatusBasedBackoff();
+    if (rowObserver instanceof ResponseQueueReader) {
```
igorbernstein2: This is a bit hacky, I wonder if there is a better way to handle this

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RetryingReadRowsOperation.java
```diff
@@ -83,22 +83,31 @@ public void run() {
...
+      // ResponseQueueReader will only read more results if prior FlatRows were retrieved from the
+      // queue.
+      ((ResponseQueueReader) rowObserver).addRequestResultMarker();
```
igorbernstein2: Shouldn't this happen after rowMerger processed the message?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RetryingReadRowsOperation.java
igorbernstein2: I think using grpc's pattern of disableAutoInboundFlowControl() & setOnReadyHandler() might make this a bit cleaner

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RetryingReadRowsOperation.java
igorbernstein2: Also, would be good to add a test for this behavior

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ResponseQueueReader.java
```diff
@@ -131,34 +138,39 @@ public int available() {
...
+
+  private void addEntry(String message, ResultQueueEntry<FlatRow> entry) {
     try {
-      completionMarkerFound.set(true);
-      resultQueue.put(ResultQueueEntry.<FlatRow> completionMarker());
+      resultQueue.put(entry);
     } catch (InterruptedException e) {
       Thread.currentThread().interrupt();
-      throw new RuntimeException("Interrupted while adding a ResultQueueEntry", e);
+      throw new RuntimeException("Interrupted while " + message, e);
```
igorbernstein2: Now that the queue is unbounded, this should never happen. So I don't think message clarification is necessary.

sduskis:
I added tests, and used a `ClientResponseObserver`.  PTAL.

PR #1480, opened 2017-08-16
Removing bigtable-hbase-1.0 through bigtable-hbase.1.3

sduskis:
Our travis builds are taking too long.  Users should move from the 1.[0-3] artifacts to the new 1.x artifacts.


sduskis:
We should do this later.

PR #1487, opened 2017-08-21
Shade tcnative & update grpc

igorbernstein2:
This PR is based on https://github.com/grpc/grpc-java/pull/2485. However, since bigtable-hbase-1.x-shaded shades grpc we should not depend on grpc-netty-shaded: this will cause conflicts when the end user has both bigtable-hbase-1.x-shaded and grpc-netty-shaded on the classpath.

This requires netty 4.1.14 & grpc 1.6.0

As a bonus this fixes relocation of service files so ManagedChannelBuilder should work now

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RetryingReadRowsOperation.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RetryingReadRowsOperation.java
@@ -17,6 +17,7 @@ package com.google.cloud.bigtable.grpc.scanner;
 
 import java.util.concurrent.ScheduledExecutorService;
 
+import javax.annotation.Nullable;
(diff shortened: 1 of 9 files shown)


ankurcha:
I believe this does not help with non-hbase `bigtable-client-core`. Currently, as beam-sdk uses grpc 1.2.0 versions of the client later than 0.9.7.1 will not be compatible (as tcnative dependencies fail to load). https://github.com/GoogleCloudPlatform/DataflowJavaSDK/issues/601

sduskis:
We're aware of the issues.  @igorbernstein2 found that the newer version of tcnative allows for shading, which is what we do to allow CloudBigtableIO work with Beam/Dataflow.  See [here](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-hbase-parent/bigtable-hbase-1.x-shaded/pom.xml#L136) for @igorbernstein2's comments about that shading.

sduskis:
Apparently, there was a bug in grpc 1.6.0.  We'll need to upgrade to 1.6.1.

PR #1494, opened 2017-08-31
Upgrading to grpc 1.6.1

sduskis:
Also using the default io.netty dependencies from grpc now that we're using newer tcnative jars.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/pom.xml
+++ b/bigtable-client-core-parent/bigtable-client-core/pom.xml
@@ -109,21 +109,6 @@ limitations under the License.
 
         <dependency>
             <groupId>io.netty</groupId>
-            <artifactId>netty-handler</artifactId>
-            <version>${netty.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>io.netty</groupId>
-            <artifactId>netty-codec-http2</artifactId>
-            <version>${netty.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>io.netty</groupId>
-            <artifactId>netty-handler-proxy</artifactId>
-            <version>${netty.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>io.netty</groupId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -58,12 +58,12 @@ limitations under the License.
         <compat.module>hbase-hadoop2-compat</compat.module>
         <junit.version>4.12</junit.version>
         <mockito.version>1.10.19</mockito.version>
-        <grpc.version>1.6.0</grpc.version>
+        <grpc.version>1.6.1</grpc.version>
         <!--NOTE: 4.1.14.Final is the first version that support shading -->
         <netty.version>4.1.14.Final</netty.version>
```
igorbernstein2: Let's remove the property as well

pom.xml
```diff
@@ -58,12 +58,12 @@ limitations under the License.
...
         <netty-tcnative-boringssl-static.version>2.0.5.Final</netty-tcnative-boringssl-static.version>
         <protobuff-java.version>3.3.1</protobuff-java.version>
-        <protoc.version>3.3.0</protoc.version>
+        <protoc.version>3.4.0</protoc.version>
```
igorbernstein2: protobuf is still 3.3.0 in 1.6.1
https://github.com/grpc/grpc-java/blob/v1.6.1/build.gradle#L96

It'll be bumped in 1.7

bigtable-client-core-parent/bigtable-client-core/pom.xml
```diff
@@ -109,21 +109,6 @@ limitations under the License.
...
-            <artifactId>netty-codec-http2</artifactId>
-            <version>${netty.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>io.netty</groupId>
-            <artifactId>netty-handler-proxy</artifactId>
-            <version>${netty.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>io.netty</groupId>
```
igorbernstein2: Good call, forgot to remove these

PR #1495, opened 2017-08-31
WIP: Opencensus

sduskis:


bigtable-client-core-parent/bigtable-client-core/pom.xml
```diff
@@ -169,6 +141,23 @@ limitations under the License.
             <version>${dropwizard.metrics.version}</version>
         </dependency>
```
bogdandrutu: Add this as well:
```xml
        <dependency>
            <groupId>io.opencensus</groupId>
            <artifactId>opencensus-api</artifactId>
            <version>${opencensus.version}</version>
        </dependency>
```

sduskis: That worked!

bigtable-client-core-parent/bigtable-client-core/pom.xml
```diff
@@ -169,6 +141,27 @@ limitations under the License.
...
+            <groupId>io.opencensus</groupId>
+            <artifactId>opencensus-exporter-trace-stackdriver</artifactId>
+            <version>${opencensus.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.opencensus</groupId>
+            <artifactId>opencensus-api</artifactId>
+            <version>${opencensus.version}</version>
+        </dependency>
+
```
igorbernstein2: Should these be <scope>runtime</scope>?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
```diff
@@ -100,6 +112,8 @@ public boolean setException(Throwable throwable) {
   protected Timer.Context operationTimerContext;
   protected Timer.Context rpcTimerContext;
 
+  private Scope operationScope;
```
bogdandrutu: This scope is good only for synchronized requests (where you do something like):

try (Scope a = ...) {
...
}

For your case create a Span (SpanBuilder#startSpan) and use:
run() {
 try (Scope s = Tracer.withSpan(span)) {
   ...
 }
}

// And on all OnX methods inherited.
onClose() {
 try (Scope s = Tracer.withSpan(span)) {
   ...
 }
}

To close the span just do Span.end(...);

sduskis: @bogdandrutu: should I include scopes for `onMessage()`?

bogdandrutu: Only on the methods that you override and you want to add annotations/events or propagate the this span as parent down.

In run() is important to do it because you want this span to be the parent of the RPC span that is created by the gRPC.

If onClose does not call any method in other class that you want to instrument and does not make any RPC call, you can skip to install it in the Scope part and simply record events directly on the Span object.

sduskis: Thanks for all the help!

sduskis:
This PR was superseded by other PRs.

PR #1498, opened 2017-09-08
Added limit to number of mutations per batch

### LINKED ISSUE #1497: BulkMutation can fail if a batch contains > 100,000 mutations
I ran into this using Dataflow (using bigtable-client-core 0.9.7.1), but I think it happens in the latest 1.0.0-pre3 release as well.

When bulk loading some very-highly dimensional data, I ran into an issue where a batch from BulkMutation would be rejected by the Bigtable instance with the error `Status{code=INVALID_ARGUMENT, description=Too many mutations: 238015 provided; limit 100000, cause=null}`,

Looking through BulkMutation, it seems that although size and # of rowkeys is checked, there's nothing to prevent a large number of mutations.  

It seems like the logic in add() needs to be inverted somewhat, where it would check if adding the current item to a batch would cause it to be > 100,000 mutations, and send the existing batch if so.

steveniemitz:
closes #1497


steveniemitz:
I signed it!

PR #1506, opened 2017-09-24
Adding tracing spans to BigtableTable and AbstractRetryingOperation.

sduskis:


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
```diff
@@ -124,44 +140,96 @@ public AbstractRetryingOperation(
...
+    String spanName = makeSpanName("Operation", rpc.getMethodDescriptor().getFullMethodName());
+    operationSpan = TRACER.spanBuilder(spanName).setRecordEvents(true).startSpan();
   }
 
   /** {@inheritDoc} */
   @Override
   public void onClose(Status status, Metadata trailers) {
-    synchronized (callLock) {
-      call = null;
+    try (NonThrowingCloseable s = TRACER.withSpan(operationSpan)) {
```
igorbernstein2: doesnt this cause the span to be ended twice? once in om finalizeStats and a second time in try-with-resources?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
```diff
@@ -124,44 +140,96 @@ public AbstractRetryingOperation(
...
+    operationTimerContext.close();
+    if (operationSpan != null) {
+      operationSpan.end(
+        EndSpanOptions.builder().setStatus(opencensusStatusFromGrpcCode(status.getCode())).build());
     }
   }
 
+  // TODO: Replace this with StatusConverter.fromGrpcStatus(status) when the opencensus 0.6.0 is
+  // released.
+  private static io.opencensus.trace.Status opencensusStatusFromGrpcCode(io.grpc.Status.Code
```
igorbernstein2: Might be better to move this to a utility class (even if its temporary)

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableTable.java
```diff
@@ -154,16 +167,21 @@ public final Configuration getConfiguration() {
   /** {@inheritDoc} */
   @Override
   public HTableDescriptor getTableDescriptor() throws IOException {
-    try (Admin admin = this.bigtableConnection.getAdmin()) {
+    try (
+        NonThrowingCloseable ss =
+            TRACER.spanBuilder("BigtableTable.getTableDescriptor").startScopedSpan();
```
igorbernstein2: Whats the difference between the spans created here and the top level spans created at the operation level?

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableTable.java
```diff
@@ -278,15 +319,19 @@ protected Result convertToResult(FlatRow row) {
...
           client.readFlatRows(hbaseAdapter.adapt(scan));
       if (hasWhileMatchFilter(scan.getFilter())) {
         return Adapters.BIGTABLE_WHILE_MATCH_RESULT_RESULT_SCAN_ADAPTER.adapt(scanner);
       }
       return Adapters.BIGTABLE_RESULT_SCAN_ADAPTER.adapt(scanner);
     } catch (Throwable throwable) {
       LOG.error("Encountered exception when executing getScanner.", throwable);
+      if (scanner != null) {
+        scanner.close();
+      }
```
igorbernstein2: Should this be pulled out into a separate PR?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
sduskis: The API is a bit confusing.  This withSpan just sets and unsets the span in a ThreadLocal.  [Newer versions of the API](https://github.com/census-instrumentation/opencensus-java/blob/b2f48c5d608c154d75ae7616c660e053f6e5c5f7/api/src/main/java/io/opencensus/trace/SpanBuilder.java#L26) has the functionality that you expect with `tracer.spanBuilder("MyChildSpan").startScopedSpan()`.

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableTable.java
sduskis: These names are more HBase-y than the Operation level.  Customers using HBase will need this translation.  This also gives us insight into how much time is spent on the HBase -> proto conversion.

sduskis:
I addressed the concerns.  PTAL at your convenience.

PR #1508, opened 2017-09-28
Enabling tracing in grpc versions that have tracing turned off.

sduskis:
Early versions of grpc don't have this setEnabled() API, and future versions won't either.  This change leverages reflection to allow us to trace with grpc 1.6, and will future-proof and past-proof the client.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -18,6 +18,7 @@ package com.google.cloud.bigtable.grpc;
 
 import java.io.Closeable;
 import java.io.IOException;
+import java.lang.reflect.Method;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.security.GeneralSecurityException;
@@ -59,6 +60,7 @@ import io.grpc.Channel;
 import io.grpc.ClientInterceptor;
 import io.grpc.ClientInterceptors;
 import io.grpc.ManagedChannel;
+import io.grpc.internal.AbstractManagedChannelImplBuilder;
 import io.grpc.internal.DnsNameResolverProvider;
 import io.grpc.internal.GrpcUtil;
 import io.grpc.netty.GrpcSslContexts;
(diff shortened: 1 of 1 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -142,6 +144,13 @@ private static void performWarmup() {
     connectionStartupExecutor.execute(new Runnable() {
       @Override
       public void run() {
+        // The first invocation of enableTracing does a reflection lookup which may be expensive.
+        enableTracing(null);
+      }
```
igorbernstein2: This is a bit confusing, I think it would be better to pay the cost of reflection than to try to prefetch the reflection methods.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -180,6 +189,30 @@ public void run() {
...
+    if (setEnableTracingMethodNotFound == null) {
+      try {
+        setEnableTracingMethod =
+            AbstractManagedChannelImplBuilder.class.getMethod("setEnableTracing", boolean.class);
+        setEnableTracingMethodNotFound = false;
+      } catch (Exception e) {
+        setEnableTracingMethodNotFound = true;
+      }
+    }
+    if (builder != null && setEnableTracingMethod != null) {
```
igorbernstein2: what if setEnableTracingMethodNotFound = false?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -180,6 +182,24 @@ public void run() {
...
+    try {
+      setEnableTracingMethod =
+          AbstractManagedChannelImplBuilder.class.getMethod("setEnableTracing", boolean.class);
+    } catch (NoSuchMethodException | SecurityException e1) {
+      return;
+    }
+    if (setEnableTracingMethod != null) {
+      try {
+        setEnableTracingMethod.invoke(builder, true);
+        System.out.println("ENABLING TRACING!!!!");
```
igorbernstein2: Please remove debug code & add a comment mentioning when this method came to be in grpc and when its no longer needed

PR #1509, opened 2017-10-02
Reverting some changes that broke head.

sduskis:
Puts and checkAnd* should throw IllegalArgumentException if the input is
incorrect.  A previous change (PR #1506) made those scenarios start
throwing IOExceptions. This PR reverts to the old behavior.


bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableTable.java
```diff
@@ -563,7 +565,7 @@ public Result increment(Increment increment) throws IOException {
 
   private IOException logAndCreateIOException(String type, byte[] row, Throwable t) {
     LOG.error("Encountered exception when executing " + type + ".", t);
-    return new IOException(
+    return new DoNotRetryIOException(
```
igorbernstein2: I dont quite understand the switch from IOException -> DoNotRetryIOException

PR #1518, opened 2017-10-23
WIP: Adding a shaded tracing artifact.

sduskis:
This artifact is about 18 MB, which isn't ideal, but it works.


sduskis:
At this point, we can't check this in.  Another shaded artifact is very confusing.  However, this means that the 1.0.0-pre4 release will only have tracing support for environments where HBase is not shaded.

sduskis:
This functionality will be available via bigable-hbase-1.x-shaded directly

PR #1522, opened 2017-10-25
Adding depracation warnings to Dataflow artifacts.

sduskis:


igorbernstein2:
LGTM, I would update the pom's description as well

sduskis:
I updated pom.xml descriptions and the readme.

PR #1530, opened 2017-11-07
Use non-deprecated Futures.addCallback function

rniemo-g:
This PR changes which Futures.addCallback() overload we call. This is because the overload we're using is being deprecated.


sduskis:
Thanks for the PR!

PR #1534, opened 2017-11-17
Improving tracing spans

sduskis:
- Using Closable where possible
- BulkMutations now report number of mutations
- Scans now report the number of rows per response and processing time.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RetryingReadRowsOperation.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RetryingReadRowsOperation.java
@@ -28,6 +28,7 @@ import com.google.cloud.bigtable.config.RetryOptions;
 import com.google.cloud.bigtable.grpc.async.AbstractRetryingOperation;
 import com.google.cloud.bigtable.grpc.async.BigtableAsyncRpc;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableMap;
 import com.google.protobuf.ByteString;
 
 import io.grpc.CallOptions;
@@ -37,6 +38,7 @@ import io.grpc.Status;
 import io.grpc.stub.ClientCallStreamObserver;
 import io.grpc.stub.ClientResponseObserver;
 import io.grpc.stub.StreamObserver;
+import io.opencensus.trace.AttributeValue;
 
 /**
  * An extension of {@link AbstractRetryingOperation} that manages retries for the readRows
@@ -169,8 +171,11 @@ public class RetryingReadRowsOperation extends
 
     ByteString previouslyProcessedKey = rowMerger.getLastCompletedRowKey();
 
+    operationSpan.addAnnotation("Got a response");
     // This may take some time. This must not block so that gRPC worker threads don't leak.
     rowMerger.onNext(message);
+    operationSpan.addAnnotation("Processed Response", ImmutableMap.of("rowCount",
+      AttributeValue.longAttributeValue(rowMerger.getRowCountInLastMessage())));
 
     ByteString lastProcessedKey = rowMerger.getLastCompletedRowKey();
     if (previouslyProcessedKey != lastProcessedKey) {
@@ -279,9 +284,8 @@ public class RetryingReadRowsOperation extends
     // Can this request be retried
     int maxRetries = retryOptions.getMaxScanTimeoutRetries();
     if (retryOptions.enableRetries() && ++timeoutRetryCount <= maxRetries) {
-      rpc.getRpcMetrics().markRetry();
       resetStatusBasedBackoff();
-      run();
+      performRetry(0);
     } else {
       throw getExhaustedRetriesException(Status.ABORTED);
     }
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RetryingReadRowsOperation.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RetryingReadRowsOperation.java
@@ -117,6 +117,7 @@ public class RetryingReadRowsOperation extends
   private int timeoutRetryCount = 0;
   private CallToStreamObserverAdapter<ReadRowsRequest> adapter;
   private StreamObserver<ReadRowsResponse> resultObserver;
+  private int totalRowsProcessed = 0;
 
   public RetryingReadRowsOperation(
       StreamObserver<FlatRow> observer,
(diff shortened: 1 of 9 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RetryingReadRowsOperation.java
```diff
@@ -169,8 +171,11 @@ public void onMessage(ReadRowsResponse message) {
 
     ByteString previouslyProcessedKey = rowMerger.getLastCompletedRowKey();
 
+    operationSpan.addAnnotation("Got a response");
     // This may take some time. This must not block so that gRPC worker threads don't leak.
     rowMerger.onNext(message);
+    operationSpan.addAnnotation("Processed Response", ImmutableMap.of("rowCount",
+      AttributeValue.longAttributeValue(rowMerger.getRowCountInLastMessage())));
```
igorbernstein2: How does this work with multiple responses? Do they accumulate or overwrite each other?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RowMerger.java
```diff
@@ -364,6 +364,7 @@ private FlatRow buildRow() {
   private ByteString lastCompletedRowKey = null;
   private RowInProgress rowInProgress;
   private boolean complete;
+  private int rowCountInLastMessage = -1;
```
igorbernstein2: should this be volatile?

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/BigtableWhileMatchResultScannerAdapter.java
```diff
@@ -33,6 +38,7 @@
  * @author sduskis
  * @version $Id: $Id
  */
+@SuppressWarnings("deprecation")
```
igorbernstein2: can you add a comment why we are suppressing deprecation as opposed to removing it?

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/BigtableWhileMatchResultScannerAdapter.java
```diff
@@ -53,32 +59,44 @@ public BigtableWhileMatchResultScannerAdapter(ResponseAdapter<FlatRow, Result> r
    * <p>adapt.</p>
    *
    * @param bigtableResultScanner a {@link com.google.cloud.bigtable.grpc.scanner.ResultScanner} object.
+   * @param span A parent span for the scan that needs to be closed when the
```
igorbernstein2: please finish the comment

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/BigtableWhileMatchResultScannerAdapter.java
```diff
@@ -53,32 +59,44 @@ public BigtableWhileMatchResultScannerAdapter(ResponseAdapter<FlatRow, Result> r
...
         if (!hasMatchingLabels(row)) {
+          close();
           return null;
         }
 
         return rowAdapter.adaptResponse(row);
       }
 
+      protected void closeSpan() {
+        try (NonThrowingCloseable s = Tracing.getTracer().withSpan(span)) {
```
igorbernstein2: s/NonThrowingCloseable/Closeable

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/BigtableResultScannerAdapter.java
```diff
@@ -50,15 +53,19 @@ public BigtableResultScannerAdapter(ResponseAdapter<T, Result> rowAdapter) {
    * @return a {@link org.apache.hadoop.hbase.client.ResultScanner} object.
    */
   public ResultScanner adapt(
-      final com.google.cloud.bigtable.grpc.scanner.ResultScanner<T> bigtableResultScanner) {
+      final com.google.cloud.bigtable.grpc.scanner.ResultScanner<T> bigtableResultScanner,
+      final Span span) {
```
igorbernstein2: Wouldn't make more sense to push this down into RetryingReadRowsOperation so that we don't have to duplicate this for callback api?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RetryingReadRowsOperation.java
sduskis: I'll add docs.  This is for a single response.  We probably should also have a different accumulated value.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RowMerger.java
sduskis: RowMerger should only be called from one thread.  We should have docs for that :)

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/BigtableWhileMatchResultScannerAdapter.java
sduskis: will do.  At a high level, NotThrowingClosable is compatible with old versions of opencensus.  Newer versions use something else. I wanted to make sure that we have the least chance of problems when working with other GCP libraries that may use older versions.

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/BigtableWhileMatchResultScannerAdapter.java
sduskis: NonThrowingCloseable does not throw exceptions, Closable does.  I can do a try/catch, which does nothing...

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/BigtableResultScannerAdapter.java
sduskis: This span is HBase specific as a wrapper for a Span created by RetryingReadRowsOperation.

sduskis:
I addressed all comments. PTAL.

PR #1559, opened 2017-12-11
Slight ReadRowsAcceptanceTest change

sduskis:
ReadRowsAcceptanceTest now creates a new ReadRowsResponse per chunk.  This is useful to ensure that RowMerger does the right thing across ReadRowsResponses.


igorbernstein2:
LGTM but in the future we should also test multiple chunks per response

PR #1560, opened 2017-12-11
Adding grpc and tcnative to bigtable-hbase-[12].x

sduskis:
The additional copies of grpc and tcnative should help solve some cases of diamond dependency problems we found with stackdriver tracing integration.

### DIFF
--- a/bigtable-hbase-parent/bigtable-hbase-1.x/pom.xml
+++ b/bigtable-hbase-parent/bigtable-hbase-1.x/pom.xml
@@ -39,6 +39,51 @@ limitations under the License.
       <version>${project.version}</version>
     </dependency>
 
+    <!-- GRPC and tcnative -->
+    <!-- 
+        These dependencies are copied to bigtable-hbase-1.x and bigtable-hbase-2.x.
+
+        The copy has to happen to ensure consistency of all io.netty and io.grpc versions
+        even if there is a different transitive dependency of grpc from other clients.
+        Specifically this should avoid problems when stackdriver tracing is used in
+        conjunction with Cloud Bigtable.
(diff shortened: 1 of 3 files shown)


bigtable-hbase-parent/bigtable-hbase-1.x/pom.xml
```diff
@@ -39,6 +39,51 @@ limitations under the License.
...
+        <groupId>io.grpc</groupId>
+        <artifactId>grpc-stub</artifactId>
+        <version>${grpc.version}</version>
+    </dependency>
+
+    <dependency>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-tcnative-boringssl-static</artifactId>
+        <version>${netty-tcnative-boringssl-static.version}</version>
+    </dependency>
```
igorbernstein2: Please add an end marker for the copied section

bigtable-hbase-parent/bigtable-hbase-2.x/pom.xml
```diff
@@ -55,6 +55,52 @@ limitations under the License.
...
+        <artifactId>grpc-stub</artifactId>
+        <version>${grpc.version}</version>
+    </dependency>
+
+    <dependency>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-tcnative-boringssl-static</artifactId>
+        <version>${netty-tcnative-boringssl-static.version}</version>
+    </dependency>
+
```
igorbernstein2: Please add an end marker comment here as well

sduskis:
The end markers were added.  Thanks for the revew!

PR #1563, opened 2017-12-14
common module updates to support HBase2.

spollapally:
2 Commits. First one has changes to common modules.


spollapally:
Submitting these changes in multiple smaller commits.

PR #1564, opened 2017-12-18
Common component changes and utilities for Hbase2 async.

spollapally:
Changes in TableAdapter are to allow extensions to work-around hbase-2 binary compatibility issues.

### DIFF
--- a/bigtable-hbase-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/FutureUtils.java
+++ b/bigtable-hbase-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/FutureUtils.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.hbase2_x;
+
+import java.util.concurrent.CompletableFuture;
+import com.google.common.util.concurrent.FutureCallback;
+import com.google.common.util.concurrent.Futures;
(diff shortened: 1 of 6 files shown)


bigtable-hbase-parent/bigtable-hbase-2.x/src/main/java/org/apache/hadoop/hbase/client/BigtableAsyncRegistry.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+package org.apache.hadoop.hbase.client;
+
+import java.util.concurrent.CompletableFuture;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.RegionLocations;
+import org.apache.hadoop.hbase.ServerName;
+
+/**
+ * Bigtable implementation of {@link AsyncRegistry}}. Hard codes required values as corresponding Bigtable
```
sduskis: Can you please add a bit more comments here about why this is necessary?

Also, `{@link AsyncRegistry}}` should be `{@link AsyncRegistry}`

bigtable-hbase-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/FutureUtils.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+package com.google.cloud.bigtable.hbase2_x;
+
+import java.util.concurrent.CompletableFuture;
+import com.google.common.util.concurrent.FutureCallback;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+
+/**
+ * Utility methods for converting Future types
```
sduskis: Can you please add a bit more comments about what frameworks are being translated, and perhaps why we need this?

bigtable-hbase-parent/bigtable-hbase-2.x/src/test/java/com/google/cloud/bigtable/hbase2_x/FutureUtilsTest.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.concurrent.TimeoutException;
+import org.hamcrest.core.IsInstanceOf;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListeningExecutorService;
+import com.google.common.util.concurrent.MoreExecutors;
+
+public class FutureUtilsTest {
```
sduskis: Please add a simple class comment that this is testing `{@link FutureUtils}`

PR #1568, opened 2017-12-18
Hbase2 async shell classes and changes to accommodate integration tests.

spollapally:
This commit has enough changes to create and integration test an AsyncConnection, very little feature code beyond that, mostly plumbing. Sets the stage to incrementally add support for Bigtable supported operations.

### DIFF
--- a/bigtable-hbase-parent/bigtable-hbase-integration-tests-2.x/src/test/java/com/google/cloud/bigtable/hbase/test_env/BigtableEnv.java
+++ b/bigtable-hbase-parent/bigtable-hbase-integration-tests-2.x/src/test/java/com/google/cloud/bigtable/hbase/test_env/BigtableEnv.java
@@ -18,7 +18,6 @@ package com.google.cloud.bigtable.hbase.test_env;
 import java.io.IOException;
 import java.util.Map.Entry;
 import java.util.Set;
-import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
@@ -112,8 +111,8 @@ class BigtableEnv extends SharedTestEnv {
     return ConnectionFactory.createConnection(configuration);
   }
   
-  public CompletableFuture<AsyncConnection> createAsyncConnection() throws IOException {
-    return ConnectionFactory.createAsyncConnection(configuration);
+  public AsyncConnection createAsyncConnection() throws Exception {
+    return ConnectionFactory.createAsyncConnection(configuration).get();
   }
 
 }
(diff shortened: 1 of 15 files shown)


bigtable-hbase-parent/bigtable-hbase-integration-tests-2.x/src/test/java/com/google/cloud/bigtable/hbase/test_env/BigtableEnv.java
```diff
@@ -98,4 +111,9 @@ protected void teardown() {
   public Connection createConnection() throws IOException {
     return ConnectionFactory.createConnection(configuration);
   }
+  
+  public CompletableFuture<AsyncConnection> createAsyncConnection() throws IOException {
```
sduskis: Why not return `AsyncConnection` instead of `CompletableFuture<AsyncConnection>`?

bigtable-hbase-parent/bigtable-hbase-integration-tests-2.x/src/test/java/com/google/cloud/bigtable/hbase/test_env/MiniClusterEnv.java
```diff
@@ -68,4 +70,19 @@ public Connection createConnection() throws IOException {
 
     return ConnectionFactory.createConnection(clientConfig);
   }
+
+  @Override
+  public CompletableFuture<AsyncConnection> createAsyncConnection() throws IOException {
```
sduskis: why not return `AsyncConnection` here?

bigtable-hbase-parent/bigtable-hbase-integration-tests-2.x/src/test/java/com/google/cloud/bigtable/hbase/test_env/BigtableEnv.java
spollapally: This would provide a way to access the future (CF) object if in case we need during tests. I do not yet have a concrete case where this might be required. I am OK to change it to AsyncConnection if you prefer.

bigtable-hbase-parent/bigtable-hbase-integration-tests-2.x/src/test/java/com/google/cloud/bigtable/hbase/test_env/MiniClusterEnv.java
spollapally: This would provide a way to access the future (CF) object if in case we need during tests. I do not yet have a concrete case where this might be required. I am OK to change it to AsyncConnection if you prefer.

spollapally:
Committed a fix for the exception messages.

bigtable-hbase-parent/bigtable-hbase-integration-tests-2.x/src/test/java/com/google/cloud/bigtable/hbase/test_env/MiniClusterEnv.java
sduskis: YAGNI.  I don't ever see a need for a CF.  If we need it, we can add CFs later.

PR #1569, opened 2017-12-19
app profiles

igorbernstein2:


igorbernstein2:
Microbenchmark of request rewriting is on the order of 25 ns on my workstation

igorbernstein2:
Ok, this is ready for review

PR #1575, opened 2017-12-21
 Incremental support for BigtableAsyncAdmin table management methods. 

spollapally:


bigtable-hbase-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/BigtableAsyncAdmin.java
```diff
@@ -63,51 +78,148 @@
...
+        } else {
+          result.add(tableAdapter2x.adapt(table));
+        }
+      }
+      return result;
+    });
+  }
+
+  private CompletableFuture<List<Table>> requestTableList() {
+    return CompletableFuture.supplyAsync(() -> {
```
sduskis: can't the builder be built outside of the CompletableFuture?

spollapally: It can execute outside in the calling thread itself, and any penalty is most likely negligible. The supplyAsync chained with thenCompose seems to provide a decent way to make most of the code async.

bigtable-hbase-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/BigtableAsyncAdmin.java
```diff
@@ -63,51 +78,148 @@
...
   }
 
   @Override
   public CompletableFuture<Void> createTable(TableDescriptor desc, Optional<byte[][]> splitKeys) {
-    throw new UnsupportedOperationException("createTable"); // TODO
+    if (desc.getTableName() == null) {
+      return FutureUtils.failedFuture(new IllegalArgumentException("TableName cannot be null"));
+    }
+
+    return CompletableFuture.supplyAsync(() -> {
```
sduskis: Anything before `FutureUtils.toCompletableFuture(bigtableTableAdminClient.createTableAsync(r.build())))`
should probably be outside of Completable future.  I found the notion of creating the request in a `CompleteableFuture` superfluous.

bigtable-hbase-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/BigtableAsyncAdmin.java
```diff
@@ -63,51 +78,148 @@
...
+        .thenAccept(r -> {
+        });
   }
 
   @Override
   public CompletableFuture<Void> createTable(TableDescriptor desc, byte[] startKey, byte[] endKey,
       int numRegions) {
-    throw new UnsupportedOperationException("createTable"); // TODO
+
+    return CompletableFuture.supplyAsync(() -> {
```
sduskis: why not build the request, and then just call `return createTable(desc, skeys)`?  Why do you need a CompletableFuture?

bigtable-hbase-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/BigtableAsyncAdmin.java
```diff
@@ -63,51 +78,148 @@
...
+      }
+      return splitKeys;
+    }).thenCompose(skeys -> createTable(desc, skeys));
   }
 
   @Override
   public CompletableFuture<Void> disableTable(TableName tableName) {
-    throw new UnsupportedOperationException("deleteTable"); // TODO
+    CompletableFuture<Void> cf = new CompletableFuture<>();
+    tableExists(tableName).thenAccept(exists -> {
```
sduskis: why not `thenCompose`?

bigtable-hbase-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/BigtableAsyncAdmin.java
```diff
@@ -63,51 +78,148 @@
...
+      cf.complete(null);
+    });
+
+    return cf;
   }
 
   @Override
   public CompletableFuture<Void> deleteTable(TableName tableName) {
-    throw new UnsupportedOperationException("deleteTable"); // TODO
+    return CompletableFuture.supplyAsync(() -> {
```
sduskis: Again, I think I prefer the builder outside of the CompletableFuture.

bigtable-hbase-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/BigtableAsyncAdmin.java
sduskis: or maybe just return null?

bigtable-hbase-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/BigtableAsyncAdmin.java
spollapally: Will remove this pattern. This change can cause exceptions to be thrown out of the CF model which clients will have to account for, but let's see how that will go.

bigtable-hbase-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/BigtableAsyncAdmin.java
spollapally: This is one of the complexity with the CF framework :) thenCompose is for cases where a CF is propagated down the chain into a another CF. I believe thenAccept (or theRun) variations are the correct forms here which take no params and return nothing.

bigtable-hbase-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/BigtableAsyncAdmin.java
sduskis: Oh... I see.  This is for exception handling.  That makes more sense.  Ok.  You can keep the pattern.  Can you please add a comment somewhere about why this pattern is used?

spollapally: Yes. I just verified not wrapping in CF causes test failures for hbaseLocalMiniClusterTestH2 profile in exception cases. Will leave it, and add some comments.

spollapally:
Added additional comments as per review.

PR #1581, opened 2017-12-22
WIP: Adding a new RowFilter builder for a more readable Filter API.

sduskis:


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/filter/RowFilters.java
```diff
@@ -0,0 +1,267 @@
...
+      this.endEdgeType = EdgeType.Closed;
+      return this;
+    }
+
+    public RowFilter build() {
+      return R.columnRange(family, start, startEdgeType, end, endEdgeType);
+    }
+  }
+
+  public final static class R {
```
igorbernstein2: I would recommend to make this a static instance instead of a static class so that users can alias it

Issue #1583, opened 2017-12-26
TableAdapter2x does not convert GcRules


sduskis:
This code is a bit broken:

```java
  public Table adapt(TableDescriptor desc) {
    Map<String, ColumnFamily> columnFamilies = new HashMap<>();
    for (ColumnFamilyDescriptor column : desc.getColumnFamilies()) {
      try {
        String columnName = column.getNameAsString();
        // TODO - This is a very shallow copy. See class comments for further details  
        HColumnDescriptor hColumnDescriptor = new HColumnDescriptor(columnName);
        ColumnFamily columnFamily = columnDescriptorAdapter.adapt(columnName, ).build();
        columnFamilies.put(columnName, columnFamily);
      } catch (Exception e) {
        LOG.error("Failed to copy ColumnFamilyDescriptor to HColumnDescriptor", e);
        throw new RuntimeException("Failed to copy ColumnFamilyDescriptor to HColumnDescriptor", e);
      }
    }
    return Table.newBuilder().putAllColumnFamilies(columnFamilies).build();
  }
}
```

`ColumnFamilyDescriptor` has garbage collection properties that need to be added to the `ColumnFamily`.  At this point, the conversion from HBase to Bigtable garbage collection policies does not happen.  That needs to be fixed.

spollapally:
Will expand the copy by calling the getters and then setters. That's probably the cleanest way for now.

PR #1584, opened 2017-12-26
Fix ColumDescriptor copy to HColumnDescriptor. Issue #1583

spollapally:


sduskis:
This should fix issue #1583.

sduskis:
It looks like `BigtableAdmin` doesn't actually use this.  There are a bunch of the HBase2 `BigtableAdmin` methods that use `TableDescriptor`s that are not implemented yet.

PR #1587, opened 2017-12-27
Fixing a subtle issue in RefreshingOAuth2CredentialsInterceptor

sduskis:
The current rate limiter approach masks any legitimate unauthenticated status codes and instead replaces the legit message with the rate limiting error message. Changing the logic to rate limit the revocations rather than the token lookups, which should be a more efficient approach.


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -325,8 +326,13 @@ HeaderCacheElement syncRefresh() {
...
           @Override
           public HeaderCacheElement call() throws Exception {
-            HeaderCacheElement newToken = refreshCredentials();
-            return updateToken(newToken);
+            try {
+              HeaderCacheElement newToken = refreshCredentials();
+              return updateToken(newToken);
+            } finally {
+              futureToken = null;
+              isRefreshing = false;
```
igorbernstein2: How come this moved from updateToken? I think it needs to be part of that synchronization block

sduskis: Sorry about that. I reverted that change.

PR #1588, opened 2017-12-28
Consolidating Table and Instance admin hosts.

sduskis:


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
```diff
@@ -120,13 +115,8 @@ private Builder(BigtableOptions original) {
...
     }
 
-    public Builder setTableAdminHost(String tableAdminHost) {
-      this.tableAdminHost = tableAdminHost;
-      return this;
-    }
-
-    public Builder setInstanceAdminHost(String instanceAdminHost) {
-      this.instanceAdminHost = instanceAdminHost;
+    public Builder setAdminHost(String tableAdminHost) {
```
kevinsi4508: String adminHost

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -185,7 +185,7 @@ public BigtableSession(BigtableOptions opts) throws IOException {
         "Opening connection for projectId %s, instanceId %s, "
         + "on data host %s, table admin host %s.",
```
kevinsi4508: "admin host %s."

PR #1592, opened 2018-01-03
Update bigtable-hbase-integration-tests to have timeout of 1800 s instead of 600 s.

kevinsi4508:
Increase test timeout from 600 s to 1800 s


igorbernstein2:
Might be better to extract it to a property at the top:

 <properties>
  <test.timeout>1800</test.timeout>
  ..

then use it in plugin config:
<forkedProcessTimeoutInSeconds>${test.timeout}</forkedProcessTimeoutInSeconds>

kevinsi4508:
close RP will try again.

PR #1594, opened 2018-01-04
Increase junit timeout for IntegrationTests

kevinsi4508:
Increase junit timeout for IntegrationTests from 10 minutes to 30.


sduskis:
The 10 minutes limit is needed for most cases.  Can  you please use a different mechanism to make this work, like a system parameter to make this work instead of a hard coded 30 minutes?

kevinsi4508:
That parameter was hardcoded.  I am trying make the test go through first and then will think of a better way.

sduskis:
This change potentially invalidates the current configuration of our integration tests.  We need to revert this change.

kevinsi4508:
Why would increasing the junit timeout invalidates the configuration of the integration tests? Can you provide details? Thanks!

sduskis:
We have to take this offline.

PR #1599, opened 2018-01-04
AbstractTests should re-use the IntegrationTest env

sduskis:
Before this change, every test class would create its own table.  This change creates a single table for all tests.


igorbernstein2:
I'm a little concerned about sharing tables across tests. At the very least the data needs to be dropped.

sduskis:
I'll truncate the table between classes... hopefully that won't cause a problem.  Our tests originally ran with a single test.

FWIW, creating a table per class is problematic in the new world.

Ideally, we have a fresh table per test case, not test class.  I think that it's less significant to truncate the table between classes rather than between test cases.

sduskis:
The table is now truncated.  PTAL.

PR #1603, opened 2018-01-05
Improving performance for bulk writes with config changes.

sduskis:
Increasing the number of requests per bulk RPC and the number of concurrent RPCs per channel.  Also, increasing the number of channels to 4x CPU count.  These settings increased throughput in a large dataflow test by about 15% and the server-side BT CPU also dropped by a few percentage points.


spollapally:
LGTM

PR #1604, opened 2018-01-12
SingleColumnValueFilterAdapter now uses an interleave.

sduskis:
`SingleColumnValueFilterAdapter` is a relatively complex HBase filter.  There were circumstances where it creates nested `Condition`s, which is problematic on the server-side.  A user used multiple `SingleColumnValueFilterAdapter`s which in turn caused severe problems.  @dmmcerlean suggested an alternative that would use `Interleave`d `Condition`s instead of nested `Condition`s.  This PR implements the suggestion.  In addition to the implementation itself, this PR attempts to document the intent of `SingleColumnValueFilterAdapter` as well as how to convert the options therein to a Cloud Bigtable `RowFilter`.

In addition, `TestSingleColumnValueFilter` was updated to test new conditions, and all `Scan`s were constrained to only the range of rows created by the test.

Also, TestSnapshots used keys that contained `TestSingleColumnValueFilter`.  The keys were updated to use a prefix of `TestSnapshots` instead.

### DIFF
--- a/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/SingleColumnValueFilterAdapter.java
+++ b/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/SingleColumnValueFilterAdapter.java
@@ -4,9 +4,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -26,7 +26,10 @@ import org.apache.hadoop.hbase.util.Bytes;
 import com.google.bigtable.v2.RowFilter;
 import com.google.bigtable.v2.RowFilter.Chain;
 import com.google.bigtable.v2.RowFilter.Condition;
+import com.google.bigtable.v2.RowFilter.Interleave;
 import com.google.cloud.bigtable.util.ByteStringer;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.protobuf.ByteString;
 
 /**
  * Adapt SingleColumnValueFilter instances into bigtable RowFilters.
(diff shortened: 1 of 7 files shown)


bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/SingleColumnValueFilterAdapter.java
```diff
@@ -51,90 +57,171 @@ public SingleColumnValueFilterAdapter(ValueFilterAdapter delegateAdapter) {
...
-                .setFamilyNameRegexFilter(
-                    Bytes.toString(quoteRegularExpression(filter.getFamily()))))
-            .addFilters(RowFilter.newBuilder()
-                .setColumnQualifierRegexFilter(
-                    ByteStringer.wrap(quoteRegularExpression(filter.getQualifier()))))
-            .addFilters(createVersionLimitFilter(filter)))
-        .build();
-  }
+  @VisibleForTesting
+  static RowFilter getColumnSpecFilter(byte[] family, byte[] qualifier, boolean latestVersionOnly)
```
igorbernstein2: nit: Maybe return the Chain.Builder to avoid the toBuilder().addFilters().build() step?

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/SingleColumnValueFilterAdapter.java
```diff
@@ -51,90 +57,171 @@ public SingleColumnValueFilterAdapter(ValueFilterAdapter delegateAdapter) {
...
+   *
+   * <pre>
+   *   condition: {
+   *      predicate: {
+   *        chain: {
+   *           family: [filter.family]
+   *           qualifier: [filter.qualifier],
+   *           value: // something interesting
+   *           // if filter.latestOnly, then add
+   *           // cells_per_column: 1
```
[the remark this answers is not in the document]
sduskis: sgtm

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/SingleColumnValueFilterAdapter.java
sduskis: The Chain needs to be reused in the `!filter.getFilterIfMissing()` case with only the family/qualifier/timestamp check and no value check.  We have to build the chain twice.

PR #1608, opened 2018-01-18
Fixing TestTruncate for HBase minicluster

sduskis:
1. removing enableTable quickly after truncate, which seems to cause the HBase minicluster some weird issues
2. 1) required that the default table is no longer used, since the default table needs to be enabled for future operations
3. fixed a typo in the test
4. Updating the HBase 2.x TruncateTable test to be the same as the 1.x test
5. Fixed AbstractTest to disable the table before truncating the table so that the table will be truncated in HBase tests.  Give the problems in 1), truncate will likely be removed in a near term PR.


spollapally:
LGTM

PR #1609, opened 2018-01-18
Removing use of DnsNameResolverProvider from BigtableSession

sduskis:
DnsNameResolverProvider is used by default in ManagedChannelBuilder.  We
had some problems relating to shading at one point which broke the
default ManagedChannelBuilder configuration.  Those problems have been
resolved, so this PR removes the explicit use of DnsNameResolverProvider.


spollapally:
LGTM

PR #1618, opened 2018-01-23
Removing Dataflow v1 artifacts.

sduskis:
We added Beam (Dataflow v2) support in the Cloud Bigtable 1.0.0 release, and we deprecated the Dataflow v1 artifacts. Removing Dataflow v1 artifacts for the 1.1.0 release.


spollapally:
LGTM

Issue #1619, opened 2018-01-23
Remove opencensus-impl from shaded artifact


sduskis:
It looks like lmax disruptor creates non-daemon threads, which stops the JVM from shutting down.  Our shaded artifact includes opencensus-impl, so we'll have to remove it until opencensus-impl is safe to use.

sduskis:
See #1615 for more details.

PR #1621, opened 2018-01-23
Removing opencensus-impl from shaded dependencies

sduskis:
See issue #1619 for more details.


igorbernstein2:
I will try to reproduce #1619 tomorrow

igorbernstein2:
This is no longer necessary, it seems like the issue disappeared once you upgraded the deps in #1537

PR #1625, opened 2018-01-25
Adding an HBase Filter that wraps a Bigtable RowFilter.

sduskis:
This powertool can be used to create filters that work well in Bigtable's dialect that cannot be expressed easily in HBase's dialect.

Also, created unit tests for serialization of BigtableFilter and TimestampRangeFilter.


sduskis:
Ugh.  I merged this without review.  Sorry about that.

PR #1628, opened 2018-01-26
Only shade on "release" Profile

sduskis:
This saves 25 seconds per build for development purposes.

This only affects the `mapreduce` and `beam-import` projects.  The `hbase-shaded` and `hbase-hadoop` artifacts are still shaded.

### DIFF
--- a/bigtable-dataflow-parent/bigtable-beam-import/pom.xml
+++ b/bigtable-dataflow-parent/bigtable-beam-import/pom.xml
@@ -101,56 +101,6 @@ limitations under the License.
     </dependency>
   </dependencies>
 
-  <build>
-    <plugins>
-      <plugin>
-        <artifactId>maven-jar-plugin</artifactId>
-        <configuration>
-          <archive>
-            <manifest>
-              <mainClass>${mainClass}</mainClass>
-            </manifest>
-          </archive>
-        </configuration>
-      </plugin>
-
-      <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>exec-maven-plugin</artifactId>
-        <version>1.6.0</version>
-        <executions>
-          <execution>
-            <goals>
-              <goal>java</goal>
-            </goals>
-          </execution>
-        </executions>
-        <configuration>
-          <mainClass>${mainClass}</mainClass>
-        </configuration>
-      </plugin>
-
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-shade-plugin</artifactId>
-        <executions>
-          <execution>
-            <phase>package</phase>
-            <goals>
-              <goal>shade</goal>
-            </goals>
-          </execution>
-        </executions>
-        <configuration>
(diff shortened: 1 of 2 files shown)


igorbernstein2:
How would you feel about inverting this instead? By default it shaded, but if a property or environment variable or even some file is present it disables the shading.

My main concern is that this would be pretty surprising to a new developer on the project:  doing a maven install does not give you the same result as downloading a prebuilt jar

sduskis:
This change would only cause missing shaded artifacts in `mapreduce` and `import` which are meant for release only.  There are already missing artifacts (javadoc, sources) in non-`release` mode.

sduskis:
I updated my comments for clarification.

sduskis:
This change saves about 20 seconds per build.

igorbernstein2:
I understand that it only affects the leaf projects. But I still think this would very surprising to someone who would want to submit a PR for the beam-import pipeline or the mapreduce job and wants to test it against the dataflow service.

sduskis:
Ack.  That makes sense.  I still think that the time optimization is still worth the tradeoff.

Our pom.xml is pretty straight forward, so on the occasions that we need to change those leaf projects, a new developer would either have to be warned, or be temporarily surprised.

igorbernstein2:
Why not put it in its own profile with a default on activation:

```
<activation>
  <property>
    <name>bigtableSkipShade</name>
    <value>!true</value>
  </property>
</activation>
```

And add the property to your ~/.m2/settings.xml?

sduskis:
I definitely hear what you're saying.  That's a good idea.  Still, it would be great to speed up my builds, builds on CI and new developer's builds.  I'm still keen on doing this as the default behavior.

igorbernstein2:
I still disagree that this should be the default behavior. I think that better ways to speed up builds are:

- targeted builds: `mvn package -pl bigtable-hbase/bigtable-hbase-1.x -am` which builds the hbase client and all of its dependencies
- [parallel builds](https://cwiki.apache.org/confluence/display/MAVEN/Parallel+builds+in+Maven+3) `mvn -T1.5C package`

But if you feel strongly about this, I won't stand in the way

sduskis:
OK.  I'm going to punt on this.  I think that we can rearrange projects to make this a lot simpler.  Now, we need to build `bigtable-client-core`, `bigtable-hbase`, `bigtable-hbase-1.x`, `bigtable-hbase-1.x-shaded`, `bigtable-hbase-1.x-hadoop`, and `bigtable-hbase-integration-tests` for any interesting change to the system.  Maybe it's worth having a simple way to either build those or perform integrations tests with just `bigtable-hbase`.  I'll close this PR for now, and I'll revisit this issue for the next release.

igorbernstein2:
You can build all of those targets with : `mvn -pl bigtable-hbase-parent/bigtable-hbase-integration-tests -am clean package`

Running tests against unshaded bigtable-hbase will take a bit of work, I added a historical comment here: https://github.com/GoogleCloudPlatform/cloud-bigtable-client/issues/1630#issuecomment-361291752

PR #1631, opened 2018-01-28
Removing SLF4JBridgeHandler from SharedTestEnvRule

sduskis:
I'm not sure what this does for us.  It's likely an artifact of a historical problem that was resolved.


igorbernstein2:
I originally I added this to limit logging from the hbase minicluster, but I think you are right and this is no longer necessary. Please remove `jul-to-slf4j` from the pom.xml as well

PR #1632, opened 2018-01-28
Simplifying creation of Table objects in integration tests.

sduskis:
- Adding a getDefaultTable() method that returns the default table
- TestImport truncates the default table, so creating a new table table for it instead.


spollapally:
LGTM

PR #1639, opened 2018-02-01
Decoupling AbstractTest and IntegrationTests.

sduskis:
`IntegrationTets` needs to set up the shared test environment, and `AbstractTest` and its subclasses need to use that environment.  This PR adds a level of indirection for the shared test environment by adding a singleton for the shared test environment.  IntegrationTests will set up the singleton, and `AbstractTest` will get that singleton.

The goal is to keep `IntegrationTests` where they are, and move `AbstractTest` and common subclasses to bigtable-hbase-integration-tests-common.


igorbernstein2:
The original reason for creating the SharedEnvRule was to allow tests to run outside of the suite. This change and all of the upcoming changes will prevent that from working. So I think that SharedEnvRule has outlived its usefulness. To avoid confusion, it might make sense to demote SharedEnvRule to a simple singleton helper

sduskis:
That does make sense.  We still do need a `Rule` for creating and removing miniclusters.

Can we rethink `SharedEnvRule` in another PR?

PR #1646, opened 2018-02-02
expose app profiles in beam

igorbernstein2:


sduskis:
I don't think that we need this.  The import tool should remain as simple as possible.  Users can write a bit of code if they really know what they're doing with App Profiles.

igorbernstein2:
As it stands the sequence file reading & writing code is vendored in the importer and I don't want to encourage our users to use it (just yet). So it's not really feasible for users write their own sequence file pipelines just to specify an app profile

sduskis:
The app profile id should not be needed for the import tool.  The import tool is meant to populate the cluster from scratch.  I don't see why we can't just use the default app_id. 

We should talk offline to see if there's anything I'm missing here.

mbrukman:
Can the import tool be used to bulk import a large new table to an existing serving cluster? Wouldn't app profiles be useful in that case to direct traffic to a particular cluster?

PR #1653, opened 2018-02-07
Reducing test dependency on Cloud Bigtable artifacts.

sduskis:
- bigtable-hbase-integration-tests-common/pom.xml no longer has a dependency on CBT
- Added a copy of the CBT Logger to bigtable-hbase-integration-tests-common
- TestBasicOps used a guava Stopwatch that is specific to the CBT dependant guava version
- BigtableEnv no longer uses BigtableOptionsFactory (specific the repackaged version) for keys, but rather uses Strings directly
- Removed TestImport which isn't going to be implemented in CBT, and requires some hbase-server dependencies.  hbase-server dependencies will also be hidden, if possible.


spollapally:
Looks good.

PR #1658, opened 2018-02-08
Auth interceptor uses consistent value in getHeader()

sduskis:
This change replaces multiple uses of the volatile headerCache to a
single use.  It also removes the need to synchronize in getHeader().


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -258,37 +258,27 @@ HeaderCacheElement getHeaderSafe() {
...
+    switch (headerCacheUnsync.getCacheState()) {
+      case Good:
+        return headerCacheUnsync;
+      case Stale:
+        asyncRefresh();
+        return headerCacheUnsync;
+      case Expired:
+      case Exception:
+        // defer the future resolution (asyncRefresh will spin up a thread that will try to acquire the lock)
+        return syncRefresh();
```
igorbernstein2: should this be asyncRefresh()?

PR #1661, opened 2018-02-08
increase batch size for import job

igorbernstein2:


igorbernstein2:
After some testing, it appears that this doesn't net any performance gains. The new 125 element/batch limit in the 1.1.0 gives enough of a boost from the 25 in 1.0.0

PR #1667, opened 2018-02-13
Release scheduledFlush under a synchronized block

udoprog:
Since the write to `scheduledFlush` is neither volatile nor written under a synchronized block I don't believe it is guaranteed to be visible to all threads.

This could have the following effects:

* Automatic flushing for `BulkMutation` is stuck, no more writes get scheduled since no threads observe a `null` value for `scheduledFlush`.
* Another flush could be scheduled before `send` has returned.


nezdolik:
Waiting for new release of bt-client-core with this fix :)

sduskis:
Is this something that needs to be done today?  If so I can do it.  If not, can this wait until Monday?  We're working through new HBase2 async features.

nezdolik:
@sduskis we can wait till Monday. Thank you

sduskis:
I just released 1.2.0 earlier today.  I had some trouble with some other changes that went into the build.  Sorry for the delay

PR #1668, opened 2018-02-13
BigtableAsyncConnection should return async object implementations.

sduskis:
I cache AsyncConnection in SharedTestEnvRule, and reuse it across async tests.
Added builders to TestAsyncConnection.


spollapally:
LGTM

PR #1673, opened 2018-02-14
Adding Increment and Append to BigtableAsyncTable

sduskis:
I refactored TestBasicAsyncOps.  I also copied the most basic case from TestAppend and TestIncrement in TestBasicAsyncOps.


spollapally:
LGTM

PR #1675, opened 2018-02-15
Implement BigtableAsyncBufferedMutator.

sduskis:
I moved most of the code from BigtableBufferedMutator to a class called BigtableBufferedMutatorHelper.  BigtableBufferedMutatorHelper returns futures, and does all of the thigns both BigtableBufferedMutator and BigtableAsyncBufferedMutator need.  BigtableBufferedMutator deferes to BigtableBufferedMutatorHelper for everything but exception management.

BigtableAsyncBufferedMutator turned out to be simpler than BigtableBufferedMutator, but has some behavior changes for flush() and close().  Users need to wait for individual CompletableFutures to complete.


spollapally:
Code looks good, but doesn't look like we have test coverage for TestAsyncBufferedMutator yet..?

spollapally:
LGTM

PR #1682, opened 2018-02-16
Adding HBase 2 CheckAndMutate methods, including async.

sduskis:
- I moved quite a bit of Check and Mutate specific functionality from `AbstractBigtableTable` into a new class: `CheckAndMutateUtil`.
- I moved the HBase 2 `TestCheckAndMutate` into the common tests project
- I copied the `TestCheckAndMutate` into `TestCheckAndMutateHBase2` for the new HBase2 methods (Only switching `CompareFilter.CompareOp` with HBase2's `CopareOperator`.
- I copied `TestCheckAndMutateHBase2` into `TestAsyncCheckAndMutateHBase` where I used the fancy new async builder methods instead of the older version, otherwise keeping the tests the same.


spollapally:
LGTM

PR #1683, opened 2018-02-17
Filter sync

igorbernstein2:
Sync with the filters that are in the upcoming veneer client. The changes are functionally split between commits to ease review. I think #7 is the only controversial one. 

1. reorder classes to make it easier to see the rest of the changes
2. cosmetic changes (doesn't api)
  - javadoc updates
  - decorate args with @Nonnull & @Nullable
  - enforce nonnull with preconditions
  - fix Preconditions import
  - minor style changes
3. Adds a common range base class to make all ranges consistent and updates `QualifierRangeFilter`, `TimestampRangeFilter` & `ValueRangeFilter` to inherit from it.
4. Make all Filters cloneable instead of just ChainFilter
5. remove range(start,end) from TimestampRangeFilter in favor of range().of()
6. overwrites all of the FilterTests with ones from the veneer client
7. removes FamilyFilter#regex(ByteString)

Justification for removing FamilyFilter#regex(ByteString):
I think this is semantically wrong, family ids are strings and should be compared to strings directly.
Furthermore, family ids are a maximum of 64 characters and the optimization lowers the copies from 3 to 2., so I think the benefit is negligible.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/data/v2/wrappers/Filters.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/data/v2/wrappers/Filters.java
@@ -15,20 +15,21 @@
  */
 package com.google.cloud.bigtable.data.v2.wrappers;
 
-import com.google.api.client.repackaged.com.google.common.base.Preconditions;
 import com.google.api.core.InternalApi;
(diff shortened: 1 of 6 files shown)


igorbernstein2:
Lets wait until the the final changes get merged into veneer. (raw needs to be renamed)

sduskis:
I'm going to release a new client today or tomorrow.  I can get rid of raw without too much difficulty.  Can we please merge today?

PR #1690, opened 2018-02-20
Removing Filters.raw()

sduskis:
ScanAdapter creates its own Filter implementation that does the same
thing as Filters.raw.


igorbernstein2:
I think I prefer the original approach. The only thing that's currently in flux is what to call it. @pongad pointed out that it should have a different naming style then the rest of the filters and suggested fromRaw, I agree with the sentiment, but think fromProto would be a better name.

igorbernstein2:
ps. I'm probably going to mark the Filter interface as InternalExtensionOnly

sduskis:
`Filters.raw()` is now `Filters.fromProto()`

PR #1694, opened 2018-02-21
Fixing CheckAndMutate tests

sduskis:
There are 3 CheckAndMutate tests:
- TestCheckAndMutate
- TestCheckAndMutateHBase2
- TestAsyncCheckAndMutate

I moved the core logic into a new class AbstractTestCheckAndMutate.  Each of the core classes performs slightly different versions of check and mutate.  TestCheckAndMutate uses the hbase 1 methods.  TestCheckAndMutateHBase2 uses the new hbase 2 methods with CompareOperator.  TestAsyncCheckAndMutate uses AsyncTable. Each concrete class only implements the checkAnd* methods which are otherwise abstract in the parent class.

Fixed a bug in SharedTestEnvRule usage to only create a single AsyncConnection.


spollapally:
LGTM

PR #1699, opened 2018-02-28
Improving error handling of ReadRows.

sduskis:
There have been some reports of ReadRows getting ReadRowsResponses, but not registering it.  Towards that end, the following changes are occurring
1) RowMerger is now final.  Calling `reset()` instead of creating a new `RowMerger()`.
2) `responseObserver` is now final.
3) Bubbling `RuntimeExceptions` to users.


sduskis:
@spollapally @igorbernstein2: ping

spollapally:
LGTM.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -486,9 +492,10 @@ public ScanHandler readFlatRows(ReadRowsRequest request, StreamObserver<FlatRow>
   }
 
   private RetryingReadRowsOperation createReadRowsRetryListener(ReadRowsRequest request,
-      StreamObserver<FlatRow> observer) {
+      StreamObserver<FlatRow> rowObserver, StreamObserver<ReadRowsResponse> responseObserver) {
```
igorbernstein2: This looks really weird, it seems like we are breaking encapsulation here

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -454,15 +454,21 @@ public int available() {
...
+
+      @Override
+      public void onError(Throwable t) {
+      }
+
+      @Override
+      public void onCompleted() {
+      }
+    };
+    RetryingReadRowsOperation operation = createReadRowsRetryListener(request, reader, responseObserver);
```
igorbernstein2: why does this need to insert a listener in the middle of the callback chain? Can the queue reader manage its own queue?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RetryingReadRowsOperation.java
```diff
@@ -247,7 +250,9 @@ protected boolean isRequestRetryable() {
   @Override
   protected boolean onOK(Metadata trailers) {
     rowMerger.onCompleted();
-    completionFuture.set("");
+    if (!completionFuture.isDone()) {
```
igorbernstein2: This is unnecessary. A future's state can only be set once

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
```diff
@@ -186,17 +186,20 @@ public void onClose(Status status, Metadata trailers) {
       } else {
         onError(status, trailers);
       }
-    } catch (IOException e) {
-      // Ignore. This is thrown Closable.close() from the TRACER.withSpan(), which actually doesn't
-      // throw anything
+    } catch (Exception e) {
+      setException(e);
```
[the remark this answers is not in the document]
sduskis: This isn't about a trace exception.  This is about catching runtime exceptions.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
sduskis: Do you mean that that having a `StreamObserver<ReadRowsResponse>` is a problem in general, or that we set it in the constructor?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
sduskis: I renamed this method to createReadRowsOperation.  Is that better?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RetryingReadRowsOperation.java
sduskis: ack.  I removed it.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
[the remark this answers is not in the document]
igorbernstein2: I missed  It looked like this was trying to catch errors from finalizeStats, I missed the onOk & onError

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
igorbernstein2: Sorry, I didn't do a very good job communicating.

This PR promotes the raw response observer to the same level as the row one. The 2 observers listen for events from 2 different layers, so there some mixing of concerns here. Also, there is nothing to tell the reader why there are 2 observers and what each one is used for.  It's also a bit arbitrary that one of them nullable...there is nothing telling the reader that the response one is optional. Before the PR there was some separation: the primary row observer was a toplevel parameter, while the optional response observer was relegated to an optional setter. 

At the very least, adding some docs and maybe better method name would help. 

I'm not sure if this would improve the code,  but maybe it would make sense to convert the response observer into a less generic signal like a `onRowBufferEmpty` (not sure about the name) method on the row observer. Another option is that the end result of receiving a notification of a response is to insert a marker into the buffer alongside the logical rows. Maybe have the operation send the marker itself?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
igorbernstein2: Yes, I think that's a better name

sduskis:
@igorbernstein2, I replaced `StreamObserver<ReadRowsResponse>` with a new interface `ReadRowsResponseObserver` which can be implemented by any `StreamObserver<FlatRow>` implementation that's passed into `RetryingReadRowsOperation`.  It still feels a bit clunky, but it's definitely better than having a separate `StreamObserver<ReadRowsResponse>`

sduskis:
This PR got a bit out of hand.  I'm going to break it down into a couple of smaller PRs so that both I and reviewers can better focus on each individual change.

PR #1700, opened 2018-02-28
The read rows timeout is now set in teh right place

sduskis:
ResponseQueueReader needs to have the proper setting for partial read timeout.


spollapally:
LGTM

PR #1705, opened 2018-03-06
Add a helper method for asyncConfiguration

spollapally:


bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConfiguration.java
```diff
@@ -85,6 +90,18 @@ public static Configuration configure(Configuration conf, String projectId, Stri
     return conf;
   }
 
+  /**
+   * <p>Configuration for getting a org.apache.hadoop.hbase.client.AsyncConnection.</p>
+   * 
+   * @param conf a {@link org.apache.hadoop.conf.Configuration} object to configure.
+   * @return the modified {@link org.apache.hadoop.conf.Configuration} object.
+   */
+  public static Configuration asyncConfigure(Configuration conf) {
```
sduskis: Please just add these to `configure` for simplicity.

spollapally: One issue in using existing configure methods is, Project and Instance params are always required (will fail if they are null). In cases like OpenTsdb they are not explicitly configured. May be moving towards builder pattern will work better, but that will be a bigger change. What do you think?

sduskis: I see.  OK then.

PR #1706, opened 2018-03-07
Fixing Bigtable HBase 2 tests.

sduskis:


spollapally:
I see couple of new method impls including enableTable, but I didn't follow the fixed test case. Is the commit missing any files?

sduskis:
I'll add tests for enable/disable :)

sduskis:
There are two fixes:

1. Use the correct Preconditions class so that the HBase MiniCluster test passes.  The HBase Minicluster does not have the hbase shaded version of Preconditions in the classpath
2. Implement the version of createTable used by by some of the new tests.

sduskis:
@spollapally, I made a significant change to enable/disable table.  Can you please take another look?

spollapally:
Changes look good. I vaguely remember Hbase impl not throwing exceptions for negative cases - when disabling or enabling tables already in that state. We should see if these are passing the Hbase integration tests, if you haven't yet.

sduskis:
HBase 2 tests are in a really bad state right now.  I'm going to merge this PR, and try to fix the HBase tests in another PR.

spollapally:
Ok

PR #1708, opened 2018-03-07
Fixing HBase 2.* minicluster test

sduskis:
There were a few problems
- `IntegrationTests` did not include `TestListTables.class`
- There was a race condition when multiple subclasses of `AbstractTestListTables` ran together relating to listing tables via a pattern.
- Multiple async admin methods throw `UnsupportedOperationException` in HBase's `Admin` and `AsyncAdmin` implementations.  Those tests should only run in Bigtable mode.
- HBase minicluster gets into a really bad state when this is run `asyncAdmin.getDescriptor(null).get()`.  The minicluster never shuts down.  This test should only run in Bigtable mode.

@pawan-qlogic, FYI.


spollapally:
LGTM

PR #1710, opened 2018-03-08
Creating a CheckAndMutateBuilder

### LINKED ISSUE #1704: CheckAndMutate silently ignores condition for null values
Extracted from https://github.com/GoogleCloudPlatform/google-cloud-go/issues/621.

It seems like https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/5c3ee47e208b4efbfaa5badceff8b403c02cb6d4/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/CheckAndMutateUtil.java#L113

will replace the condition as always true for comparisons that involve `null`

sduskis:
There was logic duplication in AbstractBigtableTable and BigtableAsyncTable.  The new CheckAndMutateUtil RequestBuilder combines the logic for both implementations.
Also, this should fix #1704. When value is null, and the CompareOp was not '==' or '!=' would do absolutely nothing, when it should have been an "non-existence" check.

### DIFF
--- a/bigtable-hbase-parent/bigtable-hbase-integration-tests-common/src/main/java/com/google/cloud/bigtable/hbase/AbstractTestCheckAndMutate.java
+++ b/bigtable-hbase-parent/bigtable-hbase-integration-tests-common/src/main/java/com/google/cloud/bigtable/hbase/AbstractTestCheckAndMutate.java
@@ -361,13 +361,6 @@ public abstract class AbstractTestCheckAndMutate extends AbstractTest {
     success = checkAndPut(rowKey, SharedTestEnvRule.COLUMN_FAMILY, qualToCheck,
         CompareOp.GREATER_OR_EQUAL, new byte[]{Byte.MIN_VALUE}, someRandomPut);
     Assert.assertTrue("4000 > MIN_BYTES should succeed", success);
-
-    if (sharedTestEnv.isBigtable()) {
-      // This doesn't work in HBase, but we need this to work in CBT
-      success = checkAndPut(rowKey, SharedTestEnvRule.COLUMN_FAMILY, qualToCheck,
-        CompareOp.NOT_EQUAL, null, someRandomPut);
-      Assert.assertTrue("4000 != null should succeed", success);
-    }
   }
 
   @Test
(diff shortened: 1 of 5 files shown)


igorbernstein2:
Can we add an integration test that can run against the minicluster & bigtable verifying this behavior?

sduskis:
Sorry about the oversight.  I had a test ready to go in another branch.

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/CheckAndMutateUtil.java
```diff
@@ -71,67 +70,129 @@ public static boolean wasMutationApplied(
...
-        requestBuilder.addAllFalseMutations(mutations);
-      } else if (CompareFilter.CompareOp.NOT_EQUAL.equals(compareOp)) {
+    public RequestBuilder ifNotExists() {
+      Preconditions.checkState(compareOp == null,
+          "ifNotExists and ifMatches are mutually exclusive");
+      this.checkNonExistence = true;
+      return this;
+    }
+
+    public RequestBuilder ifEquals(CompareOp compareOp, byte[] value) {
```
igorbernstein2: I don't think this is used

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/CheckAndMutateUtil.java
```diff
@@ -71,67 +70,129 @@ public static boolean wasMutationApplied(
...
+      return ifMatches(CompareOp.EQUAL, value);
+    }
+
+    public RequestBuilder ifMatches(CompareOp compareOp, byte[] value) {
```
igorbernstein2: I think it would be cleaner to stick to a builder style and have `setOperator` & `setOperand` or go the other way make this a DSL and add the qualifier as a parameter

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/CheckAndMutateUtil.java
```diff
@@ -71,67 +70,129 @@ public static boolean wasMutationApplied(
...
+          "ifNotExists and ifMatches are mutually exclusive");
+
+      this.compareOp = Preconditions.checkNotNull(compareOp, "compareOp is null");
+
+      // TODO (issue #1704): only NOT_EQUALS should allow null.  Everything else should use ifNotExists();
+      this.value = value;
+      return this;
+    }
+
+    public CheckAndMutateRowRequest buildForPut(Put put) throws DoNotRetryIOException {
```
igorbernstein2: It might be better to split the 2 responsibilities:
`addPut` and `build()`

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/CheckAndMutateUtil.java
sduskis: HBase uses this approach, and I think I'd like to keep it for now.   I'd like to think through what a more powerful `CheckAndMutateRequest` builder would look like.  

The `with(Put|Delete|Mutation)` API allows for more than the single mutation, which is great.  It would be awesome to be able to add more conditions as well; something like:

```
builder
   .ifQualifierMatches(q1, ...)
   .ifQualifierNotExists(q2)
   .ifCondition(randomFilter)
`

File: bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/CheckAndMutateUtil.java (hunk 3 above)
Comment (sduskis) [reply bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/CheckAndMutateUtil.java]: ack

File: bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/CheckAndMutateUtil.java (hunk 1 above)
Comment (sduskis) [reply bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/CheckAndMutateUtil.java]: Thanks for the catch.

File: bigtable-hbase-parent/bigtable-hbase-integration-tests-common/src/main/java/com/google/cloud/bigtable/hbase/AbstractTestCheckAndMutate.java [hunk 4]
```diff
@@ -366,6 +370,54 @@ public void testCompareOps() throws Exception {
...
+
+    success = checkAndPut(rowKey, SharedTestEnvRule.COLUMN_FAMILY, nullQual,
+        CompareOp.EQUAL, null, someRandomPut);
+    Assert.assertTrue("== null should succeed", success);
+
+    success = checkAndPut(rowKey, SharedTestEnvRule.COLUMN_FAMILY, nullQual,
+        CompareOp.GREATER_OR_EQUAL, new byte[]{Byte.MIN_VALUE}, someRandomPut);
+    Assert.assertFalse("> MIN_BYTES should fail", success);
+
+    if (sharedTestEnv.isBigtable()) {
```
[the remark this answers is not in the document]
sduskis: ok

sduskis:
I addressed the comments.  As per the comment on the outdated code, I would like to keep `ifMatches` as it is to reduce the chances of problematic situations where a user doesn't set both `compareOp` and an explicit value selection.

sduskis:
For now, I'm going to ask that we stay consistent with HBase 2.*'s API, at least for now.  Is that ok?

Issue #1707, opened 2018-03-07
GetAdapter should support isCheckExistenceOnly()


sduskis:
If `Get. isCheckExistenceOnly()` returns true, then `GetAdapter` should create a filter similar to `AbstractBigtableTable.exists`.  Any explicit uses of this filter should then use `GetAdapter` instead of duplicating the same logic over and over again.

PR #1712, opened 2018-03-08
Support Get.isCheckExistenceOnly().

sduskis:
HBase 1.0 had a setting for only checking existence, which is intended to have the same functionality as `Table.exists(Get)`.  `GetAdapter` should convert `Get`s with that flag to appropriate Bigtable Filter objects.  In addition, HBase 1 and 2 exist checks should just create new `Get` objects with `setCheckExistenceOnly(true)`.
This should fix issue #1707.


spollapally:
LGTM

PR #1714, opened 2018-03-08
Retrying operations now setExceptions instead of silent failures

sduskis:


spollapally:
LGTM. Is it possible to add a test for this behavior, or is it difficult to create the test scenario?

sduskis:
These exceptions are incredibly difficult to test.  There's some unexplained behaviors that users are seeing, which may possibly mean that there are silent Exception failures.

Given that there's so much uncertainty, I have no idea how to test this.

PR #1715, opened 2018-03-08
RetryingReadRowsOperation uses more final variables

sduskis:
- RowMerger is final and now has a reset method
- `CallToStreamObserverAdapter` is now final, and use required no modification.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RowMerger.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RowMerger.java
@@ -366,7 +366,7 @@ public class RowMerger implements StreamObserver<ReadRowsResponse> {
   private ByteString lastCompletedRowKey = null;
   private RowInProgress rowInProgress = null;
   private boolean complete = false;
-  private int rowCountInLastMessage = -1;
+  private Integer rowCountInLastMessage = null;
 
   /**
    * <p>Constructor for RowMerger.</p>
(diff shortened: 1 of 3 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RowMerger.java
```diff
@@ -377,6 +377,13 @@ public RowMerger(StreamObserver<FlatRow> observer) {
     this.observer = observer;
```
igorbernstein2: Might better to call reset() in the constructor and remove the variable initializers to avoid duplicating initialization

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RowMerger.java
```diff
@@ -364,8 +364,8 @@ private FlatRow buildRow() {
 
   private RowMergerState state = RowMergerState.NewRow;
   private ByteString lastCompletedRowKey = null;
-  private RowInProgress rowInProgress;
-  private boolean complete;
+  private RowInProgress rowInProgress = null;
+  private boolean complete = false;
   private int rowCountInLastMessage = -1;
```
igorbernstein2: This is probably out of scope for this PR, so feel free to ignore this question, but I was wondering why this initialized to -1 instead of 0?

sduskis: It probably should be null and not either -1 or 0.  I'll update that.

sduskis:
I made some significant changes here to better reflect what `RowMerger.reset` was actually intended to do.

I also addressed the nits.

spollapally:
LGTM.

PR #1719, opened 2018-03-09
CheckAndMutate calls with Puts should have server-side timestamps

sduskis:
In general, Puts have client side-time stamps either set by the user or set by the CBT java client.  This is done so that if there are retries, the call is idempotent.  In the case of CheckAndMutate, there are some subtle issues when client-side timestamps are set.  Since Check and mutate is never retryable, the puts should have server-side timestamps.

In addition to the change, this PR introduces a TestCheckAndMutateUtil which tests for both the server-side timestamp condition as well as some other unit tests that should have been introduced earlier.


sduskis:
I think that the main website is the right place for that kind of cross-cutting best-practice type warning.  I don't mind adding some javadoc somewhere, but I don't think that will be enough to meet the documentation needs.

PR #1720, opened 2018-03-09
Upgrade to Beam 2.3

igorbernstein2:
* Fix destination resource id generation
* Upgrade to 2.3
* Update SequenceFileIO to use a constant dynamic destination


sduskis:
LGTM

PR #1722, opened 2018-03-11
Adding tests for RefreshingOAuth2CredentialsInterceptor

sduskis:
- Making more methods private and using other methods in RefreshingOAuth2CredentialsInterceptorTest.
- Moving variables to be more consistent.


spollapally:
LGTM - the reduced scopes, avoiding using syncRefresh directly, and the improved test cases look good. The overall functionality with ratelimited background token refresh through ClientCall interceptor is a bit dense :) Specially, this call. Is it adding a chain of interceptors till the refresh succeeds or fails on a stale token?

delegate().start(new UnAuthResponseListener<>(RefreshingOAuth2CredentialsInterceptor.this,
responseListener, headerCache), headers);

sduskis:
The `UnAuthResponseListener` is a way for us to find an unusual case that pops up.  We use OAuth2 tokens to establish authentication.  Most of the time, an OAuth token has to be refreshed before it expires.  Sometimes, the server-side can revoke the token before expiration and send `UNAUTHENTICATED` statuses.  `UnAuthResponseListener` will force a refresh of the token on `UNAUTHENTICATED`, but will only do that once every 10 seconds.

PR #1723, opened 2018-03-11
AbstractRetryingOperation's call is now private.

sduskis:
`call` should be read only.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RetryingReadRowsOperation.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RetryingReadRowsOperation.java
@@ -57,22 +57,22 @@ public class RetryingReadRowsOperation extends
 
     @Override
     public void onNext(ReadRowsRequest value) {
-      call.sendMessage(value);
+      getCall().sendMessage(value);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RetryingReadRowsOperation.java
```diff
@@ -148,15 +148,13 @@ protected ReadRowsRequest getRetryRequest() {
...
       // restart the clock.
-      synchronized (callLock) {
-        super.run();
-        // pre-fetch one more result, for performance reasons.
-        adapter.request(1);
-        if (rowObserver instanceof ClientResponseObserver) {
-          ((ClientResponseObserver<ReadRowsRequest, FlatRow>) rowObserver).beforeStart(adapter);
-        }
-        lastResponseMs = clock.currentTimeMillis();
+      super.run();
```
spollapally: is synchronized (callLock) not required anymore?

sduskis: Correct.  It's not needed anymore.  There were some strange cases where call was null, but it isn't ever null anymore with this change.

spollapally:
LGTM

Issue #1724, opened 2018-03-12
Restore Dataflow v1 support


sduskis:
Some users still need Dataflow v1 support.

PR #1725, opened 2018-03-13
Re-adding Dataflow v1.

sduskis:
A user is having technical difficulties with the Beam API, so re-adding `bigtable-hbase-dataflow`.  Hopefully, this can be removed towards the end of 2018.

This is a restore of [code that was deleted](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/tree/bigtable-client-1.0.0/bigtable-dataflow-parent/bigtable-hbase-dataflow) after the 1.0.0 release.

This should fix issue #1724.


spollapally:
LGTM - as a revert of removed module, with no dependencies on other modules. Is there anything in particular you want me to double check?

PR #1726, opened 2018-03-13
   Breaking down RefreshingOAuth2CredentialsInterceptor into two classes

sduskis:
`RefreshingOAuth2CredentialsInterceptor` managed oauth2 tokens asynchronously and integrated with grpc.  There was a lot of code associated with `RefreshingOAuth2CredentialsInterceptor`, and it is difficult to tell if everything is properly tested.  Testing of the grpc specific integration was incomplete.  To fix this problem, this PR breaks down `RefreshingOAuth2CredentialsInterceptor` into two components:
    
1. `RefreshingOAuth2CredentialsInterceptor` will only handle the grpc aspects.  It will act as a ClientCall.Listener that is invoked by a gRPC call chain.  It will delegate the details of managing the auth header to OAuthCredentialsStore.  It will also manage the functionality of handling UNAUTHENTICATED RPCs, including throttling

 2. `OAuthCredentialsStore` will handle the lifecycle of an OAuth2 token
    
    There should be no public facing behavioral changes or interface changes.  `RefreshingOAuth2CredentialsInterceptor` maintains its `asyncRefresh()` method and the `TIMEOUT_SECONDS` public variable.
    
    Missing tests were introduced, and some of the tests were refactored, but remained the same functionally.


spollapally:
LGTM

PR #1727, opened 2018-03-14
RefreshingOAuth2CredentialsInterceptor now honors CallOptions Deadlines

### LINKED ISSUE #1718: RefreshingOAuth2CredentialsInterceptor should respect deadlines
Currently the RefreshingOAuth2CredentialsInterceptor is hardcoded to wait up to 15 seconds for a background thread to refresh the token. Instead it should check if a user provided deadline is given in the CallOptions and use that if it's earlier than the hardcoded one

sduskis:
This should fix #1718.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
@@ -120,13 +120,13 @@ public class RefreshingOAuth2CredentialsInterceptor implements ClientInterceptor
 
       @Override
       public void start(Listener<RespT> responseListener, Metadata headers) {
-        int timeoutSeconds = TIMEOUT_SECONDS;
+        long timeoutMs = -1;
         if (callOptions.getDeadline() != null) {
-          long deadlineMs = callOptions.getDeadline().timeRemaining(TimeUnit.MILLISECONDS);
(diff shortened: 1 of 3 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/RefreshingOAuth2CredentialsInterceptor.java
```diff
@@ -119,7 +120,13 @@ public RefreshingOAuth2CredentialsInterceptor(ExecutorService scheduler, OAuth2C
 
       @Override
       public void start(Listener<RespT> responseListener, Metadata headers) {
-        HeaderToken token = store.getHeader(TIMEOUT_SECONDS);
+        int timeoutSeconds = TIMEOUT_SECONDS;
+        if (callOptions.getDeadline() != null) {
+          long deadlineMs = callOptions.getDeadline().timeRemaining(TimeUnit.MILLISECONDS);
+          int deadlineSeconds = (int) Math.round(deadlineMs / 1000.0);
```
igorbernstein2: I think the deadlines should be kept in ms

igorbernstein2: The timeout shouldn’t be > deadline. I think would be more appropriate

sduskis: Ack.

sduskis:
I updated the code to use Ms timeouts.  I also added some tests.  PTAL.

PR #1730, opened 2018-03-14
Adding maven-enforcer-plugin with dependency convergence to core

sduskis:
Ensuring that the plugin passes for bigtable-client-core via dependency exclusions.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/pom.xml
+++ b/bigtable-client-core-parent/bigtable-client-core/pom.xml
@@ -27,7 +27,7 @@ limitations under the License.
     <packaging>jar</packaging>
     <name>${project.groupId}:${project.artifactId}</name>
     <description>
-       This project contains artifacts bigtable client needs to interact with grpc.
+        This project contains artifacts bigtable client needs to interact with grpc.
     </description>
 
     <dependencies>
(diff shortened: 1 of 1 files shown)


bigtable-client-core-parent/bigtable-client-core/pom.xml
```diff
@@ -27,23 +27,35 @@ limitations under the License.
...
                 </exclusion>
                 <exclusion>
-                    <groupId>com.google.api</groupId>
-                    <artifactId>api-common</artifactId>
+                    <groupId>io.grpc</groupId>
+                    <artifactId>grpc-protobuf</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>io.grpc</groupId>
+                    <artifactId>grpc-stub</artifactId>
```
igorbernstein2: might be easier to exclude everything from io.grpc using a * wildcard for artifact ( same for below)

bigtable-client-core-parent/bigtable-client-core/pom.xml
```diff
@@ -118,6 +194,14 @@ limitations under the License.
...
                     <artifactId>guava-jdk5</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>com.google.http-client</groupId>
+                    <artifactId>google-http-client</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>com.google.http-client</groupId>
+                    <artifactId>google-http-client-jackson2</artifactId>
+                </exclusion>
```
igorbernstein2: How these deps need to pulled out/up? Does anything other than google-auth-library-oauth2-http use http-client? if not, it might better to remove the exclusions and the toplevel deps

Issue #1703, opened 2018-03-06
FuzzyRowFilter does not behave as expected on variable-length row keys


Naveg:
We have observed that FuzzyRowFilters work as we expect them to when the length of the row key matches the length of the fuzzy key/mask, and we use FuzzyRowFilters regularly on a table that has a fixed row key length.

We recently tried to use FuzzyRowFilters on a table that has a row keys of either 20 or 24 bytes (5 or 6 integers). Our fuzzy keys and masks were 16 bytes long, intended to be matched against the first 4 (prefix) integers of the row key. Specifically, we fixed the first and fourth integers in the filter, leaving the middle two as wildcards. HBase treated this filter as we expected it to, comparing the first 16 bytes of the row key with the fuzzy data. Bigtable's behavior did not match (in our case, no rows were returned).

I did not spend much time investigating how bigtable actually _did_ treat the filter, nor could I find much documentation. Looking at FuzzyRowFilterAdapter, it seemed like the intention is the behavior we expected.

sduskis:
Can you please produce a simple test case (assuming you already have a
Table) to help us debug this problem?


Solomon Duskis |  Google Cloud Bigtable Tech Lead |  sduskis@google.com |
914-462-0531

Naveg:
Here's a short test program - we have some scaffolding that injects the `Connection`, you'll want to replace with your own.

When run against HBase, outputs:

```
[1, 2, 3, 4, 5, 6]
[1, 9, 9, 4, 9, 9]
Matched 2 keys
```

When run against Bigtable, outputs:

```
Matched 0 keys
```

```java
public class FuzzyTest {
  private final Connection hbase;

  @Inject
  public FuzzyTest(Connection hbase) {
    this.hbase = hbase;
  }

  private byte[] createKey(int... values) {
    byte[] bytes = new byte[4 * values.length];
    for (int i = 0; i < values.length; i++) {
      System.arraycopy(Bytes.toBytes(values[i]), 0, bytes, 4 * i, 4);
    }
    return bytes;
  }

  private int[] parseKey(byte[] key) {
    int[] values = new int[key.length / 4];
    for (int i = 0; i < values.length; i++) {
      values[i] = Bytes.toInt(key, 4 * i);
    }
    return values;
  }

  public void run() throws Exception {
    String tableName = "fuzzy-test";
    String familyName = "data";
[...]

PR #1731, opened 2018-03-14
Added FuzzyRowFilter test case

pawan-qlogic:
Added test case for  FuzzyRowFilter with KnownGaps in TestFilters class.


sduskis:
This will help us begin addressing #1703.

Issue #1561, opened 2017-12-12
Turn on cached channel pooling by default


sduskis:
Cloud Bigtable should have `BigtableOptions.useCachedDataPool = true` by default.

See #1540

sduskis:
I'm going to close this, since we're going to transition to google-cloud-java, and it won't have cached data pool.  @igorbernstein2, it's something to think about.

igorbernstein2:
I would like to pull this out of the client and push it into our beam connector

PR #1736, opened 2018-03-17
#1561-Turn on cached channel pooling by default

rupeshitpatekarq:
Cached channel pooling by default set to 'True'.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
@@ -87,7 +87,7 @@ public class BigtableOptions implements Serializable {
 
     private BulkOptions bulkOptions;
     private boolean usePlaintextNegotiation = false;
-    private boolean useCachedDataPool = false;
+    private boolean useCachedDataPool = true;
 
     private RetryOptions retryOptions = new RetryOptions.Builder().build();
     private CallOptionsConfig callOptionsConfig = new CallOptionsConfig.Builder().build();
@@ -299,7 +299,7 @@ public class BigtableOptions implements Serializable {
       dataChannelCount = 1;
       instanceName = null;
       usePlaintextNegotiation = false;
-      useCachedDataPool = false;
+      useCachedDataPool = true;
 
       bulkOptions = null;
       callOptionsConfig = null;
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -237,10 +237,14 @@ public class BigtableSession implements Closeable {
     // Defer the creation of both the tableAdminClient until we need them.
     }
 
-  private ChannelPool getDataChannelPool() throws IOException {
+private ChannelPool getDataChannelPool() throws IOException {
     String host = options.getDataHost();
     int channelCount = options.getChannelCount();
-    if (options.useCachedChannel()) {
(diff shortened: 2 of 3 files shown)


rupeshitpatekarq:
I signed it!

rupeshitpatekarq:
I signed it!

sduskis:
Can you please rebase this branch?

rupeshitpatekarq:
I signed it!

rupeshitpatekarq:
I signed it!

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
```diff
@@ -87,7 +87,7 @@ private static int getDefaultDataChannelCount() {
 
     private BulkOptions bulkOptions;
     private boolean usePlaintextNegotiation = false;
-    private boolean useCachedDataPool = false;
+    private boolean useCachedDataPool = true;
```
sduskis: Please make this a static final value (USE_CACHE_DATA_POOL_DEFAULT).

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -237,10 +237,14 @@ public BigtableSession(BigtableOptions opts) throws IOException {
...
-  private ChannelPool getDataChannelPool() throws IOException {
+private ChannelPool getDataChannelPool() throws IOException {
     String host = options.getDataHost();
     int channelCount = options.getChannelCount();
-    if (options.useCachedChannel()) {
+    /**
+     * #1561-useCachedDataPool is now by default true hence for unit test cases to run successfully added null credential condition.
+     * As null credentials should not be used in production.  
+    */
+    if (options.useCachedChannel() && !CredentialOptions.nullCredential().equals(options.getCredentialOptions())) {
```
sduskis: What tests fail without this option?  It would probably be a good idea to fix those test cases.

rupeshitpatekarq: TestAppProfile is calling this method and as now we are setting options.useCachedChannel() value as true hence condition becomes true now & it is trying to load channelPool on localhost which is throwing following exception:
-----------------------------------------------------------------------------------------------------------------------------------------------------------
com.google.cloud.bigtable.grpc.async.AbstractRetryingOperation info
INFO: Retrying failed call. Failure #13, got: Status{code=UNAVAILABLE, description=io exception, cause=io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: localhost/127.0.0.1:37367
------------------------------------------------------------------------------------------------------------------------------------------------------------
In **TestAppProfile.java -> setup()** method I tried to set this **useCachedDataPool = false**  while building BigtableOptions to avoid above exception but this value will be overridden by our default value (true) and eventually above exception will come again.

Earlier as options.useCachedChannel() was false hence IF block never executed during test cases.

Hence to avoid exception I have added nullCredential() check because credentials will be null during test cases executions only.

Please let me know your thought on this.

sduskis: You've exposed a bug in `BigtableOptions` which is causing the problem in TestAppProfile.   I think that we should fix our Options classes. I'll create another bug for that.  We should fix `BigtableOptions` before merging this.

sduskis: BigtableOptions is fixed.  Please rebase, and see if ` !CredentialOptions.nullCredential().equals(options.getCredentialOptions())` can be removed

PR #1741, opened 2018-03-20
HBase 2 create table now works.

sduskis:
TableDescriptor cannot be cast from HTableDescriptor.  Convert TableDescriptors to CreateTableRequests.

### DIFF
--- a/bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
+++ b/bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
@@ -361,7 +361,7 @@ public abstract class AbstractBigtableAdmin implements Admin {
     try {
       bigtableTableAdminClient.createTable(builder.build());
     } catch (Throwable throwable) {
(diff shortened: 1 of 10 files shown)


bigtable-hbase-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -328,52 +329,80 @@ public void createTable(HTableDescriptor desc) throws IOException {
...
+  }
+
+  protected ListenableFuture<Table> createTableAsync(CreateTableRequest.Builder builder,
+      final TableName tableName) throws IOException {
+    builder.setParent(bigtableInstanceName.toString());
+    ListenableFuture<Table> future = bigtableTableAdminClient.createTableAsync(builder.build());
+    return toCreateTableSettableFuture(tableName, future);
+  }
+
+  public static ListenableFuture<Table> toCreateTableSettableFuture(final TableName tableName,
```
spollapally: Code looks good, handles re usability well. The functionality of these methods is not easy to follow without looking at the code though. Do you think renaming them a bit differently helps readability? Some thing like handle*..

spollapally:
LGTM

PR #1739, opened 2018-03-19
Remove duplication 1691

pawan-qlogic:
Not able to move 3 tests cases testTimestampRangeFilter, testBigtableFilter, testFuzzyDifferentSizes. If moved test cases fails in minicluster mode.

PR #1744, opened 2018-03-21
Cleaning up TestFilter tests.

sduskis:
There were some uncaught consequences due to PR #1739.  This PR improves `TestFilters` by removing some cruft.


spollapally:
LGTM. It might be a good idea to measure test coverage as we go through the refactoring. What do you think?

sduskis:
Code coverage is usually best in the same project with Unit tests.  We're pretty bad at unit tests.

PR #1758, opened 2018-03-23
Fixing BigtableOptions

sduskis:
PR #1736 shows that the current BigtableOptions code is brittle.  useCachedDataPool was not being transfered in the manual cloning process.  This PR is a proposal for a fix, and also a request for proposlas.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
@@ -32,7 +32,8 @@ import com.google.common.base.Strings;
  * @version $Id: $Id
  */
 //TODO: Perhaps break this down into smaller options objects?
-public class BigtableOptions implements Serializable {
+//TODO: This should be @Autovalue + Builder
+public class BigtableOptions implements Serializable, Cloneable {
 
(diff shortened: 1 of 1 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
```diff
@@ -544,4 +469,12 @@ public Builder toBuilder() {
   public boolean useCachedChannel() {
     return useCachedDataPool;
   }
+
+  protected BigtableOptions clone() {
+    try {
+      return (BigtableOptions) super.clone();
```
spollapally: Wouldn't this inadvertently update the reference type fields already in use before the clone, and effectively make them mutable?

igorbernstein2:
Have you thought about using auto value?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
sduskis: You mean objects like `CallOptionsConfig` and `RetryOptions`?  Those are supposed to be immutable.

sduskis:
@igorbernstein2, yup, I added a comment to that affect.

spollapally:
LGTM

PR #1761, opened 2018-03-25
Implemented few admin methods.

pawan-qlogic:
Implemented few admin methods. After your review comments, I can implement the same for the rest of the methods.


pawan-qlogic:
ok, will do.

Issue #1773, opened 2018-03-29
ChannelPool: least outstanding request algorithm


sduskis:
Our `ChannelPool` uses a round-robbin approach to selecting a `Channel`.  It's effective in most cases, but may cause problems in other cases.   A user found that one channel was slow and overloaded and caused various problems a "least outstanding request algorithm" will reduce problems like theirs in the future.

eiiches:
Hi,
I've been in contact with the Japanese support team about the issue.

Here's my implementation of the algorithm: https://github.com/eiiches/cloud-bigtable-client/commits/feature/least-outstanding-requests-balancing-channel-pool

We've load-tested the implementation, and it worked well, but I'm still worried that there may be some corner case that the number of outstanding requests is not correctly incr/decremented.

I can open a PR if you like, but it will probably take some time to sign a Corporate CLA. So I'd also be glad if someone can implement it independently.

All that said, I believe these kind of load balancing algorithms should be ideally provided by gRPC modules to avoid re-implementing it everywhere.

igorbernstein2:
Hi eiiches,

Thanks for sharing! I've sketched out an alternative approach here: https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/1774.

It should have a lower overhead to maintaining an ordered list of loaded connections while evenly spreading the load across the pool.

Also, GRPC actually has native [load balancing ](https://github.com/grpc/grpc/blob/master/doc/load-balancing.md), which we will switch to at some point in the future.

eiiches:
Wow, I like your approach! I'll switch to your implementation when it gets merged.
Thanks!

steveniemitz:
To chime in here, I recently spent time debugging a very-high p9999 latency in our requests.  We've been at a limited-load launch, so our QPS to bigtable has been fairly low and spread across a lot of instances of our service.  It turns out the QPS was so low that our connections would be closed due to inactivity by our outgoing NAT devices.

My solution was to simply reduce the data channel count to 1, but thats not great because now I need to remember to increase it (maybe?) as we ramp up traffic.

A cool solution might be to build a pool that dynamically expands/contracts based on the exponential moving average of the load over the channels.

sduskis:
[...]

PR #1774, opened 2018-03-30
wip: use random sampling for to avoid picking overloaded channels

igorbernstein2:
But wanted to present an alternative solution to https://github.com/GoogleCloudPlatform/cloud-bigtable-client/issues/1773#issuecomment-377513896.

This approach is a bit simpler: instead of maintaining a list of connections ordered by load, I think we can use a simpler approach of using random sampling to distribute the load.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
@@ -16,6 +16,7 @@
 package com.google.cloud.bigtable.grpc.io;
 
 import java.io.IOException;
+import java.util.Random;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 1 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/ChannelPool.java
```diff
@@ -258,9 +280,24 @@ public ChannelPool(ChannelFactory factory, int count) throws IOException {
   private ManagedChannel getNextChannel() {
     int currentRequestNum = requestCount.getAndIncrement();
     int index = Math.abs(currentRequestNum % channels.size());
+
+    int minLoad = channels.get(index).getOpenStreamCount();
+
+    if (minLoad > repickThreshold) {
+      for (int i = 0; i < repickThreshold; i++) {
```
evildour: I think this should use `repickChances` instead of `repickThreshold`

igorbernstein2: yep, sorry, this was meant to be a sketch. I will update it later on this week.
I would also like to benchmark it against the current round robin scheme and a sticky version of this approach: reuse the last channel if there are no outstanding requests.

sduskis:
Closing this for now. We may resurrect this later.

Issue #1755, opened 2018-03-22
Each integration test should be runnable without a Suite


sduskis:
Right now, `IntegrationTest` will do something like:

```
  static {
    SharedTestEnvRule.setInstance(new SharedTestEnvRule() {
      @Override
      /*
       * NOTE: This code is identical to the hbase 2.x code, but this produces binary incompatible
       * results.  This code has to remain in the 1.x integration testing code-base.
       */
      public void createTable(TableName tableName) throws IOException {
        try (Admin admin = getConnection().getAdmin()) {
          LOG.info("Creating table " + tableName.getNameAsString());
          HColumnDescriptor hcd = new HColumnDescriptor(COLUMN_FAMILY).setMaxVersions(MAX_VERSIONS);
          HColumnDescriptor family2 = new HColumnDescriptor(COLUMN_FAMILY2).setMaxVersions(MAX_VERSIONS);
          admin.createTable(
              new HTableDescriptor(tableName)
                  .addFamily(hcd)
                  .addFamily(family2));
        }
      }
    });
  }
```

Ideally, we should have a better way of doing this initialization.  I think that a well known class name (something like `com.google.cloud.bigtable.hbase.test_env.TableCreator`, would allow us to navigate the differences between HBase 1 and HBase 2 table creation without tying any tests to the IntegrationTest setup.

rahulKQL:
### Observations
It is possible to run the individual test cases for both 1.x & 2.x, but after running the tests we have to remove the added annotations/dependencies as those would **break the maven build**.

### Workaround
To run the individual test cases, we have to make following changes in their respective project/class:

#### Integration-tests-1.x
After adding below two changes, existing test cases are running fine: 
`@RunWith(JUnit4.class)` 

and 

```
  @ClassRule
  public static SharedTestEnvRule sharedTestEnvRule = SharedTestEnvRule.getInstance();
```

#### Integration-tests-2.x
Along with above annotations, it requires us to add these two dependencies:
```
        <dependency>
            <groupId>org.apache.hbase</groupId>
            <artifactId>hbase-server</artifactId>
            <version>${hbase.version}</version>
            <scope>test</scope>
        </dependency>
        <dependency>
            <groupId>org.apache.hbase</groupId>
            <artifactId>hbase-testing-util</artifactId>
            <version>${hbase.version}</version>
            <scope>test</scope>
        </dependency>
```

PR #1777, opened 2018-04-03
IntegrationTests no longer is required for running tests

sduskis:
Added a new TableCreator interface that would alleviate some of the problems in running individual tests.  This should work towards a solution for #1755.


bigtable-hbase-parent/bigtable-hbase-integration-tests-common/src/main/java/com/google/cloud/bigtable/hbase/test_env/SharedTestEnvRule.java
```diff
@@ -34,31 +34,36 @@
...
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+  }
   /**
    * This class is generally a singleton, where implementation can change.  Some startup utility
    * will set this instance.
    */
-  public static void setInstance(SharedTestEnvRule instance) {
+  public synchronized static void setInstance(SharedTestEnvRule instance) {
```
spollapally: Can we remove this setInstance()? That way this class can be a better singleton with one synchronized call.

sduskis:
@spollapally: done.  I removed setInstance()

spollapally:
LGTM

PR #1784, opened 2018-04-09
WIP - Change to gather mean from a smaller recent sample to better tr…

spollapally:
…ack server latency

Use throttlinClient for all async calls, except for scan calls, instead of dataClient.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/ResourceLimiter.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/ResourceLimiter.java
@@ -5,7 +5,6 @@ import com.google.api.client.util.NanoClock;
 import com.google.cloud.bigtable.config.Logger;
 import com.google.cloud.bigtable.grpc.BigtableSessionSharedThreadPools;
 import com.google.common.annotations.VisibleForTesting;
-
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/ResourceLimiter.java
```diff
@@ -92,6 +93,7 @@ public void markCanBeCompleted(long id) {
     Long start = starTimes.remove(opId);
     if (start != null) {
       stats.markRpcComplete(clock.nanoTime() - start);
```
sduskis: Is it safe to remove the DropWizard metrics?

sduskis:
@igorbernstein2, FYI.  @spollapally found a non-dropwizard approach that makes decisions on the last 1000 RPCs instead of a 1 minute average.  What are your thoughts?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/ResourceLimiter.java
spollapally: I think we need them, as they provide broader lifetime stats using samplings, opposed to the bounded recent time-window metrics we get using DescriptiveStatistics.

igorbernstein2:
I don't have strong feelings about using rpc vs time as an indicator. I think the new dependency should be shaded in the -shaded & -hadoop artifacts. I think that drop wizard metrics will need to be eventually replaced with opencensus, so decoupling it from throttling makes sense.

igorbernstein2:
Also, it might be good to take a look at how hbase deals with flow control. The config seems to have some notion of it via `hbase.region_client.inflight_limit` and `hbase.region_client.pending_limit`.

sduskis:
Closing this for now.  We may need to open this later.

Issue #1785, opened 2018-04-10
CreateTable flaky test: "a" + RandomStringUtils.random(10, false, false)


sduskis:
A table named `"a" + RandomStringUtils.random(10, false, false)` always fails in Cloud Bigtable, but will only fail sometimes in HBase.  The percentage of failed tests is pretty high, and it causes email overload.  There seems to be an inconsistency between allowed Bigtable names and allowed HBase names that is still TBD.

For now, turn off the test so that our CI tests don't fail for HBase.  Long term, we need to identify the inconsistency, document it, and properly test it.

Issue #1770, opened 2018-03-28
Fuzzy keys with integer values above 127 match no rows


Naveg:
As in #1703, we use keys that consist of multiple integers. We have found that with the 1.2.0 client, fuzzy filters are not correctly matching values above 127. This bug does not exist on version 1.0.0, where the filter works as expected. I did not try this with any other versions.

The sample program below should output all 8 keys. On version 1.0.0, it does. On version 1.2.0, it only outputs 4 keys - those with 126 or 127 in the 3rd position.

```java
public class FuzzyTest {
  private final Connection hbase;

  @Inject
  public FuzzyTest(Connection hbase) {
    this.hbase = hbase;
  }

  private byte[] createKey(int... values) {
    byte[] bytes = new byte[4 * values.length];
    for (int i = 0; i < values.length; i++) {
      System.arraycopy(Bytes.toBytes(values[i]), 0, bytes, 4 * i, 4);
    }
    return bytes;
  }

  private int[] parseKey(byte[] key) {
    int[] values = new int[key.length / 4];
    for (int i = 0; i < values.length; i++) {
      values[i] = Bytes.toInt(key, 4 * i);
    }
    return values;
  }

  public void run() throws Exception {
    String tableName = "fuzzy-test";
    String familyName = "data";
    HTableDescriptor tableDescriptor = new HTableDescriptor(TableName.valueOf(tableName));
    tableDescriptor.addFamily(new HColumnDescriptor(familyName));
    if (!hbase.getAdmin().tableExists(TableName.valueOf(tableName))) {
      hbase.getAdmin().createTable(tableDescriptor);
    }

    Table table = hbase.getTable(TableName.valueOf("fuzzy-test"));
    List<byte[]> keys = ImmutableList.of(
        createKey(5, 10, 126, 5),
        createKey(5, 11, 126, 7),
        createKey(5, 10, 127, 5),
        createKey(5, 11, 127, 7),
        createKey(5, 10, 128, 5),
        createKey(5, 11, 128, 7),
        createKey(5, 10, 129, 5),
        createKey(5, 11, 129, 7));
    List<Put> puts = keys.stream()
        .map(key -> new Put(key)
            .addColumn(Bytes.toBytes(familyName), Bytes.toBytes(0), Bytes.toBytes(0)))
        .collect(Collectors.toList());
    table.put(puts);

    // match keys with 5 in the first position and 126/127/128/129 in the 3rd position
[...]

PR #1786, opened 2018-04-10
Fixing some test

sduskis:
TestCreateTable failed in HBse mode in many cases, so removing the test and adding issue #1785.
TestFilters failed in Bigtable in mode, since testFuzzyWithIntKeys() fails as per issue #1770.  Marking that test as a KnownGap so that the CI doesn't continuously fail.


spollapally:
LGTM

PR #1790, opened 2018-04-12
WIP: Refactoring Async admin.

sduskis:


bigtable-hbase-2.x-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/BigtableAsyncAdmin.java
```diff
@@ -162,53 +156,43 @@ public BigtableAsyncAdmin(BigtableAsyncConnection asyncConnection) throws IOExce
...
-        cf.completeExceptionally(new TableNotFoundException(tableName));
+        throw new CompletionException(new TableNotFoundException(tableName));
       } else if (disabledTables.contains(tableName)) {
-        cf.completeExceptionally(new TableNotEnabledException(tableName));
+        throw new CompletionException(new TableNotEnabledException(tableName));
       } else {
         disabledTables.add(tableName);
         LOG.warn("Table " + tableName + " was disabled in memory only.");
-        cf.complete(null);
+        return null;
```
spollapally: Rest of the changes look good. Can you take another look at cases where a null is returned instead of CompletableFuture? I think this can create expectation issues for the caller.

sduskis: You're right.  We should never return null.

sduskis: On second though, we actually are returning a `CompletableFuture<Void>`, and this `return null` is from within the `thenApply`.  That's the actual intent.

We return null all of the time from all methods that return `CompletableFuture<Void>`.

PR #1797, opened 2018-04-20
Add extension point for channel creation

steveniemitz:
This simply adds a protected method to BigtableSession that allows hooking into the channel creation process.  

Previously `protected ChannelPool createChannelPool(final String hostString, int count)` existed but was insufficient, since the fields needed to actually create the channels correctly are private to `BigtableSession`.


steveniemitz:
Wow thanks for the fast turnaround!

PR #1799, opened 2018-04-23
Added RowCellUtil class to get cell value efficiently.

rupeshitpatekarq:
@sduskis , could you please review it.

### DIFF
--- a/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/RowCell.java
+++ b/bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/RowCell.java
@@ -21,7 +21,7 @@ import org.apache.hadoop.hbase.KeyValue;
 import org.apache.hadoop.hbase.KeyValue.Type;
 import org.apache.hadoop.hbase.util.Bytes;
 
-import com.google.cloud.bigtable.hbase.util.RowCellUtil;
+import com.google.cloud.bigtable.hbase.util.CellUtils;
 
 /**
  * This implementation of {@link org.apache.hadoop.hbase.Cell} is more efficient for Bigtable scanning than {@link org.apache.hadoop.hbase.KeyValue}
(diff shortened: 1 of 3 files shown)


sduskis:
We need something a bit different.  The interface would look like this:

```
class CellUtils {
  public static byte[] getValue(Cell cell) { ... } 
}
```

It would basically do the same thing that you do in `RowCellUtil.getValueArray`, but would not require a `RowCell` object.

rupeshitpatekarq:
@sduskis , Created CellUtils class.

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/RowCell.java
```diff
@@ -139,7 +141,7 @@ public long getSequenceId() {
   /** {@inheritDoc} */
   @Override
   public byte[] getValueArray() {
-    return this.valueArray;
+    return CellUtils.getValue(this, this.valueArray);
```
sduskis: This should still be `this.valueArray`.  There is no reason for CellUtils here.

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/util/CellUtils.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * <p> Utility methods helpful for slinging {@link Cell} instances.CellUtils class to provide utilities related to RowCell class. </p>
+ * 
+ * @author rupeshit
+ *
+ */
+public class CellUtils {
+
+  public static byte[] getValue(Cell cell,byte[] valueArray) {
```
sduskis: There's no reason to have valueArray here.

bigtable-hbase-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/util/CellUtils.java
```diff
@@ -0,0 +1,40 @@
...
+ * 
+ * @author rupeshit
+ *
+ */
+public class CellUtils {
+
+  public static byte[] getValue(Cell cell) {
+    if(cell.getValueLength() >= cell.getValueArray().length) {
+      return cell.getValueArray();
+    }else if(cell.getValueLength() < cell.getValueArray().length) {
```
sduskis: Why not just "else"? 

Also, please add a space between `if` and `(`.  There should also be a space between `{` and `else`

rupeshitpatekarq:
@sduskis , please have a look now.

PR #1805, opened 2018-05-01
Move some table config validation to runtime.

kevinsi4508:


bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableIO.java
```diff
@@ -1034,15 +1037,20 @@ public void populateDisplayData(DisplayData.Builder builder) {
...
-   public static PTransform<PCollection<KV<String, Iterable<Mutation>>>, PDone>
-      writeToMultipleTables(CloudBigtableConfiguration config) {
-    validateConfig(config);
-    return new CloudBigtableWriteTransform<>(new CloudBigtableMultiTableWriteFn(config));
+  public static PTransform<PCollection<KV<String, Iterable<Mutation>>>, PDone>
+      writeToMultipleTables(final CloudBigtableConfiguration config) {
+    return new CloudBigtableWriteTransform<KV<String, Iterable<Mutation>>>(
+        new CloudBigtableMultiTableWriteFn(config)) {
+      @Override
+      public void validate(PipelineOptions options) {
```
sduskis: Can you please put the validation code directly in the `CloudBigtableWriteTransform` class (it would also mean passing the config in the constructor)?

kevinsi4508:
Putting the validation code in CloudBigtableWriteTransform seems to the
right approach. There were two reasons I didn't end up doing that:
1. The constructor is public. I am not sure if it should be.
2. There are two different config classes. I guess we can call Validate ()
on the config class if it is defined.

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableIO.java
kevinsi4508: Putting the validation code in CloudBigtableWriteTransform seems to the
right approach. There were two reasons I didn't end up doing that:
1. The constructor is public. I am not sure if it should be.
2. There are two different config classes. I guess we can call Validate ()
on the config class if it is defined.

kevinsi4508: BTW, I made CloudBigtableWriteTransform private and passed the config through the constructor.
I also updated config classes to include validate() method.

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableIO.java
```diff
@@ -850,17 +854,16 @@ public void onException(RetriesExhaustedWithDetailsException exception,
...
     public CloudBigtableSingleTableBufferedWriteFn(CloudBigtableTableConfiguration config) {
       super(config);
-      tableName = config.getTableId();
     }
 
     @StartBundle
-    public synchronized void getBufferedMutator(StartBundleContext context)
-        throws IOException {
-      mutator = createBufferedMutator(context, tableName);
+    public synchronized void getBufferedMutator(StartBundleContext context) throws IOException {
```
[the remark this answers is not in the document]
kevinsi4508: done.

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableConfiguration.java
```diff
@@ -263,4 +264,14 @@ public void populateDisplayData(DisplayData.Builder builder) {
     }
   }
 
+  protected static void checkNotNullOrEmpty(String value, String type) {
```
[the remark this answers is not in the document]
kevinsi4508: done.

PR #1806, opened 2018-05-03
WIP, for feedback.

kevinsi4508:
We would like to get rid of Scan & ReadRowsRequest in CloudBigtableScanConfiguration.
We should have parameters in String or primitive types instead, just like in CloudBigtableTableConfiguration and CloudBigtableConfiguration.
The reason is that we need parameters to be access at runtime for templatization. Scan simply doesn't work because it is not even Serializable.

If we have to keep the existing Scan and ReadRowsRequest, this PR (not complete) is one way to have both. It is ugly.


sduskis:
Let's please not do this until we speak in person.

kevinsi4508:
WIP, for feedback. 
I wasn't planning to commit it.

PR #1813, opened 2018-05-11
Convert parameters in CloudBigtableConfiguration to runtime parameters

kevinsi4508:
Though I tried to have minimum change, 'request' object in CloudBigtableScanConfiguration needs to be converted to runtime parameter as well because it depends on project ID and instance ID.

### DIFF
--- a/bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
+++ b/bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
@@ -46,15 +46,15 @@ public class CloudBigtableScanConfiguration extends CloudBigtableTableConfigurat
   private static final long serialVersionUID = 2435897354284600685L;
 
   /**
-   * Converts a {@link CloudBigtableTableConfiguration} object to a {@link
-   * CloudBigtableScanConfiguration} that will perform the specified {@link Scan} on the table.
-   *
+   * Converts a {@link CloudBigtableTableConfiguration} object to a
+   * {@link CloudBigtableScanConfiguration} that will perform the specified {@link Scan} on the
+   * table.
    * @param config The {@link CloudBigtableTableConfiguration} object.
    * @param scan The {@link Scan} to add to the configuration.
    * @return The new {@link CloudBigtableScanConfiguration}.
    */
-  public static CloudBigtableScanConfiguration fromConfig(
-      CloudBigtableTableConfiguration config, Scan scan) {
+  public static CloudBigtableScanConfiguration fromConfig(CloudBigtableTableConfiguration config,
+      Scan scan) {
     CloudBigtableScanConfiguration.Builder builder = new CloudBigtableScanConfiguration.Builder();
     config.copyConfig(builder);
     return builder.withScan(scan).build();
@@ -67,7 +67,8 @@ public class CloudBigtableScanConfiguration extends CloudBigtableTableConfigurat
     private Scan scan;
     private ReadRowsRequest request;
 
-    public Builder() {}
+    public Builder() {
+    }
 
     /**
      * Specifies the {@link Scan} that will be used to filter the table.
@@ -123,9 +124,6 @@ public class CloudBigtableScanConfiguration extends CloudBigtableTableConfigurat
 
     /**
      * {@inheritDoc}
-     *
-     * <p>Overrides {@link CloudBigtableTableConfiguration.Builder#withInstanceId(String)} so that
-     * it returns {@link CloudBigtableScanConfiguration.Builder}.
      */
     @Override
     public Builder withInstanceId(String instanceId) {
@@ -135,9 +133,6 @@ public class CloudBigtableScanConfiguration extends CloudBigtableTableConfigurat
 
     /**
      * {@inheritDoc}
-     *
-     * <p>Overrides {@link CloudBigtableTableConfiguration.Builder#withConfiguration(String,
-     * String)} so that it returns {@link CloudBigtableScanConfiguration.Builder}.
      */
     @Override
     public Builder withConfiguration(String key, String value) {
@@ -146,10 +141,8 @@ public class CloudBigtableScanConfiguration extends CloudBigtableTableConfigurat
     }
 
     /**
-     * {@inheritDoc}
-     *
-     * <p>Overrides {@link CloudBigtableTableConfiguration.Builder#withTableId(String)} so that it
-     * returns {@link CloudBigtableScanConfiguration.Builder}.
+     * {@inheritDoc} Overrides {@link CloudBigtableScanConfiguration.Builder#withTableId(String)} so
+     * that it returns {@link CloudBigtableScanConfiguration.Builder}.
      */
     @Override
     public Builder withTableId(String tableId) {
@@ -183,7 +176,6 @@ public class CloudBigtableScanConfiguration extends CloudBigtableTableConfigurat
   /**
    * Creates a {@link CloudBigtableScanConfiguration} using the specified project ID, instance ID,
    * table ID, {@link Scan} and additional connection configuration.
-   *
    * @param projectId The project ID for the instance.
    * @param instanceId The instance ID.
    * @param tableId The table to connect to in the instance.
@@ -217,9 +209,8 @@ public class CloudBigtableScanConfiguration extends CloudBigtableTableConfigurat
   }
 
   /**
-   * Gets the {@link ReadRowsRequest}. Sets the table name if it is not yet set.
-   *
-   * @return The {@link ReadRowsRequest}.
+   * Gets the {@link Scan} used to filter the table.
+   * @return The {@link Scan}.
    */
   public ReadRowsRequest getRequest() {
     return request.get();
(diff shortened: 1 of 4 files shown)


kevinsi4508:
I still need to update the ordering of those .with* methods.

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
```diff
@@ -64,11 +67,11 @@ public static CloudBigtableScanConfiguration fromConfig(CloudBigtableTableConfig
     private Scan scan;
     private ReadRowsRequest request;
 
-    public Builder() {
-    }
+    public Builder() {}
```
sduskis: please revert the formatting.

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
```diff
@@ -43,15 +46,15 @@
   private static final long serialVersionUID = 2435897354284600685L;
 
   /**
-   * Converts a {@link CloudBigtableTableConfiguration} object to a
-   * {@link CloudBigtableScanConfiguration} that will perform the specified {@link Scan} on the
-   * table.
+   * Converts a {@link CloudBigtableTableConfiguration} object to a {@link
```
sduskis: please revert the formatting

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
```diff
@@ -43,15 +46,15 @@
...
+   * CloudBigtableScanConfiguration} that will perform the specified {@link Scan} on the table.
+   *
    * @param config The {@link CloudBigtableTableConfiguration} object.
    * @param scan The {@link Scan} to add to the configuration.
    * @return The new {@link CloudBigtableScanConfiguration}.
    */
-  public static CloudBigtableScanConfiguration fromConfig(CloudBigtableTableConfiguration config,
-      Scan scan) {
+  public static CloudBigtableScanConfiguration fromConfig(
```
sduskis: please revert the formatting

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
```diff
@@ -117,6 +123,9 @@ public Builder withProjectId(String projectId) {
 
     /**
      * {@inheritDoc}
+     *
```
sduskis: please move these comment to a different PR.

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
```diff
@@ -231,14 +262,14 @@ ByteString getStopRowByteString() {
   }
 
   RowRange getRowRange() {
-    RowSet rows = request.getRows();
+    RowSet rows = getRequest().getRows();
```
sduskis: nit, please move the encapsulation to a different PR

kevinsi4508: It is either "getRequest()" or "request.get()". Since we are modifying the code anyway, I would rather do this now.

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
kevinsi4508: done.

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
kevinsi4508: done.

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
kevinsi4508: done.

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
kevinsi4508: done.

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
```diff
@@ -173,25 +182,38 @@ public CloudBigtableScanConfiguration build() {
...
+                if (!request.getTableName().isEmpty()) {
+                  return request;
+                }
+
+                BigtableInstanceName bigtableInstanceName =
+                    new BigtableInstanceName(getProjectId(), getInstanceId());
+                String fullTableName = bigtableInstanceName.toTableNameStr(getTableId());
+                return request.toBuilder().setTableName(fullTableName).build();
+              }
+            });
```
igorbernstein2: I dont think this is right...`this.request` will end up with an incomplete `ValueProvider` that will return `true` for `isAccessible` when projectId or instanceId are not accessible. I think you need to create a your own subclass of `ValueProvider` that wraps all of the relevant `ValueProvider`s and in its implementation `isAccessible` it takes an && of all its component ValueProviders.

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
```diff
@@ -265,8 +287,9 @@ public ByteKeyRange toByteKeyRange() {
   @Override
   public void populateDisplayData(DisplayData.Builder builder) {
     super.populateDisplayData(builder);
-    builder
-        .add(DisplayData.item("readRowsRequest", request.toString()).withLabel("ReadRowsRequest"));
+    if (areParametersAccessible()) {
+      builder.add(
+          DisplayData.item("readRowsRequest", request.toString()).withLabel("ReadRowsRequest"));
+    }
```
igorbernstein2: To ease debugging, maybe set the value to "Unavailable during pipeline construction" or something if the params are not accessible. It will give users something to copy/paste into a bug report

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
kevinsi4508: I don't think this matters at the moment, no? If you use CloudBigtableScanConfiguration now, there is no way you can set runtime parameters, therefore isAccessible is a non-issue.

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
kevinsi4508: I don't think user will see this at all at construction time.
I have confirmed this actually showed up correctly at runtime.

igorbernstein2: I didn't mena for "Unavailable during pipeline construction"  to show up in the normal case. I meant for it show up in case something does wrong and we need to debug. It will be easier for the user to describe this message rather than the absence of it.

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
kevinsi4508: The request object can't be modified before wrapping it in StaticValueProvider. I added TODO.

igorbernstein2: SGTM

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
kevinsi4508: Added TODO.

PR #1818, opened 2018-05-14
Templatize CloudBigtableTableConfiguration to be used in Dataflow template.

kevinsi4508:


bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableTableConfiguration.java
```diff
@@ -140,18 +186,21 @@ public void copyConfig(Builder builder) {
   @Override
   public boolean equals(Object obj) {
     return super.equals(obj)
```
sduskis: Do we need to check `areParametersAccessible()` for equals?

kevinsi4508: We know that validate() and  populateDisplayData() are called at construction time, therefore we must check areParametersAccessible(). For equals(), it should not be called a construction time.

PR #1823, opened 2018-05-21
Use RequestWithKeysValueProvider instead of NestedValueProvider as it requires CloudBigtableScanConfiguration.Builder to be serializable.

kevinsi4508:


bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
```diff
@@ -103,6 +103,54 @@ Builder withRequest(ValueProvider<ReadRowsRequest> request) {
...
+    /**
+     * Provides an updated request by adding a row range with start and end keys in the existing
+     * request.
+     */
+    private static class RequestWithKeysValueProvider
+        implements ValueProvider<ReadRowsRequest>, Serializable {
+      private final ByteString start;
+      private final ByteString stop;
+      private final ValueProvider<ReadRowsRequest> request;
+      private transient volatile ReadRowsRequest cachedRequest;
```
igorbernstein2: nit: I dont think you need transient (ReadRowsRequest is serializable) and I don't think you need volatile because this is is not threadsafe

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
```diff
@@ -103,6 +103,54 @@ Builder withRequest(ValueProvider<ReadRowsRequest> request) {
       return this;
     }
 
+    /**
+     * Provides an updated request by adding a row range with start and end keys in the existing
```
igorbernstein2: Please update docs to indicate that you are replacing the row set

kevinsi4508: done.

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
kevinsi4508: done.

PR #1831, opened 2018-05-24
Add runtime parameters metadata to be used in Dataflow import/export templates

kevinsi4508:


kevinsi4508:
I have considered adding "uploading parameter metadata file" in some automated step. I have not decided where. Ideally, it should be at the time of uploading the template to GCS.

kevinsi4508:
I am going keep it as a manual step for now. I will find out from Teleport team if there is a better way.
Ideally, I would like something like this:
mvn compile exec:java \
     -DmainClass=com.google.cloud.bigtable.beam.sequencefiles.ExportJob \
     -Dexec.args="--runner=DataflowRunner \
                  --project=stellar-mercury-540 \
                  --stagingLocation=gs://template_testing/staging \
                  --templateLocation=gs://template_testing/templates/ExportJob \
                  **--templateMetadataFile=local/file/ExportJob_metadata \**
                  --wait=false"

PR #1833, opened 2018-06-01
Fixing an exception handling case in BigtabeAsyncAdmin

sduskis:
getDescriptor needs to throw exceptions even if the exception is not `TableNotFound`


spollapally:
LGTM

Issue #1756, opened 2018-03-23
Use grpc version 1.10.1


rpeng:
Currently grpc 1.10.0 is affected by https://github.com/grpc/grpc-java/pull/4027, where we run into UnresolvedAddressException when setting GRPC_PROXY_EXP. It seems to be fixed in grpc version 1.10.1 - is it possible to make a release with the new grpc version?

sduskis:
Thanks for the update.  We'll make a new release next week.

sduskis:
@pawan-qlogic, can you please try to perform the upgrade and run all of our tests?

pawan-qlogic:
Upraded grpc version, all the test cases were successfully performed. Pull request created.

sduskis:
PR #1763 fixed this.

sduskis:
I'm going to keep this open until a new release is made.

plelevier:
Hello, 
When using google-cloud-java, it seems that they have not released any version with grpc 1.10.0  : 
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/v0.42.0/pom.xml
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/v0.42.1/pom.xml

The problem is when mixing cloud-bigtable-client 1.3.0 and google-cloud-java, maven raise a conflict between grpc 1.10.0 and 1.10.1. 
I've been able to force grpc dependencies to 1.10.1 but it is a little bit ugly and it may lead to setup problems with google libs to newcomers.

It would be great that the dependencies of cloud-bigtable-client were following the dependencies used in google-cloud-java (ie, never use some grpc version that was never referenced in google-cloud-java)

sduskis:
@plelevier: We're working on it.  Hopefully the next release will sync up all of the relevant versions. I'm hoping to depend on some of the core google-cloud-java artifacts via the [google-cloud-java BOM](https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-bom) to ensure that everything is in sync.

plelevier:
Great solution, thank you

sduskis:
We aim to make a release sometime early next week.  We have a couple things to do before the release.

PR #1838, opened 2018-06-04
Use grpc-netty-shaded instead of grpc-netty.

sduskis:
@igorbernstein2, the `google-cloud-java` BOM uses `grpc-netty-shaded`, so this would be a prereq for using the BOM.  I'm running into trouble with the `Provider` which is expecting this:

* com.google.bigtable.repackaged.io.grpc.netty.shaded.com.google.bigtable.repackaged.io.grpc.netty.NettyChannelProvider

Do you have any thoughts on how to fix the service file?

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/async/TestBulkMutationAwaitCompletion.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/async/TestBulkMutationAwaitCompletion.java
@@ -26,12 +26,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
+import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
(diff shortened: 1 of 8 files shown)


sduskis:
This should work towards fixing #1756

igorbernstein2:
Figured it out: the issue is caused by grpc-netty-shade.jar using the package name `io.grpc.netty.shaded.io.grpc.netty` and the  ServicesResourceTransformer usin greplaceAll:
https://github.com/apache/maven-shade-plugin/blob/ab82be37c5d4aa9855c03473a62cf7beed2f2b4a/src/main/java/org/apache/maven/plugins/shade/relocation/SimpleRelocator.java#L219

To workaround the issue, please the following, right before the relocation of `io.grpc`:

```
                <!-- Take special care of grpc-netty-shaded, it uses the package
                     io.grpc.netty.shaded.io.grpc.netty, which will cause the
                     ServicesResourceTransformer to replace both occurrences of io.grpc
                 -->
                <relocation>
                  <pattern>io.grpc.netty.shaded.io.grpc.netty</pattern>
                  <shadedPattern>com.google.bigtable.repackaged.io.grpc.netty</shadedPattern>
                </relocation>
```

sduskis:
@igorbernstein2, I update the pom.xml as per your suggestion, and now all tests pass!  Can you please review?

bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/async/TestBulkMutationAwaitCompletion.java
```diff
@@ -26,12 +26,7 @@
 import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
+import java.util.concurrent.*;
```
igorbernstein2: nit: please revert the wildcard

PR #1840, opened 2018-06-05
Using google-cloud-core-http for auth

sduskis:
This allows us to use a centralized configuration for all auth related dependencies.


sduskis:
This will partially fix #1756

sduskis:
As per an offline conversation with @igorbernstein2, I need to reintroduce all direct dependencies.

sduskis:
@igorbernstein2, I updated the dependencies.  PTAL.

PR #1842, opened 2018-06-05
Using google-cloud-clients for grpc versions

### LINKED ISSUE #1756: Use grpc version 1.10.1

sduskis:
In general the client will now pull in all io.grpc dependencies from google-cloud-core-grpc.

This will fix #1756

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/pom.xml
+++ b/bigtable-client-core-parent/bigtable-client-core/pom.xml
@@ -35,7 +35,14 @@ limitations under the License.
             <dependency>
                 <groupId>com.google.cloud</groupId>
                 <artifactId>google-cloud-bom</artifactId>
-                <version>${google-cloud-bom.version}</version>
+                <version>${google-cloud.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+            <dependency>
+                <groupId>com.google.cloud</groupId>
+                <artifactId>google-cloud-clients</artifactId>
+                <version>${google-cloud.version}</version>
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
@@ -64,11 +71,19 @@ limitations under the License.
             <groupId>com.google.api.grpc</groupId>
             <artifactId>grpc-google-cloud-bigtable-admin-v2</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.google.api.grpc</groupId>
+            <artifactId>proto-google-iam-v1</artifactId>
+        </dependency>
         <!-- Common -->
         <dependency>
             <groupId>com.google.code.findbugs</groupId>
             <artifactId>jsr305</artifactId>
-            <version>${jsr305.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>${slf4j.version}</version>
         </dependency>
         <dependency>
             <groupId>commons-logging</groupId>
@@ -78,14 +93,12 @@ limitations under the License.
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
-            <version>${guava.version}</version>
         </dependency>
 
         <!-- Auth -->
         <dependency>
             <groupId>com.google.auth</groupId>
             <artifactId>google-auth-library-oauth2-http</artifactId>
-            <version>${google.auth.library.version}</version>
             <exclusions>
                 <exclusion>
                     <groupId>com.google.guava</groupId>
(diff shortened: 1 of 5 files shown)


sduskis:
@garrettjonesgoogle: The goal is to align Cloud Bigtable dependencies to the same versions as `google-cloud-clients`.  Longer term, our `cloud-bigtable-core` artifact ought to be deprecated in favor of `google-cloud-bigtable`.  This gets us one step closer, and will make our users happy.

Do you have a strong no to this PR, or is your comment more of a 'caveat emptor'?

bigtable-client-core-parent/bigtable-client-core/pom.xml
```diff
@@ -35,7 +35,14 @@ limitations under the License.
             <dependency>
                 <groupId>com.google.cloud</groupId>
                 <artifactId>google-cloud-bom</artifactId>
-                <version>${google-cloud-bom.version}</version>
+                <version>${google-cloud.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
```
igorbernstein2: google-cloud-clients already imports google-cloud-bom. Does this not get propagated?

bigtable-client-core-parent/bigtable-client-core/pom.xml
```diff
@@ -204,13 +182,11 @@ limitations under the License.
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <version>${junit.version}</version>
-            <scope>test</scope>
```
igorbernstein2: I'm a bit on the fence about depending on the bom for scopes. How would you feel about keeping them?

bigtable-hbase-2.x-parent/bigtable-hbase-2.x-shaded/pom.xml
```diff
@@ -79,6 +79,14 @@ limitations under the License.
...
     <version>${opencensus.version}</version>
     <exclusions>
+        <exclusion>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-core</artifactId>
+        </exclusion>
+        <exclusion>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-core-grpc</artifactId>
+        </exclusion>
```
igorbernstein2: Can you add a note about why the exclusion are necessary?

bigtable-hbase-parent/bigtable-hbase-1.x-shaded/pom.xml
```diff
@@ -76,16 +76,20 @@ limitations under the License.
...
         <exclusion>
-            <groupId>io.netty</groupId>
-            <artifactId>*</artifactId>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-core</artifactId>
+        </exclusion>
+        <exclusion>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-core-grpc</artifactId>
         </exclusion>
```
igorbernstein2: Can you add a note about why the exclusion are necessary?

bigtable-hbase-parent/bigtable-hbase/pom.xml
```diff
@@ -38,28 +49,10 @@ limitations under the License.
...
-            <groupId>io.grpc</groupId>
-            <artifactId>grpc-core</artifactId>
-            <version>${grpc.version}</version>
-        </dependency>
-
-        <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-            <version>${guava.version}</version>
-        </dependency>
```
igorbernstein2: I don't understand why these are being removed

bigtable-client-core-parent/bigtable-client-core/pom.xml
sduskis: Interesting.  Apparently it is propagated.  I'll remove it.

bigtable-client-core-parent/bigtable-client-core/pom.xml
sduskis: Keeping it seems like a good idea.  I'll revert this change.

bigtable-hbase-parent/bigtable-hbase/pom.xml
sduskis: It was probably a mistake.  I'll reintroduce these, and just remove the versions.

sduskis:
@igorbernstein2, the comments were addressed; PTAL.

PR #1844, opened 2018-06-05
Optimized a bunch of imports

sduskis:
Mainly, removed * wildcards from imports.


spollapally:
LGTM. 

Do imports like java.util.* instead of explicit class incur additional penalties in general?

sduskis:
I don't know if there are additional penalties with `*`, but it's usually easier to understand imports that don't have it.

PR #1848, opened 2018-06-13
Use static grpc MethodDescriptor methods instead of fields

sduskis:


sduskis:
@spollapally, I can speak with you offline about this, if you'd like.  I spoke with @igorbernstein2 who agreed with this in principle.

spollapally:
I should be be good Solomon. Will go through this in the next 30 mins.

spollapally:
LGTM. Thought the string literals can be externalized, but if this is temporary and references are contained to the class, it seems fine to leave it as is.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -174,33 +179,53 @@ public BigtableDataGrpcClient(
     this.retryExecutorService = retryExecutorService;
     this.retryOptions = bigtableOptions.getRetryOptions();
 
+    // grpc static variable method descriptors are deprecated.  They are going to be replaced
+    // with static methods.  Once the new static methods exist, then we should use those.
+    // For now, use a hacky approach to get the method descrptors so that all environments work.
+    //
+    // See https://github.com/grpc/grpc-java/issues/1901 for more details.
+    ServerServiceDefinition definition = new BigtableGrpc.BigtableImplBase() {}.bindService();
```
igorbernstein2: Might be slightly nicer to use BigtableGrpc.getServiceDescriptor()

sduskis: BigtableGrpc actually has static getters.  I'm not sure why I missed that.

sduskis:
@igorbernstein2, I updated the code to use getters.  I'm not sure why I didn't see them last time I looked.

igorbernstein2:
One more thing to keep in mind: the proto artifact name is different in beam and bigtable-client-core, so both will be included on the classpath

igorbernstein2:
Maybe we can sidestep this issue entirely and just shade & relocate the stub for the beam connector?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGrpcClient.java
```diff
@@ -175,30 +176,24 @@ public BigtableDataGrpcClient(
...
-            getMutationRetryableFunction(ARE_RETRYABLE_MUTATIONS));
-    this.checkAndMutateRpc =
-        asyncUtilities.createAsyncRpc(
-            BigtableGrpc.METHOD_CHECK_AND_MUTATE_ROW,
-            getMutationRetryableFunction(Predicates.<CheckAndMutateRowRequest> alwaysFalse()));
-    this.readWriteModifyRpc =
-        asyncUtilities.createAsyncRpc(
-            BigtableGrpc.METHOD_READ_MODIFY_WRITE_ROW,
-            Predicates.<ReadModifyWriteRowRequest> alwaysFalse());
+    this.sampleRowKeysAsync = asyncUtilities.createAsyncRpc(
```
igorbernstein2: Would you mind adding:
```
try {
   BigtableGrpc.getReadRowsMethod();
} catch(NoSuchMethodError e) {
  throw new RuntimeException("Please make sure that you are using grpc-google-cloud-bigtable-v2 > xx and bigtable-protos is not on your classpath");
}
```

sduskis:
@igorbernstein2: good idea.  Done.

bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/TestBigtableTableAdminGrpcClient.java
```diff
@@ -0,0 +1,248 @@
...
+  ClientCall mockClientCall;
+
+  @Mock
+  NanoClock nanoClock;
+
+  BigtableTableAdminGrpcClient defaultClient;
+
+  @Before
+  public void setup() {
+    MockitoAnnotations.initMocks(this);
```
igorbernstein2: nit: consider using a rule:
https://static.javadoc.io/org.mockito/mockito-core/1.10.19/org/mockito/junit/MockitoRule.html

PR #1849, opened 2018-06-13
Update scm tag from bigtable-client-parent-1.3.1-SNAPSHOT to bigtable-client-parent-1.4.1-SNAPSHOT

kevinsi4508:


sduskis:
LGTM.

I think that we need to figure out how to better manage this tag, so that it gets automatically updated with every release.

kevinsi4508:
Yeap, Igor also agreed that we should make it better.

PR #1852, opened 2018-06-18
Update Import/export doc with regard to the suggested number of workers

kevinsi4508:


kevinsi4508:
Also this doc need to be updated?

https://cloud.google.com/bigtable/docs/importing-sequence-files

PR #1853, opened 2018-06-20
ScanAdapter now supports inclusive/exclusive

### LINKED ISSUE #1850: BigTable Scan withStopRow inclusive issue
Maybe I am mistaken,  but it seems that the "inclusive" boolean on the scan method is not working. Below, I would expect the scan to include "row3" because I have withStopRow("row3".getBytes(), true)), however, I only scan into row2.

    Output Received:   
    Row: row1
    Row: row2

pom below...

    <dependency>
	    <artifactId>bigtable-hbase-1.x-hadoop</artifactId>
	    <version>1.4.0</version>
    </dependency>

Java Code:

    Table table = conn.getTable(TABLE_NAME);

    table.put(new Put("row1".getBytes()).addColumn("fam".getBytes(), "qual".getBytes(), "val".getBytes()));
    table.put(new Put("row2".getBytes()).addColumn("fam".getBytes(), "qual".getBytes(), "val".getBytes()));
    table.put(new Put("row3".getBytes()).addColumn("fam".getBytes(), "qual".getBytes(), "val".getBytes()));

    for (Result r : table.getScanner(new Scan().withStartRow("row1".getBytes()).withStopRow("row3".getBytes(), true))) {
			System.out.println(("Row: " + new String(r.getRow())));
		}

sduskis:
HBase 1.4+ have the ability to set inclusive/exclusive on Scans.  The `ScanAdapter` should support explicit settings of inclusive/exclusive.  It should also be backwards compatible with versions less than 1.4.

This fixes #1850

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ScanAdapter.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ScanAdapter.java
@@ -16,6 +16,8 @@
 package com.google.cloud.bigtable.hbase.adapters.read;
 
 import static com.google.cloud.bigtable.data.v2.wrappers.Filters.FILTERS;
+
+import com.google.bigtable.v2.BigtableGrpc;
 import com.google.common.collect.Range;
 import com.google.bigtable.v2.ReadRowsRequest;
 import com.google.bigtable.v2.RowFilter;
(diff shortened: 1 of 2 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ScanAdapter.java
```diff
@@ -52,6 +54,19 @@
 public class ScanAdapter implements ReadOperationAdapter<Scan> {
 
   private static final int UNSET_MAX_RESULTS_PER_COLUMN_FAMILY = -1;
+  private static boolean OPEN_CLOSED_AVAILABLE;
```
igorbernstein2: nit: final

PR #1862, opened 2018-06-27
add a watchdog

igorbernstein2:
Fixes #1851

Extract watchdog responsibilities to an interceptor. This removes the feedback loop in ResumingStreamingResultScanner which should fix the race conditions. The watchdog is inspired by the implementation in gax, but is scoped for read rows.

PR #1870, opened 2018-07-04
Adding "Not Start" state to Watchdog

sduskis:
The master branch is broken.  It appears that there's a discrepancy between "idle" and "not started" states that were breaking the tests.  This PR adds the notion of "not started" and also has some other minor touch ups.


igorbernstein2:
Lgtm. I think the not started case used to handled by state being null, but I unwitteningly broke it addressing https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/1862#discussion_r199209495.

I think having an explicit state is better

kevinsi4508:
Not super familiar with watchdog. I don't see documentation such as state transition diagram. Up to Igor.

PR #1871, opened 2018-07-09
Updating Beam dependency

sduskis:
Also, depending on guava classes from the Cloud Bigtable repackaged project


bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableConfiguration.java
```diff
@@ -23,16 +23,16 @@
 import java.util.concurrent.TimeUnit;
 import java.util.Map.Entry;
 
+import org.apache.beam.repackaged.beam_sdks_java_io_hbase.com.google.common.base.Strings;
```
kevinsi4508: What is so special about Strings?

sduskis: I missed a commit before I pushed.  I removed `org.apache.beam.repackaged`.  PTAL.

sduskis:
@kevinsi4508, I fixed the imports.  PTAL.

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableIO.java
```diff
@@ -66,7 +64,8 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.bigtable.repackaged.com.google.bigtable.v2.SampleRowKeysResponse;
+import com.google.bigtable.repackaged.com.google.common.annotations.VisibleForTesting;
+import com.google.bigtable.repackaged.com.google.common.base.Preconditions;
```
kevinsi4508: These imports are not in the right order?

kevinsi4508:
LGTM
(check to make sure imports are in the right order)

kevinsi4508:
LGTM

byam:
When will this change going into [bigtable-hbase-beam](https://mvnrepository.com/artifact/com.google.cloud.bigtable/bigtable-hbase-beam) ?
@sduskis

sduskis:
I was hoping to get out a release today, but some things crept up.  At this point, it looks like early next week.

PR #1879, opened 2018-07-13
Fixing deprecation and other warnings.

sduskis:


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CredentialFactory.java
```diff
@@ -135,21 +135,6 @@ public static Credentials getCredentials(CredentialOptions options) throws IOExc
...
-   * Initializes OAuth2 credential using preconfigured ServiceAccount settings on the local
-   * Google Compute Engine VM. See:
-   * <a href="https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances"
-   * >Creating and Enabling Service Accounts for Instances</a>.
-   *
-   * @return a {@link com.google.auth.Credentials} object.
-   * @throws java.io.IOException if any.
-   * @throws java.security.GeneralSecurityException if any.
-   */
-  public static Credentials getCredentialFromMetadataServiceAccount()
```
kevinsi4508: Could this used publicly?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CredentialFactory.java
```diff
@@ -218,8 +208,7 @@ public static Credentials getApplicationDefaultCredential() throws IOException,
    * @throws java.io.IOException if any.
    * @throws java.security.GeneralSecurityException if any.
```
kevinsi4508: delete "@throws java.security.GeneralSecurityException if any."

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CredentialFactory.java
sduskis: Users should not use CredentialFactory directly.  If they need to replicate this logic, they can use `ComputeEngineCredentials` themselves.  This code was a copy from somewhere else that needed it, but for CBT, we don't need this.

sduskis:
Comments were addressed.  Thanks for the review.  PTAL.

kevinsi4508:
LGTM

Issue #1858, opened 2018-06-24
Client will retry until max timeout is reached, even if deadline is exceeded


steveniemitz:
I've been experimenting with using gRPC itself to set my per-call timeouts.  I've noticed that if I set my per-call deadline low, lets say in this example to 1 second, I can end up in a situation where the client submits "requests" even though they'll never actually be sent to bigtable (since the deadline is already exceeded).

Example log:
```
2018-06-24 19:22:36,317 INF AbstractRetryingOperation Retrying failed call. Failure #1, got: Status{code=DEADLINE_EXCEEDED, description=deadline exceeded after 739333109ns, cause=null} on channel 14
2018-06-24 19:22:36,325 INF AbstractRetryingOperation Retrying failed call. Failure #2, got: Status{code=DEADLINE_EXCEEDED, description=deadline exceeded: -16434248 ns from now, cause=null} on channel 15
2018-06-24 19:22:36,341 INF AbstractRetryingOperation Retrying failed call. Failure #3, got: Status{code=DEADLINE_EXCEEDED, description=deadline exceeded: -31615089 ns from now, cause=null} on channel 16
2018-06-24 19:22:36,366 INF AbstractRetryingOperation Retrying failed call. Failure #4, got: Status{code=DEADLINE_EXCEEDED, description=deadline exceeded: -56484939 ns from now, cause=null} on channel 17
2018-06-24 19:22:36,396 INF AbstractRetryingOperation Retrying failed call. Failure #5, got: Status{code=DEADLINE_EXCEEDED, description=deadline exceeded: -86724983 ns from now, cause=null} on channel 18
```

It'll then retry until it reaches the max-retry limit based on the retry policy.  It seems like the client should stop retrying once the deadline has been exceeded here.

steveniemitz:
Sorry I was kind of short on this bug description, minimally I can reproduce it with something like:

```
val btOpts = new BigtableOptions.Builder()
    .setCallOptionsConfig(new CallOptionsConfig(true, CallOptionsConfig.SHORT_TIMEOUT_MS_DEFAULT, 1000))
    .build()
val sess = new BigtableSession(btOpts)
val someBigRequest = ReadRowsRequest.newBuilder()....build()
val observer = new StreamObserver[FlatRow]() { ... }
sess.getDataClient.readFlatRows(someBigRequest, observer)
```

[...]

PR #1880, opened 2018-07-13
Making a distinction between RPC and Operation deadlines

sduskis:
Users set deadlines on logical operations.  Logical operations may consist of multiple RPCs, depending on the number of times temporary failures occur.  Each RPC has an additional concept of a deadline as a "safety net" if a user doesn't explicitly set a Deadline; this prevents hanging behavior. This PR distinguishes between the two in `AbstractRetryingOperation`.  This is a step towards fixing issue #1858.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
@@ -369,7 +369,7 @@ public abstract class AbstractRetryingOperation<RequestT, ResponseT, ResultT>
   //      RetryingReadRowsOperation like it should.
   protected boolean isStreamingRead() {
     return request instanceof ReadRowsRequest &&
-            CallOptionsFactory.ConfiguredCallOptionsFactory.isGet((ReadRowsRequest) request);
+            !CallOptionsFactory.ConfiguredCallOptionsFactory.isGet((ReadRowsRequest) request);
   }
 
   protected RequestT getRetryRequest() {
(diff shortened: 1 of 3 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
```diff
@@ -329,16 +329,36 @@ protected void run() {
...
+    return callOptions;
+  }
+
+  /**
+   * Create an {@link CallOption} that has a fail safe RPC deadline to make sure that unary operations don't hang. This
+   * will have to be overridden for streaming RPCs like read rows.
+   *
+   * @return a {@link CallOptions}
+   */
+  protected CallOptions getRpcCallOptions() {
```
kevinsi4508: I am confused. The logic here stays the same, no?

Can you document how user specified deadline is used; streaming vs non-streaming?

For each RPC:

```
               | User deadline provided            |  User deadline not provided
-----------------------------------------------------------------------------
Streaming      |    User deadline                  |  watch dog?
-----------------------------------------------------------------------------
Non-Streaming  |    User deadline                  |  UNARY_DEADLINE_MINUTES
-----------------------------------------------------------------------------

```

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RetryingReadRowsOperation.java
```diff
@@ -203,6 +204,25 @@ public void onMessage(ReadRowsResponse message) {
     }
   }
 
+  @Override
+  /**
+   * Override the default behavior of {@link AbstractRetryingOperation#getRpcCallOptions()} to ensure that deadlines
+   * are not set for streaming reads.
+   */
+  protected CallOptions getRpcCallOptions() {
```
kevinsi4508: I failed to see why this is needed? The parent class already know whether this is steaming call or not, no? Has to do with 'request' vs 'nextRequest'?

sduskis: I'll remove it.

sduskis:
@kevinsi4508, I addressed all of the comments. PTAL.

sduskis:
@igorbernstein2, do you want to take a look?

sduskis:
@kevinsi4508, @igorbernstein2 ping

kevinsi4508:
LGTM

igorbernstein2:
LGTM, but shouldn't there be a test?

PR #1881, opened 2018-07-13
Replacing GrpcUtil with ThreadUtil

sduskis:
GrpcUtil is in grpc's internal package.  Copied the functionality Cloud Bigtable needs into ThreadUtil so that we don't have to depend on grpc internal classes.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/util/ThreadUtil.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/util/ThreadUtil.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2018 Google LLC. All Rights Reserved.
+ *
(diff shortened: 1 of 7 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/util/ThreadUtil.java
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.util;
+
+import com.google.common.util.concurrent.MoreExecutors;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+
+import java.util.concurrent.ThreadFactory;
+
+public class ThreadUtil {
```
kevinsi4508: Should we mention that this is copied from GrpcUtil?

kevinsi4508:
LGTM

PR #1882, opened 2018-07-16
Replacing ByteStringer with ByteString.Output where appropriate

sduskis:
This allows ByteStringer to be moved to bigtable-hbase.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RowMerger.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RowMerger.java
@@ -15,7 +15,6 @@
  */
 package com.google.cloud.bigtable.grpc.scanner;
 
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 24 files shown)


danielweis:
This change looks good. I'd like to know why ByteStringer still needs to exist so we can see if we can obviate the usage in those contexts as well.

steveniemitz:
I'm not sure what the API stability guarantee of these libraries are, but this is a breaking change to the API surface (ByteStringer was/is public).  We use this in a bunch of places in our own code as well because its generally useful.  There are alternatives now in protobuf as well iirc, so its not the end of the world to have to refactor, but I just wanted to point it out.  Maybe you can include it as a breaking change in the release notes?

sduskis:
@steveniemitz, I guess we'll need to revert the change.  

@danielweis: `ByteStringer` was required for performance reasons.  Extracting `byte[]`s from `ByteString`s and creating `ByteString`s from `byte[]` are expensive CPU operations at the throughput that Cloud Bigtable supports.  Some users reported a 10X increase in overall throughput once we added `ByteStringer` to our read path that extracts `byte[]`.

steveniemitz:
fwiw I don't really mind having to refactor my stuff, we actually use a vendored version of this anyways from most places in our code.  I just wanted to point out that it is a public API and a breaking change by moving it.

danielweis:
@sduskis That makes sense. What I'm aiming for is whether alternative APIs like ByteString.Output will suffice for your other use cases. We introduced UnsafeByteOperations for the use cases where there are no alternatives so it's okay to use, with the caveat that if used incorrectly, your application will break in "spooky-action-at-a-distance" ways. Consequently I much prefer folks using alternative APIs wherever possible.

PR #1889, opened 2018-07-18
Retrying operations stop after deadline exceeded

sduskis:
Users can set Deadlines in a variety of ways.  If they set it, then the max retry timeout should be no more than that deadline.

This PR fixes issue #1858.


sduskis:
Closing this in favor of another approach I discussed offline with @igorbernstein2.

PR #1890, opened 2018-07-19
fix: validation of TimeRange

piroyoung:
Every valid `TimeRange` will be denied.


sduskis:
This change reflects the fact that we don't have enough Unit and Integration tests for this.  I'll add those in a separate PR.

Issue #1843, opened 2018-06-05
hbase 2 integration tests don't run.


sduskis:
The hbase 2 integration tests get this error: 

`Caused by: java.lang.IllegalArgumentException: port out of range:-1`

The HBase integration tests complete with a status of 0, but don't actually run any tests.

sduskis:
This seems to be an [HBase bug](http://mail-archives.apache.org/mod_mbox/hbase-issues/201805.mbox/%3CJIRA.13157891.1525785159000.43512.1525975740673@Atlassian.JIRA%3E).

sduskis:
I think that we should wait for an official fix from HBase when we upgrade to the next 2.* release.

PR #1891, opened 2018-07-19
Upgrading to HBase 2.1.0

sduskis:
The HBase mini cluster tests still seem to be broken.  That needs to still be investigated as per issue #1843


sduskis:
@igorbernstein2, ping.

PR #1892, opened 2018-07-20
Using gax's ExponentialRetryAlgorithm for backoffs

sduskis:


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
```diff
@@ -276,29 +283,60 @@ protected void setException(Exception exception) {
...
+    return currentBackoff != null;
+  }
+
+  /**
+   * Either a response was found, or a timeout event occurred. Reset the information relating to
+   * Status oriented exception handling.
+   */
+  protected void resetStatusBasedBackoff() {
+    this.currentBackoff = null;
+    this.failedCount = 0;
```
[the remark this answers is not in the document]
sduskis: I need to make sure that the timeout only starts after a failure, to preserve the semantics of RetryOptions.  Specifically, the timeout starts with the first failure, and not on the actual first attempt.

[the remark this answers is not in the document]
sduskis: ugh.  FailedCount is also set by Watchdog time outs.  I don't know how to replace it.

sduskis:
@igorbernstein2  I added a bunch of comments.  I don't think that I'll have great answers for you relating to proper usage of the retry algorithm.  My main goal is to achieve a similar behavior compared to the use of ExponentialBackoff.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
```diff
@@ -276,29 +284,75 @@ protected void setException(Exception exception) {
...
+
+        // How fast should the retry delay increase?
+        .setRetryDelayMultiplier(retryOptions.getBackoffMultiplier())
+
+        // What is the maximum amount of sleep time between retries?
+        // There needs to be some sane number for max retry delay, and it's unclear what that
+        // number ought to be.  20% of timeout time was chosen because some number is needed
+        .setMaxRetryDelay(toDuration(timeoutMs / 5))
+
+        // How long should we wait before giving up retries?
```
igorbernstein2: Please clarify that this starts from the first failed rpc

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
```diff
@@ -276,29 +284,75 @@ protected void setException(Exception exception) {
...
+        // How long should the sleep be between RPC failure and the next RPC retry?
+        .setInitialRetryDelay(toDuration(retryOptions.getInitialBackoffMillis()))
+
+        // How fast should the retry delay increase?
+        .setRetryDelayMultiplier(retryOptions.getBackoffMultiplier())
+
+        // What is the maximum amount of sleep time between retries?
+        // There needs to be some sane number for max retry delay, and it's unclear what that
+        // number ought to be.  20% of timeout time was chosen because some number is needed
+        .setMaxRetryDelay(toDuration(timeoutMs / 5))
```
igorbernstein2: I think it would be better to hardcode this to 1 minute

bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/async/TestRetryingMutateRowsOperation.java
```diff
@@ -179,6 +179,7 @@ public void testCompleteFailure() throws InterruptedException, TimeoutException
       underTest.getAsyncResult().get(1, TimeUnit.MINUTES);
       Assert.fail("Expecting a DEADLINE_EXCEEDED exception");
     } catch (ExecutionException e) {
+      e.printStackTrace();
```
igorbernstein2: please remove

bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/async/TestRetryingMutateRowsOperation.java
```diff
@@ -218,9 +219,8 @@ public void testPartialResponse() {
   private RetryingMutateRowsOperation createOperation(MutateRowsRequest request) {
     return new RetryingMutateRowsOperation(RETRY_OPTIONS, request, mutateRows, CallOptions.DEFAULT,
         executorService, new Metadata()) {
-      @Override
-      protected ExponentialBackOff createBackoff() {
-        return clock.createBackoff(retryOptions);
+      @Override protected ApiClock getApiClock() {
+        return clock;
```
igorbernstein2: nit: I think it would be nicer to pass it as a constructor arg

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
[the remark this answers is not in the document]
igorbernstein2: ok, let's leave it for now

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
```diff
@@ -276,29 +283,60 @@ protected void setException(Exception exception) {
    */
   protected abstract boolean onOK(Metadata trailers);
 
-  protected long getNextBackoff() {
+  protected Long getNextBackoff() {
+    if (exponentialRetryAlgorithm == null) {
+      exponentialRetryAlgorithm = createRetryAlgorithm();
```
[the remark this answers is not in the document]
igorbernstein2: ok, that makes sense

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
```diff
@@ -276,29 +283,60 @@ protected void setException(Exception exception) {
...
    *
-   * @return a {@link com.google.api.client.util.ExponentialBackOff} object.
+   * @return a {@link ExponentialRetryAlgorithm} object.
    */
+  private ExponentialRetryAlgorithm createRetryAlgorithm() {
+    RetrySettings retrySettings = RetrySettings.newBuilder()
+        .setJittered(true)
+        .setInitialRetryDelay(toDuration(retryOptions.getInitialBackoffMillis()))
+        .setRetryDelayMultiplier(retryOptions.getBackoffMultiplier())
+        .setMaxRetryDelay(toDuration(retryOptions.getMaxElapsedBackoffMillis() / 5))
```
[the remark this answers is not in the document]
igorbernstein2: I think it would be better as a const. Maybe a 60s?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
igorbernstein2: I think that test should be tweaked

sduskis: What is a reasonable max and min, do you think?

sduskis:
@igorbernstein2, I changed the backoff algorithm slightly.  The default is to not retry if the `nextRetryDelay + timePassed > timeout`.  I updated the algorithm to wait `timeout - timePassed` in that case. 

We need some sort of boundary condition in the unit test to make sure that the sleep occurred within some boundaries so that we can determine correctness.

igorbernstein2:
I’m wondering if the logic should be changed in gax instead of overlaying it here

sduskis:
@igorbernstein2, we need to move forward with some sort of logic in order to fix #1858.  It would be fine to fix gax, or fix our settings later.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RetryingReadRowsOperation.java
```diff
@@ -307,12 +301,14 @@ int getTimeoutRetryCount() {
...
-    return currentBackoff;
+  RowMerger getRowMerger() {
+    return rowMerger;
   }
 
+
   @VisibleForTesting
-  RowMerger getRowMerger() {
-    return rowMerger;
+  boolean performingRetries() {
```
igorbernstein2: Is this method necessary? can't the test just use the existing inRetryMode?

sduskis: I made `inRetryMode()` to be able to use it in the scanner package.

PR #1894, opened 2018-07-23
The Deadline timeout is now the max backoff timeout

### LINKED ISSUE #1858: Client will retry until max timeout is reached, even if deadline is exceeded

sduskis:
This fixes #1858 

This should hopeuflly be the last PR before the next release.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
@@ -42,6 +42,7 @@ import com.google.common.util.concurrent.ListenableFuture;
 
 import io.grpc.CallOptions;
 import io.grpc.ClientCall;
+import io.grpc.Deadline;
 import io.grpc.Metadata;
 import io.grpc.Status;
 import io.grpc.Status.Code;
@@ -336,7 +337,12 @@ public abstract class AbstractRetryingOperation<RequestT, ResponseT, ResultT>
    * @return a {@link ExponentialRetryAlgorithm} object.
    */
   private ExponentialRetryAlgorithm createRetryAlgorithm(ApiClock clock) {
-    int timeoutMs = retryOptions.getMaxElapsedBackoffMillis();
+    long timeoutMs = retryOptions.getMaxElapsedBackoffMillis();
+
+    Deadline deadline = getOperationCallOptions().getDeadline();
+    if (deadline != null) {
+      timeoutMs = deadline.timeRemaining(TimeUnit.MILLISECONDS);
+    }
 
     RetrySettings retrySettings = RetrySettings.newBuilder()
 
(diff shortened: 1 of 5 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
```diff
@@ -336,7 +337,12 @@ protected void resetStatusBasedBackoff() {
    * @return a {@link ExponentialRetryAlgorithm} object.
    */
   private ExponentialRetryAlgorithm createRetryAlgorithm(ApiClock clock) {
-    int timeoutMs = retryOptions.getMaxElapsedBackoffMillis();
+    long timeoutMs = retryOptions.getMaxElapsedBackoffMillis();
+
+    Deadline deadline = getOperationCallOptions().getDeadline();
```
igorbernstein2: This is a bit off...deadline should be overall timeout, rather then just the retry amount. Please add a TODO to fix it

igorbernstein2: how does this interact with grpc deadline? I'm concerned about this following case:
- deadline is set to 10s
- first rpc returns in 9s with a retryable error
- the retry mechanism schedules a retry in 2s
- 2nd rpc returns a local deadline exceeded
- the retry mechanism reschedules
- 3rd rpc returns a local deadline exceeded
...
until 10s pass after the first rpc failed

sduskis: The `CallOptions` are sent for each RPC.  If a deadline is set, the retry mechanism will only go as long as the Deadline in the `CallOptions`.

Something similar to your case:

- deadline is set to 10s
- first rpc returns in 9s with a retryable error
- The retry schedules a timeout in 10s - 9s = 1s, with an initial retry of a few ms.
- The second RPC still has roughly 1s to complete.
- 1s goes by, DEADLINE_EXCEED happens due to the CallOptions
- Retry logic sees that Timeout occurred, and stops retries.

PR #1897, opened 2018-07-26
Inlining use of GoogleNetHttpTransport

sduskis:


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CredentialFactory.java
```diff
@@ -90,9 +90,24 @@ public static void setHttpTransportFactory(HttpTransportFactory httpTransportFac
...
 
+  private static class DefaultHttpTransportFactory implements HttpTransportFactory {
+    @Override
+    public HttpTransport create() {
+      // Consider App Engine Standard
+      if (PlatformInformation.isOnGAEStandard7()) {
+        try {
+          return new UrlFetchTransport();
+        } catch (Exception ignore) {
+          // Maybe not on App Engine
```
igorbernstein2: I'm guessing this should never happen, but might be good to log a warning

sduskis: ack.  I blindly copied code, which I guess wasn't ideal :)

sduskis:
I added logging for the Exception.

PR #1898, opened 2018-07-26
Logging large rows in RowMerger

sduskis:
NOTE: Getting the size of cells seems not to cause any performance issues in a microbenchmark, but this change should go through normal nightly performance tests.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RowMerger.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RowMerger.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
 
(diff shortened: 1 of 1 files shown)


sduskis:
Interestingly, `RowMergerPerf`, a mini-benchmark for RowMerger, shows no appreciable differences for this change.

sduskis:
1. won't using `cellChunk.value` for size aggregation cause issues with situations where the value is smaller than the qualifier / timestamp?
2. The size accumulator is in the `RowInProgress` class, which gets set to null on complete/reset.

sduskis:
@igorbernstein2, I did more testing with different sized rows / cell counts with `RowMergerPerf` with and without `chunk.getSerializedSize()`.  There was no significant difference between including `chunk.getSerializedSize()` and not including it.  I would like to keep `chunk.getSerializedSize()` for completeness.  WDYT?

sduskis:
@igorbernstein2, I did more testing with large values in `RowMergerPerf`. `chunk.getSerializedSize()` doesn't seem to make an appreciable runtime difference in large rows.

PTAL.

Issue #1895, opened 2018-07-26
Allow configuration of Credentials via HBase.


hfgbarrigas:
Can we add a constructor on BigtableAsyncConnection that accepts BigtableOptions? Feels weird that I can use the BigtableOptions builder and I'm forced to pass in Configuration from hbase.

Or am I missing something?

sduskis:
In theory this is possible, but not something I'd like to do.

We think about the artifacts in `bigtable-client-core` as a stand alone client, and the HBase artifacts as adapters for the HBase interface.  The general configuration interface for HBase is a `Configuration` object, and that's what our HBase customers use either programmatically or via an `hbase-site.xml` file.  You can see the details of the conversion in the `BigtableOptionsFactory` class.

Is there a reason you don't want to use a `Configuration` object, and you want to use a `BigtableOptions` object?

hfgbarrigas:
Looking at the code on `BigtableOptionsFactory` looks like `Configuration` can only deal with credentials provided either via json, file location and if none is used will default to `applicationDefault`. Since I already have `Credentials` available (another component deals with providing credentials), seems redundant to have to fetch them again. we could perhaps have a `toConfiguration(BigtableOptions)` ?

sduskis:
`toConfiguration(BigtableOptions)` or a new Configuration won't work easily.  I'll talk with my colleagues, specifically @igorbernstein2, about some options for passing either `BigtableOptions` or some other mechanism for `Credential` passing.  Hopefully, we can think through an approach before our next release, which is imminent.

FWIW, We're changing to a new model in the near future that uses a new client [google-cloud-bigtable](https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-clients/google-cloud-bigtable).  This will impact some of our design decisions.

sduskis:
#1899 is an attempt to allow you to set a `Credentials` object.  It uses reflection, and isn't perfect, but it's a lower impact change.  WDYT, @hfgbarrigas?

PR #1899, opened 2018-07-27
User can configure a Credentials Callable.

sduskis:


sduskis:
This is an attempt to fix #1895

igorbernstein2:
I dont love this approach: it's adding yet another layer of indirection. At the end of the day we need a `Credentials` instance. `GoogleCredentials` already (de)serializes json and dispatches to both `UserCredentials`,  `ServiceAccountCredentials`. On top of that we have `CredentialFactory` that dispatches to `GoogleCredentials` from `CredentialOptions` from various sources. This will add yet another dispatcher to the mix. I would prefer to simplify the current system then add another knob to it

sduskis:
@igorbernstein2: Ack.  Indirection is quite bad in this case.

In #1895, A user requested some way to use the same `Credentials` object for multiple clients. Right now, the only published option for CBT is the HBase API where the user interface is a `Map<String, String>`. 

There are a few ways to get a `Credentials` that we discussed on #1895:

1. We can add a new constructor to our HBase `Connection` objects, and potentially add wrappers in `BigtableConfiguration`.
2. We can use a level of indirection with `String`s like we did here.
3. We can do some other well known `Singleton` type thing.

Do you have any better ideas about how to implement this in the short term?

igorbernstein2:
I think it might be easier to discuss in person, but I would love to deprecate the `CredentialOption` indirection step, have BigtableOptions expose `Credentials` directly, and have the hbase config encode the `GoogleCredentials` json format.

sduskis:
Closing this PR in favor of #1900

PR #1917, opened 2018-08-29
Add dependency for google cloud java bigtable client and deprecate filters related class 

ajaaym:
Add java veneer client dependency and remove duplicate class


sduskis:
Ugh...  I didn't realize that there's a package change.  Please reinstate the `wrappers.Filters` object, mark the class as `@Deprecated`.  In a future PR, please delegate from `wrappers.Filters` delegate to `models.Filters` as best as possible.

PR #1931, opened 2018-09-06
Adding a method to "fail fast" large rows

sduskis:


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
```diff
@@ -401,6 +401,10 @@ public BulkMutation(
   public synchronized ListenableFuture<MutateRowResponse> add(MutateRowsRequest.Entry entry) {
     Preconditions.checkNotNull(entry, "Request null");
     Preconditions.checkArgument(!entry.getRowKey().isEmpty(), "Request has an empty rowkey");
+    Preconditions.checkArgument(entry.getMutationsCount() <= MAX_NUMBER_OF_MUTATIONS, String
```
steveniemitz: Late to the party here, but this will cause a `String.format` and `toStringUtf8` call for every row added, which could add non-trivial overhead (CPU and allocations) here.  It's probably better to just use an `if` statement for the check here, and only compute the error message when needed.

PR #1934, opened 2018-09-07
Supporting Scan limits (HBase 1.4+)

### LINKED ISSUE #1933: .setLimit() has no effect on Scan?
Hi,

I've tried to use the .setLimit() on a Scan, but it still returns with all rows in the table after the startrow.

Below the code snippet, which produces this error.

```java
public static void main(String[] args) throws IOException  {
    Connection c = BigtableConfiguration.connect(projectId, instanceId);
    Table t = c.getTable(TableName.valueOf(tableName));

    Scan s = new Scan().withStartRow(Bytes.toBytes(startRow)).setLimit(5);
    ResultScanner resultScanner = t.getScanner(s);
    
    for (Result r : resultScanner) {
        System.out.print(Bytes.toString(r.getRow()) + "\t");
        System.out.println(Bytes.toString(r.value()));
    }
}
```

Gradle dependecy: 
```
dependencies {
    implementation "com.google.cloud.bigtable:bigtable-hbase-beam:1.5.0"
}
```

I tried with different versions of Bigtable dependencies but the result was the same.

Do I misuse something? 

Thanks!

sduskis:
This fixes #1933

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ScanAdapter.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ScanAdapter.java
@@ -53,7 +53,7 @@ public class ScanAdapter implements ReadOperationAdapter<Scan> {
 
   private static final int UNSET_MAX_RESULTS_PER_COLUMN_FAMILY = -1;
   private static final boolean OPEN_CLOSED_AVAILABLE = isOpenClosedAvailable();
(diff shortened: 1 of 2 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ScanAdapter.java
```diff
@@ -67,6 +68,15 @@ private static boolean isOpenClosedAvailable() {
     }
   }
 
+  private static boolean isLimitAvaoilable() {
```
ajaaym: s/Avaoilable/Available/

PR #1952, opened 2018-09-26
Code deduplication of non-async to async

rahulKQL:
Delegating call from BigtableAdmin to BigtableAsyncAdmin using Configurations.


rahulKQL:
@sduskis Could you please have a look on BigtableAdmin changes.

PR #1953, opened 2018-09-27
Adding interface for CommonCon for HBase 1.x & 2.x

rahulKQL:
Adding `CommonConnection` interface to be implemented in `AbstractBigtableConnection`& `BigtableAsyncConnection.`  
Also `connection.getRegionLocator()` which is using connection would convert into `getConnection().getRegionLocator()`. 
I have added initial implementation for this interface [here](https://github.com/rahulKQL/cloud-bigtable-client/commit/491295e11e7955e6276ef8d42bd8910dd66da92f)
@sduskis  Could you please have a look and let me know your thoughts on this.


sduskis:
Please add the header with the copyright.  Also please "implement" this interface in `AbstractBigtableConnection` and `BigtableAsyncConnection`, and use it in `BigtableAsyncAdmin`.

PR #1954, opened 2018-09-27
Fixing HBase filter test

sduskis:
This fixes a break that was caused by #1946


sduskis:
I merged this, to fix the broken tests.

PR #1956, opened 2018-09-28
WIP: Unsafe mutation split option

elisheva-qlogic:
EXPERIMENTAL
Created unsafe mutation split option for bulk mutations


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java
```diff
@@ -442,10 +447,48 @@ public void run() {
...
     return future;
   }
 
+  private ListenableFuture<MutateRowResponse> processUnsafeMutationSplit(MutateRowsRequest.Entry entry) {
+    final SettableFuture<MutateRowResponse> mutateRowResponseListenableFuture = SettableFuture.create();
+    final List<List<Mutation>> subSets = Lists.partition(entry.getMutationsList(), (int) MAX_NUMBER_OF_MUTATIONS);
+    // split subSets into several entries of 100,000
+    for (List<Mutation> mutations : subSets) {
+      ListenableFuture<MutateRowResponse> childFuture = sendDirect(entry.getRowKey(), mutations);
+      Futures.addCallback(childFuture, new FutureCallback<MutateRowResponse>() {
```
sduskis: This callback needs to be created outside the for loop, otherwise, `success` will never be more than 1.

igorbernstein2:
I thought we decided that this should be done only for the import job?

sduskis:
@elisheva-qlogic, can we close this, or is there more work to be done in this PR?

PR #1958, opened 2018-10-02
CommonConnection for async and non-async 

rahulKQL:
### Description
This change contains addition of CommonConnection, which is implemented in AbstarctBigtableConnection & BigtableAsyncConnection.

### Imapact of This PR
After this PR, BigtableAdmin & BigtableAsyncAdmin would have same argument in the constructor, so that a same connection instance can be used to create either of the Admin classes.

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
@@ -48,6 +48,7 @@ import org.apache.hadoop.hbase.ClusterStatus;
 import org.apache.hadoop.hbase.HBaseIOException;
 import org.apache.hadoop.hbase.HColumnDescriptor;
 import org.apache.hadoop.hbase.HRegionInfo;
+import org.apache.hadoop.hbase.HRegionLocation;
 import org.apache.hadoop.hbase.HTableDescriptor;
 import org.apache.hadoop.hbase.MasterNotRunningException;
 import org.apache.hadoop.hbase.NamespaceDescriptor;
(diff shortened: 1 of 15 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -700,10 +701,12 @@ public Configuration getConfiguration() {
   /** {@inheritDoc} */
   @Override
   public List<HRegionInfo> getTableRegions(TableName tableName) throws IOException {
-    List<HRegionInfo> regionInfos = new ArrayList<>();
-    for (HRegionLocation location : connection.getRegionLocator(tableName).getAllRegionLocations()) {
+    List<HRegionInfo> regionInfos = Collections.emptyList();
+    BigtableRegionLocator regionLocator = new BigtableRegionLocator(tableName, options, connection.getSession().getDataClient());
```
sduskis: Can we please add a `CommonConnection.getAllRegionInfos(TableName)` method instead?

sduskis:
@rahulKQL, a bunch of HBase 2 tests fail.  We need to fix those.

rahulKQL:
Hi @sduskis,
I ran all 4 integration-test profile, It's all building successfully in my system. Could you please re-run the **Hbase-2**  test on [code_duplication_removal](https://github.com/rahulKQL/cloud-bigtable-client/tree/code_duplication_removal) branch.

rahulKQL:
This commit now contains: 

- Changes reverted  related to the `getSampledRowKeysAdapter()`, as it is still having **compatibility issue**.
- Also added `getAllRegionsInfo()`, on `BigtableAdmin` & `BigtableAsyncAdmin`. 
- Ran test cases against actual(_node-client-performance_) HBase 1.x & 2.x both **passed**. 

@sduskis Please have a look and let me know if anything is required.

PR #1959, opened 2018-10-03
test-deduplication in ColumnFamily & ListSnapshot

rahulKQL:
### Description
To remove duplicate code from Integration-Test-1.x & 2.x

Refactoring of below test classes to remove duplication: 
 - AbstractTestColumnFamilyAdmin (commons-tests)
 - TestColumnFamilyAdmin (integration-test 1.x & 2.x)
 - AbstractTestSnapshot (commons-tests)
 - TestSnapshots (integration-test 1.x & 2.x)


sduskis:
I don't think that we should merge these.  I like the fact that the abstract methods exist in the parent `Abstract*Test` classes to force specific implementation in child implementations.

PR #1963, opened 2018-10-11
Interface to wrap google cloud java

ajaaym:
This PR contains interface related to mutate row api to wrap google cloud java.


sduskis:
Please add the default Apache license header.

sduskis:
We need to figure out good names for these.  We should try to figure out names that are not already used

PR #1973, opened 2018-10-25
Data settings addition

rahulKQL:
This commits contains: 
- A class with static method for conversion of `BigtableOptions` to `BigtableDataSettings`.
- Basic test cases for the same, more to add once finalised.

**Note:** This work is still in progress, opening PR for review purpose.


rahulKQL:
@igorbernstein2  Please review

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,174 @@
...
+            .setFlowControlSettings(
+              FlowControlSettings.newBuilder().setLimitExceededBehavior(LimitExceededBehavior.Block)
+                  .setMaxOutstandingRequestBytes(bulkMutation.getMaxMemory()).build());
+
+    long autoFlushMs = bulkMutation.getAutoflushMs();
+    if (autoFlushMs > 0) {
+      batchingSettingsBuilder.setDelayThreshold(Duration.ofMillis(autoFlushMs));
+    }
+
+    builder.bulkMutationsSettings().setBatchingSettings(batchingSettingsBuilder.build());
```
rahulKQL: Will add more mutationsSettings methods

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,174 @@
...
+          .setInitialRpcTimeout(
+            ofMillis(options.getCallOptionsConfig().getShortRpcTimeoutMs()))
+          .setMaxRpcTimeout(ofMillis(options.getCallOptionsConfig().getLongRpcTimeoutMs()))
+          .build();
+      //TODO rahulkql:  Is this correct??, as these settings are for BigtableOptions 
+      //whole but here we are just setting it for bulkMutation.
+      builder.bulkMutationsSettings().setRetrySettings(retrySettings);
+      builder.checkAndMutateRowSettings().setRetrySettings(retrySettings);
+      builder.readModifyWriteRowSettings().setRetrySettings(retrySettings);
+      builder.readRowsSettings().setRetrySettings(retrySettings);
```
igorbernstein2: * readRows should use the long timeout from CallOptionsConfig for totalTimeout
* readRows should use the ReadPartialRowTimeoutMillis as the rpc timeout

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,174 @@
...
+      builder.mutateRowSettings().setRetrySettings(retrySettings);
+      builder.sampleRowKeysSettings().setRetrySettings(retrySettings);
+    } else {
+
```
igorbernstein2: Should retries be disabled for other methods as well?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    builder.setTransportChannelProvider(InstantiatingGrpcChannelProvider.newBuilder()
+        .setChannelsPerCpu(options.getChannelCount())
+        .setMaxInboundMessageSize(retryOptions.getStreamingBufferSize()).build());
+
+    // userAgent can be set on headers.
+    String userAgent = options.getUserAgent();
+    if (null != userAgent)
+      builder.setHeaderProvider(FixedHeaderProvider.create("USER_AGENT", userAgent));
```
igorbernstein2: The header name is wrong here. Please see the logic in:
https://github.com/googleapis/gax-java/blob/master/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcHeaderInterceptor.java#L68

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
rahulKQL: Initially I have disabled retries on all settings(). Would by default enable retries for all, and then for user specified settings we can disable retries. 
does that sounds good?

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/TestBigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,72 @@
+package com.google.cloud.bigtable.hbase;
```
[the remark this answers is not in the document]
rahulKQL: Sorry for late response, Would add those in today.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
igorbernstein2: I think so. Also, please be careful: not all methods are idempotent. For example CheckAndMutate & ReadModifyWrite are not

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,174 @@
...
+      builder.mutateRowSettings().setRetrySettings(retrySettings);
+      builder.sampleRowKeysSettings().setRetrySettings(retrySettings);
+    } else {
+
+      builder.bulkMutationsSettings().setSimpleTimeoutNoRetries(Duration.ZERO);
+    }
+
+    builder.setTransportChannelProvider(InstantiatingGrpcChannelProvider.newBuilder()
+        .setChannelsPerCpu(options.getChannelCount())
+        .setMaxInboundMessageSize(retryOptions.getStreamingBufferSize()).build());
```
igorbernstein2: streamingBufferSize is a different concept from inboundMessageSize:
inboundMessageSize is the size of the grpc's internal buffer
streamingBufferSize doesnt appear to be used, but was intended to be used by RowMerger to limit the number of rows to get merged.

I think streamingBufferSize should be removed altogether

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
rahulKQL: I have added two sets of header in new commit as It could be set in headerProvider as well as ManagedChannelBuilder#userAgent.

rahulKQL:
#### In this commit:
- individual settings method, with retry check.
- default method to set RetrySettings.
- added/fixed code based on suggestions.
- added many inline comment on doubtful points.
- Added default RPC timeout as 100ms in case of No Retry

Would add test cases & removed unneeded comments once it gets reviewed.
@igorbernstein2 Could you please have a look at it

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,310 @@
...
+        FlowControlSettings.newBuilder()
+            .setMaxOutstandingRequestBytes(bulkOptions.getMaxMemory());
+
+    long autoFlushMs = bulkOptions.getAutoflushMs();
+    if (autoFlushMs > 0) {
+      batchSettingsBuilder.setDelayThreshold(Duration.ofMillis(autoFlushMs));
+    }
+    long maxIn = bulkOptions.getMaxInflightRpcs();
+    if(maxIn > 0) {
+      flowControlBuilder.setMaxOutstandingElementCount(maxIn);
```
igorbernstein2: This should be maxInflightRpcs * bulkMaxRowKeyCount

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,310 @@
...
+ * Static methods to convert an instance of {@link Configuration} or {@link BigtableOptions} to a
+ * {@link BigtableDataSettings} instance.
+ */
+public class BigtableDataSettingsFactory {
+  /** Constant <code>LOG</code> */
+  private static final Logger LOG = new Logger(BigtableDataSettingsFactory.class);
+
+  // NOTE: Included Code.UNAUTHENTICATED for default retryCodes
+  private static final Set<Code> DEFAULT_RETRY_CODES =
+      ImmutableSet.of(Code.DEADLINE_EXCEEDED, Code.UNAVAILABLE, Code.ABORTED, Code.UNAUTHENTICATED);
```
igorbernstein2: I dont think we should change the default veneer retry codes

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,310 @@
...
+
+  // NOTE: set rpcTimeout 100ms, without retries.
+  private static final Duration DEFAULT_RPC_TIMEOUT = Duration.ofMillis(100);
```
igorbernstein2: I think this should come from calloptionsconfig

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,310 @@
...
+   */
+  private static void buildSampleRowKeys(Builder builder, BigtableOptions options) {
+    if (options.getRetryOptions().enableRetries()) {
+      builder.sampleRowKeysSettings()
+          .setRetryableCodes(DEFAULT_RETRY_CODES)
+          .setRetrySettings(defaultRetrySettings(options));
+    } else {
+      builder.sampleRowKeysSettings()
+          .setSimpleTimeoutNoRetries(DEFAULT_RPC_TIMEOUT)
+          .getRetryableCodes().clear();
```
igorbernstein2: doesn't setSimpleTimeoutNoRetries already clear the codes?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,310 @@
...
+  private static void buildReadRows(Builder builder, BigtableOptions options) {
+    // TODO: set readPartialRowTimeout for watchdog timer, taken BigtableSession#setupWatchdog()
+    if(options.getRetryOptions().enableRetries()) {
+      builder.readRowsSettings()
+          .setIdleTimeout(ofMillis(options.getRetryOptions().getReadPartialRowTimeoutMillis()))
+          .setRetryableCodes(DEFAULT_RETRY_CODES)
+          .setRetrySettings(defaultRetrySettings(options));
+    } else {
+      builder.readRowsSettings()
+          .setIdleTimeout(ofMillis(options.getRetryOptions().getReadPartialRowTimeoutMillis()))
```
igorbernstein2: This is incorrect idleTimeout is the delay for the user leaving a stream open not the timeout for the server. You want to set the streamingWaitTimeout. There are 2 ways to do it:
1. the rpcTimeout of the RetrySettings gets translated into a waitTimeout
2. set it in calloptions

1 will work for when retries are enabled. However when retries are disabled, you would have to augment the calloptions at the callsite, which is awkward. Please open a ticket in gax-java to expose waitTimeout at the callable level

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
rahulKQL: oh, will update the same.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
rahulKQL: Yes, Will remove unnecessary `getRetryableCodes().clear()`

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
rahulKQL: I would open an issue stating this above scenario. Thanks

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,310 @@
...
+        .setRetryDelayMultiplier(retryOptions.getBackoffMultiplier())
+        .setMaxRetryDelay(ofMillis(retryOptions.getMaxElapsedBackoffMillis()));
+
+    // TODO: verify if maxScanTimeout would cover for maxAttempts or not.
+    retryBuilder.setMaxAttempts(retryOptions.getMaxScanTimeoutRetries());
+
+    // configurations for RPC timeouts
+    retryBuilder
+        .setInitialRpcTimeout(ofMillis(options.getCallOptionsConfig().getShortRpcTimeoutMs()))
+        .setMaxRpcTimeout(ofMillis(retryOptions.getReadPartialRowTimeoutMillis()))
```
rahulKQL: @igorbernstein2 
I set the `rpcTimeout` with `readPartialTimoutMillis`.. As we are expecting clients must have retry enabled then, would this be fine?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,265 @@
...
+   *
+   * @param configuration a {@link BigtableOptions} object.
+   * @return a {@link BigtableDataSettings} object.
+   * @throws IOException if any.
+   */
+  public static BigtableDataSettings fromBigtableOptions(final BigtableOptions options)
+      throws IOException, GeneralSecurityException {
+    if (!options.getRetryOptions().enableRetries()) {
+      throw new IllegalStateException(
+          "Retry is must for BigtableDataSettings configuration from BigtableOptions.");
```
igorbernstein2: The message is a bit awkward. Maybe something like "Disabling retries is not currently supported"

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,265 @@
...
+
+    BigtableDataSettings.Builder builder = BigtableDataSettings.newBuilder();
+
+    InstanceName instanceName = InstanceName.newBuilder().setProject(options.getProjectId())
+        .setInstance(options.getInstanceId()).build();
+    builder.setInstanceName(instanceName);
+    builder.setAppProfileId(options.getAppProfileId());
+
+    LOG.debug("endpoint host %s.", options.getDataHost());
+    LOG.debug("endpoint host %s.", options.getPort());
```
igorbernstein2: This should be logged elsewhere

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,265 @@
...
+    builder.setEndpoint(options.getDataHost() + ":" + options.getPort());
+
+    buildCredentialProvider(builder, options.getCredentialOptions());
+
+    buildBulkOptions(builder, options);
```
igorbernstein2: The naming is a bit inconsistent here, please name it after the method name like `buildBulkMutations`

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,265 @@
...
+    buildReadModifyWrite(builder, options.getCallOptionsConfig().getShortRpcTimeoutMs());
+
+    buildReadRows(builder, options);
+
+    buildMutateRow(builder, options);
+
+    buildSampleRowKeys(builder, options);
+
+    // TODO: would it map to GrpcHeaderInterceptor? or we should build userAgent
+    // using ManagedChannelBuilder AND
```
igorbernstein2: If the user agent on the Managed channel gets preserved, I think that would be cleaner. If it doesn't, add note that it doesnt

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,265 @@
...
+    builder.setHeaderProvider(
+      FixedHeaderProvider.create(GrpcUtil.USER_AGENT_KEY.name(), options.getUserAgent()));
+
+    // TODO: implementation for channelCount or channelPerCPU
+    ManagedChannelBuilder channelBuilder = ManagedChannelBuilder
+        .forAddress(options.getDataHost(), options.getPort())//
```
igorbernstein2: stray //

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,265 @@
...
+        .userAgent(options.getUserAgent());
```
igorbernstein2: Please verify if this works and remove the other approach if it does

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,265 @@
...
+    long autoFlushMs = bulkOptions.getAutoflushMs();
+    long bulkMaxRowKeyCount = bulkOptions.getBulkMaxRowKeyCount();
+    long maxInflightRpcs = bulkOptions.getMaxInflightRpcs();
+
+    if (autoFlushMs > 0) {
+      batchSettingsBuilder.setDelayThreshold(Duration.ofMillis(autoFlushMs));
+    }
+    if (maxInflightRpcs > 0) {
+      flowControlBuilder.setMaxOutstandingElementCount(maxInflightRpcs * bulkMaxRowKeyCount);
+    }
```
igorbernstein2: Please group flowController settings together

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,265 @@
...
+        .setSimpleTimeoutNoRetries(ofMillis(rpcTimeoutMs));
+  }
+
+  /**
+   * To create default RetrySettings, for BigtableDataSettings.
+   *
+   * @param builder a {@link BigtableDataSettings.Builder} object.
+   * @param bulkMutation a {@link BulkOptions} object.
+   */
+  private static RetrySettings defaultRetrySettings(BigtableOptions options) {
```
igorbernstein2: Please rename this to something more appropriate. This method doesn't represent default settings. Maybe something like `buildIdempotentRetrySettings`?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
igorbernstein2: This is mixing up different concepts. For readRows you want to set
initialRpcTimeout = maxRpcTimeout = readPartialRowTimeoutMillis

For non-streaming methods you want to set 
initialRpcTimeout = maxRpcTimeout = shortRpcTimeout

I'm guessing the easiest way to get there is have readRows not use this method

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
```diff
@@ -0,0 +1,265 @@
...
+        .setMaxRetryDelay(ofMillis(retryOptions.getMaxElapsedBackoffMillis()))
+        .setMaxAttempts(retryOptions.getMaxScanTimeoutRetries());
+
+    // configurations for RPC timeouts
+    retryBuilder
+        .setInitialRpcTimeout(ofMillis(options.getCallOptionsConfig().getShortRpcTimeoutMs()))
+        .setMaxRpcTimeout(ofMillis(retryOptions.getReadPartialRowTimeoutMillis()))
+        .setTotalTimeout(ofMillis(options.getCallOptionsConfig().getLongRpcTimeoutMs()));
+
+    // TODO: an option to set RetryOptions#allowRetriesWithoutTimestamp
```
igorbernstein2: I dont think this will ever be supported as a setting. Please throw an unsupported exception instructing users to use ajay's unsafe Mutation factory method

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableDataSettingsFactory.java
sduskis: This actually won't work.  gRPC sets overrides the user agent based on configuration that happens at the creation of the Channel.  `WithUserAgent()` is likely the only way to implement this feature.

rahulKQL:
## Changes

#### BigtableVaneerSettingsFactory
 - Added UserAgent through `HeaderProvider` with `GrpcUtil.USER_AGENT_KEY` which internally translated to ManagedChannel [here](https://github.com/googleapis/gax-java/blob/master/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L228).
 - Moved to [InstantiatingGrpcChannelProvider](https://github.com/googleapis/gax-java/blob/master/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L65) for ChannelProvider, which gives option for Headers along with channel PoolSize.
 - Overriding non-idempotent RPCs only when SHORT_RPC_TIMEOUT is other than default value(i.e. 60seconds).
 - For **BigtableTableAdminSettings**: overriding Endpoint, CredentialProvider & TransportChannelProvider configuration only.


#### TestBigtableVaneerSettingsFactory
- Added [test case](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/1973/files#diff-4acee08860ebdfaf1d83ea5d747b5508R138) which tests `BigtableDataClient` as well `BigtableTableAdminClient` with an actual Bigtable, This test case runs only when it founds "**_test.client.project.id_**" & "**_test.client.project.id_**" VM arguments.



#### Question:
My purpose for using InstantiatingGrpcChannelProvider is to be able to create channelPools, however 
according to [this comment](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/1973#discussion_r235973614), it seems ManagedChannel#WithUserAgnet() should be used to setting user-agent, So for considering multipleChannel:

Should we use the [BigtableSession#createChannelConfig](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java#L478) as it takes care of creating ChannelPool?

rahulKQL:
@sduskis, @igorbernstein2 

Please have a look at this and let me know of your thought.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
```diff
@@ -0,0 +1,312 @@
...
+
+    buildReadModifyWriteSettings(builder, options.getCallOptionsConfig().getShortRpcTimeoutMs());
+
+    buildReadRowsSettings(builder, options);
+
+    buildMutateRowSettings(builder, options);
+
+    buildSampleRowKeysSettings(builder, options);
+
+    String userAgent = BigtableVersionInfo.CORE_UESR_AGENT + "," + options.getUserAgent();
```
igorbernstein2: s/CORE_UESR_AGENT/CORE_USER_AGENT

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
```diff
@@ -0,0 +1,312 @@
...
+    HeaderProvider headers = FixedHeaderProvider.create(GrpcUtil.USER_AGENT_KEY.name(), userAgent);
```
igorbernstein2: Does this actually work or does grpc clobber it?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
```diff
@@ -0,0 +1,312 @@
...
+   * @throws IOException if any.
+   */
+  public static BigtableTableAdminSettings createTableAdminSettings(BigtableOptions options)
+      throws IOException {
+    checkState(options.getProjectId() != null, "Project ID is required");
+    checkState(options.getInstanceId() != null, "Instance ID is required");
+
+    BigtableTableAdminSettings.Builder adminBuilder = BigtableTableAdminSettings.newBuilder();
+
+    adminBuilder.setInstanceName(com.google.bigtable.admin.v2.InstanceName
```
igorbernstein2: Please add a note to stop using typesafe names

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
```diff
@@ -0,0 +1,312 @@
...
+  }
+
+  /**
+   * To build BigtableDataSettings#mutateRowSettings with default retries settings.
+   *
+   * @param builder a {@link BigtableDataSettings.Builder} object.
+   * @param options a {@link BigtableOptions} object.
+   */
+  private static void buildMutateRowSettings(Builder builder, BigtableOptions options) {
+    builder.mutateRowSettings()
```
igorbernstein2: where are the retry codes configured? (here and everywhere else)

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
```diff
@@ -0,0 +1,312 @@
...
+    // configurations for RPC timeouts
+    Duration readPartialRowTimeout = ofMillis(retryOptions.getReadPartialRowTimeoutMillis());
+    retryBuilder
+        .setInitialRpcTimeout(readPartialRowTimeout)
+        .setMaxRpcTimeout(readPartialRowTimeout)
+        .setTotalTimeout(ofMillis(options.getCallOptionsConfig().getLongRpcTimeoutMs()));
+
+    builder.readRowsSettings()
+        .setRetrySettings(retryBuilder.build());
+  }
```
igorbernstein2: Please update google-cloud-java dep and add config for single row reads

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
```diff
@@ -0,0 +1,312 @@
...
+  private static CredentialsProvider buildCredentialProvider(
+      CredentialOptions credentialOptions) throws IOException {
+    CredentialsProvider credentialsProvider = NoCredentialsProvider.create();
+
+    switch (credentialOptions.getCredentialType()) {
+      case DefaultCredentials:
+        credentialsProvider = BigtableStubSettings.defaultCredentialsProviderBuilder().build();
+        break;
+      case P12:
+      case SuppliedCredentials:
```
igorbernstein2: should this be something along the lines of 
```
credential = ((UserSuppliedCredentialOptions)credentialOptions).getCredential();
credentialsProvider = FixedCredentialsProvider.create(credential);
```

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
```diff
@@ -0,0 +1,312 @@
...
+   */
+  private static CredentialsProvider buildCredentialProvider(
+      CredentialOptions credentialOptions) throws IOException {
+    CredentialsProvider credentialsProvider = NoCredentialsProvider.create();
+
+    switch (credentialOptions.getCredentialType()) {
+      case DefaultCredentials:
+        credentialsProvider = BigtableStubSettings.defaultCredentialsProviderBuilder().build();
+        break;
+      case P12:
```
igorbernstein2: Shouldnt this do something similar to 
https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CredentialFactory.java#L169-L181

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
rahulKQL: Renamed the constant

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
rahulKQL: Updated classes to use with raw ids

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
rahulKQL: Updated [buildCredentialProvider()](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/1973/files#diff-3b1db200ad0c45bb25cfa85e47b21766R284) to use [CredentialFactory#getCredentials](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CredentialFactory.java#L129-L156)

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
rahulKQL: Thanks for the pointer, Have updated this class to utilize the same.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
rahulKQL: As of now, not overriding `RetryCodes` of `BigtableDataSettings`. 
Defaults retrycode in GCJ:
**[DEADLINE_EXCEEDED, UNAVAILABLE]** --> sampleRowKeysSettings, readRowsSettings, mutateRowSettings.
**empty** for bulkMutationsSettings, readModifyWriteRowSettings, checkAndMutateRowSettings.

Default retrycodes in [bigtable-client](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/RetryOptions.java#L50).

However, I can override these with operation specific retry codes.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
rahulKQL: I think we would need readRowSettings() in the [BigtableDataSettings](https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java) as well. For now, It is available in [EnhanccedBigtableStubSettings](https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java) only.

Let me know If I should add this change.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
rahulKQL: I think It is getting translated to both the clients. 

I have added a simple test case for BigtableDataSettigns & BigtbaleTableAdminSettings with [ActualBigtable here](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/1973/files#diff-4acee08860ebdfaf1d83ea5d747b5508R140) (which we may remove as it might not needed here). 

So It created 32 ManagedChannels under channels pool when I run it in my local machine, Also It had the persisted user-agent on those channels. (attaching snapshot from debug console).

igorbernstein2: I think this needs to be verified on the server side and there should be a test to confirm this.  Please add a todo

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
igorbernstein2: There are few more defaults for gcj in the overlay. The gapic generated client settings can't express all of the settings, so they are overlaid here:
https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java#L273-L321

This is where bulkMutations have their defaults set.

I'm trying to understand why we shouldn't copy the retry codes from cbt?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
igorbernstein2: Please do, I apologize for the omission

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
igorbernstein2: Can you add a comment about how gax-java has special handling for  the USER_AGENT_KEY:
https://github.com/googleapis/gax-java/blob/master/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcHeaderInterceptor.java#L63-L74

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
```diff
@@ -0,0 +1,298 @@
...
+    adminBuilder.setInstanceId(options.getInstanceId());
+
+    final String endpoint = options.getAdminHost() + ":" + options.getPort();
+    adminBuilder.stubSettings().setEndpoint(endpoint);
+
+    //Overriding default credential with BigtableOptions's credential.
+    adminBuilder.stubSettings()
+        .setCredentialsProvider(buildCredentialProvider(options.getCredentialOptions()));
+
+    String userAgent = BigtableVersionInfo.CORE_USER_AGENT + "," + options.getUserAgent();
```
igorbernstein2: Please add something here that would allow us to disambiguate when the hbase client is using the gcj adapter

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
rahulKQL: Added a separate test case for User-agent [TestUserAgent](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/1973/files#diff-8b4fb9399d066a371cb9b54a9a6d871cR59), verifying header with the help of HeaderServerInterceptor for [PlainText](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/1973/files#diff-8b4fb9399d066a371cb9b54a9a6d871cR91) as well as [TLS](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/1973/files#diff-8b4fb9399d066a371cb9b54a9a6d871cR122) Negotiation (_Using Client_trust, Server_Cert & Private key for the TLS user agent test case_).

Added [an inline comment here](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/1973/files#diff-6a492166d0005302ce3c542cdcf00ad5R283) about User-Agent handling.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
rahulKQL: Added [buildReadRowSettings()](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/1973/files#diff-6a492166d0005302ce3c542cdcf00ad5R174) for point reads.

As of now set it using the same configuration as readRowsSettings.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
rahulKQL: Does ["VENEER_ADAPTER"](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/1973/files#diff-6a492166d0005302ce3c542cdcf00ad5R54) sound good?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVaneerSettingsFactory.java
rahulKQL: I think for **Non-idempotent** operation we can utilize GCJ's default values, as we are overriding it with shortRpcTimeout & no retryCodes (_here both adapters have the default value of 1 min_).

In case of **Idempotent** operation default values varies alot & 
 - [CBT has retryCode](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/RetryOptions.java#L49-L54) as `{DEADLINE_EXCEEDED, UNAVAILABLE, ABORTED, UNAUTHENTICATED}` 
- [GCJ's retryCodes](https://github.com/googleapis/googleapis/blob/master/google/bigtable/v2/bigtable_gapic.yaml#L35-L39) are only `{DEADLINE_EXCEEDED, UNAVAILABLE}`.

rahulKQL:
## Summary of Client Settings changes

After many Iteration now in this PR, as of now we are:
  - Overriding both client setting's endpoint with BigtableOptions endpoint [here](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/1973/files#diff-6a492166d0005302ce3c542cdcf00ad5R83).
  - Utilizing existing [CredentialFactory#getCredentials](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/1973/files#diff-6a492166d0005302ce3c542cdcf00ad5R138) for CredentialProvider.
  - Providing ["VENEER_ADAPTER"](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/1973/files#diff-6a492166d0005302ce3c542cdcf00ad5R300) in the User-Agent to be able to distinguish GCJ adapters.
  - Created TransportChannelProvider using InstantiatingGrpcChannelProvider [here](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/1973/files#diff-6a492166d0005302ce3c542cdcf00ad5R306):
    - Kept ChannelsPerCpu to 2 (Though it BigtableOptions has 4 ChannelsPerCpu as default).
    - Kept MaxInboundMessageSize to CBT's default(i.e. 256MB).
    - For plainText Negotiation using [ChannelConfigurator](https://github.com/googleapis/gax-java/blob/master/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L450-L462) (_would this be ok?, as its not stable yet?_).
  - Created [buildIdempotentRetrySettings](https://github.com/rahulKQL/cloud-bigtable-client/blob/dataSettings_addition/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java#L260-L284) to configure generic RetrySettings.

    ### [createBigtableDataSettings](BigtableVeneerSettingsFactory)
      - As of now we are not supporting disabling the retries.
      - In case of [bulkMutationsSettings](https://github.com/rahulKQL/cloud-bigtable-client/blob/dataSettings_addition/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java#L150-L184), We override:
            - BatchingSettings with BulkOptions values.
            - RetrySettings with idempotentRetrySettings.
            - RetryCodes with BigtableOptions's Retries Code.
      - In case of [sampleRowKeysSettings](https://github.com/rahulKQL/cloud-bigtable-client/blob/dataSettings_addition/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java#L208-L216), [mutateRowSettings](https://github.com/rahulKQL/cloud-bigtable-client/blob/dataSettings_addition/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java#L218-L226) and [readRowSettings](https://github.com/rahulKQL/cloud-bigtable-client/blob/dataSettings_addition/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java#L228-L236), We overrides
             - RetrySettings with idempotentRetrySettings.
             - RetryCodes with BigtableOptions's Retries Code.
      - In case of [readRowsSettings](https://github.com/rahulKQL/cloud-bigtable-client/blob/dataSettings_addition/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java#L238-L258), We override
             - RetrySettings with idempotentRetrySettings considering streaming method according to this suggestion.
             - RetryCodes with BigtableOptions's Retries Code.
      - In case of checkAndMutateRowSettings & readModifyWriteRowSettings, We override SimpleTimeoutNoRetries when shortRpcTimeoutMs is not default.
       
    ### [createTableAdminSettings](https://github.com/rahulKQL/cloud-bigtable-client/blob/dataSettings_addition/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java#L113-L132)
      - Only Overrided [endpoint, CredentialsProvider & TransportChannelProvider](https://github.com/rahulKQL/cloud-bigtable-client/blob/dataSettings_addition/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java#L123-L129). Do we need to override any specific settings for this?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java
```diff
@@ -0,0 +1,327 @@
...
+  private static TransportChannelProvider createChannelProvider(String hostname,
+      BigtableOptions options) {
+    final String endpoint = hostname + ":" + options.getPort();
+    String userAgent = VENEER_ADAPTER + options.getUserAgent();
+
+    //InstantiatingGrpcChannelProvider has special handling for User-Agent header. It set the
+    //ChannelBuilder with provided value.
+    HeaderProvider headers = FixedHeaderProvider.create(USER_AGENT_KEY.name(), userAgent);
+
+    InstantiatingGrpcChannelProvider.Builder builder =
```
igorbernstein2: Please a TODO to refactor google-cloud-java to expose a static `defaultTransportChannelProvider` method and then customize it.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java
```diff
@@ -0,0 +1,327 @@
...
+        InstantiatingGrpcChannelProvider.newBuilder()
+            .setChannelsPerCpu(2)
+            .setHeaderProvider(headers)
+            .setEndpoint(endpoint)
+            .setPoolSize(options.getChannelCount())
+            .setMaxInboundMessageSize(MAX_MESSAGE_SIZE);
```
igorbernstein2: This seems to ignore BigtableOptions.getDataChannelCount(). Might be better to remove the per cpu channel counts and just use the absolute channel count via `setPoolSize`

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java
```diff
@@ -0,0 +1,327 @@
...
+   * To create an instance of {@link BigtableDataSettings} from {@link BigtableOptions}.
+   *
+   * @param options a {@link BigtableOptions} object.
+   * @return a {@link BigtableDataSettings} object.
+   * @throws IOException if any.
+   */
+  public static BigtableDataSettings createBigtableDataSettings(@Nonnull final BigtableOptions options) throws IOException {
+    checkState(options.getProjectId() != null, "Project ID is required");
+    checkState(options.getInstanceId() != null, "Instance ID is required");
+    checkState(options.getRetryOptions().enableRetries(), "Disabling retries is not currently supported.");
```
igorbernstein2: I don't think useCachedDataPool option is supported here, please throw an error if its enabled

igorbernstein2: useEmulator option is not covered here. Please throw an error and add a todo to support it

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java
```diff
@@ -0,0 +1,327 @@
...
+    String userAgent = VENEER_ADAPTER + options.getUserAgent();
+
+    //InstantiatingGrpcChannelProvider has special handling for User-Agent header. It set the
+    //ChannelBuilder with provided value.
+    HeaderProvider headers = FixedHeaderProvider.create(USER_AGENT_KEY.name(), userAgent);
+
+    InstantiatingGrpcChannelProvider.Builder builder =
+        InstantiatingGrpcChannelProvider.newBuilder()
+            .setChannelsPerCpu(2)
+            .setHeaderProvider(headers)
```
igorbernstein2: Shouldn't the header provider be set on the top level settings? If I understand correctly, the header provider on the channel provider is set by the client context during build: 
https://github.com/googleapis/gax-java/blob/122eda65b4857cce676b53f07d19fef59dc0cb15/gax/src/main/java/com/google/api/gax/rpc/ClientContext.java#L151

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java
```diff
@@ -0,0 +1,327 @@
...
+        .setRetryDelayMultiplier(retryOptions.getBackoffMultiplier())
+        .setMaxRetryDelay(ofMillis(retryOptions.getMaxElapsedBackoffMillis()))
+        .setMaxAttempts(retryOptions.getMaxScanTimeoutRetries());
+
+    // configurations for RPC timeouts
+    Duration shortRpcTimeout = ofMillis(options.getCallOptionsConfig().getShortRpcTimeoutMs());
+    retryBuilder
+        .setInitialRpcTimeout(shortRpcTimeout)
+        .setMaxRpcTimeout(shortRpcTimeout)
+        .setTotalTimeout(ofMillis(options.getCallOptionsConfig().getLongRpcTimeoutMs()));
```
igorbernstein2: This doesn't look right. I think that for all rpcs except ReadRows (plural), the mapping should be:
- if useTimeouts == true, then initialRpcTimeout = maxRpcTimeout = shortRpcTimeout
- if useTimeouts == false, then initialRpcTimeout = maxRpcTimeout = [6 minutes?](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java#L74)
- totalTimeout = RetryOptions.getMaxElapsedBackoffMillis
- maxAttempts = 0 (disabled) (@sduskis, am I correct that classic java client doesn't limit attempts for other rpcs?)

For readRows
- initialRpcTimeout = maxRpcTimeout = RetryOptions.readPartialRowTimeoutMillis
- if (useTimeouts) totalTimeout = longRpcTimeoutMs
- maxAttempts = RetryOptions.getMaxScanTimeoutRetries

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java
```diff
@@ -0,0 +1,327 @@
...
+  /** Creates {@link RetrySettings} for non-streaming idempotent method. */
+  private static RetrySettings buildIdempotentRetrySettings(RetrySettings retrySettings,
+      BigtableOptions options) {
+    RetryOptions retryOptions = options.getRetryOptions();
+    RetrySettings.Builder retryBuilder = retrySettings.toBuilder();
+
+    retryBuilder
+        .setInitialRetryDelay(ofMillis(retryOptions.getInitialBackoffMillis()))
+        .setRetryDelayMultiplier(retryOptions.getBackoffMultiplier())
+        .setMaxRetryDelay(ofMillis(retryOptions.getMaxElapsedBackoffMillis()))
```
igorbernstein2: getMaxElapsedBackoffMillis is the overall timeout, not the max delay. It seems like the classic client hardcodes maxRetryDelay to 1 minute:
https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java#L338

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java
```diff
@@ -0,0 +1,327 @@
...
+        .setMaxAttempts(retryOptions.getMaxScanTimeoutRetries());
```
igorbernstein2: There is no max attempts in the classic client. Please use 0 to disable

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java
```diff
@@ -0,0 +1,327 @@
...
+  }
+
+  /** To build BigtableDataSettings#readRowsSettings with default Retry settings. */
+  private static void buildReadRowsSettings(Builder builder, BigtableOptions options) {
+    RetryOptions retryOptions = options.getRetryOptions();
+
+    RetrySettings.Builder retryBuilder = builder.readRowsSettings().getRetrySettings().toBuilder()
+        .setInitialRetryDelay(ofMillis(retryOptions.getInitialBackoffMillis()))
+        .setRetryDelayMultiplier(retryOptions.getBackoffMultiplier())
+        .setMaxRetryDelay(ofMillis(retryOptions.getMaxElapsedBackoffMillis()))
```
igorbernstein2: maxRetryDelay = 1 minute

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java
```diff
@@ -0,0 +1,327 @@
...
+          .setSimpleTimeoutNoRetries(ofMillis(rpcTimeoutMs));
+    }
+  }
+
+  /**
+   * Builds BigtableDataSettings#readModifyWriteRowSettings when short timeout is other
+   * than 60_000 ms.
+   */
+  private static void buildReadModifyWriteSettings(Builder builder, long rpcTimeoutMs) {
+    if(rpcTimeoutMs != SHORT_TIMEOUT_MS_DEFAULT) {
```
igorbernstein2: why if? (here and above)

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java
```diff
@@ -0,0 +1,327 @@
...
+  /**
+   * To create an instance of {@link BigtableDataSettings} from {@link BigtableOptions}.
+   *
+   * @param options a {@link BigtableOptions} object.
+   * @return a {@link BigtableDataSettings} object.
+   * @throws IOException if any.
+   */
+  public static BigtableDataSettings createBigtableDataSettings(@Nonnull final BigtableOptions options) throws IOException {
+    checkState(options.getProjectId() != null, "Project ID is required");
+    checkState(options.getInstanceId() != null, "Instance ID is required");
```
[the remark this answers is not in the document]
rahulKQL: I have added these for an extra layer of validation. Would remove them.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java
rahulKQL: If shortTimeOutMs is at default then assuming the client did not customize BigtableOption's settings, 
that is why we are not overriding the underlying settings here(or directly using defaults of GCJ's settings).

igorbernstein2: I think you should override with the BigtableOptions in case the diverge

rahulKQL: > override with the BigtableOptions in case the diverge
By saying above comment,
do you mean we should check all configuration against their default value(present in BigtableOption) & then override only those which are different?
OR
Should we override only in case of specific operation(say non-streming operations) by checking BigtableOption's present value against BigtableDataSetting's default value?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java
rahulKQL: Removed ChannelPerCPU() config.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java
rahulKQL: Added state check for useCachedChannel:
```java
checkState(!options.useCachedChannel(), "cachedDataPool is not currently supported.");
````

Comment (rahulKQL) [reply bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java]: I hope checking against Environment variable would be sufficient:
```
String emulatorHost = System.getenv(BIGTABLE_EMULATOR_HOST_ENV_VAR);
checkState(emulatorHost == null, "Emulator Hosting is not supported yet.");
````

File: bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java (hunk 36 above)
Comment (rahulKQL) [reply bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java]: removed overriding of maxAttempts(except ReadRows), it would now be defaults to `setMaxAttempts(0)`.

File: bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java (hunk 35 above)
Comment (rahulKQL) [reply bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java]: Thanks for the reference, Have set MaxRetryDelay to 1 min using `MAX_RETRY_TIMEOUT_MS = 60_000`

File: bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java (hunk 34 above)
Comment (rahulKQL) [reply bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java]: Thanks a lot for detailed values. Have followed the same for configuring all idempotent retry settings.

File: bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java (hunk 33 above)
Comment (rahulKQL) [reply bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java]: Unfortunately, while [creating `BigtableStubSettings` in `EnhancedBigtableStub#create`](https://github.com/rahulKQL/google-cloud-java/blob/master/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java#L88-L94), we are not considering client’s `HeaderProvider` value present in the argument.

I cross checked,  only default header i.e. `x-goog-api-client` is available at ClientContext’s Xtor. We would need user-agent to be available at `TransportChannelProvider`, for that reason I created `InstantiatingGrpcChannelProvider`.
Also we had to set `ManagedChannelBuilder#usePlainText` to support plainText negotiation rpcs.

Do let me know If I should raise PR to consider HeaderProvider at `EnhancedBigtableStub#create` before ClientContext creation.

File: bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java (hunk 32 above)
Comment (igorbernstein2) [reply bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java]: Actually nevermind, the emulator should work fine. Sorry for the noise

File: bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java (hunk 33 above)
Comment (igorbernstein2) [reply bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java]: Please do, its a bug: the header provider should be copied over in the enhanced stub

File: bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java (hunk 38 above)
Comment (igorbernstein2) [reply bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java]: I just meant you should remove the if statement and always setSimpleTimeoutNoRetries disrarding if the if its the default value or not. If we decide that to change the gapic config to 30 secs, but forget to change SHORT_TIMEOUT_MS_DEFAULT we will have confusing behavior: the user will check their BigtableOptions and see SHORT_TIMEOUT_MS_DEFAULT = 1 minute, but their rpcs will exceed deadline @ 30secs

Comment (rahulKQL) [reply bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java]: Thanks for the clarification!! 
I followed the same [here](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/1973/files#diff-6a492166d0005302ce3c542cdcf00ad5R96).

File: bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java (hunk 33 above)
Comment (rahulKQL) [reply bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java]: [PR#4491](https://github.com/googleapis/google-cloud-java/pull/4491) for headerProvider addition is merged now. 

Once the next release(i.e. v0.81.0) is available, I will have this PR rebase and updated.

**Comment by igorbernstein2** [comment]
I just submitted https://github.com/googleapis/google-cloud-java/pull/4507, it should help with the transport builder

File: bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java [hunk 40]
```diff
@@ -0,0 +1,314 @@
...
+ * {@link BigtableDataSettings} or {@link BigtableTableAdminSettings} instance .
+ */
+public class BigtableVeneerSettingsFactory {
+
+  /** Constant <code>LOG</code> */
+  private static final Logger LOG = new Logger(BigtableVeneerSettingsFactory.class);
+
+  //Identifier to distinguish between CBT or GCJ adapter.
+  private static final String VENEER_ADAPTER =  BigtableVersionInfo.CORE_USER_AGENT+","
+      + "VENEER_ADAPTER,";
```
igorbernstein2: nit: Why is this capitalized?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java
```diff
@@ -0,0 +1,314 @@
...
+   * @param options a {@link BigtableOptions} object.
+   * @return a {@link BigtableDataSettings} object.
+   * @throws IOException if any.
+   */
+  public static BigtableDataSettings createBigtableDataSettings(
+      @Nonnull final BigtableOptions options) throws IOException {
+    checkState(options.getRetryOptions().enableRetries(), "Disabling retries is not currently supported.");
+    checkState(!options.useCachedChannel(), "cachedDataPool is not currently supported.");
+
+    //TODO:add configuration for emulator hosting.
```
igorbernstein2: what is the todo for?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java
```diff
@@ -0,0 +1,314 @@
...
+
+  /** Creates {@link RetrySettings} for non-streaming idempotent method. */
+  private static RetrySettings buildIdempotentRetrySettings(RetrySettings retrySettings,
+      BigtableOptions options) {
+    RetryOptions retryOptions = options.getRetryOptions();
+    CallOptionsConfig callOptions = options.getCallOptionsConfig();
+    RetrySettings.Builder retryBuilder = retrySettings.toBuilder();
+
+    if (retryOptions.allowRetriesWithoutTimestamp()) {
+      LOG.warn("Retries without Timestamp does not support yet.");
```
igorbernstein2: This should throw instead of warn

rahulKQL:
@sduskis 

Will be just rebasing and updating these three last comments

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java
rahulKQL: This was intended to do separate emulator configuration, but I think we agreed on not to have the separate configuration.

PR #1975, opened 2018-10-29
Interface additions

elisheva-qlogic:
@ajaaym Here are the additions to the IBigtableDataClient interface


sduskis:
Closing this in favor of another approach.

PR #1990, opened 2018-11-06
First key only filter strip values

mcorgan:


sduskis:
I'm going to open a different PR with tests

PR #1994, opened 2018-11-09
Fix backoff NPE when batch mutator runs out of retry attempts 

igorbernstein2:
When we migrated to gax's RetryAlgorithm, we forgot to migrate the all call sites. The bulk mutation handler was still expecting a -1 to signal end of retries, but it should've been expecting a null.


bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/async/OperationClock.java
```diff
@@ -44,6 +43,10 @@ public OperationClock() {
     this.timeNs = System.nanoTime();
   }
 
+  @VisibleForTesting
+  void setTime(long time, TimeUnit timeUnit) {
+    this.timeNs = timeUnit.toNanos(time);
+  }
   @Override
```
sduskis: Please add a new line.

PR #1995, opened 2018-11-09
strip values off cells in FirstKeyOnlyFilter

mcorgan:


sduskis:
@mcorgan: I submitted #1996 which also has unit / system tests.

PR #1996, opened 2018-11-09
Fix `KeyOnlyFilterAdapter` and `KeyOnlyFilterAdapter`

### LINKED ISSUE #1987: KeyOnlyFilter shouldn't limit cells per row
https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/3a8d06156681bf2c86a39fbb80c000a66cf05201/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/KeyOnlyFilterAdapter.java#L45

I think this is a mismatch with HBase's definition of the filter: https://hbase.apache.org/1.2/apidocs/org/apache/hadoop/hbase/filter/KeyOnlyFilter.html

"A filter that will only return the key component of each KV (the value will be rewritten as empty).
This filter can be used to grab all of the keys without having to also grab the values."

With the limit in place, it acts like a FirstKeyOnlyFilter.

### LINKED ISSUE #1989: FirstKeyOnlyFilterAdapter could strip value
https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/3a8d06156681bf2c86a39fbb80c000a66cf05201/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/FirstKeyOnlyFilterAdapter.java#L32

Could this adapter also do FILTERS.value().strip() for potentially better performance?

sduskis:
- `FirstKeyOnlyFilterAdapter` now strips value (Fixes #1989)
- `KeyOnlyFilter` returns all key/family/qualifier combinations (Fixes #1987)

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestFilters.java
+++ b/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestFilters.java
@@ -1370,9 +1370,6 @@ public abstract class AbstractTestFilters extends AbstractTest {
     Get get = new Get(rowKey).setFilter(filter);
     Result result = table.get(get);
     Assert.assertEquals("Should only return 1 keyvalue", 1, result.size());
-    // The following is true for Bigtable, but not HBase.
-//    Assert.assertEquals("Should only have 0 bytes", 0,
-//        result.rawCells()[0].getValueLength());
 
     table.close();
   }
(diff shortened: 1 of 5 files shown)


bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestFilters.java
```diff
@@ -1370,6 +1370,9 @@ public void testFirstKeyFilter() throws IOException {
     Get get = new Get(rowKey).setFilter(filter);
     Result result = table.get(get);
     Assert.assertEquals("Should only return 1 keyvalue", 1, result.size());
+    // The following is true for Bigtable, but not HBase.
+//    Assert.assertEquals("Should only have 0 bytes", 0,
+//        result.rawCells()[0].getValueLength());
```
igorbernstein2: Consider removing the comment

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/FirstKeyOnlyFilterAdapter.java
```diff
@@ -29,7 +29,10 @@
  */
 public class FirstKeyOnlyFilterAdapter extends TypedFilterAdapterBase<FirstKeyOnlyFilter> {
 
-  private static RowFilter LIMIT_ONE = FILTERS.limit().cellsPerRow(1).toProto();
+  private static RowFilter LIMIT_ONE = FILTERS.chain()
+      .filter(FILTERS.value().strip())
+      .filter(FILTERS.limit().cellsPerRow(1))
```
igorbernstein2: I'm not sure if it makes that much of a difference, but logically it would be more efficient to limit cells first

PR #1998, opened 2018-11-13
bigtable-beam-import: allow users to configure concurrency of the split probing and make it high by default

igorbernstein2:
Also, fix validation


bigtable-dataflow-parent/bigtable-beam-import/src/main/resources/log4j.properties
```diff
@@ -21,3 +21,4 @@ log4j.appender.stdout.Target=System.out
 log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
 log4j.appender.stdout.layout.ConversionPattern=%d{yyyy-MM-dd HH:mm:ss} %-5p %c{1}:%L - %m%n
 
+log4j.category.org.apache.hadoop.io.compress.CodecPool=WARN
```
sduskis: new line?

igorbernstein2: fixed, thanks for reviewing!

PR #1985, opened 2018-10-31
Admin method changes 

rahulKQL:
### Commit Contains: 
Veneer client related changes for `AbstractBigtableAdmin` which is tracked under #1974 

Adding updates for following methods:
- [x] createTableAsync
- [x] addColumn
- [x]  modifyColumn
- [x]  deleteColumn
- [x] createTable
- [x]  modifyColumns
- [x] deleteColumn

Also Added `ColumnDescriptorAdapter#buildGarbageCollectionRule` for `com.google.cloud.bigtable.admin.v2.models.GCRules.GCRule` creation.

Issue #1974, opened 2018-10-25
Update admin methods to use Veneer client.


ajaaym:
Following methods needs to be updated.

- [x] listTableNames
- [x] requestTableList
- [x] getTableDescriptor
- [x] getTableNames
- [x] createTable
- [x] createTableAsync
- [x] deleteTable
- [x] deleteTables
- ~~[ ] enableTable~~
- ~~[ ] enableTables~~
- ~~[ ] disableTable~~
- ~~[ ] disableTables~~
- ~~[ ] isTableEnabled~~
- ~~[ ] isTableDisabled~~
- [x] isTableAvailable
- [x] tableExists
- [x] addColumn
- [x] modifyColumn
- [x] deleteColumn
- [x] modifyTable
- [x] modifyColumns
- [x] deleteColumn

sduskis:
@ajaaym: I crossed out methods that are in memory only.

rahulKQL:
Have updated underlying grpc client to now use `BigtableTableAdminClientWrapper`.
Referencing related PRs:

 #2045 
- listTables
- createTable
- createTableAsync
- deleteTable
- modifyFamilies
- dropRowRange

#2089 
* getTableAsync
* deleteTableAsync
* modifyColumnFamilyAsync
* dropRowRangeAsync

#2094 
- snapshot operations.

#2095 
 - BigtableAsyncAdmin#listTablesAsync

rahulKQL:
@sduskis, Can we close this issue?
I believe all the operations are now adapted with GCJ models.

sduskis:
We are almost done.  We need to remove all references to `BigtableTableAdminClient` with the exception of `BigtableSession`.  The `BigtableSession` method should be marked as deprecated.

PR #2009, opened 2018-11-23
adding createTable & modifyTable changes

rahulKQL:
Veneer client related changes for Admin method tracked under #1974 (Previously created under #1985).

- now these methods are adapt to `v2.models. ModifyColumnFamiliesRequest` : 
  - [x] createTableAsync
  - [x] addColumn
  - [x]  modifyColumn
  - [x]  deleteColumn
  - [x]  createTable
  - [x] modifyColumns
  - [x] deleteColumn
  - [x] modifyTable

- Updated `ModifyTableBuilder` to adapt to `v2.models. ModifyColumnFamiliesRequest`, removed unused `add`,`modify`, `delete` method from it.

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/util/ModifyTableBuilder.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/util/ModifyTableBuilder.java
@@ -15,29 +15,20 @@
  */
 package com.google.cloud.bigtable.hbase.util;
 
-import com.google.bigtable.admin.v2.ModifyColumnFamiliesRequest;
-import com.google.cloud.bigtable.grpc.BigtableTableName;
-import com.google.cloud.bigtable.hbase.adapters.admin.ColumnDescriptorAdapter;
+import com.google.cloud.bigtable.admin.v2.models.ModifyColumnFamiliesRequest;
 import org.apache.hadoop.hbase.HColumnDescriptor;
 import org.apache.hadoop.hbase.HTableDescriptor;
 import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.util.Bytes;
 
-import java.util.ArrayList;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Set;
 
+import static com.google.cloud.bigtable.hbase.adapters.admin.ColumnDescriptorAdapter.buildGarbageCollectionRule;
 /**
  * Utilitiy to create {@link ModifyColumnFamiliesRequest} from HBase {@link HColumnDescriptor}s.
  */
 public class ModifyTableBuilder {
-
-  private static final ColumnDescriptorAdapter ADAPTER = new ColumnDescriptorAdapter();
-
-  public static ModifyTableBuilder create() {
-    return new ModifyTableBuilder();
-  }
 
   private static Set<String> getColumnNames(HTableDescriptor tableDescriptor) {
     Set<String> names = new HashSet<>();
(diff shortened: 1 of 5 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/util/ModifyTableBuilder.java
```diff
@@ -48,70 +39,34 @@ public static ModifyTableBuilder create() {
...
     Set<String> currentColumnNames = getColumnNames(currentTableDescriptors);
     Set<String> newColumnNames = getColumnNames(newTableDescriptor);
 
     for (HColumnDescriptor hColumnDescriptor : newTableDescriptor.getFamilies()) {
-      if (currentColumnNames.contains(hColumnDescriptor.getNameAsString())) {
-        builder.modify(hColumnDescriptor);
+      String columnName = hColumnDescriptor.getNameAsString();
+      if (currentColumnNames.contains(columnName)) {
+        request.updateFamily(columnName,
+                buildGarbageCollectionRule(hColumnDescriptor));
```
sduskis: Please use a single variable for the result of `buildGarbageCollectionRule(hColumnDescriptor)` before the `if`.

bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -109,7 +111,7 @@
    * <p>
    * Constructor for AbstractBigtableAdmin.
    * </p>
-   * @param connection a {@link org.apache.hadoop.hbase.client.AbstractBigtableConnection} object.
+   * @param connection a {@link CommonConnection} object.
```
sduskis: Thanks!

bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -618,26 +638,17 @@ public void modifyTable(TableName tableName, HTableDescriptor newDecriptor) thro
...
       return null;
     } catch (Throwable throwable) {
       throw new IOException(
-          String.format(
-              "Failed to %s column '%s' in table '%s'",
-              modificationType,
-              columnName,
-              tableName.getNameAsString()),
-          throwable);
+              "Failed to while modifying column", throwable);
```
sduskis: Please restore the details of the exception.

bigtable-hbase-2.x-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/BigtableAdmin.java
```diff
@@ -145,37 +145,44 @@ public void createTable(TableDescriptor desc, byte[][] splitKeys) throws IOExcep
...
+   *
    * Calling {@link #addColumn(TableName, ColumnFamilyDescriptor)} was causing stackoverflow.
    * Copying the same code here. //TODO - need to find a better way
    */
   @Override
   public void addColumnFamily(TableName tableName, ColumnFamilyDescriptor columnFamilyDesc)
       throws IOException {
-    modifyColumns(tableName, columnFamilyDesc.getNameAsString(), "add",
-        ModifyTableBuilder.create().add(TableAdapter2x.toHColumnDescriptor(columnFamilyDesc)));
+    ModifyColumnFamiliesRequest request =
```
sduskis: This sequence is very verbose.  Let's discuss an HBase wrapper of ModifyColumnFamiliesRequest

rahulKQL: Should that wrapper be similar to [ModifyTableAdapter](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/b78bb1519ea4c28fea47dd98555790be046433f8/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/util/ModifyTableBuilder.java#L81), which may return new `v2.models. ModifyColumnFamiliesRequest` instead of any builder or wrapper?

sduskis: The wrapper should be similar to [`CheckAndMutateUtil.RequestWrapper`](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/CheckAndMutateUtil.java#L84).  The Adapter has a clear "from" and "to" with a single method.  We need something more like a Builder.

PR #2023, opened 2018-12-11
garbage collect stale tables 15 mins after they are created

igorbernstein2:
The benefit is that this prevents concurrent test runs from interfering with each other.
It accomplishes this by encoding the current timestamp into the table name and only deleting tables whose timestamps are old than 15 minutes


sduskis:
You shouldn't need this.  Multiple tests should not be run at the same time.  Jenkins has locks for this purpose.

igorbernstein2:
I've noticed quite a few failures due to table not found in the jenkins tests.  If I was wrong, then this will make it easier to reproduce the failures when running the tests manually

PR #2025, opened 2018-12-13
Refactored value provider

rahulKQL:
Refactoring `ReadRowsRequest` creation into a new `RequestWithScanValueProvider`. Now request.get() would have two different get()
-  Inside TemplateUtil's  [RequestValueProvider#get](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/a7c42302ad53d8361f541be02c56db598795ae41/bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/TemplateUtils.java#L77).
- Inside beam/CloudBigtableScanConfiguration's [RequestWithScanValueProvide#get](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/compare/master...rahulKQL:refactored_valueProvider#diff-5836fc24866e6a80ee2b663eb55592d4R446).

### DIFF
--- a/bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
+++ b/bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
@@ -15,6 +15,7 @@
  */
 package com.google.cloud.bigtable.beam;
 
+import com.google.cloud.bigtable.hbase.BigtableExtendedScan;
 import java.io.Serializable;
 import java.util.Map;
 import java.util.Objects;
@@ -34,7 +35,6 @@ import com.google.bigtable.repackaged.com.google.protobuf.ByteString;
 import com.google.cloud.bigtable.hbase.adapters.Adapters;
 import com.google.cloud.bigtable.hbase.adapters.read.DefaultReadHooks;
 import com.google.cloud.bigtable.hbase.adapters.read.ReadHooks;
-import org.apache.hadoop.hbase.filter.Filter;
 
 /**
  * This class defines configuration that a Cloud Bigtable client needs to connect to a user's Cloud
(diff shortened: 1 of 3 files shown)


bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableScanConfiguration.java
```diff
@@ -272,12 +273,11 @@ Builder withTableId(ValueProvider<String> tableId) {
...
     public CloudBigtableScanConfiguration build() {
       if (request == null) {
-        ReadHooks readHooks = new DefaultReadHooks();
-        if (scan == null) {
+        if(scan == null){
           scan = new Scan();
         }
-        ReadRowsRequest.Builder builder = Adapters.SCAN_ADAPTER.adapt(scan, readHooks);
-        request =  StaticValueProvider.of(readHooks.applyPreSendHook(builder.build()));
+        request = new RequestWithScanValueProvider(scan.getStartRow(), scan.getStopRow(),
```
sduskis: There's a `BIgtableExtendedScan` that needs to be considered here.  This approach doesn't seem like it would work.

sduskis: I think that we will need to convert the `Scan` to a `Filter` and `RowSet`, even though we don't have all of the information that `Query` needs.

igorbernstein2:
Question, would it not be easier to just use a valueprovider for the entire configuration object?

igorbernstein2:
Does this help?
https://github.com/apache/beam/blob/master/sdks/java/io/hbase/src/main/java/org/apache/beam/sdk/io/hbase/SerializableScan.java

sduskis:
@igorbernstein2, we hade a problem at one point, but that was related to classpath issues.  Now that we use hbase-client-shaded, it ought to work.

rahulKQL:
Initial Implementation with `BigtableExtendedScan`, Considered `RowSet` of scan object in `CloudBigtableScanConfiguration` constructor, I would consider `filter` as well in the `RequestWithtTableNameValueProvider` but wanted it to be reviewed first.

rahulKQL:
@sduskis 
Whenever you have time, Please have a look and let me know about this approach. If this seems fine then I would add `Filter` to `RequestWithtTableNameValueProvider`.

Issue #2020, opened 2018-12-08
Bigtable scan of 130 million rows throws java.lang.IllegalStateException: Not started


srini6teen:
Trying to scan Bigtable which has 130 million rows using the HBase java client. After a few hours into the scan, it throws

```
INFO com.google.cloud.bigtable.grpc.async.AbstractRetryingOperation - Retrying failed call. Failure #1, got: Status{code=DEADLINE_EXCEEDED, description=Error while reading table '', cause=null} on channel 5.
```

There were a few such exceptions from which the retry recovered but eventually results in

```
java.lang.IllegalStateException: Not started
    at com.google.common.base.Preconditions.checkState(Preconditions.java:507)
    at io.grpc.internal.ClientCallImpl.request(ClientCallImpl.java:367)
    at io.grpc.PartialForwardingClientCall.request(PartialForwardingClientCall.java:34)
    at io.grpc.ForwardingClientCall.request(ForwardingClientCall.java:22)
    at io.grpc.ForwardingClientCall$SimpleForwardingClientCall.request(ForwardingClientCall.java:44)
    at io.grpc.PartialForwardingClientCall.request(PartialForwardingClientCall.java:34)
    at io.grpc.ForwardingClientCall.request(ForwardingClientCall.java:22)
    at io.grpc.ForwardingClientCall$SimpleForwardingClientCall.request(ForwardingClientCall.java:44)
    at io.grpc.PartialForwardingClientCall.request(PartialForwardingClientCall.java:34)
    at io.grpc.ForwardingClientCall.request(ForwardingClientCall.java:22)
    at com.google.cloud.bigtable.grpc.io.RefreshingOAuth2CredentialsInterceptor$1.request(RefreshingOAuth2CredentialsInterceptor.java:146)
    at io.grpc.PartialForwardingClientCall.request(PartialForwardingClientCall.java:34)
    at io.grpc.ForwardingClientCall.request(ForwardingClientCall.java:22)
    at io.grpc.ForwardingClientCall$SimpleForwardingClientCall.request(ForwardingClientCall.java:44)
    at com.google.cloud.bigtable.grpc.io.Watchdog$WatchedCall.request(Watchdog.java:198)
    at io.grpc.PartialForwardingClientCall.request(PartialForwardingClientCall.java:34)
    at io.grpc.ForwardingClientCall.request(ForwardingClientCall.java:22)
[...]

PR #2026, opened 2018-12-13
Moving `AbstractRetryingOperation.call` to a `CallController`

sduskis:
Moving `RetryingReadRowsOperation.CallToStreamObserverAdapter` to a new class called `CallController` for better control of the call variable.  Moved `AbstractRetryingOperation.call` to `CallController` as well.

Next steps: protect `call` with the appropriate locking to prevent issues like #2020.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/CallWrapper.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/CallWrapper.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2018 Google LLC. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/CallWrapper.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Override public void sendMessage(Object message) {
+    }
+  };
+
+  private boolean autoFlowControlEnabled = true;
+  private ClientCall<RequestT, ResponseT> call = NULL_CALL;
+
+  synchronized void resetCall() {
+    call = NULL_CALL;
```
igorbernstein2: If I understand this correctly. This is intended to be called when we get a transient error. If thats the case, we need to retain the current request count, otherwise the stream will stall:

- something calls request(1)
- transient error occurs
- this gets called, and the request for the message is lost
- no message gets pulled to the listener

sduskis: Those are all good points.  Again, this PR is just a refactor so that we can better focus on issues like the ones you're describing.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/CallWrapper.java
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  @Override
+  public void onNext(RequestT value) {
+    call.sendMessage(value);
+  }
+
+  @Override
+  public void onError(Throwable t) {
+    call.cancel("Cancelled by client with StreamObserver.onError()", t);
```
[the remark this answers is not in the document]
sduskis: This is called.  I'll document the usages.

PR #2034, opened 2018-12-19
Adding implementation for readRows

rahulKQL:
Added new operation `IBigtableDataClient#readRows(ReadRowsRequest)` which currently return `ResultScanner<Row>`, with intend to delegate `BigtableDataClient#readFlatRow` for vaneer client.

Reason for raising this PR:
- `FlatRow` and `v2.models.Row` has identical fields.
- Vanner client has all read operation with `v2.models.Row`.

**Assumption:** `RowCell` has same ordering as `FlatRow.Cell`.


igorbernstein2:
I think it would be better to keep the FlatRow abstraction in cloud-bigtable-client and instead pas a custom [RowAdapter](https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/RowAdapter.java#L26) to [readRowsCallable](https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java#L598).

This would keep the existing api stable w/o any performance loss

igorbernstein2:
See here for an example implementation:
https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/DefaultRowAdapter.java

rahulKQL:
Thanks @igorbernstein2 for review. I have added `FlatRowModelAdapter.java` which we could use for `BigtableDataClient#readRowsCallable` later when BigtableDataClient vaneer client is ready.

When get a chance please have a look!!

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowModelAdapter.java
```diff
@@ -0,0 +1,103 @@
...
+
+import com.google.cloud.bigtable.data.v2.models.RowAdapter;
+import com.google.common.collect.ImmutableList;
+import com.google.protobuf.ByteString;
+import java.util.List;
+
+/**
+ * Adapter for {@link RowAdapter} that uses {@link FlatRow}'s to represent logical rows.
+ */
+public class FlatRowModelAdapter implements RowAdapter<FlatRow> {
```
igorbernstein2: s/FlatRowModelAdapter/FlatRowAdapter

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowModelAdapter.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @Override
+    public void finishCell() {
+      this.cells.add(
+          new FlatRow.Cell(this.family, this.qualifier, this.timestamp, this.value, this.labels));
+    }
+
+    @Override
+    public FlatRow finishRow() {
+      return new FlatRow(this.currentKey, this.cells.build());
```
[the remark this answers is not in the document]
rahulKQL: Hi @igorbernstein2,

I have added changes for family name sorting. Please have a look, whenever you have time.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowAdapter.java
```diff
@@ -0,0 +1,150 @@
...
+        combined.addAll(familyCellList);
+      }
+
+      return new FlatRow(this.currentKey, combined.build());
+    }
+
+    @Override
+    public void reset() {
+      this.currentKey = null;
+      this.cells = new TreeMap<>();
```
igorbernstein2: Please keep this.cell near the related fields (previousFamily & cell count) below

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowAdapter.java
```diff
@@ -0,0 +1,150 @@
...
+
+    @Override
+    public FlatRow createScanMarkerRow(ByteString rowKey) {
+      return new FlatRow(rowKey, ImmutableList.<Cell>of());
+    }
+  }
+
+  @Override
+  public boolean isScanMarkerRow(FlatRow row) {
+    return row.getCells() == null || row.getCells().isEmpty();
```
igorbernstein2: getCells() can never be null

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowAdapter.java
```diff
@@ -0,0 +1,150 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.grpc.scanner;
+
+import com.google.cloud.bigtable.data.v2.models.RowAdapter;
+import com.google.cloud.bigtable.grpc.scanner.FlatRow.Cell;
+import com.google.common.base.Objects;
```
igorbernstein2: Please use the standard java Objects:
https://docs.oracle.com/javase/7/docs/api/java/util/Objects.html

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowAdapter.java
```diff
@@ -0,0 +1,150 @@
...
+
+  @Override
+  public RowBuilder<FlatRow> createRowBuilder() {
+    return new FlatRowAdapter.FlatRowBuilder();
+  }
+
+  public class FlatRowBuilder implements RowBuilder<FlatRow> {
+    private ByteString currentKey;
+    private Map<String, List<Cell>> cells = new TreeMap<>();
+    private List<Cell> familyCells = null;
```
igorbernstein2: Please group these with the related fields previousFamily & totalCellCount. Also please add a comment to help the reader disambiguate between them

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowAdapter.java
igorbernstein2: Also, shouldn't familyCells be reset?

bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/scanner/TestFlatRowAdapter.java
```diff
@@ -0,0 +1,161 @@
...
+import com.google.common.collect.ImmutableList;
+
+import com.google.protobuf.ByteString;
+
+import java.util.List;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class TestFlatRowAdapter {
```
igorbernstein2: Please add `@RunWith` annotation

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowAdapter.java
rahulKQL: Thanks, Have moved the `cells` map & reset the `currentFamilyCells` list as well,

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowAdapter.java
rahulKQL: Added [this](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/2034/files#diff-e2e90efec4bfb8d9914f43602303a3dfR47) comment for the map, Changed `familyCells` to `currentFamilyCells` to make it more descriptive.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowAdapter.java
rahulKQL: `FlatRow` can be created without `List<FlatRow.Cell>`, there is no null check in its [constructor](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRow.java#L195). 

Would this class be only used by internal APIs? shall I remove this null check & add `@InternalApi` annotation?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowAdapter.java
```diff
@@ -0,0 +1,159 @@
...
+
+  public class FlatRowBuilder implements RowBuilder<FlatRow> {
+    private ByteString currentKey;
+    private String family;
+    private ByteString qualifier;
+    private List<String> labels;
+    private long timestamp;
+    private ByteString value;
+
+    //Map to hold family & cells record in lexicographical order.
```
igorbernstein2: Please update comment to explain that cells for the family are buffered in current family cells, list and that cells holds lists for all families.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowAdapter.java
igorbernstein2: FlatRows should only be instantiated here or the old row merger. We haven’t been good about marking apis as internal, but I think the ctor for flatrow should be internal

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowAdapter.java
rahulKQL: Updated comment for the same.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowAdapter.java
rahulKQL: Removed the null check for `row.getCells()` as it is would never be null here.

PR #2035, opened 2018-12-27
IBigtableDataClient remaining method Implementations

rahulKQL:
## What this PR contains

I am considering two steps for migrating to ApiFuture:
  **1st step** would be to adapt to GCJ `v2.models` for all operation than in 
   **2nd step** we can migrate to `ApiFuture`,

In this PR we have adapted to `ListenableFuture` in `IBigtableDataClient`.

 - Added implementation for these methods:
  1. mutateRowAsync
  2. readModifyWriteRow
  3. readModifyWriteRowAsync,
  4. checkAndMutateRow
  5. checkAndMutateRowAsync
- Added utility from `CheckAndMutationUtil#wasMutationApplied` in `BigtableDataClientWrappperImpl`, as `CheckAndMutationUtil` is in child module.
- Added `convert()` for conversion between `com.google.bigtable.v2.Row` to `v2.Models.Row`.


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataClientWrapper.java
```diff
@@ -46,39 +60,112 @@ public BigtableDataClientWrapper(BigtableDataClient bigtableDataClient,
...
 
+  /** {@inheritDoc} */
   @Override
-  public ApiFuture<Void> mutateRowAsync(RowMutation rowMutation) {
-    throw new UnsupportedOperationException("Not implemented yet");
+  public ListenableFuture<Void> mutateRowAsync(RowMutation rowMutation) {
+    ListenableFuture<MutateRowResponse> response =
+        delegate.mutateRowAsync(rowMutation.toProto(requestContext));
+    return  Futures.transform(response, new Function<MutateRowResponse, Void>() {
+      @NullableDecl
```
igorbernstein2: What's NullableDecl?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataClientWrapper.java
```diff
@@ -46,39 +60,112 @@ public BigtableDataClientWrapper(BigtableDataClient bigtableDataClient,
...
+            && !response.getPredicateMatched());
+  }
+
+  /**
+   * This method converts instances of {@link com.google.bigtable.v2.Row} to {@link Row}.
+   *
+   * @param row an instance of {@link com.google.bigtable.v2.Row} type.
+   * @return {@link Row} an instance of {@link Row}.
+   */
+  private static Row convert(com.google.bigtable.v2.Row row) {
```
igorbernstein2: Please name this method a bit more specific

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataClientWrapper.java
```diff
@@ -46,39 +60,112 @@ public BigtableDataClientWrapper(BigtableDataClient bigtableDataClient,
...
+   */
+  private static boolean wasMutationApplied(CheckAndMutateRowRequest request,
+      CheckAndMutateRowResponse response) {
+
+    // If we have true mods, we want the predicate to match.
+    // If we have false mods, we do not want the predicate to match.
+    return (request.getTrueMutationsCount() > 0
+        && response.getPredicateMatched()) || (
+        request.getFalseMutationsCount() > 0
+            && !response.getPredicateMatched());
```
igorbernstein2: This looks odd. Why do we need to transform it? It becomes ambiguous when a conditional mutation has both true & false mutations

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataClientWrapper.java
rahulKQL: According to javadocs
>Identical to {@code @Nullable}, but can only be written at declaration locations.

It was an autogenerated annotation for `ApiFunction`, would change this to `@Nullable` for consistency across the repository.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataClientWrapper.java
rahulKQL: If predicate matches for true mutations then we would not look for false mutation, as current implementation for [grpc.BigtableDataClient#checkAndMutateRow](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/1d1636f898cbde475c279d2673bc7efa4c503309/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataClient.java#L88) is following this on [hbase1.x](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/36216cc9f2c5f9a71e361501c4d32b2c7f078c14/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java#L459) & [hbase2.x](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/82b9bb6db4e861161f164cd5e514ec2d8f9e96c8/bigtable-hbase-2.x-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/BigtableAsyncTable.java#L238).

I have just looked up GAX impl for the [checkAndMutate](https://github.com/googleapis/google-cloud-java/blob/bed08ffd27d1d232f229df20bc9f01da322131a8/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/CheckAndMutateRowCallable.java#L42) and now considering predicate match stautus.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataClientWrapper.java
rahulKQL: update method name to `transformResponse()`

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataClientWrapper.java
sduskis: @igorbernstein2, All we care about is whether or not there was a mutation applied.  If there was a true mutation that was applied, then we're done; there's no need to check false mutations at that point.  Is there an objection to this?

PR #2044, opened 2019-01-08
Fixing compilation problems in master

sduskis:
The problems were caused by #2043


rahulKQL:
Thank you for the correction, I was working on the [same.](https://github.com/rahulKQL/cloud-bigtable-client/commit/8f8c2aec73dbc0c146b2bf86f7c9731025d18eab).

sduskis:
I merged multiple PRs at once into master, which may have caused the problem.  Luckily, the fix was easy.

PR #2046, opened 2019-01-14
Shading now uses include instead of exclude

sduskis:
`<exclude>`s make it very difficult to ensure that jars added to the shaded uber jars are all relocated.  The move from `<exclude>` to `<incclude>` will cause some friction in the future, but hopefully this will reduce friction caused by the lack of relocation.

There were a couple of un-relocated includes, which prompted this change.

### DIFF
--- a/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-shaded/README.md
+++ b/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-shaded/README.md
@@ -1,0 +1,21 @@
+This artifact is used in situations where dependencies don't converge properly.  Hadoop is one of
+those environments.  Hadoop uses really old versions of netty, protobuf and guava.  The simplest
+way to deal with that problem is to have a shaded jar.
+
+** NOTE TO DEVELOPERS:
+Shading can complicate dependency upgrades.  If you upgrade something, and your testing find that
+there's a `ClassNotFound`, or something similar, you will have update the pom.xml
+of this project in the `org.apache.maven.plugins:maven-shade-plugin` section:
+
+1. You'll have to add an `<include>` for your dependency.  One way to  find the right package is
+by runing `mvn clean package` on this project, and looking through the excluded libraries.
+
+2. Add a `<relocation>` for the new inclusion
+
+3. Confirm that the new inclusion is indeed relocated, and that no other classes from other have
+been added to the shaded jar.  You can do that by running the following command:
+
+> jar -tf target/bigtable-hbase-*-shaded-*-SNAPSHOT.jar | grep class | grep -v repackaged |  grep -v hbase
+
+That should only return classes that start with `com/google/cloud/bigtable/metrics/`.  Any other
+classes have to be relocated
--- a/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-shaded/pom.xml
+++ b/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-shaded/pom.xml
@@ -149,24 +149,60 @@ limitations under the License.
                 <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer" />
               </transformers>
               <artifactSet>
-                <excludes>
-                  <!-- exclude user visible deps -->
-                  <exclude>io.dropwizard.metrics:metrics-core</exclude>
-                  <exclude>commons-logging:commons-logging</exclude>
-                  <exclude>com.google.code.findbugs:jsr305</exclude>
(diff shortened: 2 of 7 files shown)


bigtable-hbase-1.x-parent/bigtable-hbase-1.x-shaded/README.md
```diff
@@ -0,0 +1,21 @@
+This artifact is used in situations where dependencies don't converge properly.  Hadoop is one of
+those environments.  Hadoop uses really old versions of netty, protobuf and guava.  The simplest
```
igorbernstein2: This isn't quite right. It implies that users should use the -shaded variant for hadoop, when they should be using the -hadoop variant. I think the description should be updated describe a general legacy application that uses both old versions of protobuf. And in a last sentence mention that hadoop is a special case that we handle via the -hadoop artifact

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-shaded/pom.xml
```diff
@@ -149,24 +149,60 @@ limitations under the License.
...
+                  <include>io.opencensus:opencensus-contrib-grpc-metrics</include>
+                  <include>io.opencensus:opencensus-contrib-grpc-util</include>
+                  <include>io.opencensus:opencensus-contrib-http-util</include>
+                  <include>io.opencensus:opencensus-contrib-monitored-resource-util</include>
+                  <include>io.opencensus:opencensus-contrib-zpages</include>
+                  <include>io.opencensus:opencensus-exporter-trace-stackdriver</include>
+                  <include>io.opencensus:opencensus-impl-core</include>
+                  <include>io.opencensus:opencensus-impl</include>
+                  <include>org.checkerframework:checker-compat-qual</include>
+                  <include>org.threeten:threetenbp</include>
```
igorbernstein2: would it make sense to use wildcards? 
```
<include>io.grpc:*</include>
<include>io.opencensus:*</include>
<include>io.opencensus:*</include>
<include>com.google.api:*</include>
<include>com.google.http-client:*</include>
// etc
```

bigtable-hbase-2.x-parent/bigtable-hbase-2.x-shaded/README.md
```diff
@@ -0,0 +1,21 @@
+This artifact is used in situations where dependencies don't converge properly.  Hadoop is one of
+those environments.  Hadoop uses really old versions of netty, protobuf and guava.  The simplest
+way to deal with that problem is to have a shaded jar.
```
igorbernstein2: see comment in the 1.x-shaded readme

igorbernstein2:
please confirm that the effective changes are intended:

Previously the following were being included, bu after the changes they are excluded:
[INFO] Excluding com.google.api.grpc:proto-google-cloud-trace-v1:jar:0.17.0 from the shaded jar.
[INFO] Excluding com.google.errorprone:error_prone_annotations:jar:2.2.0 from the shaded jar.
[INFO] Excluding com.google.j2objc:j2objc-annotations:jar:1.1 from the shaded jar.
[INFO] Excluding com.google.oauth-client:google-oauth-client:jar:1.27.0 from the shaded jar.
[INFO] Excluding javax.annotation:javax.annotation-api:jar:1.2 from the shaded jar.
[INFO] Excluding joda-time:joda-time:jar:2.9.2 from the shaded jar.
[INFO] Excluding org.apache.httpcomponents:httpclient:jar:4.5.5 from the shaded jar.
[INFO] Excluding org.apache.httpcomponents:httpcore:jar:4.4.9 from the shaded jar.
[INFO] Excluding org.codehaus.mojo:animal-sniffer-annotations:jar:1.17 from the shaded jar.

igorbernstein2:
If new dependencies are being exposed, then they must be manually promoted:
https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-hadoop/pom.xml#L58

sduskis:
@igorbernstein2, I should probably leave a note that we don't want to manually promote those dependencies.

igorbernstein2:
I'm not sure what you mean. Won't things fail if for example proto-google-cloud-trace-v1 isn't shaded nor added as a dependency?

sduskis:
@igorbernstein2, I don't think so.  OpenCensus should use trace v2, but I'll double check that.

sduskis:
@igorbernstein2, it looks like there are some declared, but unused dependencies in `google-cloud-java`.  I'll submit a separate PR just for the package renaming, and then com back to the include/exclude PR later once I fix the gcj deps

PR #2048, opened 2019-01-17
Repackaging some shaded libraries

sduskis:


bigtable-hbase-1.x-parent/bigtable-hbase-1.x-shaded/pom.xml
```diff
@@ -246,6 +247,18 @@ limitations under the License.
                   <pattern>org.apache.commons.codec</pattern>
                   <shadedPattern>com.google.bigtable.repackaged.org.apache.commons.codec</shadedPattern>
                 </relocation>
+                <relocation>
+                  <pattern>org.threeten</pattern>
+                  <shadedPattern>com.google.bigtable.repackaged.org.threeten</shadedPattern>
+                </relocation>
+                <relocation>
+                  <pattern>org.checkerframework</pattern>
+                  <shadedPattern>com.google.bigtable.repackaged.org.checkerframework</shadedPattern>
```
igorbernstein2: I'm a little concerned about shading the checkerframework. I dont really know how it works, but I'm guessing it uses annotations with some kind of tool that does static analysis. If my understanding is correct then shading it would break the static analysis. 
Is this dependency coming from gcj? I'm wondering how it's used.

igorbernstein2: I think exclude it.

sduskis: grpc added the `checkerframework` dependency.  I'll ask the diamond dependency group about this.

sduskis: Hmm... Guava has it as well.  I'll exclude it from shading, and manually promote the dependency.

igorbernstein2: For future: I wonder if it can be demoted to the provided scope? If either of those projects only use it the annotations, then we should send PRs to both to change the scope and avoid dealing with it in the client

sduskis:
I found another problem... I looks like we're pulling in `io.grpc` 0.18.0, even though `google-cloud-clients` uses 0.17.1.  That doesn't sound great.  We need to grack the grpc version from gcj

PR #2056, opened 2019-01-23
WIP: Added ModelRowAdapter for models.Row to Result

rahulKQL:
## What Change this PR contains
- Adding `ModelRowAdapter`: To convert `models.Row` to `hbase.client.Result`.
- Added static reference of this in `Adapters.java`.

**Note:** _This Adapter has a similar implementation as existing RowAdapter, Once all references of existing `RowAdapter` is migrated then we can remove `RowAdapter`._

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ModelRowAdapter.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ModelRowAdapter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Google LLC. All Rights Reserved.
+ * Copyright 2019 Google LLC.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ModelRowAdapter.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ModelRowAdapter.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2019 Google LLC. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.hbase.adapters.read;
+
+import com.google.cloud.bigtable.data.v2.models.Row;
+import com.google.cloud.bigtable.hbase.adapters.ResponseAdapter;
+import com.google.cloud.bigtable.hbase.util.ByteStringer;
+import com.google.cloud.bigtable.hbase.util.TimestampConverter;
+import java.util.SortedSet;
+import java.util.TreeSet;
+import org.apache.hadoop.hbase.Cell;
+import org.apache.hadoop.hbase.KeyValue;
+import org.apache.hadoop.hbase.client.Result;
+import org.apache.hadoop.hbase.util.Bytes;
+
+/**
+ * Adapt between a {@link Row} and an hbase client {@link Result}.
+ */
+public class ModelRowAdapter implements ResponseAdapter<Row, Result> {
+
+  @Override
+  public Result adaptResponse(Row response) {
+    if (response == null) {
+      return new Result();
+    }
+
+    SortedSet<Cell> hbaseCells = new TreeSet<>(KeyValue.COMPARATOR);
+    byte[] rowKey = ByteStringer.extract(response.getKey());
+    for (com.google.cloud.bigtable.data.v2.models.RowCell rowCell : response.getCells()) {
+      // Cells with labels are for internal use, do not return them.
+      // TODO(kevinsi4508): Filter out targeted {@link WhileMatchFilter} labels.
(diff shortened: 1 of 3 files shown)


rahulKQL:
**Reason for WIP:**  Would raise another PR to update existing [RowAdapter](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/RowAdapter.java) which will convert `models.Row` to `Result`.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ModelRowAdapter.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+public class ModelRowAdapter implements ResponseAdapter<Row, Result> {
+
+  @Override
+  public Result adaptResponse(Row response) {
+    if (response == null) {
+      return new Result();
+    }
+
+    SortedSet<Cell> hbaseCells = new TreeSet<>(KeyValue.COMPARATOR);
```
sduskis: This will slow things down.  We need to sort the `Row`'s cells more efficiently.  We should have an algorithm for that somewhere.

sduskis: If that's true, the we should remove the `SortedSet`/`TreeSet` and just use a `List`.

sduskis: We've had users who read rows that have ~1 million cells per row (I'm not sure why...), and this type of operation is costly at that scale.

rahulKQL: @igorbernstein2 Sorry for confusing PR titles, We implemented [FlatRowAdapter](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowAdapter.java#L31) to get FlatRow result using readRow callable. However this change is to create Adapter for convertion of gcj's models.Row to hbase's Result.


@sduskis Thanks for information & suggestion, I would update this to use `List`

rahulKQL:
I have updated `SortedSet` to `ImmutableList` to as records received from gcj's `readRows` would already be sorted.

@sduskis @igorbernstein2 Please have a look

rahulKQL:
> Is this model adapter only used for read rows? or will it be used for mutateRow, ?checkAndMutateRow, readAndModifyRow, etc?

yes, Our intent is to use this adapter for all operations involving models.Row (In case of FlatRow we would be using an existing [read.FlatRowAdapter](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/FlatRowAdapter.java)).

igorbernstein2:
This PR has surfaced an deficiency in google-cloud-java client. Not having the cells ordered by default creates a lot of confusion. For example, ReadModifyWrite returns a row that doesn't use a RowAdapter so it will not be sorted. To avoid guessing I think we should just move the sorting to google-cloud-java.

I put together PR here:
https://github.com/googleapis/google-cloud-java/pull/4421

rahulKQL:
Thanks for adding this sorting feature in GCJ!!

This would make cbt flow easy for converting `Row` into `hbase.Result` without  any concerns on sorting.
However, As of now we are wrapping existing native CBT implementation using GCJ's model, So until we be moves 100% to GCJ's adapter(i.e. #1973) we should keep this sorting in ModelRowAdapter.

Please let me know your thought on this.

rahulKQL:
@sduskis @igorbernstein2 

I know we had already discussed ModelRowAdapter in this PR, But I have put together another approach in #2061 to make existing `RowAdpater`, adapt to GCJ's `models.Row` [here](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/2061/files#diff-cdbae492d0f6e5586b7bd64061dc8ba6).

As I mentioned [above](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/2056#issuecomment-459819984), we would have to keep TreeSet in RowAdapter till we start using GCJ's adapter(as of now we are just using its model classes, not the actual client).

Please have a look and let me know of your thoughts.

sduskis:
Would it be ok to keep a `FlatRow` instead of using `model.Row` as the output of reads?

rahulKQL:
I think we should be ok to use `FlatRow`, as we have its separate adapter i.e. [FlatRowAdapter](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowAdapter.java).

[here](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/pull/2034#issuecomment-448615681) the conversation about the same in #2034.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ModelRowAdapter.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    ImmutableList.Builder<Cell> hbaseCellBuilder = ImmutableList.builder();
+    byte[] rowKey = ByteStringer.extract(response.getKey());
+    for (com.google.cloud.bigtable.data.v2.models.RowCell rowCell : response.getCells()) {
+      // Cells with labels are for internal use, do not return them.
+      // TODO(kevinsi4508): Filter out targeted {@link WhileMatchFilter} labels.
+      if (rowCell.getLabels().size() > 0) {
+        continue;
+      }
```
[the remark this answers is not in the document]
sduskis: This is unnecessary code.  Please remove the comment and the if statement.

rahulKQL:
Could we close this PR?
As we are updating existing `RowAdapter` to convert `models.Row` to `Result`, which previously this PR's `ModelRowAdapter` meant to do.

PR #2057, opened 2019-01-24
adding Retry#getStatusCode & unit test

rahulKQL:
Adding Retry#getStatusCode to be used in BigtableDataSettingFactory for translation of any specific RetryCodes to BigtableDataSettings.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/RetryOptions.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/RetryOptions.java
@@ -23,7 +23,6 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableSet;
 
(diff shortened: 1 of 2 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/RetryOptions.java
```diff
@@ -337,7 +336,16 @@ public int getMaxScanTimeoutRetries() {
...
     return statusToRetryOn.contains(code);
   }
-  
+
+  /**
+   * Should returns {@link Set} of available {@link Status.Code}.
+   *
+   * @return a {@link Set}.
+   */
+  public Set<Status.Code> getStatusCodes() {
```
igorbernstein2: this should be a bit more descriptive:

getRetryableStatusCodes()

Also, the description is misleading, these aren't available status codes

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/RetryOptions.java
```diff
@@ -337,7 +336,16 @@ public int getMaxScanTimeoutRetries() {
   public boolean isRetryable(Status.Code code) {
     return statusToRetryOn.contains(code);
   }
-  
+
+  /**
+   * Should returns {@link Set} of {@link Status.Code} based on which RPC retries.
```
igorbernstein2: I think this would be better: `Returns {@link Set} of {@link Status.Code} that allow RPC retries.`

rahulKQL: Thanks for the input. Updated Javadoc with the same.

PR #2063, opened 2019-02-06
Switching many requests from BigtableDataClient to the wrapper

sduskis:


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableRegionLocator.java
```diff
@@ -54,12 +54,13 @@
 
   protected final TableName tableName;
   private ListenableFuture<List<HRegionLocation>> regionsFuture;
-  private final BigtableDataClient client;
+  private final BigtableDataClientWrapper client;
```
rahulKQL: Should we use `IBigtableDataClient` here(and everywhere else) instead of using the `BigtableDataClientWrapper`?

sduskis: @rahulKQL, good catch.  I replaced `BigtableDataClientWrapper` where ever I could with `IBigtableDataClient`.

sduskis:
The Stackdriver version of guava is 20, which doesn't have a method relied on by gcb.

PR #2066, opened 2019-02-06
Using the clientWrapper for checkAnd* in AbstractBigtableTable

sduskis:


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/CheckAndMutateUtil.java
```diff
@@ -62,20 +63,20 @@ public void applyPreSendHook(Query query) {
...
    * @return a boolean.
    */
   public static boolean wasMutationApplied(
-      CheckAndMutateRowRequest request,
-      CheckAndMutateRowResponse response) {
+      ConditionalRowMutation request,
+      Boolean predicateMatched) {
 
+    CheckAndMutateRowRequest proto =
+        request.toProto(RequestContext.create("SomeProject", "Some Instance", ""));
```
igorbernstein2: Should we add hasTrueMutations() and hasFalseMutations() to the veneer model?
Or maybe checkAndMutateRow should return an enum with 3 possible outcomes:  true mutations applied, false mutations applied, nothing happened?

sduskis: I think that new `has()` methods in the model would be ideal.

PR #2068, opened 2019-02-06
Removing deprecated uses of gcp InstanceName

sduskis:
Also, adding a `BigtableOptions.getRequestContext()` for convenience

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
@@ -301,6 +301,8 @@ public class BigtableOptions implements Serializable, Cloneable {
   private boolean useCachedDataPool;
 
   private BigtableInstanceName instanceName;
+
+  // TODO: remove the `transient` modifier once RequestContext is Serializable.
   private transient RequestContext requestContext;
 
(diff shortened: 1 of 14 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
```diff
@@ -405,12 +409,27 @@ public int getChannelCount() {
...
   public BigtableInstanceName getInstanceName() {
     return instanceName;
   }
 
+  /**
+   * <p>Getter for the field <code>instanceName</code>.</p>
+   *
+   * @return a {@link RequestContext} object.
+   */
+  public RequestContext getRequestContext() {
```
igorbernstein2: This context can only be used for the data client. Maybe rename as such. Also, the docs are a bit off

igorbernstein2: Also consider keeping the RequestContext in client implementation (maybe BigtableSession?) , since its not really a configureable option

sduskis: I moved `RequestContext` to `BigtableSession` and renamed.

PR #2080, opened 2019-02-07
Removing WhileMatchFilter logic from RowAdapter.

sduskis:
See #2056 for context

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/FlatRowAdapter.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/FlatRowAdapter.java
@@ -57,7 +57,7 @@ public class FlatRowAdapter implements ResponseAdapter<FlatRow, Result> {
     String previousFamily = null;
     for (FlatRow.Cell cell : cells) {
       // Cells with labels are for internal use, do not return them.
-      // TODO(kevinsi4508): Filter out targeted {@link WhileMatchFilter} labels.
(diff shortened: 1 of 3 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/FlatRowAdapter.java
```diff
@@ -56,6 +56,8 @@ public Result adaptResponse(FlatRow flatRow) {
     byte[] previousFamilyBytes = null;
     String previousFamily = null;
     for (FlatRow.Cell cell : cells) {
+      // Cells with labels are for internal use, do not return them.
+      // TODO(kevinsi4508): Filter out targeted {@link WhileMatchFilter} labels.
```
igorbernstein2: Is this the right place for this? It seems like a filter specific thing, but FlatRowAdapter seems to be generic. Would a ReadHook be a better place?

sduskis: `ReadHook` updates the `ReadRowsRequest`, not the results.  We would need to figure out some other mechanism.

This PR at least brings attention to the problem, so if it's ok, I'd like to proceed with merging this code, and figure out a better alternative later.

igorbernstein2: Please add a TODO

PR #2088, opened 2019-02-07
Upgrading to google-cloud-clients 0.80.0-alpha

sduskis:
This doesn't change other critical dependencies, but does bring in a useful tool for the conversion to using google-cloud-bigtable models.


sduskis:
@igorbernstein2, FYI, I'm merging this.

PR #2095, opened 2019-02-12
hbase2 TableAdminClient update for listTables

rahulKQL:
Updated `listTablesAsync` which previously returned List of protobuf Table 
to IBigtableTableAdminClient#listTablesAsync.
It now returns List of `tableId` then we fetch `models.Table` for matched tableNames.


bigtable-hbase-2.x-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/BigtableAsyncAdmin.java
```diff
@@ -223,14 +222,25 @@ public BigtableAsyncAdmin(CommonConnection asyncConnection) throws IOException {
...
+    CompletableFuture<List<CompletableFuture<Table>>> futureTables =
+        bigtableTableAdminClient.listTablesAsync().thenApply(r ->
+            r.stream()
+                .filter(t -> !tableNamePattern.isPresent() ||
+                    tableNamePattern.get().matcher(t).matches())
+            .map(q -> bigtableTableAdminClient.getTableAsync(q)
+                .exceptionally(ex -> {
+                  LOG.error("Error while fetching Table", ex);
+                  return null;
+                }))
```
rahulKQL: Reason for adding `exceptionally`:

Test cases were sharing tables and in that process, some of the tables were getting disabled parallelly before it getTableAsync calls finishes, To handle that we have added exceptionally.

If this doesn't look right then Another solution would be to add an operation to fetch List of `models.Table` in one synchronized call(which is currently not present in GCJ's `BigtableTableAdminClient`). 
Do let me know If you think I should add that.

sduskis: This masks a legitimate error.  We can't do this.  We need to fix what ever the underlying problem is.

ajaaym: I think best way to fix this is to add method in GCJ to return list of table proto in addition to name. Currently GCJs list ignores all information and just returns list of string.

sduskis: I played around with this.  There are weird states that this gets into.  `FAILED_PRECONDITION"`if you try to `getTable` that's being created, or `NOT_FOUND` if `listTables` happened before a delete.  This sounds like a tricky issue that @igorbernstein2 and I should work on.

For this PR, I would suggest removing the new functionality to actually get descriptors, and just use the wrapper.  We should also add a TODO to get the full descriptor.

sduskis: @ajaaym, the service only returns a name for `ListTables`, so GCJ does the right thing.

rahulKQL: @sduskis 
I have updated listTables, It now returns List of TableDescriptors(containing only table-name). 
Also added TODO to get full descriptors.

Please have a look.

PR #2099, opened 2019-02-14
Remove HBase 2 BigtableDataClient

sduskis:


rahulKQL:
One side Ques: 
Should we also follow same pattern for [hbase2_x.BigtableTableAdminClient](https://github.com/rahulKQL/cloud-bigtable-client/blob/master/bigtable-hbase-2.x-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/BigtableTableAdminClient.java).

PR #2114, opened 2019-03-01
Migrated admin operation to GCJ BigtableTableAdminClient

rahulKQL:
Replace all occurence of `IBigtableTableAdminClient` with `BigtableTableAdminClient`. 
Uupdated all of snapshotOperation to utilize `BaseBigtableTableAdminClient`.

**Note:** Please have a look at `BaseBigtableTableAdminClient` Implementation for `hbase2_x.BigtableAdmin#L136` & `BigtableAsyncAdmin#listSnapshots `.


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -384,16 +396,30 @@ public synchronized BigtableTableAdminClient getTableAdminClient() throws IOExce
   }
 
   /**
-   * <p>Getter for the field <code>adminClientWrapper</code>.</p>
+   * <p>Getter for the field <code>adminClientV2</code>.</p>
```
sduskis: We actually don't need this method.  There needs to be a new method that looks at a new flag in `BigtableOptions` and either choose either the `wrapper` or `v2` implementation.

PR #2117, opened 2019-03-04
Adding x-goog-api-client

sduskis:


igorbernstein2:
Should these be set for the admin RPCs as well?

sduskis:
The headers are set for admin as well. See [BigtableSession.createChannelPool](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java#L420)

igorbernstein2:
Sorry, misread the code

PR #2118, opened 2019-03-06
Fix a couple of Throttler issues

igorbernstein2:
The common path to invoking the ThrottlingClientInterceptor:
call = channel.newCall(...)
call.sendMessage(request)
call.halfClose()

However if the resource limiter is interrupted while the interceptor is
waiting to send a message, the delegate call will not be set, causing an
NPE.

To fix this issue, the ForwardedClientCall was remove and all forwarded
calls check for null delegates. The impl for halfClose is skipped
entirely because only UnaryCalls are supported.

The other issue addressed here is an unhandled cancellation of the call
before sending a message.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/ThrottlingClientInterceptor.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/ThrottlingClientInterceptor.java
@@ -22,18 +22,18 @@ import io.grpc.CallOptions;
 import io.grpc.Channel;
 import io.grpc.ClientCall;
 import io.grpc.ClientInterceptor;
-import io.grpc.ForwardingClientCall;
 import io.grpc.ForwardingClientCallListener.SimpleForwardingClientCallListener;
 import io.grpc.Metadata;
 import io.grpc.MethodDescriptor;
 import io.grpc.Status;
+import java.util.concurrent.CancellationException;
+import javax.annotation.Nullable;
 
 /**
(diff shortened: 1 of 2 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/ThrottlingClientInterceptor.java
```diff
@@ -98,14 +129,14 @@ public void onClose(io.grpc.Status status, Metadata trailers) {
...
-        delegate().sendMessage(message);
+        delegateCall.start(markCompletionListener, headers);
+        delegateCall.request(numMessagesRequested);
+        delegateCall.sendMessage(message);
       }
 
       @Override
-      protected ClientCall<ReqT, RespT> delegate() {
-        return delegateCall;
+      public void halfClose() {
```
sduskis: Should we propgate the `halfClose()`, or is it irrelevant for our use case?

sduskis:
The tests fail: 
>  TestAppProfile.testBulkMutation:197 NullPointer

igorbernstein2:
Fixed & added a test to make sure it doesn't break again

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/ThrottlingClientInterceptor.java
igorbernstein2: I just put the halfClose as part of the sendMessage since its a unary call

PR #2119, opened 2019-03-07
Removing the new_key > older_key comparison exception

sduskis:
... and replacing it with a warning.


igorbernstein2:
Can you describe the motivation for this?

PR #2120, opened 2019-03-07
minor log tweaks

igorbernstein2:


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -546,6 +547,9 @@ public ManagedChannel create() throws IOException {
   public static ManagedChannel createNettyChannel(String host,
       BigtableOptions options, ClientInterceptor ... interceptors) throws SSLException {
 
+    LOG.info("Creating new channel for %s", host);
+    LOG.debug(Throwables.getStackTraceAsString(new Throwable()));
```
sduskis: wait... this is expensive.  Please add an if statement around this.

igorbernstein2: Updated, PTAL

sduskis: we can also add a new method in the `Logger` to do this.

igorbernstein2: I really don't have strong feelings on this. What did you have in mind?

sduskis: First off, why do you want this?

sduskis: Can you please remove this line, and we'll revisit this in a separate PR?

PR #2122, opened 2019-03-07
Refactoring the timestamp logic.

sduskis:
Removing TimeUnit, since it has weird issues when a Put "timestamp" isn't actually a millisecond timestamp.  TimeUnit conversions use an `int` in their conversion, and weird overflow issues happen in that case.


bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/util/ByteStringComparatorTest.java
```diff
@@ -17,21 +17,26 @@
...
+    compare("aa", "aac");
   }
 
+  @Test
   public void testUnsigned() {
     compare(ByteString.copyFrom(new byte[] { 0x7f }),
       ByteString.copyFrom(new byte[] { (byte) 0x80 }));
   }
 
+  protected void compare(String a, String b) {
```
igorbernstein2: why protected?

bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/util/ByteStringComparatorTest.java
```diff
@@ -17,21 +17,26 @@
...
+    compare(ByteString.copyFromUtf8(a), ByteString.copyFromUtf8(b));
+  }
+
   protected void compare(ByteString a, ByteString b) {
-    Assert.assertEquals(-1, underTest.compare(a, b));
-    Assert.assertEquals(1, underTest.compare(b, a));
+    Assert.assertTrue(underTest.compare(a, b) < 0);
+    Assert.assertTrue(underTest.compare(b, a) > 0);
     Assert.assertEquals(0, underTest.compare(b, b));
```
igorbernstein2: The last assert is different from the first 2

bigtable-hbase-2.x-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/BigtableAsyncAdmin.java
```diff
@@ -350,7 +349,7 @@ public BigtableAsyncAdmin(CommonConnection asyncConnection) throws IOException {
    * {@inheritDoc} */
   @Override
   public CompletableFuture<Void> restoreSnapshot(String snapshotName) {
-    boolean takeFailSafeSnapshot = BigtableConstants.DEFAULT_SNAPSHOT_RESTORE_TAKE_FAILSAFE_SNAPSHOT;
+    boolean takeFailSafeSnapshot = false;
```
igorbernstein2: I think its safe to inline this var

PR #2125, opened 2019-03-15
Refactored readModifyWrite from BulkMutation

rahulKQL:
- Refactored `#readModifyWrite` from `BulkMutation` & `IBulkMutation`.
- Added `BulkMutation#register` to be able to add future tasks in the ongoing batch.
- Refactored `BigtableBufferedMutatorHelper` to add `IBigtableDataClient` to and redirected the `#readModifyWrite` call.
- Fixed unit test cases.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/core/IBulkMutation.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/core/IBulkMutation.java
@@ -28,12 +28,12 @@ import com.google.common.util.concurrent.ListenableFuture;
 public interface IBulkMutation {
 
   /**
-   * Send any outstanding {@link RowMutation} and wait until all requests are complete.
+   * Sends any outstanding {@link RowMutation} and wait until all requests are complete.
    */
   void flush() throws InterruptedException;
 
   /**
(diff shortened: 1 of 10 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/core/IBulkMutation.java
```diff
@@ -50,12 +52,10 @@
...
-   * {@link ReadModifyWriteRow}. This method may block if
+   * Adds a future task as RPC operation. This method may block if
    * {@link OperationAccountant#registerOperation(ListenableFuture)} blocks.
    *
-   * @param request The {@link ReadModifyWriteRow} to send.
-   * @return a {@link ApiFuture} which can be listened to for completion events.
+   * @param future a {@link ApiFuture} which would be listened for completion events.
    */
-  ApiFuture<Row> readModifyWrite(ReadModifyWriteRow request);
+  void register(final ApiFuture<?> future);
```
sduskis: Please remove `register` and use an `OperationAccountant` in `BigtableBufferedMutatorHelper`

rahulKQL: Thanks for the suggestion, Have implemented the same.

PR #2128, opened 2019-03-20
Addition of BigtableDataGCJClient with GCJ adapters

rahulKQL:
- Apart from [AbstractTestFilter#testInterleaveNoDuplicateCells](https://github.com/googleapis/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestFilters.java#L2011-L2033) all integration test case are passing for both H1 & H2. Currently marked that test case as ignored will update once resolved.
- Now returning QualifierFilter instance instead of ChainFilter in ColumnPrefixFilterAdapter.java. (Might have mistakenly returned ChainFilter).
- Added RowResultScanner to translate ServerStream into ResultScanner(will add test case).
- Have created StreamObserverAdapter similar to ApiStreamObserver from GCJ.

Test Cases can be invoked with:
```
mvn clean verify \
       -PbigtableIntegrationTest \
       -Dgoogle.bigtable.project.id=grass-clump-479 \
       -Dgoogle.bigtable.instance.id=node-client-performance \
       -Dgoogle.bigtable.use.gcj.client=true
```

Note: The GCJ returns duplicate values when using multiple filters with `interleave`, If the same set of filters used with Filters#chain none it works fine.

To demonstrate duplication from `Filter#interleave`, I have added one example in [`TestBigtableVeneerClientFactory`](https://github.com/googleapis/cloud-bigtable-client/compare/master...rahulKQL:gcj_data_client_impl#diff-a9e488e7dc2d8858b2c25e7a08c272a0R194).

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGCJClient.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGCJClient.java
@@ -17,6 +17,8 @@ package com.google.cloud.bigtable.grpc;
 
 import com.google.api.core.ApiFuture;
 import com.google.api.gax.rpc.ServerStream;
+import com.google.api.gax.rpc.StateCheckingResponseObserver;
+import com.google.api.gax.rpc.StreamController;
 import com.google.cloud.bigtable.core.IBigtableDataClient;
 import com.google.cloud.bigtable.core.IBulkMutation;
 import com.google.cloud.bigtable.data.v2.BigtableDataClient;
@@ -154,11 +156,35 @@ public class BigtableDataGCJClient implements IBigtableDataClient, AutoCloseable
   @Override
   public void readFlatRowsAsync(Query request, StreamObserver<FlatRow> observer) {
     //TODO: figure out to convert StreamObserver to ResponseObserver.
-    //delegate.readRowsCallable(new FlatRowAdapter()).call(request, observer);
+    delegate.readRowsCallable(new FlatRowAdapter()).call(request, new StreamObserverAdapter<>(observer));
   }
 
   @Override
   public void close() throws Exception {
     delegate.close();
+  }
+
+  // Inspired from ApiStreamObserverAdapter of GCJ
+  static class StreamObserverAdapter<T> extends StateCheckingResponseObserver<T> {
+    private final StreamObserver<T> delegate;
+
+    StreamObserverAdapter(StreamObserver<T> delegate) {
+      this.delegate = delegate;
+    }
+
+    protected void onStartImpl(StreamController controller) {
+    }
+
+    protected void onResponseImpl(T response) {
+      this.delegate.onNext(response);
+    }
+
+    protected void onErrorImpl(Throwable t) {
+      this.delegate.onError(t);
+    }
+
+    protected void onCompleteImpl() {
+      this.delegate.onCompleted();
+    }
   }
 }
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGCJClient.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGCJClient.java
@@ -32,10 +32,7 @@ import com.google.cloud.bigtable.grpc.scanner.FlatRow;
 import com.google.cloud.bigtable.grpc.scanner.FlatRowAdapter;
 import com.google.cloud.bigtable.grpc.scanner.ResultScanner;
 import com.google.cloud.bigtable.grpc.scanner.RowResultScanner;
-import com.google.common.collect.ImmutableList;
 import io.grpc.stub.StreamObserver;
-import java.io.IOException;
-import java.util.Iterator;
 import java.util.List;
 
 /**
@@ -118,36 +115,6 @@ public class BigtableDataGCJClient implements IBigtableDataClient, AutoCloseable
     return new RowResultScanner<>(stream, new FlatRow[0]);
   }
 
-  private ResultScanner<FlatRow> readFlatRowsV2(Query request){
-    List<FlatRow> flatRows = delegate.readRowsCallable(new FlatRowAdapter()).all().call(request);
-    final Iterator<FlatRow> flatRowIterator = flatRows.iterator();
-    return new ResultScanner<FlatRow>(){
-
-      @Override
-      public void close() throws IOException {
-      }
-
-      @Override
-      public FlatRow next() throws IOException {
-        return flatRowIterator.next();
-      }
-
-      @Override
-      public FlatRow[] next(int count) throws IOException {
-        ImmutableList.Builder<FlatRow> builder = ImmutableList.builder();
-        for (int i = 0; flatRowIterator.hasNext() && i < count; i++) {
-          builder.add(flatRowIterator.next());
-        }
-        return builder.build().toArray(new FlatRow[0]);
-      }
-
-      @Override
-      public int available() {
-        throw new UnsupportedOperationException("can not support available");
-      }
-    };
-  }
-
   @Override
   public ApiFuture<List<FlatRow>> readFlatRowsAsync(Query request) {
     return delegate.readRowsCallable(new FlatRowAdapter()).all().futureCall(request);
@@ -155,8 +122,8 @@ public class BigtableDataGCJClient implements IBigtableDataClient, AutoCloseable
 
   @Override
   public void readFlatRowsAsync(Query request, StreamObserver<FlatRow> observer) {
-    //TODO: figure out to convert StreamObserver to ResponseObserver.
-    delegate.readRowsCallable(new FlatRowAdapter()).call(request, new StreamObserverAdapter<>(observer));
+    delegate.readRowsCallable(new FlatRowAdapter()).call(request,
+        new StreamObserverAdapter<FlatRow>(observer));
   }
 
   @Override
@@ -164,8 +131,13 @@ public class BigtableDataGCJClient implements IBigtableDataClient, AutoCloseable
     delegate.close();
   }
 
-  // Inspired from ApiStreamObserverAdapter of GCJ
-  static class StreamObserverAdapter<T> extends StateCheckingResponseObserver<T> {
+  /**
+   * To wrap stream of native CBT client's {@link StreamObserver} onto GCJ
+   * {@link com.google.api.gax.rpc.ResponseObserver}.
+   *
+   * <p>Note:Inspired from {@link com.google.api.gax.rpc.ApiStreamObserverAdapter} of GCJ.
+   * */
+  private static class StreamObserverAdapter<T> extends StateCheckingResponseObserver<T> {
     private final StreamObserver<T> delegate;
 
     StreamObserverAdapter(StreamObserver<T> delegate) {
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGCJClient.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGCJClient.java
@@ -1,0 +1,164 @@
+/*
+ * Copyright 2019 Google LLC. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.grpc;
+
+import com.google.api.core.ApiFuture;
+import com.google.api.gax.rpc.ServerStream;
+import com.google.cloud.bigtable.core.IBigtableDataClient;
+import com.google.cloud.bigtable.core.IBulkMutation;
+import com.google.cloud.bigtable.data.v2.BigtableDataClient;
+import com.google.cloud.bigtable.data.v2.models.ConditionalRowMutation;
+import com.google.cloud.bigtable.data.v2.models.KeyOffset;
+import com.google.cloud.bigtable.data.v2.models.Query;
+import com.google.cloud.bigtable.data.v2.models.ReadModifyWriteRow;
+import com.google.cloud.bigtable.data.v2.models.Row;
+import com.google.cloud.bigtable.data.v2.models.RowMutation;
+import com.google.cloud.bigtable.grpc.scanner.FlatRow;
+import com.google.cloud.bigtable.grpc.scanner.FlatRowAdapter;
+import com.google.cloud.bigtable.grpc.scanner.ResultScanner;
+import com.google.cloud.bigtable.grpc.scanner.RowResultScanner;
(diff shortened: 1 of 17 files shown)


rahulKQL:
@sduskis,

For the [testInterleaveNoDuplicateCells](https://github.com/googleapis/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestFilters.java#L2011-L2033) issue I tried discarding Cells with labels from FlatRowAdapter but doing that makes [WhileMatchFilter allow every cell](https://github.com/googleapis/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/BigtableWhileMatchResultScannerAdapter.java#L116-L137). 
Then I tried [adding labels into cell comparison](https://github.com/googleapis/cloud-bigtable-client/pull/2128/files#diff-e2e90efec4bfb8d9914f43602303a3dfR171), that works as expected. 

As of now, I am not sure about performance overhead with comparing labels of each cell (_To keep performance in check I only checked for the difference in size of the labels as order does not matter to us_).

There is one more approach( to check each cell with labels with "-in" & "-out" and discard others. But that sounds more verbose.

Please have a review and let me know your thought on this.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGCJClient.java
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  @Override
+  public ResultScanner<FlatRow> readFlatRows(Query request) {
+    final ServerStream<FlatRow> stream =
+        delegate.readRowsCallable(new FlatRowAdapter()).call(request);
+    return new RowResultScanner<>(stream, new FlatRow[0]);
+  }
+
+  private ResultScanner<FlatRow> readFlatRowsV2(Query request){
```
sduskis: Is `readFlatRowsV2` required?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGCJClient.java
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  @Override
+  public ApiFuture<List<FlatRow>> readFlatRowsAsync(Query request) {
+    return delegate.readRowsCallable(new FlatRowAdapter()).all().futureCall(request);
+  }
+
+  @Override
+  public void readFlatRowsAsync(Query request, StreamObserver<FlatRow> observer) {
+    //TODO: figure out to convert StreamObserver to ResponseObserver.
```
sduskis: Is this `//TODO` still valid?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -250,6 +253,10 @@ public BigtableSession(BigtableOptions opts) throws IOException {
     dataClient =
         new BigtableDataGrpcClient(dataChannel, sharedPools.getRetryExecutor(), options);
     dataClient.setCallOptionsFactory(callOptionsFactory);
+    BigtableDataSettings dataSettings =
```
sduskis: Please delay creating the `dataSettings` and `dataGCJClient` until the client is actually needed.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -351,6 +358,9 @@ public RequestContext getDataRequestContext() {
    * @return a {@link IBigtableDataClient} object.
    */
   public IBigtableDataClient getClientWrapper() {
+    if(options.useGCJClient()){
+      return dataGCJClient;
+    }
```
sduskis: Please use `else` here.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowAdapter.java
```diff
@@ -102,10 +106,9 @@ public void cellValue(ByteString value) {
     public void finishCell() {
       if (!Objects.equals(this.family, this.previousFamily)) {
         previousFamily = this.family;
-        currentFamilyCells = new ArrayList<>();
+        currentFamilyCells = new TreeSet<>(compareCells());
```
sduskis: This should really be a List, and not a `TreeSet`

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowAdapter.java
```diff
@@ -121,8 +124,9 @@ public void finishCell() {
      */
     @Override
     public FlatRow finishRow() {
-      ImmutableList.Builder<FlatRow.Cell> combined = ImmutableList.builderWithExpectedSize(totalCellCount);
-      for (List<FlatRow.Cell> familyCellList : cells.values()) {
+      ImmutableList.Builder<FlatRow.Cell> combined =
+          ImmutableList.builderWithExpectedSize(totalCellCount);
+      for (Set<FlatRow.Cell> familyCellList : cells.values()) {
```
sduskis: Please compare the current and previous for the following:

```
FlatRow.Cell previous;
for(FlatRow.Cell c : cells.values) {
  if(previous == null || !c.getLables().isEmpty() || !keysMatch(c, previous)){
    combined.add(c);
  }
}
```

bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/config/TestBigtableVeneerSettingsFactory.java
```diff
@@ -133,8 +134,15 @@ private void initializeClients() throws IOException{
   }
 
   /**
-   * This test runs only if it finds "test.client.project.id" & "test.client.project.id"
-   * VM arguments. Then it calls to an actual Bigtable Table & performs the checks below:
+   * This test runs only if it founds below env variables,
```
sduskis: s/founds below env variables/finds the environment variables listed below`

bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/config/TestBigtableVeneerSettingsFactory.java
```diff
@@ -133,8 +134,15 @@ private void initializeClients() throws IOException{
...
+   * Then it calls to an actual Bigtable Table & performs the checks below:
```
sduskis: Please don't use `&`, since the comments are valid HTML.  Please use the word `and`

bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/config/TestBigtableVeneerSettingsFactory.java
```diff
@@ -183,20 +191,35 @@ public void testWithActualTables() throws Exception{
       dataClient.mutateRow(rowMutation);
       LOG.info("Successfully Mutated");
 
-      Query query = Query.create(TABLE_ID);
+      Query query = Query.create(TABLE_ID)
```
sduskis: This seems to be part of a separate PR.

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestFilters.java
```diff
@@ -2009,6 +2009,7 @@ private static void assertNextEquals(ResultScanner scanner, byte expectedKey[])
   }
 
   @Test
+  //TODO(rahulkql): failing due to duplicate Cells when used with interleave
```
sduskis: Please remove this comment.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/ColumnPrefixFilterAdapter.java
```diff
@@ -45,9 +45,8 @@ public Filter adapt(FilterAdapterContext context, ColumnPrefixFilter filter)
     ByteString.Output outputStream = ByteString.newOutput(prefix.length * 2);
     ReaderExpressionHelper.writeQuotedExpression(outputStream, prefix);
     outputStream.write(ReaderExpressionHelper.ALL_QUALIFIERS_BYTES);
-    Filter qualifierFilter = FILTERS.qualifier().regex(outputStream.toByteString());
 
-    return FILTERS.chain().filter(qualifierFilter);
+    return FILTERS.qualifier().regex(outputStream.toByteString());
```
sduskis: Please make this change in a different PR.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestFilters.java
```diff
@@ -37,7 +37,7 @@
 import static com.google.cloud.bigtable.hbase.test_env.SharedTestEnvRule.COLUMN_FAMILY;
 
 public class TestFilters extends AbstractTestFilters {
-
```
sduskis: Please make this change in a different PR.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -265,10 +268,18 @@ public BigtableSession(BigtableOptions opts) throws IOException {
 
     BigtableClientMetrics.counter(MetricLevel.Info, "sessions.active").inc();
 
-    // Defer the creation of both the tableAdminClient until we need them.
-    //TODO(rahulkql): Need to identify which resources to initialized for GCJ's adapter case.
-    this.adminSettings = BigtableVeneerSettingsFactory.createTableAdminSettings(options);
-    this.baseAdminSettings = BaseBigtableTableAdminSettings.create(adminSettings.getStubSettings());
+
+    //TODO(rahulkql): Need to identify which resources to initialized for GCJ's adapter case
+    if(options.useGCJClient()){
```
rahulKQL: @sduskis 
Have included `dataSettings` & `dataClient` creation in xor of `BigtableSession` only in case of `options.useGCJClient.` 
As moving it to getClientWrapper would introduce a chain of updates to handle `IOException`.

If needed, I could create a separate PR on which we can move these into `BigtableSession#getClientWrapper` & `BigtableSession#getAdminClientWrapper`.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowAdapter.java
```diff
@@ -122,13 +122,23 @@ public void finishCell() {
...
+          }
+          previous = c;
+        }
       }
 
       return new FlatRow(this.currentKey, combined.build());
     }
 
+    private boolean keysMatch(Cell c, Cell previous) {
+      return c.equals(previous);
```
rahulKQL: @sduskis 
I had a doubt here, Using Cell#equals would be ok? Or shall I compare specific field i.e. labels only?

sduskis: Please only use the qualifier and timestamp, since the row key and family are definitely going to be the same.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
sduskis: Please wrap the `IOException` in a `RuntimeException` for now.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowAdapter.java
rahulKQL: I have updated to compare with qualifier & timestamp... but it seems labels comparison are needed for the case of `WhileMatchFilter`.

rahulKQL:
@sduskis 

Thanks for the suggestions, I have made update Please have a look

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/FlatRowAdapter.java
```diff
@@ -121,14 +121,27 @@ public void finishCell() {
...
+        }
       }
 
       return new FlatRow(this.currentKey, combined.build());
     }
 
+    private boolean keysMatch(Cell current, Cell previous) {
+      return current.getTimestamp() == previous.getTimestamp()
+          && Objects.equals(current.getQualifier(), previous.getQualifier())
+          && Objects.equals(current.getLabels(), previous.getLabels());
```
sduskis: Please remove the `getLabels()` comparison.

sduskis:
@rahulKQL, if you think that this PR is ready, please remove "WIP: " from the title.

rahulKQL:
@sduskis 

Thanks for pointing it out, This PR is ready for another round of review.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RowResultScanner.java
```diff
@@ -0,0 +1,88 @@
...
+        break;
+      }
+      resultList.add(row);
+    }
+    return resultList.toArray(arr);
+  }
+
+  @Override
+  public int available() {
+    throw new UnsupportedOperationException("this operation is not supported");
```
sduskis: @igorbernstein2, is there anyway to get the in-memory count of available rows?

igorbernstein2: No, the  gcj client only parses rows on demand. Closest you can do is `iterator.hasNext() ? 1 : 0;`
Remind me how this is used?

rahulKQL: `ResultScanner#availble` returns the number of rows available without blocking the network. 
Should we use `stream.isReceiveReady() ? 1 : 0` ? as this will be a non-blocking call.
@sduskis 
Please let me know if we should return a value or continue to throw an exception.

sduskis: Please return a value.

rahulKQL: Have updated this method with `stream.isReceiveReady() ? 1 : 0`

PR #2131, opened 2019-03-26
Exclude slf4j-log4j from Beam

sduskis:
The `slf4j-log4j` dependency used by the Beam HBase dependency caused an infinite loop in Beam.


igorbernstein2:
I don't quite understand the underlying issue here. What contributes to the infinite loop?

igorbernstein2:
Spoke offline, the underlying issue is that dataflow workers already have log4j-to-slf4j on the classpath and beam-sdks-java-io-hbase brings in slf4-log4j12, which causes an infinite loop in logging. So it needs to be excluded

PR #2141, opened 2019-04-02
Create the right data client in BigtableSession

sduskis:
We only need one client open at a time - either the `cloud-bigtable-client` or `google-cloud-bigtable` clients.  Let's make sure that we do that correctly in the constructor.

NOTE: The throttling data client is used for BulkMutation even if the options say that google-cloud-java ought to be used.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -669,7 +669,7 @@ public class BigtableSession implements Closeable {
       if (dataGCJClient != null) {
         dataGCJClient.close();
       }
-    } catch (Exception e) {
+    } catch (Exception ex) {
       throw new IOException("Could not close the data client", ex);
     }
     try {
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -633,20 +633,6 @@ public class BigtableSession implements Closeable {
   public synchronized void close() throws IOException {
     watchdog.stop();
 
-    if (managedChannels.isEmpty()) {
-      return;
-    }
-
-    try {
-      if (dataGCJClient != null) {
-        dataGCJClient.close();
-      }
-      if (adminGCJClient != null) {
-        adminGCJClient.close();
-      }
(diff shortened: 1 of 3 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -242,17 +250,37 @@ public BigtableSession(BigtableOptions opts) throws IOException {
...
     CallOptionsFactory.ConfiguredCallOptionsFactory callOptionsFactory =
         new CallOptionsFactory.ConfiguredCallOptionsFactory(options.getCallOptionsConfig());
-    dataClient =
-        new BigtableDataGrpcClient(dataChannel, sharedPools.getRetryExecutor(), options);
-    dataClient.setCallOptionsFactory(callOptionsFactory);
+
+    if (options.useGCJClient()) {
+      BigtableDataSettings dataSettings =
+          BigtableVeneerSettingsFactory.createBigtableDataSettings(options);
+      this.dataGCJClient = new BigtableDataGCJClient(
```
igorbernstein2: Why not merge dataGCJClient and dataClient into a single member variable of type IBigtableDataClient?

sduskis: We still need to have access to the underlying `dataClient` directly for the users that use it directly.  Eventually, we may be able to deprecate the getter, but for now, we'll have to keep it around.

igorbernstein2: with this pr that getter will return null? If so, I think it would be better to use IBigtableDataClient, and in getter check if the underlying client is the classic client and return it. If its gcj to throw a NotImplemented exception

sduskis: +1 on cleanup.  NotImplement exception will cause our integration tests to fail.  Would a warning be enough for now?

igorbernstein2: sure

sduskis: Can we do the refactoring now (i.e. just move the creation of GCJ into the constructor), and do the other cleanup in another PR?

igorbernstein2: ok

PR #2142, opened 2019-04-03
Deprecated getDataClient & createBulkMutation

rahulKQL:
- Marked BigtableSession#getDataClient & BigtableSession#createBulkMutation as deprecated. 
 - Updated classic client usage to wrappers.


rahulKQL:
@sduskis 

With this change, I have just delegated the client's call to wrappers. Please have a look,

sduskis:
Please don't deprecate the clients just yet.  The test change is a good thing!

rahulKQL:
Thanks for the comments, I have reverted deprecation related change.

PR #2144, opened 2019-04-05
Integration test for filtering rows using RegEx

### LINKED ISSUE #2008: Hbase FilterList as MUST_PASS_ONE with two RegexStringComparator RowFilters returns everything
As per https://stackoverflow.com/questions/53395833/hbase-filterlist-as-must-pass-one-with-two-rowfilters-returns-everything 

I have a column family called odds_api with the number of rows with the rowkey: /bt=??/bm=??/mk=??/se=??:

```
/bt=1/bm=MN/mk=344/se=23394/odds_api
/bt=1/bm=BY/mk=344/se=23394/odds_api
/bt=1/bm=SN/mk=344/se=23394/odds_api
/bt=1/bm=BB/mk=344/se=23394/odds_api
/bt=1/bm=SF/mk=344/se=23394/odds_api
/bt=1/bm=XY/mk=344/se=23394/odds_api
```
I want to filter based on a list of bm values aka filter based on bm=SF,BB,MN.

To do this I created a filterList of MUST_PASS_ONE with 1 to many rowFilters (depending on how many values the user requests)

```
public ResultScanner scan(String id , List<String> bms) throws BigTableGetException {
    try{
        Table table = connection.getTable(TableName.valueOf(this.tableName));
        Scan scan = new Scan();
        scan.addFamily(Bytes.toBytes("odds_api"));
        FilterList mainFilterList = new FilterList(FilterList.Operator.MUST_PASS_ONE);

        bms.stream()
                .forEach(bm -> {
                    mainFilterList.addFilter(new RowFilter(CompareOp.EQUAL, new RegexStringComparator("/bt="+id+"/bm="+bm+".*")));

                });
        System.out.println("this is the filter list " + mainFilterList.toString());

        scan.setFilter(mainFilterList);
        return table.getScanner(scan);
    }catch (IOException ex){
        throw new BigTableGetException("Failed to get rows in BigTable", ex);
    }
}
```
(I know -- pointless stream for a forEach!)

This works fine when only one bm is specified, print statement:

`this is the filter list FilterList OR (1/1): [RowFilter (EQUAL, /bt=1/bm=B4.*)]
`
however if more than one is specified then it returns everything, print statement:

`this is the filter list FilterList OR (2/2): [RowFilter (EQUAL, /bt=1/bm=B4.*), RowFilter (EQUAL, /bt=1/bm=PP.*)]`
In fact, if I enter two or more 'incorrect' bm values it still returns everything!

I have also tried to move the bt=? filter out to a separate MUST_PASS_ALL filter and then have another list of row filters for bm=?

`this is the filter list FilterList AND (2/2): [FilterList AND (1/1): [RowFilter (EQUAL, .*bt=1)], FilterList OR (2/2): [RowFilter (EQUAL, .*/bm=B4.*), RowFilter (EQUAL, .*/bm=PP.*)]]`
Same problems.

hbase versions:

```
<dependency>
    <artifactId>bigtable-hbase-1.x</artifactId>
    <version>1.4.0</version>
</dependency>

<dependency>
    <artifactId>bigtable-hbase-1.x-hadoop</artifactId>
    <version>1.4.0</version>
</dependency>
```

rahulKQL:
Fixes #2008

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestFilters.java
+++ b/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestFilters.java
@@ -2126,33 +2126,34 @@ public abstract class AbstractTestFilters extends AbstractTest {
             "/firstKey=1/secKey=ABC/thirdKey=222",
             "/firstKey=2/secKey=CD/thirdKey=333",
             "/firstKey=3/secKey=BT/thirdKey=444",
-            "/firstKey=4/secKey=BTL/thirdKey=555",
-            "/firstKey=4/secKey=AD/thirdKey=666",
+            "/firstKey=4/secKey=AD/thirdKey=555",
+            "/firstKey=4/secKey=BTL/thirdKey=666",
             "/firstKey=5/secKey=SET/thirdKey=777",
             "/firstKey=11/secKey=AA/thirdKey=888"
         };
 
     List<Put> puts = new ArrayList<>();
+    String rowPrefix = dataHelper.randomString("testrow-");
     byte[] qualA = dataHelper.randomData("test-regex");
     for (int i = 0; i < rowKeys.length; i++) {
       String indexStr = String.valueOf(i);
       puts.add(
-          new Put(rowKeys[i].getBytes()).addColumn(COLUMN_FAMILY, qualA, Bytes.toBytes(indexStr)));
+          new Put(Bytes.toBytes(rowPrefix + rowKeys[i])).addColumn(COLUMN_FAMILY, qualA,
+              Bytes.toBytes(indexStr)));
     }
     table.put(puts);
 
-    // Expected rows should be sorted.
-    int[] expected = { 5, 4 };
+    int[] expected = { 4, 5 };
     String[] conditions = { "/firstKey=4.*" };
-    assertRowKeysWithRegex(table, conditions, rowKeys, expected);
+    assertRowKeysWithRegex(table, conditions, rowPrefix, rowKeys, expected);
 
     expected = new int[] { 0, 1 };
     conditions = new String[] { ".*secKey=AB.*" };
-    assertRowKeysWithRegex(table, conditions, rowKeys, expected);
+    assertRowKeysWithRegex(table, conditions, rowPrefix, rowKeys, expected);
 
     expected = new int[] { 0, 6 };
     conditions = new String[] { ".*/thirdKey=111.*", ".*/thirdKey=777.*" };
-    assertRowKeysWithRegex(table, conditions, rowKeys, expected);
+    assertRowKeysWithRegex(table, conditions, rowPrefix, rowKeys, expected);
   }
 
   protected final void assertMatchingRow(Result result, byte[] key) {
@@ -2217,8 +2218,8 @@ public abstract class AbstractTestFilters extends AbstractTest {
     return table;
   }
 
-  private void assertRowKeysWithRegex(Table table, String[] rowRegEx, String[] rowKeys,
-      int[] expected) throws IOException {
+  private void assertRowKeysWithRegex(Table table, String[] rowRegEx,
+      String rowPrefix, String[] rowKeys, int[] expected) throws IOException {
     FilterList filtersList = new FilterList(FilterList.Operator.MUST_PASS_ONE);
     for (String con : rowRegEx) {
       filtersList.addFilter(new RowFilter(CompareOp.EQUAL, new RegexStringComparator(con)));
@@ -2228,7 +2229,8 @@ public abstract class AbstractTestFilters extends AbstractTest {
     try (ResultScanner scanner = table.getScanner(scan)) {
       Result[] results = scanner.next(10);
       for (int i = 0; i < results.length; i++) {
-        Assert.assertArrayEquals(results[i].getRow(), rowKeys[expected[i]].getBytes());
+        Assert.assertArrayEquals(results[i].getRow(),
+            Bytes.toBytes(rowPrefix + rowKeys[expected[i]]));
       }
     }
   }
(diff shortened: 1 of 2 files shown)


bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestFilters.java
```diff
@@ -2118,6 +2118,43 @@ public void testFuzzyWithIntKeys() throws Exception {
...
+            "/firstKey=5/secKey=SET/thirdKey=777",
+            "/firstKey=11/secKey=AA/thirdKey=888"
+        };
+
+    List<Put> puts = new ArrayList<>();
+    byte[] qualA = dataHelper.randomData("test-regex");
+    for (int i = 0; i < rowKeys.length; i++) {
+      String indexStr = String.valueOf(i);
+      puts.add(
+          new Put(rowKeys[i].getBytes()).addColumn(COLUMN_FAMILY, qualA, Bytes.toBytes(indexStr)));
```
sduskis: Please create a single random byte string to put before these row keys.  This test might be run more than once.

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestFilters.java
```diff
@@ -2118,6 +2118,43 @@ public void testFuzzyWithIntKeys() throws Exception {
...
+    }
+    table.put(puts);
+
+    // Expected rows should be sorted.
+    int[] expected = { 5, 4 };
```
sduskis: shouldn't this be `{4, 5}` ?

rahulKQL: The test dataset I created was not sorted So returned output is in sorted order.
I would change the order of test data as it may create confusion.

sduskis: oh... I see.  Thanks for the clarification.

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestFilters.java
rahulKQL: Thanks for the feedback, I have added a random Prefix to rows.

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestFilters.java
```diff
@@ -2180,6 +2218,23 @@ protected Table addDataForTesting(String rowKeyPrefix, byte[] qualA)
     return table;
   }
 
+  private void assertRowKeysWithRegex(Table table, String[] rowRegEx,
+      String rowPrefix, String[] rowKeys, int[] expected) throws IOException {
+    FilterList filtersList = new FilterList(FilterList.Operator.MUST_PASS_ONE);
+    for (String con : rowRegEx) {
+      filtersList.addFilter(new RowFilter(CompareOp.EQUAL, new RegexStringComparator(con)));
```
sduskis: Shouldn't this be `new RegexStringComparator(rowPrefix + con)`?

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestFilters.java
```diff
@@ -2180,6 +2223,24 @@ protected Table addDataForTesting(String rowKeyPrefix, byte[] qualA)
...
+    }
+    Scan scan = new Scan();
+    scan.setFilter(filtersList);
+    try (ResultScanner scanner = table.getScanner(scan)) {
+      Result[] results = scanner.next(10);
+      for (int i = 0; i < results.length; i++) {
+        Assert.assertArrayEquals(results[i].getRow(),
+            Bytes.toBytes(rowPrefix + rowKeys[expected[i]]));
+      }
+    }
```
sduskis: please add another test to make sure that the `scanner.next()` is empty.

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestFilters.java
```diff
@@ -2180,6 +2223,24 @@ protected Table addDataForTesting(String rowKeyPrefix, byte[] qualA)
...
+      String rowPrefix, String[] rowKeys, int[] expected) throws IOException {
+    FilterList filtersList = new FilterList(FilterList.Operator.MUST_PASS_ONE);
+    for (String condition : rowRegEx) {
+      filtersList.addFilter(new RowFilter(CompareOp.EQUAL,
+          new RegexStringComparator(rowPrefix + condition)));
+    }
+    Scan scan = new Scan();
+    scan.setFilter(filtersList);
+    try (ResultScanner scanner = table.getScanner(scan)) {
+      Result[] results = scanner.next(10);
```
sduskis: `scanner.next(rowKeys.length)` might be better than `scanner.next(10)`

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestFilters.java
rahulKQL: Thanks for the comment, Added a check for the scanner.next().

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestFilters.java
rahulKQL: Thanks for the suggestion, Address the same.

rahulKQL:
@sduskis 

I have address all the feedback, Please have a look.

PR #2148, opened 2019-04-12
Cleanup of throttled data client creation.

sduskis:


sduskis:
@rahulKQL, can you please take a look>?

PR #2149, opened 2019-04-12
Adding Integration test for Beam

rahulKQL:
### What this PR contains
- Test cases are meant to test the functionality of `CloudBigtableTableConfiguration` & `CloudBigtableScanConfiguration`.
- This test currently performs writing & reading of 10_000 rows. This can be provided through the system parameter as most of the options are configurable.

_**NOTE:** Currently complete test takes about 19 mins to finish._

### DIFF
--- a/bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/it/CloudBigtableBeamITTest.java
+++ b/bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/it/CloudBigtableBeamITTest.java
@@ -1,0 +1,277 @@
+/*
+ * Copyright 2019 Google LLC. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.beam.it;
+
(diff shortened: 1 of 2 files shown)


bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/it/CloudBigtableBeamITTest.java
```diff
@@ -0,0 +1,283 @@
...
+        .apply("Count", Count.<Result>globally());
+
+    PAssert.thatSingleton(count).isEqualTo(TOTAL_ROW_COUNT);
+    return p;
+  }
+
+  @Test
+  public void testBeamReadAndWrite() throws Exception {
+    // Created Executors for read jobs to run in parallel.
+    ExecutorService executor = Executors.newCachedThreadPool();
```
rahulKQL: Reason for adding executor is to be able to run `BigtableIO.Read` pipeline in parallel.
Currently facing Shaded Library related issue. Would add it to here once resolved.

sduskis: Please don't use an executor.  Please run the write test, wait for it to complete, and then run the read test, or tests.  There's no need for an executor with a single write, and then a single read.  Once we have the multiple reads, you still don't need an executor, of you start both jobs by calling run on each pipeline, and then waiting for the two tests to finish.

rahulKQL: Then we have to make sure read test always run after write test completes. It would not be independent that's the reason I am running write & read in a single test case.

sduskis: The tests should be dependent on each other.

sduskis: The way you wrote it, you're basically waiting until the write is complete before you do the read test.

PR #2153, opened 2019-04-22
Added `RandomStringUtils` Test with prevetted tableName

### LINKED ISSUE #1785: CreateTable flaky test: "a" + RandomStringUtils.random(10, false, false)

rahulKQL:
Fixes #1785 

The reason for test failure for `RandomStringUtils#random` was the difference in checks present for the table name. Please see [here](https://github.com/apache/hbase/blob/branch-1.4/hbase-common/src/main/java/org/apache/hadoop/hbase/TableName.java#L189-L195) for Hbase1 TableName and [here](https://github.com/apache/hbase/blob/rel/2.1.4/hbase-common/src/main/java/org/apache/hadoop/hbase/TableName.java#L206-L211) for Hbase2 TableName.

- `Hbase1`  expects character from the alphabetic character between `[a-zA-Z0-9.]`. It doesn't support any other characters, This is similar in the case of Cloud Bigtable as well.

- `Hbase2` seems to accept multi-lingual alphabetic character even of different language. For example`"a웷忀꺡힟"` works in **Hbase2** but not with **CloudBigtable**.

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestCreateTable.java
+++ b/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestCreateTable.java
@@ -21,6 +21,7 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import org.apache.commons.lang.RandomStringUtils;
 import org.apache.hadoop.hbase.HRegionLocation;
 import org.apache.hadoop.hbase.TableExistsException;
 import org.apache.hadoop.hbase.TableName;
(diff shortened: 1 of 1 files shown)


bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestCreateTable.java
```diff
@@ -135,6 +130,27 @@ private boolean contains(TableName[] tableNames, TableName tableNameTocheck) {
     return false;
   }
 
+  private void assertBadTableName(String tableName){
+    boolean failed = false;
+    try {
+      createTable(TableName.valueOf(tableName));
+    } catch (Exception ex) {
```
sduskis: Add this in the try block:

```
 Assert.fail("Should fail as table name: '" + tableName + "'");
```

and remove the `assertTrue`

rahulKQL:
Question: Should we add a warning log or some comments in BigtableAdmin about these differences?

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestCreateTable.java
rahulKQL: Addressed feedback. Please have an fresh look.

rahulKQL:
@sduskis If you have some time then Please have a look.

PR #2157, opened 2019-04-29
Support for skipping the emulator maven plugin

tommybo-entur:
We need to skip execution in some of the jobs in our delivery pipeline.

### DIFF
--- a/bigtable-test/bigtable-emulator-maven-plugin/README.md
+++ b/bigtable-test/bigtable-emulator-maven-plugin/README.md
@@ -65,3 +65,5 @@ Usage:
   ```java
       Connection connection = BigtableConfiguration.connect("fakeproject", "fakeinstace");
    ```
+- If `maven.test.skip` is set to true the emulator will not start. The emulator can also be controlled directly 
+  by setting the boolean property `bigtable.emulator.skip`.
--- a/bigtable-test/bigtable-emulator-maven-plugin/src/main/java/com/google/cloud/bigtable/test/emulator/Start.java
+++ b/bigtable-test/bigtable-emulator-maven-plugin/src/main/java/com/google/cloud/bigtable/test/emulator/Start.java
@@ -40,8 +40,15 @@ public class Start extends AbstractMojo {
   @Parameter(readonly = true, defaultValue = "bigtable.emulator.endpoint")
   private String propertyName;
 
+  @Parameter(defaultValue = "${maven.test.skip}", property = "bigtable.emulator.skip")
+  private boolean skip;
+
 
   public void execute() throws MojoExecutionException {
+    if(skip) {
+      getLog().info("bigtable.emulator.skip resolved to true. Skipping excution.");
+      return;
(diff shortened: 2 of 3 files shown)


rahulKQL:
@tommybo-entur Thanks for the PR.

Could you please help me to understand the use case for this change.
From what I get is that you want to enable the emulator plugin optionally in your jobs. If this is correct then, Can you not add this plugin in any profile and enable that profile when you want to enable emulator.

tommybo-entur:
It's the opposite. In our release pipeline we've got some jobs that doesn't require the tests to run. These jobs can run in a thin/small docker container as long as we also skip the emulator.
To obtain the same process using profiles I need to configure a profile that needs to be enabled for normal runs and disabled on occation. I generally find profiles to be a very verbose and counterintuitive way to control the build flow, while having a skip option (as many other stock maven plugin's) is super easy. Also, as the plugin's purpose is to simplify testing it makes sense to me to tie this to the 'maven.test.skip'.

sduskis:
@tommybo-entur, can you please sign the CLA?

tommybo-entur:
Of course. I'm just checking with my organisation whether we already have a google group. Will know tomorrow.

tommybo-entur:
A new CLA for our organisation has been submitted so I'm currently waiting for my superior to sign.

tommybo-entur:
It has been signed now.

igorbernstein2:
This change looks good to me. But I would highly encourage you to consider migrating to 
https://github.com/googleapis/google-cloud-java/tree/master/google-cloud-testing/google-cloud-bigtable-emulator

The new integration:
- supports JUnit Rules
- allows for direct programmatic control
- bundles the emulator binary

Example usage:
```java
@Rule
public final BigtableEmulatorRule bigtableEmulator = BigtableEmulatorRule.create();
private BigtableSession session;
private Connection connection;

@Setup
public void setup() {
  // for -core
  BigtableOptions opts = new BigtableOptions.Builder()
        .enableEmulator("localhost:" + bigtableEmulator.getPort())
        .setUserAgent("fake")
        .setProjectId("fakeproject")
        .setInstanceId("fakeinstance")
        .build();
   this.session = new BigtableSession(opts);

  // for hbase
  Configuration config = BigtableConfiguration.configure("fakeproject", "fakeinstance");
  config.set("google.bigtable.emulator.endpoint.host", "localhost:" + bigtableEmulator.getPort());
  this.connection = BigtableConfiguration.connect(config);
}
```

Alternatively, if you don't use JUnit, you can use the Emulator wrapper directly:
https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-testing/google-cloud-bigtable-emulator/src/main/java/com/google/cloud/bigtable/emulator/v2/Emulator.java

This new integration should be a lot less magical and easier to manage.

tommybo-entur:
Thanks for the tip. I'll be sure to pass it on to our developers. Does indeed look more manageable. :-)

PR #2160, opened 2019-05-10
Introduce different timeouts for mutateRows & read rows

### LINKED ISSUE #2156: Scans and timeouts
We have enabled our client to use the timeout feature. We have set a timeout of 15 seconds for both long and short rpc values since that works well for our standard Get and Put calls and our overall application level timeouts. We have noticed some unexpected behavior for our long running scans. We can demonstrate with this code
```
            Scan scan = new Scan();
            try (HTableInterface table = connection.getTable(table);
                 ResultScanner scanner = table.getScanner(scan)) {

                for (Result r : scanner) {
                    Thread.sleep(100);
                }
            }
```
After 15 seconds we see a `Could not complete RPC. Failure #0, got: Status{code=DEADLINE_EXCEEDED,` exception log message. After reading more entries and exhausting the current entries in the underly ResponseQueueReader, we see an exception is thrown. 
```
Exception in thread "main" java.lang.RuntimeException: com.google.bigtable.repackaged.com.google.cloud.bigtable.grpc.io.IOExceptionWithStatus: Error in response stream
Caused by: com.google.bigtable.repackaged.com.google.cloud.bigtable.grpc.io.IOExceptionWithStatus: Error in response stream
	at com.google.bigtable.repackaged.com.google.cloud.bigtable.grpc.scanner.ResultQueueEntry$ExceptionResultQueueEntry.getResponseOrThrow(ResultQueueEntry.java:100)
	at com.google.bigtable.repackaged.com.google.cloud.bigtable.grpc.scanner.ResponseQueueReader.getNextMergedRow(ResponseQueueReader.java:105)
	at com.google.bigtable.repackaged.com.google.cloud.bigtable.grpc.scanner.ResponseQueueReader.getNextMergedRow(ResponseQueueReader.java:111)
	at com.google.bigtable.repackaged.com.google.cloud.bigtable.grpc.scanner.ResumingStreamingResultScanner.next(ResumingStreamingResultScanner.java:78)
	at com.google.bigtable.repackaged.com.google.cloud.bigtable.grpc.scanner.ResumingStreamingResultScanner.next(ResumingStreamingResultScanner.java:35)
	at com.google.cloud.bigtable.hbase.adapters.read.BigtableResultScannerAdapter$1.next(BigtableResultScannerAdapter.java:66)
	at com.sift.db.bigtable.BigtableScanner$1.hasNext(BigtableScanner.java:66)
	... 1 more
Caused by: com.google.bigtable.repackaged.io.grpc.StatusRuntimeException: DEADLINE_EXCEEDED: deadline exceeded after 14934826494ns
	at com.google.bigtable.repackaged.io.grpc.Status.asRuntimeException(Status.java:517)
	at com.google.bigtable.repackaged.com.google.cloud.bigtable.grpc.async.AbstractRetryingOperation.onError(AbstractRetryingOperation.java:206)
	at com.google.bigtable.repackaged.com.google.cloud.bigtable.grpc.async.AbstractRetryingOperation.onClose(AbstractRetryingOperation.java:171)
	at com.google.bigtable.repackaged.com.google.cloud.bigtable.grpc.scanner.RetryingReadRowsOperation.onClose(RetryingReadRowsOperation.java:176)
	at com.google.bigtable.repackaged.com.google.cloud.bigtable.grpc.io.ChannelPool$InstrumentedChannel$2.onClose(ChannelPool.java:210)
	at com.google.bigtable.repackaged.io.grpc.PartialForwardingClientCallListener.onClose(PartialForwardingClientCallListener.java:39)
	at com.google.bigtable.repackaged.io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:23)
	at com.google.bigtable.repackaged.io.grpc.ForwardingClientCallListener$SimpleForwardingClientCallListener.onClose(ForwardingClientCallListener.java:40)
	at com.google.bigtable.repackaged.com.google.cloud.bigtable.grpc.io.Watchdog$WatchedCall$1.onClose(Watchdog.java:176)
...
```
It appears a user cannot iterate through the ResultScanner object for longer than the long rpc timeout value.

The behavior I would except is have the timeout on the individual ReadRows calls to get the next batch of rows. If one of those is over the timeout, then throw an exception. But reading the results of a scan can continue over the 15 second threshold. Is there a fix that can be made to implement this behavior?

It appears the scan operation will use long rpc timeout value. We don't want to raise this value because it's also used for put(Put[]) requests which we would like to keep within 15 seconds.

cc @dmwaura

rahulKQL:
Fixes #2156 

Introduced two new timeout fields in the `CallOptiosConfig` for each long request operation (i.e. `mutateRows` & `readRows`). This fields will also be defaults to `600 seconds` same as current `longRpcTimeoutMs`.

Deprecated`longRpcTimeoutMs`, If the user continues to use this then it can be translated into new fields.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
@@ -109,8 +109,8 @@ public class CallOptionsConfig implements Serializable {
      *
      * @param longRpcTimeoutMs timeout value in milliseconds.
      * @return a {@link Builder} object, for chaining
-     *
-     * @deprecated Please use {@link #setMutateRpcTimeoutMs(int)} or {@link #setReadRowsRpcTimeoutMs(int)}.
+     * @deprecated Please use {@link #setMutateRpcTimeoutMs(int)} or {@link
+     *     #setReadRowsRpcTimeoutMs(int)}.
      */
     @Deprecated
     public Builder setLongRpcTimeoutMs(int longRpcTimeoutMs) {
@@ -121,11 +121,13 @@ public class CallOptionsConfig implements Serializable {
 
     /**
      * The amount of milliseconds to wait before issuing a client side timeout for mutation RPCs.
+     *
      * @param mutateRpcTimeoutMs timeout value in milliseconds.
      * @return a {@link Builder} object, for chaining
      */
     public Builder setMutateRpcTimeoutMs(int mutateRpcTimeoutMs) {
-      Preconditions.checkArgument(mutateRpcTimeoutMs > 0, "Long Timeout ms has to be greater than 0");
+      Preconditions.checkArgument(
+          mutateRpcTimeoutMs > 0, "Long Timeout ms has to be greater than 0");
       this.mutateRpcTimeoutMs = mutateRpcTimeoutMs;
       return this;
     }
@@ -133,11 +135,13 @@ public class CallOptionsConfig implements Serializable {
     /**
      * The amount of milliseconds to wait before issuing a client side timeout for readRows
      * streaming RPCs.
+     *
      * @param readStreamRpcTimeoutMs timeout value in milliseconds.
      * @return a {@link Builder} object, for chaining
      */
-    public Builder setReadRowsRpcTimeoutMs(int readStreamRpcTimeoutMs){
-      Preconditions.checkArgument(readStreamRpcTimeoutMs > 0, "Read Stream Timeout ms has to be greater than 0");
+    public Builder setReadRowsRpcTimeoutMs(int readStreamRpcTimeoutMs) {
+      Preconditions.checkArgument(
+          readStreamRpcTimeoutMs > 0, "Read Stream Timeout ms has to be greater than 0");
       this.readRowsRpcTimeoutMs = readStreamRpcTimeoutMs;
       return this;
     }
@@ -159,7 +163,6 @@ public class CallOptionsConfig implements Serializable {
    * @param useTimeout a boolean.
    * @param unaryRpcTimeoutMs an int.
    * @param longRpcTimeoutMs an int.
-   *
    * @deprecated Please use {@link #builder()}
    */
   @Deprecated
@@ -221,8 +224,8 @@ public class CallOptionsConfig implements Serializable {
    * Getter for the field <code>longRpcTimeoutMs</code>.
    *
    * @return an int.
-   *
-   * @deprecated Please use {@link #getMutateRpcTimeoutMs()} or {@link #getReadStreamRpcTimeoutMs()}.
+   * @deprecated Please use {@link #getMutateRpcTimeoutMs()} or {@link
+   *     #getReadStreamRpcTimeoutMs()}.
    */
   @Deprecated
   public int getLongRpcTimeoutMs() {
@@ -230,7 +233,7 @@ public class CallOptionsConfig implements Serializable {
   }
 
   /**
-   * <p>Getter for the field <code>mutateRpcTimeoutMs</code>.</p>
+   * Getter for the field <code>mutateRpcTimeoutMs</code>.
    *
    * @return an int.
    */
@@ -239,7 +242,7 @@ public class CallOptionsConfig implements Serializable {
   }
 
   /**
-   * <p>Getter for the field <code>readStreamRpcTimeoutMs</code>.</p>
+   * Getter for the field <code>readStreamRpcTimeoutMs</code>.
    *
    * @return an int.
    */
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
@@ -105,7 +105,8 @@ public class CallOptionsConfig implements Serializable {
     }
 
     /**
-     * The amount of milliseconds to wait before issuing a client side timeout for long RPCs.
+     * The amount of milliseconds to wait before issuing a client side timeout for long running
+     * RPCs.
      *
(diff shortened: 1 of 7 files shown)


rahulKQL:
@igorbernstein2 @sduskis 
Please have a look.

sduskis:
@kolea2, can you please take a look?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
```diff
@@ -105,34 +109,86 @@ public Builder setShortRpcTimeoutMs(int shortRpcTimeoutMs) {
...
 
+    /**
+     * The amount of milliseconds to wait before issuing a client side timeout for mutation RPCs.
+     *
+     * @param mutateRpcTimeoutMs timeout value in milliseconds.
+     * @return a {@link Builder} object, for chaining
+     */
+    public Builder setMutateRpcTimeoutMs(int mutateRpcTimeoutMs) {
+      Preconditions.checkArgument(
+          mutateRpcTimeoutMs > 0, "Long Timeout ms has to be greater than 0");
```
kolea2: should mutateRpcTimeoutMs be an int or a long?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/CallOptionsFactory.java
```diff
@@ -93,24 +93,24 @@ public ConfiguredCallOptionsFactory(CallOptionsConfig config) {
...
     }
 
     /**
-     * @param request
-     * @return true if this is a {@link MutateRowsRequest} or a {@link ReadRowsRequest} that's a
-     *     scan.
+     * @param request an RPC request.
+     * @return timeout in milliseconds as per the Request type.
      */
-    public static boolean isLongRequest(Object request) {
```
kolea2: Deleting this would be a breaking change - can we keep it or deprecate it?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
```diff
@@ -258,9 +258,24 @@
    * If timeouts are set, how many milliseconds should pass before a DEADLINE_EXCEEDED for a long
    * read? Currently, this feature is experimental.
    */
+  @Deprecated
```
kolea2: please add to the javadoc what to use instead

bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/TestCallOptionsFactory.java
```diff
@@ -100,7 +108,8 @@ public void run() {
                 CallOptionsConfig.builder()
                     .setUseTimeout(true)
                     .setShortRpcTimeoutMs((int) TimeUnit.SECONDS.toMillis(100))
-                    .setLongRpcTimeoutMs((int) TimeUnit.SECONDS.toMillis(1000))
+                    .setMutateRpcTimeoutMs((int) TimeUnit.SECONDS.toMillis(1000))
```
kolea2: can we add more tests with just .setMutateRpcTimeoutMs, one with just .setReadRwosRpcTimeoutsMs, and one with no settings?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
sduskis: I think "Long Timeout" referred to "long rpc" vs. "short rpc" timeouts.  Timeouts should be `int`s.

rahulKQL: @sduskis Thanks for the explanation

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
rahulKQL:
```java
  /**
   *  @deprecated Please use {@link #BIGTABLE_MUTATE_RPC_TIMEOUT_MS_KEY} or {@link
   *     #BIGTABLE_READ_RPC_TIMEOUT_MS_KEY} based on long operation.
   */
```

I believe this deprecated comment should be enough for the user to be able to distinguish new timeout key.

bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/TestCallOptionsFactory.java
rahulKQL: added more tests to cover these scenarios.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/CallOptionsFactory.java
rahulKQL: Thanks for pointing it out. I have reverted this method and added a deprecation message.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
kolea2: Thanks! In that case, can we add that clarification to the error message? That wasn't obvious to me when first reading.

rahulKQL:
@kolea2 Updated the error messages to be explanative. Please have a fresh look.

PR #2162, opened 2019-05-20
Introduce `google-java-format` plugin for consistent formmating

rahulKQL:
This change adds a plugin in root pom.xml to format the files consistent across other google java repository. Formatting commands would be:
- To check format:

        mvn -X com.coveo:fmt-maven-plugin:check

- To enforce formatting:

        mvn -X com.coveo:fmt-maven-plugin:format

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ infrequent product and client library announcements.
    -pl bigtable-hbase-2.x-parent/bigtable-hbase-2.x-integration-tests \
    -am clean install
    ```
-3. Please run `mvn -X com.coveo:fmt-maven-plugin:format` to check and enforce consistent java formatting to all of your code changes.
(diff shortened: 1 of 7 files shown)


rahulKQL:
@igorbernstein2 @kolea2 
Q: Should we consider adding formatting plugin to this repo? 
As the total number of affected files was `436` when ran at the root level, So I have only added files from `bigtable-hbase-1x` module in this PR.

igorbernstein2:
I think this is a great addition. But I think this PR should be split as follows:
- This PR should add the plugin and update development docs: https://github.com/googleapis/cloud-bigtable-client#developing-and-testing
- PR 2 should reformat the entire repo
- PR 3 should add presubmit checks

rahulKQL:
@igorbernstein2 Please have a look

PR #2168, opened 2019-05-29
Introducing google-cloud-bigtable-emulator

### LINKED ISSUE #1957: no way to programatically set credentials when using cbt emulator
Using BIGTABLE_EMULATOR_HOST is a reasonable solution for trying things out, but when using something like testcontainers to unit test with cbt emulator in a container with an unpredictable port it's not possible to set environment variables.

It would be very helpful if there was a way to set credentials programatically as can be done with BigtableOptions in the normal client, or GoogleCredentials and similar with storage and other products.

rahulKQL:
Fixes #1957 

## What this PR contains:
- Added instruction to programmatically set host & port in case of running this client in containers.
- Introduced new Google-cloud-emulator (inspired by [this](https://github.com/googleapis/cloud-bigtable-client/pull/2157#issuecomment-491359555) comment).

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,6 @@ infrequent product and client library announcements.
    mvn clean verify -PhbaseLocalMiniClusterTestH2
    ```
 
-
    You can run those commands at the top of the project, or you can run them at the appropriate integration-tests project.  
    
    Developer's NOTE: You can build the project faster by running the following command, and then run the integration test command from the appropriate integration test directory:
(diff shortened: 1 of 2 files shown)


rahulKQL:
@igorbernstein2 Whenever get a chance, Please have a look.

igorbernstein2:
I think we should start directing users to https://github.com/googleapis/google-cloud-java/tree/master/google-cloud-testing/google-cloud-bigtable-emulator.

Docker adds too much complexity

rahulKQL:
@igorbernstein2 I have removed the docker related instruction from this PR. Please have a fresh look.

igorbernstein2:
I think we should move the docs for connecting to the emulator to the main readme. And change the docs in the maven plugin to direct people to use the gcj emulator

PR #2171, opened 2019-06-04
Changed Throwable logs to debug level to reduce unwanted logs in build

rahulKQL:
Line#643 was creating a lot of exception log in the build-log, It turns out this is originated from [hbase-2x-config](https://github.com/googleapis/cloud-bigtable-client/blob/master/bigtable-hbase-2.x-parent/bigtable-hbase-2.x/src/main/resources/log4j.properties#L24).

@igorbernstein2, Please let me know your view on this. Not sure if just a cleaner build log is good enough reason for this change.


rahulKQL:
@igorbernstein2 Whenever get a chance, Please have a look.

PR #2174, opened 2019-06-17
CreateTable doesn't call getTable()

sduskis:
CreateTable shouldn't need to call getTable()

@rahulKQL can you please take a look>?


rahulKQL:
Looking good to me except I think we may have to run formatting command.

Just a side Question: Should I include formatting check(`mvn -X com.coveo:fmt-maven-plugin:check`).
 in the build lifecycle?  And add a separate profile with `applyGoogleJavaFormat`?

PR #2178, opened 2019-06-21
Updated `AbstractBigtableTable#put` to include latency metrics

rahulKQL:
adding this fix as I am not sure the reason behind having an unused Timer.


sduskis:
The `mutateRow` method already has an OpenCensus timer.  I'm not sure where the client is going in terms of DropWizard relating timing.  I think that @igorbernstein2 should make a decision about DropWizard before we add more timings.

igorbernstein2:
I'm working on instrumenting gcj client with opencensus now. The gcj client will not support dropwizard metrics directly, but it might be possible add a compatibility layer by implementing an opencensus exporter to dropwizard. I would like to experiment with this approach before adding new metrics to the hbase layer. If this is possible then I would like instrument the hbase client with opencensus, and then add the dropwizard exporter. If this isn't possible then I would like to fill out the hbase layer with dropwizard metrics and then migrate the -core metrics like mtuateRow to opencensus.
So I would vote for folding off merging this for now

rahulKQL:
Closing this PR for now; This change will be addressed once the approach for instrumentation is finalized.

PR #2181, opened 2019-06-24
Updating GCJ dependency with latest GCJ version

rahulKQL:
After updating to `google-cloud-java:0.97.0`, I started to see ClassNotFoundException while creating GCJ connection:
```java
java.lang.ClassNotFoundException: io.grpc.alts.ComputeEngineChannelBuilder
```

For this, we require `grpc-alts` which is transitively coming from `gax-grpc`
```sh
[INFO] |  +- com.google.api:gax-grpc:jar:1.46.1:compile
[INFO] |  |  +- com.google.api:gax:jar:1.46.1:compile
[INFO] |  |  +- org.threeten:threetenbp:jar:1.3.3:compile
[INFO] |  |  \- io.grpc:grpc-alts:jar:1.21.0:compile
[INFO] |  |     +- io.grpc:grpc-grpclb:jar:1.21.0:compile
[INFO] |  |     \- org.apache.commons:commons-lang3:jar:3.5:compile
```

To fix this issue, I've excluded the only other grpc jar(i.e `grpc-grpclb`) from this instead of excluding all grpc jars. I hope this should not create any unknown dependency problem for `bigtable-core`.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/pom.xml
+++ b/bigtable-client-core-parent/bigtable-client-core/pom.xml
@@ -169,7 +169,7 @@ limitations under the License.
             <exclusions>
                 <exclusion>
                     <groupId>io.grpc</groupId>
-                    <artifactId>*</artifactId>
+                    <artifactId>grpc-grpclb</artifactId>
                 </exclusion>
             </exclusions>
         </dependency>
--- a/bigtable-client-core-parent/bigtable-client-core/pom.xml
+++ b/bigtable-client-core-parent/bigtable-client-core/pom.xml
@@ -166,12 +166,6 @@ limitations under the License.
         <dependency>
             <groupId>com.google.api</groupId>
             <artifactId>gax-grpc</artifactId>
-            <exclusions>
-                <exclusion>
-                    <groupId>io.grpc</groupId>
-                    <artifactId>grpc-grpclb</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
 
         <!-- Metrics -->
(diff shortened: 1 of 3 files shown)


rahulKQL:
@igorbernstein2, @sduskis Please take a look and let me know if we should handle `grpc-alts` differently.

igorbernstein2:
Please leave the exclusion as is and the required dep along with the other deps here:
https://github.com/googleapis/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/pom.xml#L164

rahulKQL:
I initially added `grpc` dependency directly in the core, but it seems we have to maintain the version number(which could easily get inconsistent) as `grpc-alts` is not present in google-cloud-clients BOM.

Please let me know If we are ok to go forward with adding the version?

igorbernstein2:
I think version should be added to [google-cloud-clients/pom.xml](https://github.com/googleapis/google-cloud-java/blob/6197bb58b2c8b5cf7c9a6bd7aa0781db29ce58d1/google-cloud-clients/pom.xml#L313)

rahulKQL:
I would update this PR with next release of GCJ, which would include changes present in [#5553](https://github.com/googleapis/google-cloud-java/pull/5553).

igorbernstein2:
Giving this a bit more thought, I think the reason the exclusion was added was because we had a mismatch between gcj grpc version and hbase. At this point we should be in sync and can probably remove the exclusion and the explicit versions

rahulKQL:
@igorbernstein2  If we would be in sync so we don't need to provide any explicit version. 

Should we keep this exclusion only(as we don't need this dependency) and allow `grpc-alts` transitively. Please let me know your thoughts.
```XML
        <dependency>
            <groupId>com.google.api</groupId>
            <artifactId>gax-grpc</artifactId>
            <exclusions>
                <exclusion>
                    <groupId>io.grpc</groupId>
                    <artifactId>grpc-grpclb</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
```

igorbernstein2:
I think that exclusion should be removed as well

rahulKQL:
@igorbernstein2 Thanks!! for confirming. Have removed the `<exclusions>` tag. Please take a fresh look.

PR #2186, opened 2019-07-19
Fix for long pauses and unauthentication issue with batch endpoint

rahulKQL:
**Issue:** When running batch operation, dataflow jobs are taking long pauses in case of dataflow jobs using `CloudBigtableIO` and logging a lot of un-authentication issue.

**Fix:** Now when the client is used with `batch-bigtable.googleapis.com` host, then only OAuth2 authentication would be used instead of using JWT token.

**Note:** This is a temporary workaround until the internal team fixes the issue. This commit may be reverted once fix is implemented.

@igorbernstein2 @kolea2


igorbernstein2:
This is unfortunate. We ended up working on the same thing.
I took a slightly different approach here: #2187.  Which will allow us to keep JWT enabled and avoids modifying any public method signatures. Unless you have strong feelings about this approach, I'd prefer to go with #2187

PR #2190, opened 2019-07-31
Updating GCJ and Guava version

rahulKQL:
- Added `Assert#assume` to cover IT test with `useGCJClient` flag.
- Increase RPC timeout to 20seconds(It was failing on my system).
- Updated GCJ client version.
- Updated `fmt-format-plugin` to auto-format instead of format check(This will not affect pre-submit format check present for travis).

CC: @igorbernstein2 @kolea2 
Please have a look

### DIFF
--- a/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestAuth.java
+++ b/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestAuth.java
@@ -66,7 +66,8 @@ public class TestAuth extends AbstractTest {
 
     // Prevent the test from hanging if auth fails
     config.set("google.bigtable.rpc.use.timeouts", "true");
-    config.set("google.bigtable.rpc.timeout.ms", "20000");
+    config.set("google.bigtable.rpc.timeout.ms", "10000");
+    config.set("google.bigtable.grpc.channel.count", "1");
 
     // Create a new connection using JWT auth & batch settings
     try (Connection connection = BigtableConfiguration.connect(config)) {
--- a/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestAuth.java
+++ b/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestAuth.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2019 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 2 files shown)


rahulKQL:
@igorbernstein2, @kolea2  If you have a few mins time, then Please review this PR.

pom.xml
```diff
@@ -233,7 +233,7 @@ limitations under the License.
                     <execution>
                         <phase>process-sources</phase>
                         <goals>
-                            <goal>check</goal>
```
igorbernstein2: please leave this as check, I don't like the idea of maven updating the source files w/o me asking it to

pom.xml
```diff
@@ -58,9 +58,9 @@ limitations under the License.
         <hamcrest.version>1.3</hamcrest.version>
 
         <!-- google-cloud-java related dependency versions -->
-        <google-cloud.version>0.97.0-alpha</google-cloud.version>
+        <google-cloud.version>0.102.0-alpha</google-cloud.version>
         <opencensus.version>0.21.0</opencensus.version>
-        <guava.version>27.1-android</guava.version>
+        <guava.version>28.0-android</guava.version>
```
igorbernstein2: Do we still need to specify this? I would think that we inherit it from google-cloud-clients

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestAuth.java
```diff
@@ -48,7 +66,7 @@ public void testBatchJwt() throws IOException {
 
     // Prevent the test from hanging if auth fails
     config.set("google.bigtable.rpc.use.timeouts", "true");
-    config.set("google.bigtable.rpc.timeout.ms", "10000");
+    config.set("google.bigtable.rpc.timeout.ms", "20000");
```
igorbernstein2: Why did this need to increase?

rahulKQL: Not sure why but it kept failing on with ~5-6 extra seconds. 
It might be because low bandwidth(~50Mbps) or could be low connection boot time(bigtable regions were all `us-central1-f`).

pom.xml
rahulKQL: Reverted

pom.xml
rahulKQL: Unfortunately yes, We are referring this version in [bigtable-beam-import](https://github.com/googleapis/cloud-bigtable-client/blob/master/bigtable-dataflow-parent/bigtable-beam-import/pom.xml#L48) & [bigtable-emulator](https://github.com/googleapis/cloud-bigtable-client/blob/master/bigtable-test/bigtable-emulator-maven-plugin/pom.xml#L52).

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestAuth.java
igorbernstein2: can you try changing the connection pool size to 1 and see if that solves the problem?

pom.xml
igorbernstein2: Ok, can you add a comment to avoid future confusion?

rahulKQL: Added comment and grouped guava version with the beam.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestAuth.java
rahulKQL: Thanks for the suggestion, It works fine with the connection pool size of 1.
Reverted back the timeout and added connection pool config.

PR #2191, opened 2019-07-31
Updating Hbase version to latest( 1x: 1.4.10, 2x: 2.2.0)

rahulKQL:
Latest version of Hbase-2.2.0([release note](https://apache.org/dist/hbase/2.2.0/RELEASENOTES.md)) has introduced new operations.

Some of the additive methods introduced in hbase 2.x:
```
BigtableAdmin#createTableAsync(TableDescriptor tableDescriptor)
BigtableAsyncAdmin#listTableDescriptors(List<TableName> list)
BigtableAsyncAdmin#restoreSnapshot(String s, boolean b, boolean b1)
BigtableAsyncAdmin#cloneSnapshot(String s, TableName tableName, boolean b)

BigtableAsyncTableRegionLocator#getRegionLocations(byte[] bytes, boolean b)
BigtableAsyncTableRegionLocator#getAllRegionLocations()
BigtableAsyncTableRegionLocator#clearRegionLocationCache()

BigtableConnection#clearRegionLocationCache()

BigtableAsyncConnection#isClosed()
BigtableAsyncConnection#setRetryPauseForCQTBE(long l, TimeUnit timeUnit)
BigtableAsyncConnection#setMaxKeyValueSize(int i)
BigtableAsyncConnection#clearRegionLocationCache()
```

After discussion, we came to a conclusion that we should implement these three methods(Other methods can be implemented based on usage and user request).

```
BigtableAdmin#createTableAsync(TableDescriptor tableDescriptor)
BigtableAsyncAdmin#listTableDescriptors(List<TableName> list)
BigtableAsyncConnection#isClosed()
```

### DIFF
--- a/bigtable-hbase-2.x-parent/bigtable-hbase-2.x-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/async/TestAsyncAdmin.java
+++ b/bigtable-hbase-2.x-parent/bigtable-hbase-2.x-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/async/TestAsyncAdmin.java
@@ -22,9 +22,11 @@ import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 
 import com.google.cloud.bigtable.hbase.test_env.SharedTestEnvRule;
+import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.TableNotDisabledException;
 import org.apache.hadoop.hbase.TableNotEnabledException;
@@ -91,6 +93,10 @@ public class TestAsyncAdmin extends AbstractAsyncTest {
           .get();
       assertEquals(true, asyncAdmin.tableExists(tableName).get());
 
+      // create one more table
+      TableName anotherTableName = sharedTestEnv.newTestTableName();
+      sharedTestEnv.createTable(anotherTableName);
+
       // test listTableNames all
       List<TableName> allTableNames = asyncAdmin.listTableNames().get();
       assertTrue("listTableNames-all should list atleast one table", allTableNames.size() > 0);
@@ -128,6 +134,19 @@ public class TestAsyncAdmin extends AbstractAsyncTest {
           patTableDescriptors.stream().anyMatch(e -> tableName.equals(e.getTableName())));
       // TODO: Verify why this test fails. getColumnFamilies() array is empyty
       // assertEquals(10, patTableDescriptors.get(0).getColumnFamilies()[0].getTimeToLive());
+
+      // test listTableDescriptors by List<TableName>
+      List<TableDescriptor> listDescriptor =
+          asyncAdmin.listTableDescriptors(Arrays.asList(tableName, anotherTableName)).get();
+      assertTrue(
+          "listTableDescriptors(list) should list at least two table", listDescriptor.size() > 1);
+      assertTrue(
+          "listTableDescriptors(list) should contain both the table ",
+          listDescriptor.stream()
+              .map(e -> e.getTableName().getNameAsString())
+              .collect(Collectors.toList())
+              .containsAll(
+                  Arrays.asList(tableName.getNameAsString(), anotherTableName.getNameAsString())));
 
       // test getTableDescriptor
       TableDescriptor tableDescriptor = asyncAdmin.getDescriptor(tableName).get();
--- a/bigtable-hbase-2.x-parent/bigtable-hbase-2.x-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/async/TestAsyncAdmin.java
+++ b/bigtable-hbase-2.x-parent/bigtable-hbase-2.x-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/async/TestAsyncAdmin.java
@@ -16,10 +16,13 @@
 package com.google.cloud.bigtable.hbase.async;
 
 import static com.google.cloud.bigtable.hbase.test_env.SharedTestEnvRule.COLUMN_FAMILY;
+import static com.google.cloud.bigtable.hbase.test_env.SharedTestEnvRule.COLUMN_FAMILY2;
+import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 import com.google.cloud.bigtable.hbase.test_env.SharedTestEnvRule;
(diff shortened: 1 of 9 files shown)


bigtable-hbase-2.x-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/BigtableAsyncAdmin.java
```diff
@@ -259,6 +263,27 @@ public BigtableAsyncAdmin(CommonConnection asyncConnection) throws IOException {
     return listTables(Optional.of(pattern));
   }
 
+  @Override
+  public CompletableFuture<List<TableDescriptor>> listTableDescriptors(List<TableName> list) {
```
igorbernstein2: Please add an integration test for testing null & empty list cases and run them using the minicluster profile and bigtable profile to make sure that the beahvior matches

igorbernstein2: Also check if the returned list should contain column families

bigtable-hbase-2.x-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/BigtableAsyncAdmin.java
```diff
@@ -264,23 +266,18 @@ public BigtableAsyncAdmin(CommonConnection asyncConnection) throws IOException {
...
+                tableNames.stream()
                     .filter(in -> t.contains(in.getNameAsString()))
-                    .map(
-                        m ->
-                            com.google.bigtable.admin.v2.Table.newBuilder()
-                                .setName(bigtableInstanceName.toTableNameStr(m.getNameAsString()))
-                                .build())
-                    .map(Table::fromProto)
-                    .map(tableAdapter2x::adapt)
+                    .map(tbName -> getDescriptor(tbName).join())
```
igorbernstein2: There is a race condition here: the table might be deleted between a list and get operations Make sure to catch the not found error and remove the table from the list

bigtable-hbase-2.x-parent/bigtable-hbase-2.x-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/async/TestAsyncAdmin.java
```diff
@@ -136,18 +139,38 @@ public void testCreateTable_harness() throws Exception {
       // assertEquals(10, patTableDescriptors.get(0).getColumnFamilies()[0].getTimeToLive());
 
       // test listTableDescriptors by List<TableName>
+      try {
+        asyncAdmin.listTableDescriptors(null).get();
+        fail("listTableDescriptors should throw an exception");
+      } catch (Exception e) {
+        Assert.assertTrue(e instanceof NullPointerException);
```
igorbernstein2: Please be consistent in your use of static imports

bigtable-hbase-2.x-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/BigtableAsyncAdmin.java
rahulKQL: Thanks for detailed feedback, I have addressed this. Please take a look.

bigtable-hbase-2.x-parent/bigtable-hbase-2.x-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/async/TestAsyncAdmin.java
rahulKQL: Addressed

bigtable-hbase-2.x-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/BigtableAsyncAdmin.java
```diff
@@ -259,6 +265,34 @@ public BigtableAsyncAdmin(CommonConnection asyncConnection) throws IOException {
...
+                tableNames.stream()
+                    .filter(inputTableName -> t.contains(inputTableName.getNameAsString()))
+                    .map(
+                        tbName -> {
+                          try {
+                            return getDescriptor(tbName).join();
+                          } catch (CompletionException ex) {
+                            LOG.warn(
+                                "Table not found while fetching details for %s",
+                                ex, tbName.getNameAsString());
```
igorbernstein2: I don’t think you can assume that an error means the table is not found. I think you should examine the error and check if it’s a statusruntimeexception with a not found status or a notfoundexception from gax

rahulKQL: As [BigtableAsyncAdmin#getDescirptor](https://github.com/googleapis/cloud-bigtable-client/blob/master/bigtable-hbase-2.x-parent/bigtable-hbase-2.x/src/main/java/com/google/cloud/bigtable/hbase2_x/BigtableAsyncAdmin.java#L278-L296) is handling `Status.Code.NOT_FOUND`, than re throwing exception for case other than `Status.Code.NOT_FOUND` ok?

```java
toCompletableFuture(bigtableTableAdminClient.listTablesAsync())
        .thenApply(
            t ->
                tableNames.stream()
                    .filter(inputTableName -> t.contains(inputTableName.getNameAsString()))
                    .map(
                        tbName -> {
                          try {
                            return getDescriptor(tbName).join();
                          } catch (CompletionException ex) {
                            if(ex.getCause() instanceof  TableNotFoundException){
                              LOG.warn(
                                  "Table not found while fetching details for %s",
                                  ex, tbName.getNameAsString());
                              return null;
                            }
                            throw ex;
                          }
                        })
                    .filter(Objects::nonNull)
                    .collect(Collectors.toList()));
```

igorbernstein2: That's even better!

rahulKQL: Thanks for the quick response, I have updated the code. Please have a look.

igorbernstein2: I don't think you need to log anything...no reason to panic the user. But otherwise  LGTM

PR #2232, opened 2019-09-04
Updating GCJ to get new Batcher API related classes

rahulKQL:


rahulKQL:
As discussed, I would update the GCJ version to `0.108` tomorrow (after being released today).

PR #2233, opened 2019-09-04
Small refactoring on core & hbase module

rahulKQL:
- Depreacated `BigtableDataClien#getDataClient` which is replaced by `BigtableSession#getDataClientWrapper` which has implementation for both (existing as well GCJ client).
- Removed only usage of BigtableDataClient in `TestAppProfile`.
- Removed unused `HBaseRequestAdapter#getTableNameString`.
- Removed package qualifier from `BigtableSession` & `BigtableDataGCJClient` javadocs.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -216,8 +216,8 @@ public class BigtableSession implements Closeable {
   /**
    * Constructor for BigtableSession.
    *
-   * @param opts a {@link com.google.cloud.bigtable.config.BigtableOptions} object.
-   * @throws java.io.IOException if any.
+   * @param opts a {@link BigtableOptions} object.
+   * @throws IOException if any.
    */
   public BigtableSession(BigtableOptions opts) throws IOException {
     this.options = opts;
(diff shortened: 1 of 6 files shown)


bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/TestAppProfile.java
```diff
@@ -145,17 +145,6 @@ public void testMutateRow() throws Exception {
     Assert.assertEquals(req2.getAppProfileId(), "my-app-profile");
   }
 
-  @Test
-  public void testMutateRows() throws Exception {
```
kolea2: instead of deleting, should we replace the test with the non deprecated method `getDataClientWrapper()`?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -411,9 +413,11 @@ public IBigtableDataClient getDataClientWrapper() {
   /**
    * createBulkMutation.
    *
-   * @param tableName a {@link com.google.cloud.bigtable.grpc.BigtableTableName} object.
-   * @return a {@link com.google.cloud.bigtable.grpc.async.BulkMutation} object.
+   * @deprecated Please use {@link #createBulkMutationWrapper(BigtableTableName)}.
+   * @param tableName a {@link BigtableTableName} object.
+   * @return a {@link BulkMutation} object.
    */
+  @Deprecated
```
igorbernstein2: I don't think we can deprecate this.

bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/TestAppProfile.java
igorbernstein2: We should keep both apis. Please undeprecate the old nonwrapped clients  ie getTableAdminClient and please mark all of the wrapper clients as internal api. They should be used solely by the hbase layer

PR #2234, opened 2019-09-04
Updated HBase BulkMutation to use new Batcher API through GCJ client

rahulKQL:
This change will enable `BigtableBufferMutator` to use GCJ's `Batcher<>` class through IBulkMutation when user set `useGCJClient`/`google.bigtable.use.gcj.client` flag.

### Summary of changes it contians: 
 - Had to update `IBigtableDataClient#createBulkMutationBatcher(tableID)` to accept tableID which is need for `Batcher` creation.
 - Updated `IBulkMutation` to adapt against `RowMutationEntry`.
 - Now `BulkMutationGCJClient` uses `Batcher<>` to perform batching.
 - Updated `HBaseRequestAdapter` to adapt against `RowMutationEntry` & also incorporate `RowMutationEntry#createUnsafe()`. 
 - Fixed Junit for the Batching related changes and added test case in IT test class `TestBufferedMutator`.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/core/IBigtableDataClient.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/core/IBigtableDataClient.java
@@ -71,7 +71,7 @@ public interface IBigtableDataClient {
   ApiFuture<Row> readModifyWriteRowAsync(ReadModifyWriteRow readModifyWriteRow);
 
   /** Creates {@link IBulkMutation} batcher. */
-  IBulkMutation createBulkMutationBatcher();
+  IBulkMutation createBulkMutationBatcher(String tableId);
 
   /**
    * Mutate a row atomically dependent on a precondition.
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/core/IBulkMutation.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/core/IBulkMutation.java
@@ -17,11 +17,12 @@ package com.google.cloud.bigtable.core;
 
 import com.google.api.core.ApiFuture;
 import com.google.api.core.InternalApi;
-import com.google.cloud.bigtable.data.v2.models.RowMutation;
+import com.google.cloud.bigtable.data.v2.models.RowMutationEntry;
 import java.io.IOException;
 
 /**
- * Interface to support batching multiple {@link RowMutation} request into a single grpc request.
+ * Interface to support batching multiple {@link RowMutationEntry} request into a single grpc *
+ * request.
  *
  * <p>For internal use only - public for technical reasons.
  */
@@ -29,18 +30,18 @@ import java.io.IOException;
 public interface IBulkMutation extends AutoCloseable {
 
   /**
-   * Adds a {@link RowMutation} to the underlying IBulkMutation mechanism.
+   * Adds a {@link RowMutationEntry} to the underlying IBulkMutation mechanism.
    *
-   * @param rowMutation The {@link RowMutation} to add.
+   * @param rowMutation The RowMutationEntry which holds row mutation details.
    * @return a {@link ApiFuture} of type {@link Void} will be set when request is successful
    *     otherwise exception will be thrown.
    */
-  ApiFuture<Void> add(RowMutation rowMutation);
+  ApiFuture<Void> add(RowMutationEntry rowMutation);
 
-  /** Sends any outstanding {@link RowMutation}s, present in the current batch. */
+  /** Sends any outstanding entry, present in the current batch but doesn't wait for response. */
   void sendUnsent();
 
-  /** Sends any outstanding {@link RowMutation} and wait until all requests are complete. */
+  /** Sends any outstanding RowMutationEntry and blocks until all requests are complete. */
   void flush() throws InterruptedException;
 
   /** Closes this bulk Mutation and prevents from mutating any more elements */
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGCJClient.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGCJClient.java
@@ -73,8 +73,8 @@ public class BigtableDataGCJClient implements IBigtableDataClient, AutoCloseable
   }
 
   @Override
-  public IBulkMutation createBulkMutationBatcher() {
-    return new BulkMutationGCJClient(delegate.newBulkMutationBatcher());
+  public IBulkMutation createBulkMutationBatcher(String tableId) {
+    return new BulkMutationGCJClient(delegate.newBulkMutationBatcher(tableId));
   }
 
   @Override
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutationGCJClient.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutationGCJClient.java
@@ -17,56 +17,52 @@ package com.google.cloud.bigtable.grpc.async;
 
 import com.google.api.core.ApiFuture;
 import com.google.api.core.InternalApi;
-import com.google.cloud.bigtable.config.Logger;
+import com.google.api.gax.batching.Batcher;
 import com.google.cloud.bigtable.core.IBulkMutation;
-import com.google.cloud.bigtable.data.v2.models.BulkMutationBatcher;
-import com.google.cloud.bigtable.data.v2.models.BulkMutationBatcher.BulkMutationFailure;
-import com.google.cloud.bigtable.data.v2.models.RowMutation;
-import com.google.cloud.bigtable.util.ApiFutureUtil;
+import com.google.cloud.bigtable.data.v2.models.RowMutationEntry;
 import com.google.common.base.Preconditions;
 import java.io.IOException;
-import java.util.concurrent.TimeoutException;
 
 /**
- * This class is meant to replicate existing {@link BulkMutation} while translating calls to
- * Google-Cloud-Java's {@link BulkMutationBatcher} api.
+ * This class is meant to replicate existing {@link BulkMutation} while translating calls to *
+ * Google-Cloud-Java's {@link Batcher} api.
  *
  * <p>For internal use only - public for technical reasons.
  */
 @InternalApi("For internal usage only")
 public class BulkMutationGCJClient implements IBulkMutation {
 
-  private static Logger LOG = new Logger(BulkMutationGCJClient.class);
+  private final Batcher<RowMutationEntry, Void> bulkMutateBatcher;
 
-  private final BulkMutationBatcher bulkMutateBatcher;
-  private final OperationAccountant operationAccountant;
-
-  public BulkMutationGCJClient(BulkMutationBatcher bulkMutateBatcher) {
+  public BulkMutationGCJClient(Batcher<RowMutationEntry, Void> bulkMutateBatcher) {
     this.bulkMutateBatcher = bulkMutateBatcher;
-    this.operationAccountant = new OperationAccountant();
   }
 
   /** {@inheritDoc} */
   @Override
-  public synchronized ApiFuture<Void> add(RowMutation rowMutation) {
+  public synchronized ApiFuture<Void> add(RowMutationEntry rowMutation) {
     Preconditions.checkNotNull(rowMutation, "mutation details cannot be null");
-    final ApiFuture<Void> response = bulkMutateBatcher.add(rowMutation);
-    operationAccountant.registerOperation(ApiFutureUtil.adapt(response));
-    return response;
+    return bulkMutateBatcher.add(rowMutation);
   }
 
   /** {@inheritDoc} */
   @Override
   public void sendUnsent() {
-    LOG.info("This operation will be implemented once the underlying API has this feature.");
+    try {
+      // TODO(rahulkql): Please use Batcher.sendOutstanding once below change is available
+      //  https://github.com/googleapis/gax-java/pull/786
+      bulkMutateBatcher.flush();
+    } catch (InterruptedException ex) {
+      throw new RuntimeException("Could not complete RPC for current Batch", ex);
+    }
   }
 
   /** {@inheritDoc} */
   @Override
   public void flush() {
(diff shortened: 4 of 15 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/core/IBigtableDataClient.java
```diff
@@ -67,7 +67,7 @@
   ApiFuture<Row> readModifyWriteRowAsync(ReadModifyWriteRow readModifyWriteRow);
 
   /** Creates {@link IBulkMutation} batcher. */
-  IBulkMutation createBulkMutationBatcher();
```
igorbernstein2: Please mark this as InternalOnly

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/core/IBulkMutation.java
```diff
@@ -16,28 +16,34 @@
...
 import com.google.api.core.ApiFuture;
-import com.google.cloud.bigtable.data.v2.models.RowMutation;
+import com.google.cloud.bigtable.data.v2.models.RowMutationEntry;
 
 /**
- * Interface to support batching multiple {@link RowMutation} request into a single grpc request.
+ * Interface to support batching multiple {@link RowMutationEntry} request into a single grpc
+ * request.
  */
 public interface IBulkMutation {
```
igorbernstein2: Please mark this class as InternalOnly

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableDataGCJClient.java
```diff
@@ -69,8 +69,8 @@ public Row readModifyWriteRow(ReadModifyWriteRow readModifyWriteRow) {
   }
 
   @Override
```
igorbernstein2: Please mark this entire class as InternalOnly

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutationGCJClient.java
```diff
@@ -16,35 +16,36 @@
...
+import com.google.cloud.bigtable.data.v2.models.RowMutationEntry;
 import com.google.cloud.bigtable.util.ApiFutureUtil;
 import com.google.common.base.Preconditions;
 
 /**
  * This class is meant to replicate existing {@link BulkMutation} while translating calls to
- * Google-Cloud-Java's {@link BulkMutationBatcher} api.
+ * Google-Cloud-Java's {@link Batcher} api.
  */
 public class BulkMutationGCJClient implements IBulkMutation {
```
igorbernstein2: InternalOnly

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutationGCJClient.java
```diff
@@ -16,35 +16,36 @@
...
 
   /** {@inheritDoc} */
   @Override
-  public synchronized ApiFuture<Void> add(RowMutation rowMutation) {
+  public synchronized ApiFuture<Void> add(RowMutationEntry rowMutation) {
     Preconditions.checkNotNull(rowMutation, "mutation details cannot be null");
     final ApiFuture<Void> response = bulkMutateBatcher.add(rowMutation);
+    // TODO(rahulkql): Still using operationAccount to give user isFlush() functionality. which
+    //  is not present in Batcher.
     operationAccountant.registerOperation(ApiFutureUtil.adapt(response));
```
igorbernstein2: Please avoid using operationAccountant. It seems like this functionality is only used for the jvm shutdown hook to make sure there are no outstanding batches. We have this functionality in gcj already. Would it be possible move the shutdown hook to BulkMutationWrapper?

rahulKQL: By shutdown hook did you meant [this](https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutationBatcher.java#L72-L84)? 

Also, I believe we don't need to wait till all the elements are finished as `Batcher#close` is going to flush it out. Please let me know If I am missing something here.

To provide `BulkMutationGCJClient#isFlush()` functionality I have used an AtomicLong#compare(0, 0), which I think should work. Please let me know if there could be a better approach?

bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/async/TestBulkMutationGCJClient.java
```diff
@@ -17,114 +17,125 @@
...
+            new Answer() {
+              @Override
+              public Object answer(InvocationOnMock invocationOnMock) {
+                future.set(null);
+                future2.set(null);
+                return null;
+              }
+            })
+        .when(batcher)
+        .flush();
```
rahulKQL: Should we make actuals calls to verify this OR this mock test would suffice?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutationGCJClient.java
igorbernstein2: If I'm not mistaken, the only place that needs isFlushed() is:

https://github.com/googleapis/java-bigtable-hbase/blob/c4e970d347746f11aff082d3e87787b7e74eb59f/bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java#L59-L77

And since we already check for unclosed batchers in gcj, I'm proposing removing the operation accountant from the gcj wrapper.

Please let me know if I'm missing something

rahulKQL: Yes, this was that is the only place where it has been used. I did remove it from `BulkMutationGCJClient` and replaced it with AtomicLong check. Please let me know if that looks good to you.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/core/IBigtableDataClient.java
```diff
@@ -69,7 +69,7 @@
...
+  IBulkMutation createBulkMutationBatcher(String tableId);
```
kolea2: we should be careful here as this is breaking. Can we deprecate existing functionality and use this and an overloaded method for now?

rahulKQL: To provide backward compatibility for existing flow, we might need to add an author `IBulkMutation` class (let's say `BulkMutationGCJClientV2`). As the existing wrapper expects `BulkMutationBatcher` and in this change, we are migrating it to `Batcher<>`.
https://github.com/googleapis/java-bigtable-hbase/blob/276834cde845c9a7efee81fb682f5185d5d979b9/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutationGCJClient.java#L42-L45

I think `IBigtableDataClient` suppose to be an internal interface. If anyone is using that then it would be great for them to migrate to GCJ veneer library instead of using bigtable-core client.

If you feel we should not break this internal API then we can add flow with existing methods.

igorbernstein2: Agreed. These wrappers only exist to abstract the api for the hbase client. All of them should be marked as `@InternalOnly`

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/core/IBulkMutation.java
```diff
@@ -17,28 +17,29 @@
...
-   * @param rowMutation The {@link RowMutation} to add.
+   * @param rowMutation The RowMutationEntry which holds row mutation details.
    * @return a {@link ApiFuture} of type {@link Void} will be set when request is successful
    *     otherwise exception will be thrown.
    */
-  ApiFuture<Void> add(RowMutation rowMutation);
+  ApiFuture<Void> add(RowMutationEntry rowMutation);
 
-  /** Sends any outstanding {@link RowMutation}s, present in the current batch. */
+  /** Sends asynchronously any outstanding RowMutationEntry, present in the current batch. */
```
igorbernstein2: It synchronously sends, but doesn't wait for the responses

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutationGCJClient.java
```diff
@@ -17,54 +17,48 @@
...
-    return response;
+    return bulkMutateBatcher.add(rowMutation);
   }
 
   /** {@inheritDoc} */
   @Override
   public void sendUnsent() {
-    LOG.info("This operation will be implemented once the underlying API has this feature.");
+    try {
+      bulkMutateBatcher.flush();
```
igorbernstein2: shouldnt this be non-blocking?

bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/async/TestBulkMutationGCJClient.java
igorbernstein2: I think it should suffice

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/HBaseRequestAdapter.java
```diff
@@ -145,10 +146,10 @@ public void adapt(Delete delete, MutationApi<?> mutationApi) {
    * adapt.
    *
    * @param delete a {@link Delete} object.
-   * @return a {@link RowMutation} object.
+   * @return a {@link RowMutationEntry} object.
```
igorbernstein2: Please mark this class as InternalOnly

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutationGCJClient.java
rahulKQL: Do you think I should execute it in the separate thread something like:
```java
  /** {@inheritDoc} */
  @Override
  public void sendUnsent() {
    new Thread() {
      @Override
      public void run() {
        try {

          bulkMutateBatcher.flush();
        } catch (InterruptedException ex) {
          throw new RuntimeException("Could not complete RPC for current Batch", ex);
        }
      }
    }.start();
  }

```

igorbernstein2: I just meant you should use sendOutstanding() instead of flush. Unless I'm mistaken this method is used for async operations in hbase that return a future and we just needed something start the request

rahulKQL:
Thanks for detailed instruction, attaching these details for information purpose.

After running the stress test, these are my observations

| No. of rows/thread | Bigtable-core | GCJ-Client |
| ------------------ | ------------- | ---------- |
|     100_000        |      52_450   |    53_787  |
|    10*100_000      |     196_777   |    91_832  |
|    25*100_000      |     432_179   |   190_677  |
|    40*100_000      |     677_109   |   288_978  |
|    50*100_000      |     858_070   |   363_713  | 

Final_batching_chart

sduskis:
@rahulKQL, just to clarify, are you saying that GCJ is 2X more performant than Bigtable-core?

rahulKQL:
It certainly seems 😃.

Have just added the test case I ran for the observation. @sduskis Would you please take a look?

Also, adding the script log for the case of 2500K & 4000K(forgot to copy the logs for 50 as I ran that first).
[row_40_100_000.log](https://github.com/googleapis/java-bigtable-hbase/files/3633014/row_40_100_000.log)
[row_25_100_000.log](https://github.com/googleapis/java-bigtable-hbase/files/3633015/row_25_100_000.log)

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutationGCJClient.java
igorbernstein2: ping

igorbernstein2:
Thats good news. Please update the sendOutstanding issue and we can merge this

rahulKQL:
@igorbernstein2 Thanks for the clarification, I have removed the stress test from this PR and added a TODO.

Please have a look and let me know if I can merge this.

igorbernstein2:
LGTM, but we need to wait for @kolea2 to approve

rahulKQL:
@kolea2 If you get time, please take a look.

igorbernstein2:
Hi Rahul,

We discussed this last week and decided that we need exercise a bit more caution before merging this. The current plan is:
1. Annotate all classes & interfaces in this repo with appropriate gax visibility annotations (`@InternalOnly`, `@InternalExtensionOnly, etc)
2. create a separate branch `v2.x`
3. merge this PR into that branch
4. finish the migration to gcj
5. release 2.0-alpha with a runtime flag to switch between implementations
6. release 2.0-beta with gcj only

igorbernstein2:
I just opened PRs for the annotations: #2245, #2246, #2247. Once we merge those, we will create a branch where we can merge this PR

igorbernstein2:
can you re-target  this PR to https://github.com/googleapis/java-bigtable-hbase/tree/bigtable-client-2.x?

rahulKQL:
@igorbernstein2 Thanks for the notifying me, Have rebased and updated the target branch to `bigtable-client-2.x`.

PR #2235, opened 2019-09-04
Marked GCJ wrappers InternalOnly

rahulKQL:
All wrapper classes which wrap GCJ client is now be marked with `@InternalApi`, Also removed deprecation from `BigtableSession#getTableAdminClient`.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/core/IBigtableDataClient.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/core/IBigtableDataClient.java
@@ -16,6 +16,7 @@
 package com.google.cloud.bigtable.core;
 
 import com.google.api.core.ApiFuture;
+import com.google.api.core.InternalApi;
 import com.google.cloud.bigtable.data.v2.models.ConditionalRowMutation;
 import com.google.cloud.bigtable.data.v2.models.KeyOffset;
 import com.google.cloud.bigtable.data.v2.models.Query;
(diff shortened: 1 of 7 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/core/IBigtableDataClient.java
```diff
@@ -67,6 +68,7 @@
   ApiFuture<Row> readModifyWriteRowAsync(ReadModifyWriteRow readModifyWriteRow);
 
   /** Creates {@link IBulkMutation} batcher. */
```
igorbernstein2: This whole interface should be InternalOnly

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -478,6 +479,7 @@ public synchronized BigtableTableAdminClient getTableAdminClient() throws IOExce
    * @return a {@link BigtableTableAdminClientWrapper} object.
    * @throws java.io.IOException if any.
    */
+  @InternalApi("For internal usage only")
   public synchronized IBigtableTableAdminClient getTableAdminClientWrapper() throws IOException {
```
igorbernstein2: The interface IBigtableTableAdminClient should also be InternalOnly

rahulKQL: Thanks for the feedback, I have marked `IBigtableTableAdminClient` and `IBigtableDataClient` as `@InternalApi`. Please have a look.

PR #2248, opened 2019-10-04
Adding Channel Caching functionality to GCJ Client + tests

j-clap:


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -707,4 +742,15 @@ public synchronized void close() throws IOException {
...
+  /**
+   * Getter for the field <code>gcjClientContext</code>.
+   *
+   * @return a HashMap of Connection Name (String) to {@link ClientContext} object.
+   */
+  public HashMap<String, ClientContext> getGCJClientContext() {
+    Preconditions.checkState(
+        options.useGCJClient(), "Client Context is only available for GCJ Client.");
+    return gcjClientContext;
+  }
```
rahulKQL: I m a little confused about these `clientContexts`, Could you please describe me where/how are we planning to use this?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -707,4 +742,15 @@ public synchronized void close() throws IOException {
   public BigtableOptions getOptions() {
     return this.options;
   }
+
+  /**
+   * Getter for the field <code>gcjClientContext</code>.
+   *
+   * @return a HashMap of Connection Name (String) to {@link ClientContext} object.
+   */
+  public HashMap<String, ClientContext> getGCJClientContext() {
```
igorbernstein2: Does this need to be public? If so, can you mark this as `@InternalApi`. See this PR for an example:#2245

igorbernstein2: I think you should just inline the getter. I don't think it brings much value here

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -240,6 +246,35 @@ public BigtableSession(BigtableOptions opts) throws IOException {
...
+        builder
+            .stubSettings()
+            .setExecutorProvider(
+                FixedExecutorProvider.create(
+                    gcjClientContext.get(options.getDataHost()).getExecutor()))
+            .setTransportChannelProvider(
+                FixedTransportChannelProvider.create(
+                    Objects.requireNonNull(
+                        gcjClientContext.get(options.getDataHost()).getTransportChannel())))
+            .build();
```
igorbernstein2: InstantiatingChannelProvider uses a reference to the credentials when being constructed:
https://github.com/googleapis/gax-java/blob/master/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L235

So I think you have to reuse the cached credentials as well:
```
.setCredentialsProvider(FixedCredentialsProvider.create(cachedClientContext.getCredentials())
```

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -240,6 +246,35 @@ public BigtableSession(BigtableOptions opts) throws IOException {
...
+        }
+
+        BigtableDataSettings.Builder builder = dataSettings.toBuilder();
+
+        // Add the executor and transport channel to the settings/options
+        builder
+            .stubSettings()
+            .setExecutorProvider(
+                FixedExecutorProvider.create(
+                    gcjClientContext.get(options.getDataHost()).getExecutor()))
```
igorbernstein2: Can you extract the fetching of the gcjContext?

```

ClientContext cachedCtx = cachedClientContexts.get(options.getDataHost());

...
setExecutorProvider(FixedExecutorProvider.create(cachedCtx.getExecutor())

```

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -99,6 +103,8 @@
   private static final Logger LOG = new Logger(BigtableSession.class);
   // TODO: Consider caching channel pools per instance.
   private static ManagedChannel cachedDataChannelPool;
+  // Hash map containing cached connections to specific destination hosts for GCJ client
+  private static HashMap<String, ClientContext> gcjClientContext = new HashMap<>();
```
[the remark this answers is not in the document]
igorbernstein2: Also please rename the variable to something like `cachedGcjClientContexts` to indicate that it contains many contexts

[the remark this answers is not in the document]
igorbernstein2: I don't want to block this PR on this, but in the future we should make the ClientContexts reference counted, so that when BigtableSession.close() is called we can start cleaning them up. Would you mind adding a `// TODO: ` comment?

igorbernstein2:
Thanks for putting this together!

PR #2252, opened 2019-10-10
Adding benchmark test case to compare Bigtable-core & GCJ client

rahulKQL:
This change adds 4 benchmark test case around Bigtable-hbase client.
- PointRead just reads a single row from Bigtable, BulkRead operation reads ~1 gig of data.
- Similarly pointWrite writes a single row and BulkWrites to ~1 gig of data to "benchmark_writes" table.
- Read operations performs of persistent table named according to it's parameters(If table not found then creates and populates data using BenchmarkDataUtils).
- Writes operations deletes it and then recreate it.


bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/Benchmarks.java
```diff
@@ -0,0 +1,249 @@
...
+  private String projectId;
+
+  @Param("")
+  private String instanceId;
+
+  @Param("100")
+  private int cellSize;
+
+  @Param("10")
+  private int cellsPerRow;
```
igorbernstein2: Consider combining these into a String called dataShape, whose values encode the number cells and value size. This will allow us to run the benchmarks using multiple patterns

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/Benchmarks.java
```diff
@@ -0,0 +1,249 @@
...
+    table.close();
+    connection.close();
+  }
+
+  private Connection createConnection() {
+    Configuration config = BigtableConfiguration.configure(projectId, instanceId);
+    config.set(BIGTABLE_USE_GCJ_CLIENT, String.valueOf(useGcj));
+    config.set(BigtableOptionsFactory.BIGTABLE_USE_BATCH, String.valueOf(useBatch));
+    config.set(
+        BigtableOptionsFactory.BIGTABLE_USE_CACHED_DATA_CHANNEL_POOL, String.valueOf(useBatch));
```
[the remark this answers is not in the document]
rahulKQL: Yes, It does would remove those

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/Benchmarks.java
rahulKQL: Thanks for the suggestion, I'll update this accordingly.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/Benchmarks.java
```diff
@@ -0,0 +1,249 @@
...
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.infra.Blackhole;
+
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@Fork(value = 1)
+@Warmup(iterations = 3)
+@Measurement(iterations = 6)
```
kolea2: should we do more than 6?

igorbernstein2: We need to figure these out experimentally: run the tests and see how many iterations it takes for the results to stabilize. We can play with these parameters once the benchmarks are updated

igorbernstein2:
@rahulKQL can you give a status update on this? Is it ready for me to play with?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BigtableBenchmark.java
```diff
@@ -0,0 +1,211 @@
...
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.infra.Blackhole;
+
+@BenchmarkMode(Mode.AverageTime)
```
igorbernstein2: Please add doc and make sure to mention what tables are being used for what

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BigtableBenchmark.java
```diff
@@ -0,0 +1,211 @@
...
+          put.addColumn(
+              COL_FAMILY,
+              Bytes.toBytes("qualifier-" + String.format(SUFFIX_FMT, cellInd)),
+              currentTimestamp,
+              cellValue);
+        }
+        putBuilder.add(put);
+
+        // Sending a lot of rows in a single batch randomly OOM exception
+        if (sendBatchInd % ROWS_PER_BATCH == 0 || rowInd == (jobConfig.totalRows - 1)) {
```
igorbernstein2: This benchmark should only send a single batch. Let JMH call this multiple times instead

igorbernstein2: This benchmark should only send a single batch. Let JMH call this multiple times instead

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BigtableBenchmark.java
```diff
@@ -0,0 +1,211 @@
...
+
+  // TODO(rahulkql): yet to be fully tested.
+  @Benchmark
+  public void bulkWrite() throws IOException {
+    try (BufferedMutator bufferedMutator =
+        connection.getBufferedMutator(TableName.valueOf(BENCHMARK_WRITE_TABLE))) {
+
+      String rowPrefix = getRowPrefix(jobConfig.cellsPerRow, jobConfig.cellSize);
+      byte[] cellValue = getRandomBytes(jobConfig.cellSize);
+      ImmutableList.Builder<Put> putBuilder = ImmutableList.builderWithExpectedSize(ROWS_PER_BATCH);
```
igorbernstein2: Why do we need an intermediate list? can't we just add directly to the buffered mutator?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BigtableBenchmark.java
```diff
@@ -0,0 +1,211 @@
...
+              cellValue);
+        }
+        putBuilder.add(put);
+
+        // Sending a lot of rows in a single batch randomly OOM exception
+        if (sendBatchInd % ROWS_PER_BATCH == 0 || rowInd == (jobConfig.totalRows - 1)) {
+          bufferedMutator.mutate(putBuilder.build());
+
+          // Does this makes sense?
+          bufferedMutator.flush();
```
igorbernstein2: I dont think so

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BigtableBenchmark.java
```diff
@@ -0,0 +1,211 @@
...
+
+          putBuilder = ImmutableList.builderWithExpectedSize(ROWS_PER_BATCH);
+        }
+      }
+    }
+  }
+
+  public static class BenchmarkJobConfig {
```
igorbernstein2: nit: private

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BigtableBenchmark.java
```diff
@@ -0,0 +1,211 @@
...
+
+  @Param("")
+  private String instanceId;
+
+  @Param({
+    "cellsPerRow/10/cellSize/100",
+    "cellsPerRow/1/cellSize/1024",
+    "cellsPerRow/100/cellSize/1"
+  })
+  private String benchmarkConfig;
```
igorbernstein2: I think a better name would `rowShape`

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BigtableBenchmark.java
```diff
@@ -0,0 +1,211 @@
...
+
+  @Param({"true", "false"})
+  private boolean useGcj;
+
+  @Param("false")
+  private boolean useBatch;
+
+  private Connection connection;
+
+  private BenchmarkJobConfig jobConfig;
```
igorbernstein2: I think this should be named `RowShapeParams rowShapeParams`

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BigtableBenchmark.java
```diff
@@ -0,0 +1,211 @@
...
+    connection = BenchmarkDataUtils.createConnection(projectId, instanceId, useBatch, useGcj);
+    jobConfig = new BenchmarkJobConfig(benchmarkConfig);
+
+    createAndFillTable(connection, jobConfig);
+    createTableToWrite(connection, TableName.valueOf(BENCHMARK_WRITE_TABLE));
+  }
+
+  @Benchmark
+  public void pointRead(Blackhole blackhole) throws IOException {
+    Table table = connection.getTable(jobConfig.tableName);
```
igorbernstein2: I think you can pull this out to setup

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BigtableBenchmark.java
```diff
@@ -0,0 +1,211 @@
...
+    String rowPrefix = getRowPrefix(jobConfig.cellsPerRow, jobConfig.cellSize);
```
igorbernstein2: I don't think the row key needs to be related to the row shape

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BigtableBenchmark.java
```diff
@@ -0,0 +1,211 @@
...
+    Result result = table.get(new Get(Bytes.toBytes(rowPrefix + String.format(SUFFIX_FMT, 1))));
+
+    for (Cell cell : result.listCells()) {
+      blackhole.consume(cell);
+    }
+  }
+
+  @Benchmark
+  public void bulkScans(Blackhole blackhole) throws IOException {
+    String rowPrefix = getRowPrefix(jobConfig.cellsPerRow, jobConfig.cellSize);
```
igorbernstein2: I don't think row keys need to be related to shape...each dataShape should have its own table

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BigtableBenchmark.java
```diff
@@ -0,0 +1,211 @@
...
+      for (Cell cell : result.rawCells()) {
+        blackhole.consume(cell);
+      }
+    }
+  }
+
+  // TODO(rahulkql): yet to be fully tested.
+  @Benchmark
+  public void pointWrite() throws IOException {
+    Table table = connection.getTable(TableName.valueOf(BENCHMARK_WRITE_TABLE));
```
igorbernstein2: pull this out to setup

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BenchmarkDataUtils.java
```diff
@@ -0,0 +1,130 @@
...
+    try (Admin admin = connection.getAdmin()) {
+
+      if (admin.tableExists(writeToTableName)) {
+        // Deleting the existing table to create a fresh table to write to.
+        admin.deleteTable(writeToTableName);
+      }
+      admin.createTable(
+          new HTableDescriptor(writeToTableName)
+              .addFamily(new HColumnDescriptor(COL_FAMILY).setMaxVersions(1)));
+    }
```
rahulKQL: @igorbernstein2 Do you think `Admin#truncateTable `of an existing table is better than deleting and creating `benchmark_write` table?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BigtableBenchmark.java
```diff
@@ -0,0 +1,211 @@
...
+        connection.getBufferedMutator(TableName.valueOf(BENCHMARK_WRITE_TABLE))) {
+
+      String rowPrefix = getRowPrefix(jobConfig.cellsPerRow, jobConfig.cellSize);
+      byte[] cellValue = getRandomBytes(jobConfig.cellSize);
+      ImmutableList.Builder<Put> putBuilder = ImmutableList.builderWithExpectedSize(ROWS_PER_BATCH);
+      long currentTimestamp = System.currentTimeMillis();
+
+      for (int rowInd = 0, sendBatchInd = 0;
+          rowInd < jobConfig.totalRows;
+          rowInd++, sendBatchInd++) {
```
igorbernstein2: I think this should only send 100 rows

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BigtableBenchmark.java
```diff
@@ -0,0 +1,211 @@
...
+
+        // zero padded row-key
+        Put put = new Put(Bytes.toBytes(rowPrefix + String.format(SUFFIX_FMT, rowInd)));
```
igorbernstein2: Also the prefix for each batch should be random

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BigtableBenchmark.java
igorbernstein2: Also I think RowShapeParams would be a better name

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/resources/log4j.properties
```diff
@@ -0,0 +1,6 @@
+log4j.rootLogger=DEBUG, stdout
```
igorbernstein2: I think INFO would be more appropriate

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BenchmarkDataUtils.java
igorbernstein2: I don't think we need to bother with truncation. Just set a GC Policy to keep only the latest version and constrain the key space to < 1 million

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BigtableBenchmark.java
rahulKQL: Actually i was flushing after 10K ross that's why I created an intermediate builder. Now removed it as we are only sending 100 rows and JMH is iterating it multiple times.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BenchmarkDataUtils.java
rahulKQL: I tried to add a constraint to it, but it seems we need to include `hbase-server` dependencies to get `BaseConstraint`. I was planning to follow [this(hbase constraints docs)](https://hbase.apache.org/devapidocs/org/apache/hadoop/hbase/constraint/package-summary.html#usage) but thought to confirm with you before including these dependencies.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BigtableBenchmark.java
```diff
@@ -0,0 +1,211 @@
...
+    table.put(put);
+  }
+
+  // TODO(rahulkql): yet to be fully tested.
+  @Benchmark
+  public void bulkWrite() throws IOException {
+    try (BufferedMutator bufferedMutator =
+        connection.getBufferedMutator(TableName.valueOf(BENCHMARK_WRITE_TABLE))) {
+
+      String rowPrefix = getRowPrefix(jobConfig.cellsPerRow, jobConfig.cellSize);
```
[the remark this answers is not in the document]
rahulKQL: by saying randomize prefix, do you mean to say randomizes between "key-10-cells-100-bytes-0000001" to "key-10-cells-100-bytes-9999999"? or should we implement complete random prefix on every run?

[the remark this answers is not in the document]
igorbernstein2: Constrained random, I would like to keep the data size stable. So random keys within the existing sequential keys

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BenchmarkDataUtils.java
igorbernstein2: I'm not sure how BaseConstraint is relevant here. I just meant the mutation benchmarks should do something like:
`String.format("some-const-prefix-%d", random.nextInt(1_000_000))`

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BenchmarkSetupUtils.java
```diff
@@ -0,0 +1,148 @@
...
+          put.addColumn(
+              COL_FAMILY,
+              Bytes.toBytes("qualifier-" + String.format("%06d", cellInd)),
+              timestamp,
+              cellValue);
+        }
+        bufferedMutator.mutate(put);
+
+        if (rowInd % 1000 == 0) {
+          bufferedMutator.flush();
```
igorbernstein2: I dont think this is necessary...let buffered mutator do its own flushing

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/src/main/java/com/google/cloud/bigtable/hbase1_x/BigtableBenchmark.java
```diff
@@ -110,8 +114,8 @@ public void setUp() throws IOException {
   /** Reads randomly one row and iterator over the cells. */
   @Benchmark
   public void pointRead(Blackhole blackhole) throws IOException {
-    int randomRowNum = new Random().nextInt(100_000);
-    byte[] rowKey = Bytes.toBytes(ROW_PREFIX + String.format("%010d", randomRowNum));
+    int randomRowNum = new Random().nextInt(10_000);
```
igorbernstein2: Please instantiate Random in setUp and reuse it

PR #2260, opened 2019-10-30
Fix Travis build status and landing page [skip ci]

mbrukman:
Update the name of the repo to match current state, which fixes both issues


kolea2:
Thank you!

PR #2267, opened 2019-11-06
[Part-2] IT test cases of Admin API for HBase 1 and HBase 2

rahulKQL:
This change includes addresses the gap between this Bigtable client and HBase API for the below methods:

- Admin#listTableNames(Pattern)
- Admin#getTableDescriptorsByTableName(List)(It is being called by below operations)
  - Admin#getTableDescriptors(List)
  - Admin#listTables(Pattern)
  - Admin#enableTables(Pattern)
  - Admin#disableTables(Pattern)
  - Admin#deleteTables(Pattern)
- Admin#isTableDisabled(TableName)
  - Admin#isTableEnabled(TableName)

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestListTables.java
+++ b/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestListTables.java
@@ -15,6 +15,10 @@
  */
 package com.google.cloud.bigtable.hbase;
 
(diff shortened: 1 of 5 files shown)


igorbernstein2:
Ideally we would close the gap, what kind of differences have you seen?

rahulKQL:
The IT test case present in this PR are currently annotated with `@Ignore` due to all of these methods are calling the `Admin#getTableDescritor(TableName)` which translates into `IBigtableTableAdminClient#getTable`.

It seems the getTable's still has this bug which causing each test case to fail with `TableNotFoundException` or `FAILED_PRECONDITION`.

PR #2271, opened 2019-11-18
feat: enable directpath

igorbernstein2:
Add experimental support for DirectPath.

Whitelisted customers can enable the feature by setting the environment variable `GOOGLE_CLOUD_ENABLE_DIRECT_PATH=bigtable`

This PR also contains an inactive integration test that ensures DirectPath fallback works. This test only run when that environment variable is set. Currently there is no execution that has that env var set. Once a kokoro environment is setup for DirectPath, we can add a profile that will enable this test.

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
@@ -15,9 +15,6 @@
  */
 package com.google.cloud.bigtable.hbase;
 
-import static com.google.cloud.bigtable.config.BigtableOptions.BIGTABLE_ADMIN_HOST_DEFAULT;
-import static com.google.cloud.bigtable.config.BigtableOptions.BIGTABLE_DATA_HOST_DEFAULT;
-import static com.google.cloud.bigtable.config.BigtableOptions.BIGTABLE_PORT_DEFAULT;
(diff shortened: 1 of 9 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
```diff
@@ -15,9 +15,6 @@
  */
 package com.google.cloud.bigtable.hbase;
 
-import static com.google.cloud.bigtable.config.BigtableOptions.BIGTABLE_ADMIN_HOST_DEFAULT;
```
kolea2: for my understanding - why are these getting deleted? are they getting set elsewhere already for default hosts?

igorbernstein2: bigtable-client-core already has defaults, we were double setting the defaults here. This leads to duplicating the logic in bigtable-client-core and bigtable-hbase. WIth directpath, this logic is more complex than a simple const. So this PR only overrides the default if it's actually specified.

kolea2: SGTM. Can we add a test asserting the overrides are getting set properly?

igorbernstein2: done

PR #2285, opened 2019-11-21
test: Fix for Gaps b/w HBase and Bigtable Admin API

rahulKQL:
Reopened the PR from #2267.
This change addresses the gap between this Bigtable & HBase Data API.

This also introduces a private method `Admin.getTableDescriptorIgnoreFailure`, which returns TableDescriptor[]. It is to address the `TableNotFoundException` & `FAILED_PRECONDITION` failure while fetching Table details of the list(HBase also throws TableNotFoundException when table not found).

This change updates below methods:
- listTables()
- deleteTables()
- enableTables()
- disableTables()
- isTableEnabled()
- isTableDisabled()
- getTableDescriptors()
- getTableDescriptorsByTableName()

Note: Have run IT test with both clients for H1 & H2 implementation.

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestListTables.java
+++ b/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestListTables.java
@@ -27,7 +27,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
 import java.util.regex.Pattern;
-import org.apache.hadoop.hbase.HTableDescriptor;
 import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.TableNotFoundException;
 import org.apache.hadoop.hbase.client.Admin;
@@ -193,9 +192,8 @@ public abstract class AbstractTestListTables extends AbstractTest {
     try (Admin admin = getConnection().getAdmin()) {
       sharedTestEnv.createTable(sharedTestEnv.newTestTableName());
       Exception actualError = null;
-      try {
-        HTableDescriptor[] descriptors = admin.listTables((Pattern) null);
-        assertTrue(descriptors.length > 0);
+      try {;
+        assertTrue(admin.listTables((Pattern) null).length > 0);
       } catch (Exception e) {
         actualError = e;
       }
@@ -207,6 +205,7 @@ public abstract class AbstractTestListTables extends AbstractTest {
         actualError = e;
       }
       assertNotNull(actualError);
+      assertTrue(actualError instanceof NullPointerException);
       actualError = null;
 
       try {
@@ -237,6 +236,7 @@ public abstract class AbstractTestListTables extends AbstractTest {
         actualError = e;
       }
       assertNotNull(actualError);
+      assertTrue(actualError instanceof NullPointerException);
       actualError = null;
 
       try {
--- a/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestListTables.java
+++ b/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestListTables.java
@@ -17,7 +17,6 @@ package com.google.cloud.bigtable.hbase;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
 import java.io.IOException;
@@ -191,14 +190,10 @@ public abstract class AbstractTestListTables extends AbstractTest {
   public void testListTablesWithEmptyElement() throws IOException {
     try (Admin admin = getConnection().getAdmin()) {
       sharedTestEnv.createTable(sharedTestEnv.newTestTableName());
-      Exception actualError = null;
-      try {;
-        assertTrue(admin.listTables((Pattern) null).length > 0);
-      } catch (Exception e) {
-        actualError = e;
-      }
-      assertNull(actualError);
 
+      assertTrue(admin.listTables((Pattern) null).length > 0);
+
+      Exception actualError = null;
       try {
         admin.listTables((String) null);
       } catch (Exception e) {
@@ -206,14 +201,8 @@ public abstract class AbstractTestListTables extends AbstractTest {
       }
       assertNotNull(actualError);
       assertTrue(actualError instanceof NullPointerException);
-      actualError = null;
 
-      try {
-        assertEquals(0, admin.listTables("").length);
-      } catch (Exception e) {
-        actualError = e;
-      }
-      assertNull(actualError);
+      assertEquals(0, admin.listTables("").length);
     }
   }
 
@@ -221,15 +210,11 @@ public abstract class AbstractTestListTables extends AbstractTest {
   public void testListTableNamesWithEmptyElement() throws IOException {
     try (Admin admin = getConnection().getAdmin()) {
       sharedTestEnv.createTable(sharedTestEnv.newTestTableName());
-      Exception actualError = null;
-      try {
-        TableName[] tableNames = admin.listTableNames((Pattern) null);
-        assertTrue(tableNames.length > 0);
-      } catch (Exception e) {
-        actualError = e;
-      }
-      assertNull(actualError);
 
+      TableName[] tableNames = admin.listTableNames((Pattern) null);
+      assertTrue(tableNames.length > 0);
+
+      Exception actualError = null;
       try {
         admin.listTableNames((String) null);
       } catch (Exception e) {
@@ -237,14 +222,8 @@ public abstract class AbstractTestListTables extends AbstractTest {
       }
       assertNotNull(actualError);
       assertTrue(actualError instanceof NullPointerException);
-      actualError = null;
 
-      try {
-        assertEquals(0, admin.listTableNames("").length);
-      } catch (Exception e) {
-        actualError = e;
-      }
-      assertNull(actualError);
+      assertEquals(0, admin.listTableNames("").length);
     }
   }
 
--- a/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestListTables.java
+++ b/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestListTables.java
@@ -15,6 +15,11 @@
  */
 package com.google.cloud.bigtable.hbase;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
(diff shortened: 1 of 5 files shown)


rahulKQL:
@igorbernstein2 If possible, Please review this as well.

bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -172,9 +173,33 @@ public boolean tableExists(String tableName) throws IOException {
     return response;
   }
 
+  // Workaround for #2016, BigtableTableAdminClient#getTable throws TableNotFoundException(due to
+  // two step details fetch) and an intermediate exception(FAILED_PRECONDITION).
+  //
+  // This method is only being called when user sends `null` or `""` string to listTable(Pattern)
+  // which in being used by deleteTables(), enableTables(), disableTables(),
+  // getTableDescriptorByTableNames().
```
igorbernstein2: I dont think explanation is quite right, I think this would be more appropriate:

```
BigtableTableAdmin#listTables doesn't include table details (like column families), so the table descriptors must be fetched into 2 phases:
1. list all of the names 
2. fetch the table details

Due to the non-atomic listing, tables in the list might disappear before fetching the details. This method will suppress those tables
```

bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -172,9 +173,33 @@ public boolean tableExists(String tableName) throws IOException {
...
+  private HTableDescriptor[] getTableDescriptorsIgnoreFailure(TableName[] tableNames) {
+    List<HTableDescriptor> descriptors = new ArrayList<>();
+    for (TableName tableName : tableNames) {
+      try {
+        descriptors.add(getTableDescriptor(tableName));
+      } catch (IOException ex) {
+        // Do we need better error message around it.
```
igorbernstein2: Please be more specific in what errors this will suppress.

bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -172,9 +173,33 @@ public boolean tableExists(String tableName) throws IOException {
...
+        LOG.warn(
```
igorbernstein2: This is an expected scenario, no need to alarm the user. Please remove the logging

bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -172,9 +173,33 @@ public boolean tableExists(String tableName) throws IOException {
...
+  }
+
   /** {@inheritDoc} */
   @Override
   public HTableDescriptor[] listTables(Pattern pattern) throws IOException {
+    if (pattern == null) {
+      return getTableDescriptorsIgnoreFailure(listTableNames());
+    }
     // NOTE: We don't have systables.
     return getTableDescriptors(listTableNames(pattern));
```
igorbernstein2: Why shouldn't this suppress the errors from the race condition?

rahulKQL: I thought to keep `getTableDescriptors()` as it is. So far no user has logged any issue for `listTables(Pattern)`. Another thought behind this was If it is valid pattern beside `".*"` then it would certainly fetch small of a number of tables.

If you think we should guard getTableDescriptors() for race condition then I can raise separate PR considering that is a Public method.

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestListTables.java
```diff
@@ -182,6 +187,67 @@ public void testNotFound() throws IOException {
     }
   }
 
+  @Test
+  public void testListTablesWithEmptyElement() throws IOException {
+    try (Admin admin = getConnection().getAdmin()) {
+      sharedTestEnv.createTable(sharedTestEnv.newTestTableName());
+      Exception actualError = null;
+      try {;
```
igorbernstein2: I think ; is a typo

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestListTables.java
```diff
@@ -182,6 +187,67 @@ public void testNotFound() throws IOException {
...
+      assertNull(actualError);
+    }
+  }
+
+  @Test
+  public void testListTableNamesWithEmptyElement() throws IOException {
+    try (Admin admin = getConnection().getAdmin()) {
+      sharedTestEnv.createTable(sharedTestEnv.newTestTableName());
+      Exception actualError = null;
+      try {
```
igorbernstein2: I don't think the try is necessary here. Also please move actualError after the assertTrue

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestListTables.java
```diff
@@ -182,6 +187,67 @@ public void testNotFound() throws IOException {
...
+      assertNotNull(actualError);
+      assertTrue(actualError instanceof NullPointerException);
+      actualError = null;
+
+      try {
+        assertEquals(0, admin.listTableNames("").length);
+      } catch (Exception e) {
+        actualError = e;
+      }
+      assertNull(actualError);
```
igorbernstein2: I don't think the try is necessary here

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/TestAdminOps.java
```diff
@@ -0,0 +1,148 @@
...
+      assertTrue(actualError instanceof NullPointerException);
+      actualError = null;
+
+      try {
+        assertEquals(0, admin.enableTables("").length);
+      } catch (Exception e) {
+        actualError = e;
+      }
+      assertNull(actualError);
+    }
```
igorbernstein2: try isnt necessary here

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/TestAdminOps.java
```diff
@@ -0,0 +1,148 @@
...
+  public void testGetTableDescriptorsByTableNameWithNullAndEmptyList() throws IOException {
+    try (Admin admin = getConnection().getAdmin()) {
+      Exception actualError = null;
+      try {
+        HTableDescriptor[] descriptor = admin.getTableDescriptorsByTableName(null);
+        assertTrue(descriptor.length > 0);
+      } catch (Exception e) {
+        actualError = e;
+      }
+      assertNull(actualError);
```
igorbernstein2: please remove try

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/TestAdminOps.java
```diff
@@ -0,0 +1,148 @@
...
+
+      try {
+        HTableDescriptor[] descriptor =
+            admin.getTableDescriptorsByTableName(ImmutableList.<TableName>of());
+        assertTrue(descriptor.length > 0);
+      } catch (Exception e) {
+        actualError = e;
+      }
+      assertNull(actualError);
```
igorbernstein2: please remove try

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/TestAdminOps.java
```diff
@@ -0,0 +1,148 @@
...
+      assertTrue(actualError instanceof NullPointerException);
+      actualError = null;
+
+      try {
+        HTableDescriptor[] descriptors = admin.getTableDescriptors(ImmutableList.<String>of());
+        assertTrue(descriptors.length > 0);
+      } catch (Exception e) {
+        actualError = e;
+      }
+      assertNull(actualError);
```
igorbernstein2: please remove unnecessary try

bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -172,9 +174,37 @@ public boolean tableExists(String tableName) throws IOException {
...
+  private HTableDescriptor[] getTableDescriptorsIgnoreFailure(TableName[] tableNames)
+      throws IOException {
+    List<HTableDescriptor> descriptors = new ArrayList<>();
+    for (TableName tableName : tableNames) {
+      try {
+        descriptors.add(getTableDescriptor(tableName));
+      } catch (IOException ex) {
+        // This will suppress TableNotFoundException and FailedPreconditionException as both of
+        // these can occur due to race condition.
+        if (!(ex instanceof TableNotFoundException
```
igorbernstein2: Please add a comment that TableNotFoundException comes from the Hbase layer and the FailedPrecondition originates from the bigtable client layer.

bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
igorbernstein2: It still vulnerable to the same race conditions. Please fix this.

getTableDescriptors(tableNames) is different in that the caller is asking for specific tables, so it should throw an error.

igorbernstein2:
ping

bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
rahulKQL: I believe by race condition you mean when listTables() is fetching a large number of TableDescriptor using getTable(). 

Should this be an acceptable(ignoring failure even if the user has provided prefix/regex):
```java
  @Override
  public HTableDescriptor[] listTables() throws IOException {
    // NOTE: We don't have systables.
    return getTableDescriptorsIgnoreFailure(listTableNames());
  }
```

rahulKQL:
This PR is ready for fresh rounds of review, PTAL

Issue #2253, opened 2019-10-15
Add integration tests for consistent null & empty string handling


igorbernstein2:
We should make sure that we have integration tests that ensure that Bigtable adapter handles nulls the same as HBase minicluster

rahulKQL:
Finished this task with #2266, #2285, #2289.

PR #2289, opened 2019-11-28
test: updates snapshot operation with sanitary checks

rahulKQL:
This change includes sanitary checks to Snapshot operation.

### What is the problem:
If an empty string is being passed on to snapshot operations(#snapshotTable, #listSnapshot) then it fails with `Unimplemented method found` exception message.
```
com.google.bigtable.repackaged.com.google.common.util.concurrent.UncheckedExecutionException: com.google.bigtable.repackaged.io.grpc.StatusRuntimeException: UNIMPLEMENTED: Method not found.
```

### Impact of this change:
Now snapshot operations would not fail with `""` anymore. Also, These methods will behave the same as the HBase API when passing an empty string or `null`.

---
Towards #2253

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestSnapshot.java
+++ b/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestSnapshot.java
@@ -20,6 +20,7 @@ import static com.google.cloud.bigtable.hbase.test_env.SharedTestEnvRule.COLUMN_
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
 
 import java.io.IOException;
 import java.util.ArrayList;
@@ -222,6 +223,7 @@ public abstract class AbstractTestSnapshot extends AbstractTest {
       actualError = ex;
     }
     assertNotNull(actualError);
+    assertTrue(actualError instanceof NullPointerException);
     actualError = null;
 
     try {
@@ -250,6 +252,7 @@ public abstract class AbstractTestSnapshot extends AbstractTest {
       actualError = ex;
     }
     assertNotNull(actualError);
+    assertTrue(actualError instanceof NullPointerException);
     actualError = null;
 
     try {
--- a/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestSnapshot.java
+++ b/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestSnapshot.java
@@ -215,7 +215,7 @@ public abstract class AbstractTestSnapshot extends AbstractTest {
   }
 
   @Test
-  public void testListSnapshotsWithNullAndEmptyString() {
+  public void testListSnapshotsWithNullAndEmptyString() throws IOException {
     Exception actualError = null;
     try {
       listSnapshotsSize((String) null);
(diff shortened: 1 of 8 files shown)


bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestSnapshot.java
```diff
@@ -190,6 +193,117 @@ public void testListTableSnapshotWithSingleArgs() throws Exception {
...
+    }
+    assertNotNull(actualError);
+    actualError = null;
+
+    try {
+      snapshot(null, tableName);
+    } catch (Exception ex) {
+      actualError = ex;
+    }
+    assertNotNull(actualError);
```
kolea2: can we make this a more specific assertion, like error message or exception type?

rahulKQL: Thanks for the review, I have updated assertions with exception-type for most of them.

Have left assertion for those methods which throws different exception with HBase1 & HBase2 implementation(eg. `Admin#snapshot(String, TableName)`)

rahulKQL:
This PR is ready for fresh rounds of review. PTAL

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestSnapshot.java
```diff
@@ -190,6 +194,119 @@ public void testListTableSnapshotWithSingleArgs() throws Exception {
...
+    try {
+      listSnapshotsSize((String) null);
+    } catch (Exception ex) {
+      actualError = ex;
+    }
+    assertNotNull(actualError);
+    assertTrue(actualError instanceof NullPointerException);
+    actualError = null;
+
+    try {
```
igorbernstein2: here and every else, please remove unnecessary try blocks

rahulKQL: Thanks for the review, I have removed the extra try-catch block. PTAL

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestSnapshot.java
```diff
@@ -190,6 +194,77 @@ public void testListTableSnapshotWithSingleArgs() throws Exception {
...
+      deleteSnapshots(null);
+
+      deleteSnapshots(Pattern.compile(""));
+
+      deleteTableSnapshots(null, null);
+
+      deleteTableSnapshots(Pattern.compile(""), Pattern.compile(""));
+    } catch (Exception ex) {
+      actualError = ex;
+    }
```
igorbernstein2: try/not necessary here

bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -837,6 +838,12 @@ public void snapshot(String snapshotName, TableName tableName)
    * @throws IOException if any.
    */
   protected Operation snapshotTable(String snapshotName, TableName tableName) throws IOException {
+    Preconditions.checkNotNull(snapshotName);
+    Preconditions.checkNotNull(tableName);
+    if (snapshotName.isEmpty()) {
+      // this mimic the HBase behavior
```
igorbernstein2: Please describe the behavior

PR #2303, opened 2019-12-13
chore: cleanup connection initialization

igorbernstein2:
Currently we have 6 methods named some variation of createChannel, which makes it very difficult to understand program flow. In the spirit of https://xkcd.com/927/, this is adding yet another way to do it :). However it re-routes all invocations to the new way. So now everything calls a single method and all of the old methods are unused and marked as deprecated.

The new approach is still not a 100% clean (interceptor creation is still spread out), but it's clean enough for me to add DirectPath functionality.

New flow:
- Individual channels are created in using createNettyChannel
- ChannelPools are created using createRawDataChannelPool
- all interceptors except the watchdog are created in `createInterceptors`. 
- The constructor for BigtableSession handles channel pool caching
- Only raw channels are cached
- constructor combines raw channel with interceptors

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -646,7 +646,6 @@ public class BigtableSession implements Closeable {
   @InternalApi("For internal usage only")
   public static BigtableInstanceClient createInstanceClient(BigtableOptions options)
       throws IOException, GeneralSecurityException {
-    // 1b
     return new BigtableInstanceGrpcClient(createChannelPool(options.getAdminHost(), options));
   }
 
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
@@ -280,8 +280,8 @@ public class BigtableSession implements Closeable {
               // Data channel requires a watchdog
               .add(setupWatchdog())
               .build();
-      Channel dataChannel = ClientInterceptors.intercept(rawDataChannelPool,
-          dataChannelInterceptors);
+      Channel dataChannel =
+          ClientInterceptors.intercept(rawDataChannelPool, dataChannelInterceptors);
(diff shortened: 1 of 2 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -305,31 +331,48 @@ public BigtableSession(BigtableOptions opts) throws IOException {
...
+    clientInterceptorsList.add(setupWatchdog());
+
+    return clientInterceptorsList;
+  }
+
+  @InternalApi
+  static List<ClientInterceptor> createInterceptors(BigtableOptions options) throws IOException {
+    ImmutableList.Builder<ClientInterceptor> interceptors = ImmutableList.builder();
+
+    // TODO: instanceName should never be null
```
kolea2: should we throw here instead?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -355,86 +398,67 @@ private WatchdogInterceptor setupWatchdog() {
...
-   *
-   * <p>For internal use only - public for technical reasons.
+   * @deprecated Channel creation is now considered an internal implementation detail channel
+   *     creation methods will be removed from the public surface in the future
    */
+  @Deprecated
   @InternalApi("For internal usage only")
   public static ManagedChannel createChannelPool(final String host, final BigtableOptions options)
       throws IOException, GeneralSecurityException {
+    // 1a
```
kolea2: ?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
igorbernstein2: In the future yes. But for now no. This PR should retain existing behavior

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
igorbernstein2: oops, leftover from tracing the code

PR #2304, opened 2019-12-13
chore: Prepare to cleanup channel management

igorbernstein2:
This PR doesn't actually change functionality, it simply groups all of the channel creation related methods together to make it easier to reason about

This is in preparation for #2303


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -567,84 +646,10 @@ protected ManagedChannel createManagedPool(String host, int channelCount) throws
   @InternalApi("For internal usage only")
   public static BigtableInstanceClient createInstanceClient(BigtableOptions options)
       throws IOException, GeneralSecurityException {
+    // 1b
```
rahulKQL: I believe this line is a typo

igorbernstein2: fixed

PR #2308, opened 2019-12-16
deps: fix hbaseLocalMiniClusterTestH2 tests

kolea2:
When fixing the 'used but undeclared' dependency issues, the version of hadoop-common was set to 2.7.4 (when it was 2.8.5). This caused the hbaseLocalMiniClusterTestH2 tests to fail because they could not initialize.

### DIFF
--- a/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/pom.xml
+++ b/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/pom.xml
@@ -323,7 +323,7 @@ limitations under the License.
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-common</artifactId>
-            <version>${hadoop.version.1}</version>
+            <version>${hbase1-hadoop.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -73,7 +73,8 @@ limitations under the License.
         <hbase.version.1>1.4.10</hbase.version.1>
         <hbase.version.2>2.2.0</hbase.version.2>
         <hbase.version>${hbase.version.1}</hbase.version>
-        <hadoop.version>2.7.4</hadoop.version>
+        <hadoop.version.1>2.7.4</hadoop.version.1>
+        <hadoop.version.2>2.8.5</hadoop.version.2>
```
igorbernstein2: Can we rename these to `hbase1-hadoop.version` & `hbase2-hadoop.version`? To stay consistent with hbase-guava & hbase-protobuf?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/pom.xml
```diff
@@ -323,7 +323,7 @@ limitations under the License.
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-common</artifactId>
-            <version>${hadoop.version}</version>
```
igorbernstein2: Shouldn't this be set for bigtable-hbase-2.x-integration-tests as well?

kolea2: hmm, let me take a closer look

PR #2345, opened 2020-01-08
chore(deps): Upgrade to the latest version of google-cloud-bigtable by switching to google-cloud-bigtable-deps-bom

igorbernstein2:
google-cloud-clients should no longer be used. Instead we are using a bigtable specific bom

- [x] Wait for google-cloud-bigtable-deps-bom > 1.9.0 to be released 
- [x] update this PR to the correct version of the bom
- [x] make sure that unshaded tcnative doesn't get pulled in


igorbernstein2:
@kolea2 @rahulKQL this is ready for review

PR #2356, opened 2020-01-13
chore: fix gcj connection cache test

igorbernstein2:
This refactors the TestConnectionCache test to be a unit test instead of an integration test. The original integration test required multiple hardcoded endpoints to run, but that didn't work when running tests against non-prod endpoints. This PR converts that test into a unit test and targets an in memory fake instead.

This should fix the integration test failures


bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/TestBigtableSessionConnectionCache.java
```diff
@@ -0,0 +1,147 @@
...
+          .setCredentialOptions(CredentialOptions.nullCredential())
+          .setUseGCJClient(true)
+          .setUseCachedDataPool(true)
+          .setPort(fakeServer.getPort())
+          .setUserAgent("test");
+
+      Map<String, ClientContext> context = getGcjContext(optionsBuilder, connectionEndpoints[0]);
+      Assert.assertEquals(context.size(), 1);
+      Assert.assertTrue(context.containsKey(connectionEndpoints[0]));
+      ClientContext Context_0 = context.get(connectionEndpoints[0]);
```
kolea2: nit - lowercase `context_0`?

kolea2: (and other similar variables)

PR #2357, opened 2020-01-14
chore(deps): fix hamcrest dependencies

igorbernstein2:
junit depends on hamcrest-core 1.x, but we depend on hamcrest 2.x. This causes a mismatch
of classes to be on the classpath. This moves our code to us hamcrest 1.x, which is named
hamcrest-library


igorbernstein2:
I think that the integration failure is just a flake to a concurrent run

PR #2359, opened 2020-01-15
chore: temporarily disable 2.x IT tests

kolea2:


igorbernstein2:
I think we can fix the flakiness via https://github.com/googleapis/java-bigtable-hbase/pull/2360.
If that doesnt work, lets come back to this

PR #2362, opened 2020-01-16
chore: update integration tests to use unshaded artifacts

igorbernstein2:
Migrate integration tests to use hbase-shaded-testing-util, which allows us to use unshaded versions of bigtable-hbase-{1,2}.x.

### DIFF
--- a/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/pom.xml
+++ b/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/pom.xml
@@ -167,7 +167,6 @@ limitations under the License.
                   <includes>
                     <include>**/TestKnownHBaseGap.java</include>
                   </includes>
-                  <trimStackTrace>false</trimStackTrace>
                   <reportNameSuffix>local-mini-cluster</reportNameSuffix>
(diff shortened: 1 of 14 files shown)


bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/pom.xml
```diff
@@ -323,68 +363,106 @@ limitations under the License.
...
       <groupId>com.google.code.findbugs</groupId>
       <artifactId>jsr305</artifactId>
       <version>${jsr305.version}</version>
       <scope>test</scope>
     </dependency>
+
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>${hbase-guava.version}</version>
```
kolea2: was this the wrong version?

igorbernstein2: No, but the context has changed.
Previously integration tests would use a shaded version of bigtable client and an expanded version of hbase minicluster, so we had to use hbase's version of guava.
Now bigtable is expanded and hbase is shaded, so we get to use bigtable's version of guava

PR #2364, opened 2020-01-20
Change deprecated example in README.

jpharaoh-google:
Calling new BigtableOptions.Builder() directly is deprecated, as per the docs:

https://cloud.google.com/bigtable/docs/hbase-client/javadoc/com/google/cloud/bigtable/config/BigtableOptions.Builder

Builder()
Deprecated.
Please use the BigtableOptions.builder() instead.

We should update the example in README according to these instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ infrequent product and client library announcements.
     @Setup
     public void setup() {
       // for bigtable-core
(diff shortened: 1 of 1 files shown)


igorbernstein2:
Thanks!

PR #2374, opened 2020-02-05
feat: adding bulkRead wrapper for core client

rahulKQL:
## Background
Currently BulkRead API returns row details with FlatRow as a logical bigtable row. Going forward we are going to introduce BulkRead API which uses gax's Batcher API to resolve the input entries.

## What this PR contains:
This PR introduces a simple wrapper b/w existing `BulkRead` with veneer's model(i.e. models.Row). Also, updated `BatchExecutor` to adapt against model Row.

**Note:** I will raise a follow up PR with BulkReadGCJClient which would perform batching with Gax's Batcher API when `google-cloud-bigtable` version updates to `1.10.0`.


igorbernstein2:
New wrappers should live in bigtable-hbase

igorbernstein2:
As discussed offline, we should not be modifying bigtable-client-core to establish an abstraction layer for veneer

PR #2383, opened 2020-02-14
refactor: add utility for converting Configuration to Bigtable Settings

rahulKQL:
## Background
As of right now, We have only one way of user provided configuration porting i.e. `BigtableOptionsFactory#fromConfiguration`. This utility converts Hbase `Configuration` to `BigtableOptions`, which is present in `bigtable-core-client`.

## What this PR contains
This change introduces two new utility in the same class to convert `Configuration` to veneer's `BigtableDataSettings` & `BigtableTableAdminSettings`.

**Note:** _We need to copy over most of the constants from [BigtableOptions](https://github.com/googleapis/java-bigtable-hbase/blob/41a3d98cb4561917ac91e1ea9207b99e1f3e0067/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java#L53-L80), [CallOptionsConfig](https://github.com/googleapis/java-bigtable-hbase/blob/41a3d98cb4561917ac91e1ea9207b99e1f3e0067/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java#L32-L47), [BulkOptions](https://github.com/googleapis/java-bigtable-hbase/blob/41a3d98cb4561917ac91e1ea9207b99e1f3e0067/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BulkOptions.java#L33-L99) [RetryOptions](https://github.com/googleapis/java-bigtable-hbase/blob/41a3d98cb4561917ac91e1ea9207b99e1f3e0067/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/RetryOptions.java#L56-L94) to `bigtable-hbase` module to maintain seamless user experience(considering some of the users are relying on these default values)._

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
@@ -15,15 +15,20 @@
  */
 package com.google.cloud.bigtable.hbase;
 
+import static com.google.cloud.bigtable.config.BigtableOptions.BIGTABLE_ADMIN_HOST_DEFAULT;
+import static com.google.cloud.bigtable.config.BigtableOptions.BIGTABLE_DATA_CHANNEL_COUNT_DEFAULT;
+import static com.google.cloud.bigtable.config.BigtableOptions.BIGTABLE_PORT_DEFAULT;
 import static com.google.cloud.bigtable.config.BulkOptions.BIGTABLE_ASYNC_MUTATOR_COUNT_DEFAULT;
 import static com.google.cloud.bigtable.config.BulkOptions.BIGTABLE_BULK_AUTOFLUSH_MS_DEFAULT;
 import static com.google.cloud.bigtable.config.BulkOptions.BIGTABLE_BULK_MAX_REQUEST_SIZE_BYTES_DEFAULT;
 import static com.google.cloud.bigtable.config.BulkOptions.BIGTABLE_BULK_MAX_ROW_KEY_COUNT_DEFAULT;
 import static com.google.cloud.bigtable.config.BulkOptions.BIGTABLE_MAX_INFLIGHT_RPCS_PER_CHANNEL_DEFAULT;
+import static com.google.cloud.bigtable.config.BulkOptions.BIGTABLE_MAX_MEMORY_DEFAULT;
 import static com.google.cloud.bigtable.config.CallOptionsConfig.LONG_TIMEOUT_MS_DEFAULT;
 import static com.google.cloud.bigtable.config.CallOptionsConfig.SHORT_TIMEOUT_MS_DEFAULT;
 import static com.google.cloud.bigtable.config.CallOptionsConfig.USE_TIMEOUT_DEFAULT;
 import static com.google.cloud.bigtable.config.RetryOptions.DEFAULT_BACKOFF_MULTIPLIER;
+import static com.google.cloud.bigtable.config.RetryOptions.DEFAULT_ENABLE_GRPC_RETRIES_SET;
 import static com.google.cloud.bigtable.config.RetryOptions.DEFAULT_INITIAL_BACKOFF_MILLIS;
 import static com.google.cloud.bigtable.config.RetryOptions.DEFAULT_MAX_ELAPSED_BACKOFF_MILLIS;
 import static com.google.cloud.bigtable.config.RetryOptions.DEFAULT_MAX_SCAN_TIMEOUT_RETRIES;
@@ -461,7 +466,7 @@ public class BigtableOptionsFactory {
 
     long maxMemory =
         configuration.getLong(
-            BIGTABLE_BUFFERED_MUTATOR_MAX_MEMORY_KEY, BulkOptions.BIGTABLE_MAX_MEMORY_DEFAULT);
+            BIGTABLE_BUFFERED_MUTATOR_MAX_MEMORY_KEY, BIGTABLE_MAX_MEMORY_DEFAULT);
     bulkOptionsBuilder.setMaxMemory(maxMemory);
 
     if (configuration.getBoolean(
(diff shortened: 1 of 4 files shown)


igorbernstein2:
Please split this up into 2 classes:


class BigtableOptionsAdapter {
   BigtableOptionsAdapter(Configuration);
   BigtableOptions getBigtableOptions();
}

class BigtableVeneerSettingsAdapter {
  BigtableVeneerSettingsAdapter(Configuration);
  BigtableDataClientSettings getDataSettings();
  BigableTableAdminSettings getTableAdminSettings();
  BigtableInstanceAdminSettings getInstanceAdminSettings();
}

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
```diff
@@ -550,11 +599,451 @@ private static RetryOptions createRetryOptions(Configuration configuration) {
...
+            buildIdempotentRetrySettings(
+                stubSettings.sampleRowKeysSettings().getRetrySettings(), configuration));
+
+    String emulatorHost = configuration.get(BIGTABLE_EMULATOR_HOST_KEY);
+    if (!Strings.isNullOrEmpty(emulatorHost)) {
+      try {
+        int lastIndexOfCol = emulatorHost.lastIndexOf(":");
+        int port = Integer.parseInt(emulatorHost.substring(lastIndexOfCol + 1));
+
+        String hostPort = emulatorHost.substring(0, lastIndexOfCol) + ":" + port;
```
igorbernstein2: veneer now has support for emulators, can we use it?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
```diff
@@ -550,11 +599,451 @@ private static RetryOptions createRetryOptions(Configuration configuration) {
...
+        stubSettings
+            .setCredentialsProvider(NoCredentialsProvider.create())
+            .setEndpoint(hostPort)
+            .setTransportChannelProvider(
+                InstantiatingGrpcChannelProvider.newBuilder()
```
igorbernstein2: If we can't reuse veneer's emulator support, then we should at least reuse the existing grpc channel provider and avoid duplicating these settings

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
```diff
@@ -550,11 +599,451 @@ private static RetryOptions createRetryOptions(Configuration configuration) {
...
+            configuration.get(BIGTABLE_SERVICE_ACCOUNT_P12_KEYFILE_LOCATION_KEY);
+        Preconditions.checkState(
+            !isNullOrEmpty(keyFileLocation),
+            "Key file location must be specified when setting service account email");
+        LOG.debug("Using p12 keyfile: %s", keyFileLocation);
+
+        credentials = getCredentialFromPrivateKeyServiceAccount(serviceAccount, keyFileLocation);
+      } else {
+        LOG.debug("Using default credentials.");
+        credentials = GoogleCredentials.getApplicationDefault();
```
igorbernstein2: This will break directpath. We should not override the default credentials unless the user specified something in specific

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
```diff
@@ -550,11 +599,451 @@ private static RetryOptions createRetryOptions(Configuration configuration) {
...
+
+    RetrySettings.Builder retryBuilder = originalRetrySettings.toBuilder();
+    if (configuration.getBoolean(ALLOW_NO_TIMESTAMP_RETRIES_KEY, false)) {
+      throw new UnsupportedOperationException("Retries without Timestamp is not supported yet.");
+    }
+
+    boolean useTimeOut = configuration.getBoolean(BIGTABLE_USE_TIMEOUTS_KEY, USE_TIMEOUT_DEFAULT);
+    int shortRpcTimeoutMs =
+        configuration.getInt(BIGTABLE_RPC_TIMEOUT_MS_KEY, SHORT_TIMEOUT_MS_DEFAULT);
+    Duration rpcTimeout = ofMillis(useTimeOut ? shortRpcTimeoutMs : RPC_DEADLINE_MS);
```
igorbernstein2: This looks wrong. If I understand correctly, useTimeout = false means that deadlines are supposed to be disabled, but here you are setting it to 6 minutes?

igorbernstein2:
This is a huge class, I think we should extract the veneer config to a separate class and mark that class as internal only

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
rahulKQL: Oh, I followed the existing conversion from BigtableOptions to BigtableDataSettings.
https://github.com/googleapis/java-bigtable-hbase/blob/41a3d98cb4561917ac91e1ea9207b99e1f3e0067/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableVeneerSettingsFactory.java#L275-L277
If this seems wrong then I can quickly open a PR to fix that as well

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
rahulKQL: reusing the buildPlainTextChannelProvider() for emulator setup.

igorbernstein2:
As discussed offline, this should:
- live in a different package
- should not inherit from BigtableOptionsFactory, instead there should be 3 new classes: a base, a variant for veneer and a variant for classic. The base should expose settings that hbase needs and the variants will add the functionality similar to here. 
- have instance methods to get hbase layer options

Please reopen this when its restructure to fit with the plan we discussed

PR #2390, opened 2020-02-27
chore(deps): update beam version to latest i.e. 2.19.0

### LINKED ISSUE #2379: Upgrade to latest version of beam

rahulKQL:
Fixes #2379 

Have verified through [CloudBigtableBeamITTest](https://github.com/googleapis/java-bigtable-hbase/blob/41a3d98cb4561917ac91e1ea9207b99e1f3e0067/bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/it/CloudBigtableBeamITTest.java) with **DataflowRunner** & **DirectRunner**.

### DIFF
--- a/bigtable-dataflow-parent/bigtable-hbase-beam/pom.xml
+++ b/bigtable-dataflow-parent/bigtable-hbase-beam/pom.xml
@@ -85,14 +85,6 @@ limitations under the License.
       <version>${slf4j.version}</version>
     </dependency>
 
-    <!--  TODO: remove this dependency when upgraded through transitive dependency (beam-sdks-java-core)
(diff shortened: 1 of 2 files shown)


igorbernstein2:
Can you also double check that all of the transitive dependencies in bigtable-hbase-beam are strictly newer than the transitive dependencies in a project that only uses beam-sdks-java-core, beam-sdks-java-io-hbase, beam-runners-google-cloud-dataflow-java

rahulKQL:
@igorbernstein2 Thanks for the pointer.

I just cross-checked the transitive dependencies of `bigtable-hbase-beam` are the same version as that of `bigtable-beam-import`, which is the only other module which has the mentioned dependencies.

PR #2392, opened 2020-02-28
deps: manually update dependency org.mockito:mockito-core to v3.3.0

igorbernstein2:
#2388  + manual fixes


bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/async/TestMutateRowsRequestManager.java
```diff
@@ -91,7 +89,6 @@ private static ProcessingStatus send(
 
   @Before
   public void setup() {
-    MockitoAnnotations.initMocks(this);
```
kolea2: does this need `@Rule`? (similar to implementation in next file)

igorbernstein2: nope, this class doesn't use `@Mock` annotations

kolea2: got it, thanks!

PR #2393, opened 2020-03-02
chore: adds BigtableHBaseSettings with classic and veneer implementation 

rahulKQL:
This change introduces `BigtableHBaseSettings`, which is then extended for specific implementation(i.e. classic and veneer):
  - `wrappers.BigtableHBaseSettings`:  abstract layer for settings.
  - `wrappers.classic.BigtableHBaseClassicSettings`
  - `wrappers.veneer.BigtableHBaseVeneerSettings`

Along with those, It includes getters for `BigtablOptions.Builder` and exposes `BigtableExtendedConfiguration` as internal public class for technical purpose.

**Assumption:** I think we should keep the BigtableOptionsFactory(this contains configuration keys) in its current state, reason being it is exists in `bigtable-hbase` module and publicly exposed. we would be deprecating & removing the `BigtableOptionsFactory#fromConfiguration()` utility in the future.
updated buildCredentials to preserve user provided credentials

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
@@ -153,6 +153,13 @@ public class BigtableOptions implements Serializable, Cloneable {
       options.useCachedDataPool = false;
       options.useGCJClient = false;
 
+      options.bulkOptions =
+          BulkOptions.builder()
+              .setMaxInflightRpcs(
+                  BulkOptions.BIGTABLE_MAX_INFLIGHT_RPCS_PER_CHANNEL_DEFAULT
+                      * options.dataChannelCount)
+              .build();
+
       options.retryOptions = new RetryOptions.Builder().build();
       options.callOptionsConfig = CallOptionsConfig.builder().build();
       // CredentialOptions.defaultCredentials() gets credentials from well known locations, such as
@@ -230,9 +237,17 @@ public class BigtableOptions implements Serializable, Cloneable {
       return this;
     }
 
+    public RetryOptions getRetryOptions() {
+      return options.retryOptions;
+    }
+
     public Builder setBulkOptions(BulkOptions bulkOptions) {
       options.bulkOptions = bulkOptions;
       return this;
+    }
+
+    public BulkOptions getBulkOptions() {
+      return options.bulkOptions;
     }
 
     public Builder setUsePlaintextNegotiation(boolean usePlaintextNegotiation) {
@@ -256,6 +271,10 @@ public class BigtableOptions implements Serializable, Cloneable {
     public Builder setCallOptionsConfig(CallOptionsConfig callOptionsConfig) {
       options.callOptionsConfig = callOptionsConfig;
       return this;
+    }
+
+    public CallOptionsConfig getCallOptionsConfig() {
+      return options.callOptionsConfig;
     }
 
     public Builder setUseBatch(boolean useBatch) {
@@ -318,11 +337,7 @@ public class BigtableOptions implements Serializable, Cloneable {
     }
 
     public BigtableOptions build() {
-      if (options.bulkOptions == null) {
-        int maxInflightRpcs =
-            BulkOptions.BIGTABLE_MAX_INFLIGHT_RPCS_PER_CHANNEL_DEFAULT * options.dataChannelCount;
-        options.bulkOptions = BulkOptions.builder().setMaxInflightRpcs(maxInflightRpcs).build();
-      } else if (options.bulkOptions.getMaxInflightRpcs() <= 0) {
+      if (options.bulkOptions.getMaxInflightRpcs() <= 0) {
         int maxInflightRpcs =
             BulkOptions.BIGTABLE_MAX_INFLIGHT_RPCS_PER_CHANNEL_DEFAULT * options.dataChannelCount;
         options.bulkOptions =
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
@@ -337,6 +337,7 @@ public class BigtableOptions implements Serializable, Cloneable {
     }
 
     public BigtableOptions build() {
+      Preconditions.checkNotNull(options.bulkOptions, "bulk options cannot be null");
       if (options.bulkOptions.getMaxInflightRpcs() <= 0) {
         int maxInflightRpcs =
             BulkOptions.BIGTABLE_MAX_INFLIGHT_RPCS_PER_CHANNEL_DEFAULT * options.dataChannelCount;
--- a/bigtable-client-core-parent/bigtable-hbase/pom.xml
+++ b/bigtable-client-core-parent/bigtable-hbase/pom.xml
@@ -135,12 +135,6 @@ limitations under the License.
     <dependency>
       <groupId>com.google.auth</groupId>
       <artifactId>google-auth-library-oauth2-http</artifactId>
-      <exclusions>
-        <exclusion>
-          <groupId>com.google.http-client</groupId>
-          <artifactId>google-http-client-jackson2</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
 
     <!-- TODO: remove this is direct dep, its only used for its Clock interface -->
--- a/bigtable-client-core-parent/bigtable-hbase/pom.xml
+++ b/bigtable-client-core-parent/bigtable-hbase/pom.xml
@@ -123,6 +123,26 @@ limitations under the License.
       <version>${commons-logging.version}</version>
     </dependency>
 
+    <!-- For veneer client settings configurations -->
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.api</groupId>
+      <artifactId>gax-grpc</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.auth</groupId>
+      <artifactId>google-auth-library-oauth2-http</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>com.google.http-client</groupId>
+          <artifactId>google-http-client-jackson2</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+
     <!-- TODO: remove this is direct dep, its only used for its Clock interface -->
     <dependency>
       <groupId>com.google.http-client</groupId>
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/BigtableHBaseSettings.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/BigtableHBaseSettings.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2020 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.hbase.wrappers;
+
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_USE_GCJ_CLIENT;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.INSTANCE_ID_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.PROJECT_ID_KEY;
+import static com.google.common.base.Strings.isNullOrEmpty;
+
+import com.google.api.core.InternalApi;
+import com.google.cloud.bigtable.hbase.BigtableOptionsFactory;
+import com.google.cloud.bigtable.hbase.util.Logger;
+import com.google.cloud.bigtable.hbase.wrappers.classic.BigtableHBaseClassicSettings;
+import com.google.cloud.bigtable.hbase.wrappers.veneer.BigtableHBaseVeneerSettings;
+import com.google.common.base.Preconditions;
+import java.io.IOException;
+import org.apache.hadoop.conf.Configuration;
+
+/** For internal use only - public for technical reasons. */
+@InternalApi("For internal usage only")
+public abstract class BigtableHBaseSettings {
+
+  protected static final Logger LOG = new Logger(BigtableOptionsFactory.class);
+
+  private final Configuration configuration;
+  private final String projectId;
+  private final String instanceId;
+
+  public static BigtableHBaseSettings create(Configuration configuration) throws IOException {
+    if (configuration.getBoolean(BIGTABLE_USE_GCJ_CLIENT, false)) {
+      return new BigtableHBaseVeneerSettings(configuration);
+    } else {
+      return new BigtableHBaseClassicSettings(configuration);
+    }
+  }
+
+  public BigtableHBaseSettings(Configuration configuration) {
+    this.configuration = new Configuration(configuration);
+    this.projectId = getRequiredValue(PROJECT_ID_KEY, "Project ID");
+    this.instanceId = getRequiredValue(INSTANCE_ID_KEY, "Instance ID");
+  }
+
+  public Configuration getConfiguration() {
+    return configuration;
+  }
+
+  public String getProjectId() {
+    return projectId;
+  }
+
+  public String getInstanceId() {
+    return instanceId;
+  }
+
+  public abstract String getDataHost();
+
+  public abstract String getAdminHost();
+
+  public abstract int getPort();
+
+  public abstract int getBulkMaxRowCount();
+
+  protected String getRequiredValue(String key, String displayName) {
+    String value = configuration.get(key);
+    Preconditions.checkArgument(
+        !isNullOrEmpty(value), String.format("%s must be supplied via %s", displayName, key));
+    return value;
+  }
+}
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/BigtableHBaseSettings.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/BigtableHBaseSettings.java
@@ -73,6 +73,8 @@ public abstract class BigtableHBaseSettings {
 
   public abstract int getBulkMaxRowCount();
 
+  public abstract long getBatchingMaxRequestSize();
+
   protected String getRequiredValue(String key, String displayName) {
     String value = configuration.get(key);
     Preconditions.checkArgument(
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
@@ -132,6 +132,11 @@ public class BigtableHBaseClassicSettings extends BigtableHBaseSettings {
     return bigtableOptions.getBulkOptions().getBulkMaxRowKeyCount();
   }
 
+  @Override
+  public long getBatchingMaxRequestSize() {
+    return bigtableOptions.getBulkOptions().getMaxMemory();
+  }
+
   public BigtableOptions getBigtableOptions() {
     return bigtableOptions;
   }
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
@@ -1,0 +1,375 @@
+/*
+ * Copyright 2020 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.hbase.wrappers.classic;
+
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.ADDITIONAL_RETRY_CODES;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.ALLOW_NO_TIMESTAMP_RETRIES_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.APP_PROFILE_ID_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_ADMIN_HOST_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_ASYNC_MUTATOR_COUNT_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_BUFFERED_MUTATOR_ENABLE_THROTTLING;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_BUFFERED_MUTATOR_MAX_MEMORY_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_BUFFERED_MUTATOR_THROTTLING_THRESHOLD_MILLIS;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_BULK_AUTOFLUSH_MS_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_BULK_MAX_REQUEST_SIZE_BYTES;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_BULK_MAX_ROW_KEY_COUNT;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_DATA_CHANNEL_COUNT_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_HOST_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_LONG_RPC_TIMEOUT_MS_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_MUTATE_RPC_TIMEOUT_MS_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_NULL_CREDENTIAL_ENABLE_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_PORT_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_READ_RPC_TIMEOUT_MS_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_RPC_TIMEOUT_MS_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_SERVICE_ACCOUNT_EMAIL_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_SERVICE_ACCOUNT_JSON_KEYFILE_LOCATION_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_SERVICE_ACCOUNT_JSON_VALUE_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_SERVICE_ACCOUNT_P12_KEYFILE_LOCATION_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_USE_BATCH;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_USE_BULK_API;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_USE_CACHED_DATA_CHANNEL_POOL;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_USE_PLAINTEXT_NEGOTIATION;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_USE_TIMEOUTS_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.CUSTOM_USER_AGENT_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.ENABLE_GRPC_RETRIES_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.ENABLE_GRPC_RETRY_DEADLINEEXCEEDED_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.INITIAL_ELAPSED_BACKOFF_MILLIS_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.MAX_ELAPSED_BACKOFF_MILLIS_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.MAX_INFLIGHT_RPCS_KEY;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.MAX_SCAN_TIMEOUT_RETRIES;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.READ_BUFFER_SIZE;
+import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.READ_PARTIAL_ROW_TIMEOUT_MS;
+import static com.google.common.base.Strings.isNullOrEmpty;
+
+import com.google.api.core.InternalApi;
+import com.google.auth.Credentials;
+import com.google.cloud.bigtable.config.BigtableOptions;
+import com.google.cloud.bigtable.config.BulkOptions;
+import com.google.cloud.bigtable.config.CallOptionsConfig;
+import com.google.cloud.bigtable.config.CredentialOptions;
+import com.google.cloud.bigtable.config.RetryOptions;
+import com.google.cloud.bigtable.hbase.BigtableExtendedConfiguration;
+import com.google.cloud.bigtable.hbase.BigtableOptionsFactory;
+import com.google.cloud.bigtable.hbase.wrappers.BigtableHBaseSettings;
+import com.google.common.base.Preconditions;
+import io.grpc.Status;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.util.VersionInfo;
+
+/** For internal use only - public for technical reasons. */
+@InternalApi("For internal usage only")
+public class BigtableHBaseClassicSettings extends BigtableHBaseSettings {
+
+  // Keeping the original configuration, this could be an instance of BigtableExtendedConfiguration.
+  private final Configuration configuration;
+  private final BigtableOptions bigtableOptions;
+
+  public BigtableHBaseClassicSettings(Configuration configuration) throws IOException {
+    super(configuration);
+    this.configuration = configuration;
+    BigtableOptions.Builder bigtableOptionsBuilder = BigtableOptions.builder();
+
+    bigtableOptionsBuilder.setProjectId(getProjectId()).setInstanceId(getInstanceId());
+
+    String appProfileId = configuration.get(APP_PROFILE_ID_KEY);
+    if (!isNullOrEmpty(appProfileId)) {
+      bigtableOptionsBuilder.setAppProfileId(appProfileId);
+    }
+
+    buildBulkOptions(bigtableOptionsBuilder);
+    buildChannelOptions(bigtableOptionsBuilder);
+    buildClientCallOptions(bigtableOptionsBuilder);
+    buildCredentialOptions(bigtableOptionsBuilder);
+    buildRetryOptions(bigtableOptionsBuilder);
+
+    String emulatorHost = configuration.get(BigtableOptionsFactory.BIGTABLE_EMULATOR_HOST_KEY);
+    if (!isNullOrEmpty(emulatorHost)) {
+      bigtableOptionsBuilder.enableEmulator(emulatorHost);
+    }
+
+    String useBatchStr = configuration.get(BIGTABLE_USE_BATCH);
+    if (!isNullOrEmpty(useBatchStr)) {
+      bigtableOptionsBuilder.setUseBatch(Boolean.parseBoolean(useBatchStr));
+    }
+
+    this.bigtableOptions = bigtableOptionsBuilder.build();
+  }
+
+  @Override
+  public String getDataHost() {
+    return bigtableOptions.getDataHost();
+  }
+
+  @Override
+  public String getAdminHost() {
+    return bigtableOptions.getAdminHost();
+  }
+
+  @Override
+  public int getPort() {
+    return bigtableOptions.getPort();
+  }
+
+  @Override
+  public int getBulkMaxRowCount() {
+    return bigtableOptions.getBulkOptions().getBulkMaxRowKeyCount();
+  }
+
+  public BigtableOptions getBigtableOptions() {
+    return bigtableOptions;
+  }
+
+  // ************** Private Helpers **************
+  private void buildBulkOptions(BigtableOptions.Builder builder) {
+    BulkOptions.Builder bulkOptionsBuilder = builder.getBulkOptions().toBuilder();
+
+    String asyncMutatorCount = configuration.get(BIGTABLE_ASYNC_MUTATOR_COUNT_KEY);
+    if (!isNullOrEmpty(asyncMutatorCount)) {
+      bulkOptionsBuilder.setAsyncMutatorWorkerCount(Integer.parseInt(asyncMutatorCount));
+    }
+
+    String useBulkApiStr = configuration.get(BIGTABLE_USE_BULK_API);
+    if (!isNullOrEmpty(useBulkApiStr)) {
+      bulkOptionsBuilder.setUseBulkApi(Boolean.parseBoolean(useBulkApiStr));
+    }
+
+    String bulkMaxRowKeyCountStr = configuration.get(BIGTABLE_BULK_MAX_ROW_KEY_COUNT);
+    if (!isNullOrEmpty(bulkMaxRowKeyCountStr)) {
+      bulkOptionsBuilder.setBulkMaxRowKeyCount(Integer.parseInt(bulkMaxRowKeyCountStr));
+    }
+
+    String bulkMaxRequestSize = configuration.get(BIGTABLE_BULK_MAX_REQUEST_SIZE_BYTES);
+    if (!isNullOrEmpty(bulkMaxRequestSize)) {
+      bulkOptionsBuilder.setBulkMaxRequestSize(Long.parseLong(bulkMaxRequestSize));
+    }
+
+    String autoFlushMs = configuration.get(BIGTABLE_BULK_AUTOFLUSH_MS_KEY);
+    if (!isNullOrEmpty(autoFlushMs)) {
+      bulkOptionsBuilder.setAutoflushMs(Long.parseLong(autoFlushMs));
+    }
+
+    String maxInFlightRpcStr = configuration.get(MAX_INFLIGHT_RPCS_KEY);
+    if (!isNullOrEmpty(maxInFlightRpcStr)) {
+      bulkOptionsBuilder.setMaxInflightRpcs(Integer.parseInt(maxInFlightRpcStr));
+    }
+
+    String maxMemoryStr = configuration.get(BIGTABLE_BUFFERED_MUTATOR_MAX_MEMORY_KEY);
+    if (!isNullOrEmpty(maxMemoryStr)) {
+      bulkOptionsBuilder.setMaxMemory(Long.parseLong(maxMemoryStr));
+    }
+
+    if (Boolean.parseBoolean(configuration.get(BIGTABLE_BUFFERED_MUTATOR_ENABLE_THROTTLING))) {
+      String bufferedMutatorThresholdMsStr =
+          configuration.get(BIGTABLE_BUFFERED_MUTATOR_THROTTLING_THRESHOLD_MILLIS);
+
+      if (!isNullOrEmpty(bufferedMutatorThresholdMsStr)) {
+        int mutatorThresholdMs = Integer.parseInt(bufferedMutatorThresholdMsStr);
+        LOG.info(
+            "Bigtable mutation latency throttling enabled with threshold %d", mutatorThresholdMs);
+        bulkOptionsBuilder.enableBulkMutationThrottling();
+        bulkOptionsBuilder.setBulkMutationRpcTargetMs(mutatorThresholdMs);
+      }
+    }
+
+    builder.setBulkOptions(bulkOptionsBuilder.build());
+  }
+
+  private void buildChannelOptions(BigtableOptions.Builder builder) {
+
+    String dataHostOverride = configuration.get(BIGTABLE_HOST_KEY);
+    if (!isNullOrEmpty(dataHostOverride)) {
+      LOG.debug("API Data endpoint host %s.", dataHostOverride);
+      builder.setDataHost(dataHostOverride);
+    }
+
+    String adminHostOverride = configuration.get(BIGTABLE_ADMIN_HOST_KEY);
+    if (!isNullOrEmpty(adminHostOverride)) {
+      LOG.debug("Admin endpoint host %s.", adminHostOverride);
+      builder.setAdminHost(adminHostOverride);
+    }
+
+    String portOverrideStr = configuration.get(BIGTABLE_PORT_KEY);
+    if (!isNullOrEmpty(portOverrideStr)) {
+      builder.setPort(Integer.parseInt(portOverrideStr));
+    }
+
+    String usePlaintextStr = configuration.get(BIGTABLE_USE_PLAINTEXT_NEGOTIATION);
+    if (!isNullOrEmpty(usePlaintextStr)) {
+      builder.setUsePlaintextNegotiation(Boolean.parseBoolean(usePlaintextStr));
+    }
+
+    String channelCountStr = configuration.get(BIGTABLE_DATA_CHANNEL_COUNT_KEY);
+    if (!isNullOrEmpty(channelCountStr)) {
+      builder.setDataChannelCount(Integer.parseInt(channelCountStr));
+    }
+
+    // This is primarily used by Dataflow where connections open and close often. This is a
+    // performance optimization that will reduce the cost to open connections.
+    String useCachedDataPoolStr = configuration.get(BIGTABLE_USE_CACHED_DATA_CHANNEL_POOL);
+    if (!isNullOrEmpty(useCachedDataPoolStr)) {
+      builder.setUseCachedDataPool(Boolean.parseBoolean(useCachedDataPoolStr));
+    }
+
+    // This information is in addition to bigtable-client-core version, and jdk version.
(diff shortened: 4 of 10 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/BigtableHBaseSettings.java
```diff
@@ -0,0 +1,103 @@
...
+  public String getInstanceId() {
+    return instanceId;
+  }
+
+  public String getDataHost() {
+    String hostNameStr = configuration.get(BIGTABLE_HOST_KEY);
+    return isNullOrEmpty(hostNameStr) ? BigtableConstants.BIGTABLE_DATA_HOST_DEFAULT : hostNameStr;
+  }
+
+  public String getAdminHost() {
```
igorbernstein2: same as above

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/BigtableHBaseSettings.java
```diff
@@ -0,0 +1,103 @@
...
+    String adminHostStr = configuration.get(BIGTABLE_ADMIN_HOST_KEY);
+    return isNullOrEmpty(adminHostStr)
+        ? BigtableConstants.BIGTABLE_ADMIN_HOST_DEFAULT
+        : adminHostStr;
+  }
+
+  public int getPort() {
```
igorbernstein2: same as above

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/BigtableHBaseSettings.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  public int getBulkMaxRowCount() {
+    String bulkMaxRowKeyCountStr = configuration.get(BIGTABLE_BULK_MAX_ROW_KEY_COUNT);
+    return isNullOrEmpty(bulkMaxRowKeyCountStr)
+        ? BigtableConstants.BIGTABLE_BULK_MAX_ROW_KEY_COUNT_DEFAULT
+        : Integer.valueOf(bulkMaxRowKeyCountStr);
+  }
+
+  protected String getValue(String key, String type) {
```
igorbernstein2: `getRequiredValue(String key, String displayName)`

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
```diff
@@ -0,0 +1,348 @@
...
+
+  public BigtableHBaseClassicSettings(Configuration configuration) {
+    super(configuration);
+  }
+
+  // <editor-fold desc="Public API">
+  /** For internal use only - public for technical reasons. */
+  @InternalApi("For internal usage only")
+  public BigtableOptions getBigtableOptions() throws IOException {
+    BigtableOptions.Builder bigtableOptionsBuilder = BigtableOptions.builder();
```
igorbernstein2: I think this should be done eagerly in the constructor

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
```diff
@@ -0,0 +1,348 @@
...
+@InternalApi("For internal usage only")
+public class BigtableHBaseClassicSettings extends BigtableHBaseSettings {
+
+  public BigtableHBaseClassicSettings(Configuration configuration) {
+    super(configuration);
+  }
+
+  // <editor-fold desc="Public API">
+  /** For internal use only - public for technical reasons. */
+  @InternalApi("For internal usage only")
```
igorbernstein2: This whole class is internal only, no need to mark the methods

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
```diff
@@ -0,0 +1,348 @@
...
+  public BigtableOptions getBigtableOptions() throws IOException {
+    BigtableOptions.Builder bigtableOptionsBuilder = BigtableOptions.builder();
+
+    bigtableOptionsBuilder.setProjectId(getValue(PROJECT_ID_KEY, "Project " + "ID"));
+    bigtableOptionsBuilder.setInstanceId(getValue(INSTANCE_ID_KEY, "Instance ID"));
```
igorbernstein2: I think you should delegate to the base impl for these

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
```diff
@@ -0,0 +1,348 @@
...
+    if (portOverrideStr != null) {
+      bigtableOptionsBuilder.setPort(Integer.parseInt(portOverrideStr));
+    }
+
+    String usePlaintextStr = configuration.get(BIGTABLE_USE_PLAINTEXT_NEGOTIATION);
+    if (usePlaintextStr != null) {
+      bigtableOptionsBuilder.setUsePlaintextNegotiation(Boolean.parseBoolean(usePlaintextStr));
+    }
+
+    setBulkOptions(bigtableOptionsBuilder);
```
igorbernstein2: The naming implies that this is a setter, which its not.
I think configureBulkOptions() is more appropriate

same for below

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
```diff
@@ -0,0 +1,348 @@
...
+
+    String useBatchStr = configuration.get(BIGTABLE_USE_BATCH);
+    if (useBatchStr != null) {
+      bigtableOptionsBuilder.setUseBatch(Boolean.parseBoolean(useBatchStr));
+    }
+
+    String useGcjClientStr = configuration.get(BIGTABLE_USE_GCJ_CLIENT);
+    if (useGcjClientStr != null) {
+      bigtableOptionsBuilder.setUseGCJClient(Boolean.parseBoolean(useGcjClientStr));
+    }
```
igorbernstein2: I dont think this is necessary anymore

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
```diff
@@ -0,0 +1,348 @@
...
+    return bigtableOptionsBuilder.build();
+  }
+  // </editor-fold>
+
+  // <editor-fold desc="Private Helpers">
+  private void setChannelOptions(BigtableOptions.Builder builder) throws IOException {
```
igorbernstein2: configureChannelOptions

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
```diff
@@ -0,0 +1,348 @@
...
+    setCredentialOptions(builder);
+
+    builder.setRetryOptions(createRetryOptions());
```
igorbernstein2: retry options are not really channel options. Maybe this should be moved to the caller

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
```diff
@@ -0,0 +1,348 @@
...
+    builder.setUserAgent(agentBuilder.toString());
+  }
+
+  private void setBulkOptions(BigtableOptions.Builder bigtableOptionsBuilder) {
+    BulkOptions.Builder bulkOptionsBuilder = BulkOptions.builder();
+
+    int asyncMutatorCount =
+        configuration.getInt(
+            BIGTABLE_ASYNC_MUTATOR_COUNT_KEY, BIGTABLE_ASYNC_MUTATOR_COUNT_DEFAULT);
+    bulkOptionsBuilder.setAsyncMutatorWorkerCount(asyncMutatorCount);
```
igorbernstein2: Although it's not as important as in the hostname case, I think you should avoid double defining the defaults and follow the pattern:
```
value =  config.get("key")
if (value != null) {
  options.setValue(value);
}
```

Same for everything below.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
```diff
@@ -0,0 +1,348 @@
...
+    agentBuilder.append("hbase-").append(VersionInfo.getVersion());
+    String customUserAgent = configuration.get(CUSTOM_USER_AGENT_KEY);
+    if (customUserAgent != null) {
+      agentBuilder.append(',').append(customUserAgent);
+    }
+    builder.setUserAgent(agentBuilder.toString());
+  }
+
+  private void setBulkOptions(BigtableOptions.Builder bigtableOptionsBuilder) {
+    BulkOptions.Builder bulkOptionsBuilder = BulkOptions.builder();
```
igorbernstein2: I think you should start with the instance of BulkOptions that already existing in bigtableOptionsBuilder as opposed to create a new instance

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
```diff
@@ -0,0 +1,348 @@
...
+          configuration.getInt(
+              BIGTABLE_BUFFERED_MUTATOR_THROTTLING_THRESHOLD_MILLIS,
+              BulkOptions.BIGTABLE_BULK_THROTTLE_TARGET_MS_DEFAULT));
+    }
+
+    bigtableOptionsBuilder.setBulkOptions(bulkOptionsBuilder.build());
+  }
+
+  private void setCredentialOptions(BigtableOptions.Builder builder) throws FileNotFoundException {
+    if (configuration.getBoolean(
```
igorbernstein2: This is important for directpath: we should not overwrite the default credentials if the user has not explicitly set an override

igorbernstein2:
@rahulKQL ping

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
```diff
@@ -0,0 +1,387 @@
...
+        builder.setCredentialOptions(
+            CredentialOptions.p12Credential(serviceAccount, keyFileLocation));
+      } else {
+        LOG.debug("Using default credentials.");
+        builder.setCredentialOptions(CredentialOptions.defaultCredentials());
+      }
+    } else if (Boolean.parseBoolean(configuration.get(BIGTABLE_NULL_CREDENTIAL_ENABLE_KEY))) {
+      builder.setCredentialOptions(CredentialOptions.nullCredential());
+      LOG.info("Enabling the use of null credentials. This should not be used in production.");
+    } else {
```
igorbernstein2: I think we should drop the else let it use default credentials

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,588 @@
...
+@InternalApi("For internal usage only")
+public class BigtableHBaseVeneerSettings extends BigtableHBaseSettings {
+
+  // Identifier to distinguish between CBT or GCJ adapter.
+  private static final String VENEER_ADAPTER =
+      BigtableVersionInfo.CORE_USER_AGENT + "," + "veneer-adapter,";
+
+  private final Configuration configuration;
+  private final BigtableDataSettings.Builder dataSettingsBuilder;
+  private final BigtableTableAdminSettings.Builder adminBuilder;
```
igorbernstein2: Please eagerly build these settings

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
rahulKQL: I am assuming user could override credentials through `BigtableExtendedConfiguration#withCredentials` OR Service Account(JSON file/Email & P12 key location). If neither of them presents then use the default one.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/BigtableConstants.java
```diff
@@ -0,0 +1,30 @@
...
+class BigtableConstants {
+
+  static final String BIGTABLE_DATA_HOST_DEFAULT = "bigtable.googleapis.com";
+
+  static final String BIGTABLE_ADMIN_HOST_DEFAULT = "bigtableadmin.googleapis.com";
+
+  static final int BIGTABLE_PORT_DEFAULT = 443;
+
+  static final int BIGTABLE_BULK_MAX_ROW_KEY_COUNT_DEFAULT = 125;
+}
```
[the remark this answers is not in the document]
rahulKQL: Removed all the defaults from these settings class.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/BigtableHBaseSettings.java
```diff
@@ -0,0 +1,103 @@
...
+  public static BigtableHBaseSettings create(Configuration configuration) {
+    if (configuration.getBoolean(BIGTABLE_USE_GCJ_CLIENT, false)) {
+      return new BigtableHBaseVeneerSettings(configuration);
+    } else {
+      return new BigtableHBaseClassicSettings(configuration);
+    }
+  }
+
+  public BigtableHBaseSettings(Configuration configuration) {
+    this.configuration = configuration;
```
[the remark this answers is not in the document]
rahulKQL: If we are cloning the instance of configuration at BigtableHBaseSettings level, then we have to keep the original configuration in classic & veneer.

Because there is possibility that a user could override the Credentials using `BigtableExtendedConfiguration`, I hope having configuration at parent & child is fine.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
rahulKQL: Instead of `configureBulkOptions()` I went with `buildBulkOptions()` to have similarity between classic & veneer private helper method names. I hope that should be fine.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
rahulKQL: For this, I had to add getter in `BigtableOptions.Builder`, I have added those changes in this same PR. If you think we should add those as a separate then I can create a PR for that rightaway.

rahulKQL:
I have tried my best to address your feedback comments. PTAL!

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/BigtableHBaseSettings.java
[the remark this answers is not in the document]
igorbernstein2: Ah, good point, please add a comment so it doesn't get changed in the future

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
rahulKQL: In case of veneer we are eagerly building Data & Admin Settings, the InstanceAdmin does not support Emulator so we will initialize it on it's first call.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
```diff
@@ -0,0 +1,396 @@
...
+
+  private void buildCredentialOptions(BigtableOptions.Builder builder)
+      throws FileNotFoundException {
+
+    // This preserves user defined Credentials
+    if (configuration instanceof BigtableExtendedConfiguration) {
+      Credentials credentials = ((BigtableExtendedConfiguration) configuration).getCredentials();
+      builder.setCredentialOptions(CredentialOptions.credential(credentials));
+
+      // Here default is true, if its false then none of the service account would be accounted.
```
igorbernstein2: This comment is not true...
the default is that configuration doesn't have the key, so the value is null which parseBoolean will interpret as false. 

I wonder if we can just ignore BIGTABLE_SERVICE_ACCOUNT_JSON_VALUE_KEY ? and make this:

if (instanceof BigtableExtendedConfiguration) {
...
} else if (BIGTABLE_NULL_CREDENTIAL_ENABLE_KEY) {
...
} else if (BIGTABLE_SERVICE_ACCOUNT_JSON_VALUE_KEY) {
} else if (BIGTABLE_SERVICE_ACCOUNT_JSON_VALUE_KEY) {
} else if (BIGTABLE_SERVICE_ACCOUNT_EMAIL_KEY) {
} 

can you think of any negative reprecussions?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
```diff
@@ -0,0 +1,396 @@
...
+    if (!isNullOrEmpty(readPartialRowTimeoutMs)) {
+      LOG.debug("gRPC read partial row timeout (millis): %d", readPartialRowTimeoutMs);
+      retryOptionsBuilder.setReadPartialRowTimeoutMillis(Integer.parseInt(readPartialRowTimeoutMs));
+    }
+
+    String streamingBufferSize = configuration.get(READ_BUFFER_SIZE);
+    if (!isNullOrEmpty(streamingBufferSize)) {
+      LOG.debug("gRPC read buffer size (count): %d", streamingBufferSize);
+      retryOptionsBuilder.setStreamingBufferSize(Integer.parseInt(streamingBufferSize));
+    }
```
igorbernstein2: Please add a todo to remove this, its not used

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,592 @@
...
+
+  // Identifier to distinguish between CBT or GCJ adapter.
+  private static final String VENEER_ADAPTER =
+      BigtableVersionInfo.CORE_USER_AGENT + "," + "veneer-adapter,";
+
+  // Keeping the original configuration, this could be an instance of BigtableExtendedConfiguration.
+  private final Configuration configuration;
+  private final BigtableDataSettings dataSettings;
+  private final BigtableTableAdminSettings tableAdminSettings;
+  private BigtableInstanceAdminSettings instanceAdminSettings;
```
igorbernstein2: Please mark this a final & `@Nullable`, and add a comment that this will be null when using the emulator because the emulator doesn't implement the instance admin api

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,592 @@
...
+
+  public BigtableHBaseVeneerSettings(Configuration configuration) throws IOException {
+    super(configuration);
+    this.configuration = configuration;
+    this.dataSettings = buildBigtableDataSettings().build();
+    this.tableAdminSettings = buildBigtableTableAdminSettings().build();
```
igorbernstein2: I think buildBigtable*Settings should just return built settings and you can the InstanceAdmin one return null if the emulator is enabled

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,592 @@
...
+    return dataSettings;
+  }
+
+  /** Utility to convert {@link Configuration} to {@link BigtableTableAdminSettings}. */
+  public BigtableTableAdminSettings getTableAdminSettings() {
+    return tableAdminSettings;
+  }
+
+  /** Utility to convert {@link Configuration} to {@link BigtableInstanceAdminSettings}. */
+  public BigtableInstanceAdminSettings getInstanceAdminSettings() throws IOException {
```
igorbernstein2: Please build this is eagerly, but duplicate the emulator check in the getter

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,592 @@
...
+
+    stubSettings.setHeaderProvider(buildHeaderProvider());
+
+    if (configuration instanceof BigtableExtendedConfiguration
+        || Boolean.parseBoolean(configuration.get(BIGTABLE_USE_SERVICE_ACCOUNTS_KEY))) {
+      stubSettings.setCredentialsProvider(buildCredentialProvider());
+
+    } else if (Boolean.parseBoolean(configuration.get(BIGTABLE_NULL_CREDENTIAL_ENABLE_KEY))) {
+      stubSettings.setCredentialsProvider(NoCredentialsProvider.create());
+    }
```
igorbernstein2: Please move this to configureCredentials(StubSettings), which will check if an override is necessary and if not it won't touch the credential provider

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,592 @@
...
+
+    if (Boolean.parseBoolean(configuration.get(BIGTABLE_USE_PLAINTEXT_NEGOTIATION))) {
```
igorbernstein2: please group this with endpoint overrides and move it to a helper method

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
rahulKQL: I assume you meant this(Basically to ignore `BIGTABLE_SERVICE_ACCOUNT_JSON_VALUE_KEY`)
```java
    if(instance instanceof BigtableExtendedConfiguration) {
      ...
    } else if(BIGTABLE_NULL_CREDENTIAL_ENABLE_KEY) {
      ...
    } else if(BIGTABLE_SERVICE_ACCOUNT_JSON_KEYFILE_LOCATION_KEY){
      ...
    } else if(BIGTABLE_SERVICE_ACCOUNT_EMAIL_KEY){
      ...
    }
``` 

then with this a user will not be able to pass key-value pairs as a string within the configuration instance. Although we can always advise users to move towards providing key file location. I believe we should be fine.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
```diff
@@ -0,0 +1,396 @@
...
+    if (!isNullOrEmpty(port)) {
+      bigtableOptionsBuilder.setPort(Integer.parseInt(port));
+    }
+
+    String appProfileId = configuration.get(APP_PROFILE_ID_KEY);
+    if (!isNullOrEmpty(appProfileId)) {
+      bigtableOptionsBuilder.setAppProfileId(appProfileId);
+    }
+
+    String dataHostOverride = configuration.get(BIGTABLE_HOST_KEY);
```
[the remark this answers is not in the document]
rahulKQL: My apologies, It was an overlook.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
```diff
@@ -0,0 +1,396 @@
...
+  private void buildBulkOptions(BigtableOptions.Builder builder) {
+    BulkOptions.Builder bulkOptionsBuilder = builder.getBulkOptions().toBuilder();
+
+    String asyncMutatorCount = configuration.get(BIGTABLE_ASYNC_MUTATOR_COUNT_KEY);
+    if (!isNullOrEmpty(asyncMutatorCount)) {
+      bulkOptionsBuilder.setAsyncMutatorWorkerCount(Integer.parseInt(asyncMutatorCount));
+    }
+
+    bulkOptionsBuilder.setUseBulkApi(
+        Boolean.parseBoolean(configuration.get(BIGTABLE_USE_BULK_API)));
```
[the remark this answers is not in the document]
rahulKQL: updated this to follow isNullOrEmpty pattern

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,570 @@
...
+    if (!isNullOrEmpty(shortRpcTimeoutStr)) {
+      // rpcTimeout & totalTimeout for non-retry operations.
+      Duration shortRpcTimeout = ofMillis(Long.parseLong(shortRpcTimeoutStr));
+
+      stubSettings.checkAndMutateRowSettings().setSimpleTimeoutNoRetries(shortRpcTimeout);
+
+      stubSettings.readModifyWriteRowSettings().setSimpleTimeoutNoRetries(shortRpcTimeout);
+    }
+
+    buildBulkMutationsSettings(stubSettings);
```
igorbernstein2: For the per method RPC settings I think it will be a lot cleaner to structure it like:

buildBulkMutationsSettings(stubSettings.bulkMutationSettings())
buildBulkReadRowsSettings(stubSettings.bulkReadRowSettings())
...
buildIdempotentCallSettings(stubSettings.mutateRowSettings())
buildIdempotentCallSettings(stubSettings. readRowSettings())
...
buildNonIdempotentCallSettings(stubSettings.checkAndMutateRowSettings())
...
void buildIdempotentUnaryCallSettings(UnaryCallSettings<?,?> callSettings) {
  // Configure the callSettings using INITIAL_ELAPSED_BACKOFF_MILLIS_KEY, BIGTABLE_USE_TIMEOUTS_KEY, ENABLE_GRPC_RETRIES_KEY, ADDITIONAL_RETRY_CODES, etc
}

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,570 @@
...
+            buildRetryCodes(stubSettings.sampleRowKeysSettings().getRetryableCodes()))
+        .setRetrySettings(
+            buildIdempotentRetrySettings(stubSettings.sampleRowKeysSettings().getRetrySettings()));
+
+    buildEmulatorSettings(stubSettings);
+
+    return dataBuilder;
+  }
+
+  private BigtableTableAdminSettings.Builder buildBigtableTableAdminSettings() throws IOException {
```
igorbernstein2: same as above, please return built object

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,570 @@
...
+    String channelCount = configuration.get(BIGTABLE_DATA_CHANNEL_COUNT_KEY);
+    if (!isNullOrEmpty(channelCount)) {
+      channelBuilder.setPoolSize(Integer.parseInt(channelCount));
+    }
+
+    return channelBuilder.build();
+  }
+
+  /** Creates {@link Set} of {@link StatusCode.Code} from {@link Status.Code} */
+  private Set<StatusCode.Code> buildRetryCodes(Set<StatusCode.Code> retryableCodes) {
```
igorbernstein2: I think this method should be split:
- extract extractRetryCodesFromConfig() that simply parses ADDITIONAL_RETRY_CODES from the config
- then use that function and ENABLE_GRPC_RETRIES_KEY  in buildIdempotentCallSettings()

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,570 @@
...
+
+    return statusCodeBuilder.build();
+  }
+
+  /** Creates {@link RetrySettings} for non-streaming VENEER_ADAPTER method. */
+  private RetrySettings buildIdempotentRetrySettings(RetrySettings originalRetrySettings) {
+    RetrySettings.Builder retryBuilder = originalRetrySettings.toBuilder();
+
+    if (configuration.getBoolean(ALLOW_NO_TIMESTAMP_RETRIES_KEY, false)) {
+      throw new UnsupportedOperationException("Retries without Timestamp is not supported yet.");
```
igorbernstein2: Is this still true? I remember you adding unsafe methods to veneer

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
```diff
@@ -153,6 +153,7 @@ public Builder() {
       options.useCachedDataPool = false;
       options.useGCJClient = false;
 
+      options.bulkOptions = BulkOptions.builder().build();
```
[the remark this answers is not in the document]
rahulKQL: If the user does not override it then this initialized with the same value in `build()`.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
rahulKQL: I did add `RowMuatationEntry#createUnsafe` in the veneer but It is not supported on other mutations, also there is no way to set this property on BigtableDataSettings.

That is why throw UnsupportedOperationException at this moment.

igorbernstein2: ok

bigtable-client-core-parent/bigtable-hbase/pom.xml
```diff
@@ -119,6 +119,26 @@ limitations under the License.
...
+    <dependency>
+      <groupId>com.google.auth</groupId>
+      <artifactId>google-auth-library-oauth2-http</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>com.google.http-client</groupId>
+          <artifactId>google-http-client-jackson2</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
```
rahulKQL: @igorbernstein2 
These three dependencies needed to be declared. It seems veneer client settings are using classes from these three. I hope this should be fine. WDYT?

igorbernstein2: yes, but why the exclude?

rahulKQL: This is due to a conflict happens in `bigtable-hbase-1.x-benchmark` module between google-http-client-jackson2's 1.34.0 & 1.34.2 versions.
```sh
WARNING] Rule 2: org.apache.maven.plugins.enforcer.RequireUpperBoundDeps failed with message:
Failed while enforcing RequireUpperBoundDeps. The error(s) are [
Require upper bound dependencies error for com.google.http-client:google-http-client-jackson2:1.34.0 paths to dependency are:
+-com.google.cloud.bigtable:bigtable-hbase-1.x-benchmarks:1.14.1-SNAPSHOT
  +-com.google.cloud.bigtable:bigtable-hbase:1.14.1-SNAPSHOT
    +-com.google.auth:google-auth-library-oauth2-http:0.20.0
      +-com.google.http-client:google-http-client-jackson2:1.34.0
and
+-com.google.cloud.bigtable:bigtable-hbase-1.x-benchmarks:1.14.1-SNAPSHOT
  +-com.google.cloud.bigtable:bigtable-hbase-1.x:1.14.1-SNAPSHOT
    +-com.google.cloud.bigtable:bigtable-client-core:1.14.1-SNAPSHOT
      +-com.google.cloud:google-cloud-core:1.93.1
        +-com.google.http-client:google-http-client-jackson2:1.34.2
and
+-com.google.cloud.bigtable:bigtable-hbase-1.x-benchmarks:1.14.1-SNAPSHOT
  +-com.google.cloud.bigtable:bigtable-hbase-1.x:1.14.1-SNAPSHOT
    +-com.google.cloud.bigtable:bigtable-client-core:1.14.1-SNAPSHOT
      +-com.google.cloud:google-cloud-core-http:1.93.1
        +-com.google.api-client:google-api-client:1.30.9
          +-com.google.http-client:google-http-client-jackson2:1.34.2
```

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,580 @@
...
+    buildReadRowsSettings(stubSettings);
+
+    buildNonIdempotentCallSettings(stubSettings.checkAndMutateRowSettings());
+    buildNonIdempotentCallSettings(stubSettings.readModifyWriteRowSettings());
+
+    buildIdempotentCallSettings(stubSettings.readRowSettings());
+    buildIdempotentCallSettings(stubSettings.mutateRowSettings());
+    buildIdempotentCallSettings(stubSettings.sampleRowKeysSettings());
+
+    buildEmulatorSettings(stubSettings);
```
igorbernstein2: can we move to line 192 after credential configuration...it affects similar things as credentials & endpoints

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,580 @@
...
+    if (configuration.getBoolean(ALLOW_NO_TIMESTAMP_RETRIES_KEY, false)) {
+      throw new UnsupportedOperationException("Retries without Timestamp is not supported yet.");
+    }
+
+    String initialElapsedBackoffMsStr = configuration.get(INITIAL_ELAPSED_BACKOFF_MILLIS_KEY);
+    if (!isNullOrEmpty(initialElapsedBackoffMsStr)) {
+      retryBuilder.setInitialRetryDelay(ofMillis(Long.parseLong(initialElapsedBackoffMsStr)));
+    }
+
+    if (Boolean.parseBoolean(configuration.get(BIGTABLE_USE_TIMEOUTS_KEY))) {
```
igorbernstein2: This isn't quite right:
BIGTABLE_USE_TIMEOUTS_KEY is used to enable deadlines. By default deadlines are disabled in bigtable-client-core. However they are mandatory in veneer.
So the correct mapping should be:
- unset in config => disabled in veneer
- false in config => disabled in veneer
- true in config => enabled in veneer
Now the question is what does disabled mean? I guess we should just set it to a very high number. Maybe create a constant for 12 hours and name it something EFFECTIVELY_DISABLED_DEADLINE_DURATION?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,580 @@
...
+
+    long bulkMaxRowKeyCount = batchMutateBuilder.build().getElementCountThreshold();
+
+    String maxInflightRpcStr = configuration.get(MAX_INFLIGHT_RPCS_KEY);
+    if (!isNullOrEmpty(maxInflightRpcStr) && Integer.parseInt(maxInflightRpcStr) > 0) {
+
+      int maxInflightRpcCount = Integer.parseInt(maxInflightRpcStr);
+      FlowControlSettings.Builder flowControlBuilder =
+          FlowControlSettings.newBuilder()
+              // TODO: verify if it should be channelCount instead of maxRowKeyCount
```
igorbernstein2: I think the current impl is correct `maxInflightRpcCount * bulkMaxRowKeyCount`

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,580 @@
...
+
+  private <Request, Response> void buildIdempotentCallSettings(
+      UnaryCallSettings.Builder<Request, Response> unaryCallSettings) {
+
+    ImmutableSet.Builder<StatusCode.Code> retryCodeBuilder = ImmutableSet.builder();
+    if (configuration.getBoolean(ENABLE_GRPC_RETRIES_KEY, true)) {
+      retryCodeBuilder
+          .addAll(extractRetryCodesFromConfig())
+          .addAll(unaryCallSettings.getRetryableCodes());
+    }
```
igorbernstein2: shouldn't this have an else?
ie if config.set(ENABLE_GRPC_RETRIES_KEY, false). Then shouldn't we disable retries? ie clear retry codes

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,580 @@
...
+      batchMutateBuilder.setFlowControlSettings(flowControlBuilder.build());
+    }
+
+    String requestByteThresholdStr = configuration.get(BIGTABLE_BULK_MAX_REQUEST_SIZE_BYTES);
+    if (!isNullOrEmpty(requestByteThresholdStr)) {
+      batchMutateBuilder.setRequestByteThreshold(Long.valueOf(requestByteThresholdStr));
+    }
+
+    ImmutableSet.Builder<StatusCode.Code> retryCodes = ImmutableSet.builder();
+    if (configuration.getBoolean(ENABLE_GRPC_RETRIES_KEY, true)) {
```
igorbernstein2: same as above...user cant disable  retries

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
igorbernstein2: argh, sorry I mis copy/pasted
I meant can we ignore `BIGTABLE_USE_SERVICE_ACCOUNTS_KEY`, we should keep BIGTABLE_SERVICE_ACCOUNT_JSON_VALUE_KEY

rahulKQL:
@igorbernstein2 Thanks a lot for your time to help me to fix and organize these changes.

I have added all the changes, PTAL!

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,594 @@
...
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.util.VersionInfo;
+import org.threeten.bp.Duration;
+
+/** For internal use only - public for technical reasons. */
+@InternalApi("For internal usage only")
+public class BigtableHBaseVeneerSettings extends BigtableHBaseSettings {
+
+  // Identifier to distinguish between CBT or GCJ adapter.
+  // TODO: find better name to identify veneer client
```
igorbernstein2: I think the user agent should look something like `hbase-x.y.z,java-bigtable-hbase-a.b.c`. I dont think we need to include veneer-adapter. `java-bigtable-hbase-a.b.c` implies that this is hbase adapter backed by veneer

Please create a helper`BIgtableHbaseVersion` that reads the version from bigtable-hbase/src/main/resources/com/google/cloud/bigtable/hbase/bigtable-hbase-version.properties

It should use the same mechanic as BigtableVersion (please ignore my previous comments about GaxProperties)

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,594 @@
...
+
+    configureHeaderProvider(stubSettings);
+
+    configureCredentialProvider(stubSettings);
+
+    return instanceAdminBuilder.build();
+  }
+
+  private void configureConnection(StubSettings.Builder stubSettings, String endpointKey) {
+    String adminHostOverride = configuration.get(endpointKey);
```
igorbernstein2: s/adminHostOverride/hostOverride

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,594 @@
...
+
+      retryCodes
+          .addAll(extractRetryCodesFromConfig())
+          .addAll(stubSettings.readRowsSettings().getRetryableCodes());
+
+      stubSettings.readRowsSettings().setRetryableCodes(retryCodes.build());
+
+      String initialElapsedBackoffMsStr = configuration.get(INITIAL_ELAPSED_BACKOFF_MILLIS_KEY);
+      if (!isNullOrEmpty(initialElapsedBackoffMsStr)) {
+        retryBuilder.setInitialRetryDelay(ofMillis(Long.parseLong(initialElapsedBackoffMsStr)));
```
igorbernstein2: Please check if initialRetryDelay > default maxRetryDelay. It it is, then set the max delay to initial? with a todo fix it in the future

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,594 @@
...
+      unaryCallSettings.setRetryableCodes(Collections.<StatusCode.Code>emptySet());
+    } else {
+
+      retryCodeBuilder
+          .addAll(extractRetryCodesFromConfig())
+          .addAll(unaryCallSettings.getRetryableCodes());
+
+      String initialElapsedBackoffMsStr = configuration.get(INITIAL_ELAPSED_BACKOFF_MILLIS_KEY);
+      if (!isNullOrEmpty(initialElapsedBackoffMsStr)) {
+        retryBuilder.setInitialRetryDelay(ofMillis(Long.parseLong(initialElapsedBackoffMsStr)));
```
igorbernstein2: Same as above, please make sure to increase the default max retry if initial is greater

igorbernstein2:
Also, please make this a chore, this isn't an end user feature

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
```diff
@@ -318,11 +337,7 @@ public Builder enableEmulator(String host, int port) {
     }
 
     public BigtableOptions build() {
-      if (options.bulkOptions == null) {
-        int maxInflightRpcs =
-            BulkOptions.BIGTABLE_MAX_INFLIGHT_RPCS_PER_CHANNEL_DEFAULT * options.dataChannelCount;
-        options.bulkOptions = BulkOptions.builder().setMaxInflightRpcs(maxInflightRpcs).build();
-      } else if (options.bulkOptions.getMaxInflightRpcs() <= 0) {
+      if (options.bulkOptions.getMaxInflightRpcs() <= 0) {
```
kolea2: can we still add a check for null and throw if it is not set?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
```diff
@@ -0,0 +1,380 @@
...
+    configureRetryOptions(bigtableOptionsBuilder);
+
+    String emulatorHost = configuration.get(BigtableOptionsFactory.BIGTABLE_EMULATOR_HOST_KEY);
+    if (!isNullOrEmpty(emulatorHost)) {
+      bigtableOptionsBuilder.enableEmulator(emulatorHost);
+    }
+
+    String useBatchStr = configuration.get(BIGTABLE_USE_BATCH);
+    if (!isNullOrEmpty(useBatchStr)) {
+      bigtableOptionsBuilder.setUseBatch(Boolean.parseBoolean(useBatchStr));
```
kolea2: please add a check here to ensure `useBatchStr` is a valid value

igorbernstein2: I don't think we should do that...hadoop's configuration class doesn't get valid boolean values, neither does parseBoolean.

kolea2: What is `useBatchStr` expected to evaluate to?

rahulKQL: `useBatchStr` is expected to be a boolean, the default value for this is `false`.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/BigtableOptions.java
rahulKQL: Thanks for the review, Added a Precondition check for bulk options.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
rahulKQL: Thanks again for the suggestion!

Have added this check to make sure initialRetryDelay & maxRetryDelay is same,
```java
  if (initialElapsedBackoffMs > retryBuilder.getMaxRetryDelay().toMillis()) {
    // TODO: fix this scenario by maybe introducing maxRetryDelayMillis directly
    retryBuilder.setMaxRetryDelay(ofMillis(initialElapsedBackoffMs));
  }
```

rahulKQL:
Sorry for the late entry to add a new getter.

I figured that we will have to expose [BufferedMutator#getWriteBufferSize](https://github.com/googleapis/java-bigtable-hbase/blob/cd64248427222d7f7faa2088db3c1c751a56d56a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java#L107-L111) to end-user, So it's best to have getter present in `BigtableHBaseSettings` class instead of type casting.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,620 @@
...
+  @Nullable private final BigtableInstanceAdminSettings instanceAdminSettings;
+
+  private final String dataHost;
+  private final String adminHost;
+  private final int port;
+  private final int bulkMaxRowKeyCount;
+  private final long batchingMaxMemory;
+  private final boolean isChannelPoolCachingEnabled;
+
+  public BigtableHBaseVeneerSettings(Configuration configuration) throws IOException {
```
igorbernstein2: Can you double check that this handles all of the same options as BigtableOptionsFactory? I just noticed that `ENABLE_GRPC_RETRY_DEADLINEEXCEEDED_KEY` is not handled here

rahulKQL: Thanks for pointing it out!! As discussed have updated this PR with more config keys.
Please have a look.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,695 @@
...
+
+    String appProfileId = configuration.get(APP_PROFILE_ID_KEY);
+    if (!isNullOrEmpty(appProfileId)) {
+      dataBuilder.setAppProfileId(appProfileId);
+    }
+
+    // added this check here to fail fast
+    if (Boolean.parseBoolean(configuration.get(BIGTABLE_BUFFERED_MUTATOR_ENABLE_THROTTLING))) {
+      throw new UnsupportedOperationException("Buffered mutator throttling is not supported.");
+    }
```
igorbernstein2: please move it to configureBulkMutationSettings

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -0,0 +1,695 @@
...
+          .addAll(extractRetryCodesFromConfig())
+          .addAll(stubSettings.readRowsSettings().getRetryableCodes());
+
+      stubSettings.readRowsSettings().setRetryableCodes(retryCodes.build());
+
+      String initialElapsedBackoffMsStr = configuration.get(INITIAL_ELAPSED_BACKOFF_MILLIS_KEY);
+      if (!isNullOrEmpty(initialElapsedBackoffMsStr)) {
+
+        long initialElapsedBackoffMs = Long.parseLong(initialElapsedBackoffMsStr);
+        if (initialElapsedBackoffMs > retryBuilder.getMaxRetryDelay().toMillis()) {
```
igorbernstein2: please move this to after `retryBuilder.setInitialRetryDelay`

rahulKQL:
@kolea2 Could you please have a look, It seems merging is blocked until your approval.

kolea2:
done, thank you!

PR #2397, opened 2020-03-04
feat: extend RowCell to include labels field

rahulKQL:
`RowCell` is an extension of HBase's Cell. This `labels` field introduced in this PR is intended to use for WhileMatchFilter implementation.

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/RowCell.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/RowCell.java
@@ -24,19 +24,12 @@ import org.apache.hadoop.hbase.KeyValue.Type;
 import org.apache.hadoop.hbase.util.Bytes;
 
 /**
- * This implementation of {@link Cell} is more efficient for Bigtable scanning than {@link
- * KeyValue}.
+ * RowCell is an alternative implementation of {@link KeyValue}. Unlike KeyValue, RowCell stores
+ * each value separately, which minimizes the number of copies that need to be made when unpacking
+ * the results from the Cloud Bigtable api.
  *
- * <p>RowCell is pretty straight forward.
- *
- * <ul>
- *   <li>Each Array() method returns the array passed in in the constructor.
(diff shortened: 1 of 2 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/RowCell.java
```diff
@@ -204,6 +236,10 @@ public int getTagsLength() {
     return Bytes.copy(this.rowArray);
   }
 
+  public List<String> getLabels() {
```
igorbernstein2: Please add some javadoc
```
Internal labels that were applied by a transformer filter. These labels are meant to be consumed by the bigtable-hbase scanner when implementing adapters like the WhileMatchFilterAdapter.
```

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/RowCell.java
```diff
@@ -22,11 +24,19 @@
...
 /**
- * This implementation of {@link org.apache.hadoop.hbase.Cell} is more efficient for Bigtable
- * scanning than {@link org.apache.hadoop.hbase.KeyValue} . RowCell is pretty straight forward. Each
- * *Array() method returns the array passed in in the constructor. Each *Offset() method returns 0.
- * Each *Length() returns the length of the array. This implementation is a few microseconds quicker
- * thank KeyValue, which makes a big performance difference for large scans.
+ * This implementation of {@link Cell} is more efficient for Bigtable scanning than {@link
+ * KeyValue}.
+ *
+ * <p>RowCell is pretty straight forward.
```
kolea2: nit - I would remove this

igorbernstein2: lol, I think something along these lines would be more helpful:

```
RowCell is an alternative implementation of KeyValue. Unlike KeyValue, RowCell stores each value separately, which minimizes the number of copies that need to be made when unpacking the results from the Cloud Bigtable api
```

PR #2398, opened 2020-03-04
refactor: move RowKeyWrapper from core to hbase module

rahulKQL:
This class is used before RPC, we need this class for various filter calculation.
 -  None of the class from `bigtable-core-client` referenced `RowKeyWrapper`, So removing it from source location.
 - RowKeyWrapper refers `ByteStringComparator#compare` for moved the compare logic in it.


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/util/RowKeyWrapper.java
```diff
@@ -34,6 +34,36 @@ public ByteString getKey() {
 
   @Override
   public int compareTo(RowKeyWrapper o) {
-    return ByteStringComparator.INSTANCE.compare(key, o.key);
+    return compare(key, o.key);
```
igorbernstein2: If you are inlining the compare, then we should a have unit test for it

igorbernstein2:
@rahulKQL can you resolve the conflict?

PR #2399, opened 2020-03-04
refactor: move RowKeyUtil from core to HBase module

rahulKQL:
Copied over `RowKeyUtil` from `bigtable-core-client` to `bigtable-hbase` utility. Also removed it from the source package.


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/util/RowKeyUtil.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.google.cloud.bigtable.util;
+package com.google.cloud.bigtable.hbase.util;
```
igorbernstein2: I don't think you can move this class. It's marked as InternalExtensionOnly. Instead, I think you have to copy it & add it your list of classes to deprecate in the future

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/util/RowKeyUtil.java
```diff
@@ -0,0 +1,66 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.hbase.util;
+
+import com.google.api.core.InternalExtensionOnly;
+import java.util.Arrays;
+
+@InternalExtensionOnly
```
igorbernstein2: Please make this InternalOnly

rahulKQL: Thanks for pointing it out. Updated with the same. PTAL

PR #2401, opened 2020-03-09
deps: add maven-enforcer-plugin

kolea2:
Prerequisite for #2382

### DIFF
--- a/bigtable-dataflow-parent/bigtable-beam-import/pom.xml
+++ b/bigtable-dataflow-parent/bigtable-beam-import/pom.xml
@@ -171,6 +171,14 @@ limitations under the License.
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <configuration>
+          <!--  todo align dependencies -->
+          <skip>true</skip>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
         <configuration>
           <source>1.8</source>
--- a/bigtable-dataflow-parent/bigtable-beam-import/pom.xml
+++ b/bigtable-dataflow-parent/bigtable-beam-import/pom.xml
@@ -173,7 +173,7 @@ limitations under the License.
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-enforcer-plugin</artifactId>
         <configuration>
-          <!--  todo align dependencies -->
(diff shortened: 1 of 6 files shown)


bigtable-dataflow-parent/bigtable-beam-import/pom.xml
```diff
@@ -169,6 +169,14 @@ limitations under the License.
 
   <build>
     <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <configuration>
+          <!--  todo align dependencies -->
```
igorbernstein2: nit. to stay consistent please use TODO(github name): message

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/pom.xml
```diff
@@ -103,6 +103,37 @@ limitations under the License.
 
   <build>
     <plugins>
+      <plugin>
```
igorbernstein2: Please add a comment that you are overriding the default config and why

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-shaded/pom.xml
```diff
@@ -48,6 +48,11 @@ limitations under the License.
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
+        <groupId>org.slf4j</groupId>
+        <artifactId>slf4j-api</artifactId>
+        <version>${slf4j.version}</version>
+      </dependency>
```
igorbernstein2: I think you need to an entry for the other slf4j dependency here as well

bigtable-hbase-2.x-parent/bigtable-hbase-2.x-hadoop/pom.xml
```diff
@@ -104,6 +104,25 @@ limitations under the License.
 
   <build>
     <plugins>
+      <plugin>
```
igorbernstein2: Please add a comment that you are overwriting the default config and why

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-shaded/pom.xml
kolea2: good catch

kolea2: actually, because we changed the version to 1.7.25, this is no longer an issue, but can add for consistency

PR #2410, opened 2020-03-12
build: disable current releasing logic in preparation for release-please

kolea2:
Towards #2382

### DIFF
--- a/bigtable-dataflow-parent/bigtable-hbase-beam/pom.xml
+++ b/bigtable-dataflow-parent/bigtable-hbase-beam/pom.xml
@@ -165,48 +165,6 @@ limitations under the License.
         </plugins>
       </build>
     </profile>
-    <profile>
-      <id>release</id>
-      <build>
-        <plugins>
-          <plugin>
-            <groupId>org.apache.maven.plugins</groupId>
-            <artifactId>maven-javadoc-plugin</artifactId>
-            <configuration>
-              <windowtitle>
-                Apache Beam + Cloud Bigtable Connector ${project.version} API
-              </windowtitle>
-              <doctitle>
-                Apache Beam + Cloud Bigtable Connector ${project.version} API
-              </doctitle>
-              <overview>../overview.html</overview>
-              <bottom><![CDATA[<br>]]></bottom>
-
-              <offlineLinks>
-                <offlineLink>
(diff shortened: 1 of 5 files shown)


bigtable-dataflow-parent/bigtable-hbase-beam/pom.xml
```diff
@@ -165,48 +165,6 @@ limitations under the License.
         </plugins>
       </build>
     </profile>
-    <profile>
-      <id>release</id>
-      <build>
-        <plugins>
-          <plugin>
-            <groupId>org.apache.maven.plugins</groupId>
-            <artifactId>maven-javadoc-plugin</artifactId>
```
igorbernstein2: I dont think you can just remove this...the javadoc config needs to exist somewhere. Maybe move it to pluginManagement?

bigtable-hbase-1.x-parent/pom.xml
```diff
@@ -53,32 +53,5 @@ limitations under the License.
         <module>bigtable-hbase-1.x-hadoop</module>
       </modules>
     </profile>
-    <profile>
-      <id>release</id>
-      <build>
-        <plugins>
-          <plugin>
-            <groupId>org.apache.maven.plugins</groupId>
```
igorbernstein2: same as above

bigtable-dataflow-parent/bigtable-hbase-beam/pom.xml
kolea2: Thank you, meant to ask about this. I'm thinking I should model it to how java-bigtable does javadoc, given they will eventually have the same parent and configuration? https://github.com/googleapis/java-bigtable/blob/master/pom.xml#L215

igorbernstein2: I think moving the config to the root pom is fine, but please add that in this PR.
When you say same parent & configuration, do you mean google-cloud-shared-config?

kolea2: Sorry for being unclear, I simply meant the structure - I'm planning on leaving the configs in the same pom files. And yes to question 2

kolea2:
@googlebot

PR #2412, opened 2020-03-12
build: add and run synthtool

kolea2:
Towards #2382 

Note - the only handwritten files are `synth.py` and `.repo-metadata.json`. All other files are generated as a result of running `synthtool`.

### DIFF
--- a/.repo-metadata.json
+++ b/.repo-metadata.json
@@ -1,0 +1,13 @@
+{
+  "name": "bigtable-hbase",
+  "name_pretty": "Google Cloud Bigtable Hbase Client",
+  "product_documentation": "https://cloud.google.com/bigtable",
+  "client_documentation": "https://googleapis.dev/java/java-bigtable-hbase/latest/",
+  "issue_tracker": "https://issuetracker.google.com/savedsearches/559782",
+  "release_level": "ga",
+  "language": "java",
+  "repo": "googleapis/java-bigtable-hbase",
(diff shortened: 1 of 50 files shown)


.repo-metadata.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "bigtable-hbase",
```
igorbernstein2: what is this name used for? Just pattern matching this looks different from the name in the native client:

In the native client:
repo name: java-bigtable
metadata name: google-cloud-bigtable

Here:
repo name: java-bigtable-hbase
metadata name: bigtable-hbase

.repo-metadata.json
```diff
@@ -0,0 +1,13 @@
...
+  "name_pretty": "Google Cloud Bigtable Hbase Client",
```
igorbernstein2: This also doesn't seem in line with the native client

kolea2: This is only used for README generation, which we are skipping

.repo-metadata.json
kolea2: This was only used for generating the bug report template. https://github.com/googleapis/java-bigtable-hbase/pull/2412/commits/30c6aa041f70e56c5b9412aef03e4d0f7880b76b#diff-74c2649c999338ed7a2833d79d4b25c3R20. I've cleaned that up a little and added it to exclude from the generation.

PR #2419, opened 2020-03-13
deps: fix undeclared used dependencies

### LINKED ISSUE #2382: Enable release-please

kolea2:
Towards #2382

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/pom.xml
+++ b/bigtable-client-core-parent/bigtable-client-core/pom.xml
@@ -261,14 +261,6 @@ this is not used directly, but upgrading due to transitive vulnerabilities in ol
     </resources>
     <plugins>
       <plugin>
-        <!-- skip for bigtable-client-core to allow for upcoming refactoring-->
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>clirr-maven-plugin</artifactId>
-        <configuration>
-          <skip>true</skip>
-        </configuration>
-      </plugin>
-      <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
--- a/bigtable-client-core-parent/bigtable-client-core/pom.xml
+++ b/bigtable-client-core-parent/bigtable-client-core/pom.xml
@@ -261,6 +261,14 @@ this is not used directly, but upgrading due to transitive vulnerabilities in ol
     </resources>
     <plugins>
       <plugin>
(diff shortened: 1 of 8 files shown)


bigtable-client-core-parent/bigtable-client-core/pom.xml
```diff
@@ -260,6 +260,14 @@ this is not used directly, but upgrading due to transitive vulnerabilities in ol
       </resource>
     </resources>
     <plugins>
+      <plugin>
+        <!-- skip for bigtable-client-core to allow for upcoming refactoring-->
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>clirr-maven-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
+        </configuration>
```
igorbernstein2: Not for this PR, but I think it would be really nice to write a maven plugin that generates an [ignoredDifferencesFile](https://www.mojohaus.org/clirr-maven-plugin/check-mojo.html#ignoredDifferencesFile) for clirr that contains everything annotated with InternalApi.

But for now, it seems like we are disabling clirr for every module. Why not just put it in the root pom?

bigtable-hbase-2.x-parent/bigtable-hbase-2.x/pom.xml
```diff
@@ -84,6 +84,10 @@ limitations under the License.
       <groupId>io.grpc</groupId>
       <artifactId>grpc-stub</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-api</artifactId>
+    </dependency>
```
igorbernstein2: Please separate changes that might change the behavior of the client from ones that dont. ie the clirr plugin doesn't affect endusers, while adding new dependencies will.

bigtable-client-core-parent/bigtable-client-core/pom.xml
kolea2: 👍 that sounds like a good plan. I'd prefer to have it granularly by each module, so that when we're done with the refactorings, we can simply remove it rather than adding it back in with other exclusions (like for shaded poms). Let me know if you disagree!

bigtable-hbase-2.x-parent/bigtable-hbase-2.x/pom.xml
kolea2: sg, will fix deps first

kolea2:
@igorbernstein2, the dependency presubmit check is currently failing because I assume it is using an older version of `dependency:analyze` - the changes in this PR are correct. Once we add in the parent shared config, this error will resolve itself.

PR #2421, opened 2020-03-16
chore: adds adapter to transform protobuf Row to hbase Result

rahulKQL:
This change adds new RowAdapter, to transform protobuf RowAdapter to hbase's Result object.

Note: If any `com.google.bigtable.v2.Cell`s contains `label`s then we are treating those cells as internal only.


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ProtoRowAdapter.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.List;
+import org.apache.hadoop.hbase.client.Result;
+
+/**
+ * Adapt between a {@link Row} and an hbase client {@link Result}.
+ *
+ * <p>For internal use only - public for technical reasons.
+ */
+@InternalApi("For internal usage only")
+public class ProtoRowAdapter implements ResponseAdapter<Row, Result> {
```
rahulKQL: I hope this name is ok to distinguish between different types of Row. As RowAdapter is already present in the same package.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/Adapters.java
```diff
@@ -62,6 +63,8 @@
           new BigtableWhileMatchResultScannerAdapter(FLAT_ROW_ADAPTER);
   /** Constant <code>GET_ADAPTER</code> */
   public static final GetAdapter GET_ADAPTER = new GetAdapter(SCAN_ADAPTER);
+  /** Constant <code>PROTO_ROW_ADAPTER</code> */
+  public static final ProtoRowAdapter PROTO_ROW_ADAPTER = new ProtoRowAdapter();
```
igorbernstein2: I don't think this belongs here. Conversion to Result should be private to the `DataClientWrapper`. `Adapters` is a layer above

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ProtoRowAdapter.java
```diff
@@ -0,0 +1,76 @@
...
+    byte[] rowKeyBytes = ByteStringer.extract(row.getKey());
+
+    for (Family family : row.getFamiliesList()) {
+      byte[] familyBytes = ByteStringer.extract(family.getNameBytes());
+
+      for (Column column : family.getColumnsList()) {
+        byte[] qualifierBytes = ByteStringer.extract(column.getQualifier());
+
+        for (Cell cell : column.getCellsList()) {
+          // Cells with labels are treated as internal cells, not sending them to user.
```
igorbernstein2: I thought we agreed that labels need to be able to propagate out of the wrapper and the labels will get filtered out by the whilescanadapter? ie. the work you started here:
https://github.com/googleapis/java-bigtable-hbase/pull/2397

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ProtoRowAdapter.java
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  private static RowCell toRowCell(byte[] rowKey, byte[] family, byte[] qualifier, Cell cell) {
+    return new RowCell(
+        rowKey,
+        family,
+        qualifier,
+        // Bigtable timestamp has more granularity than HBase one. It is possible that Bigtable
+        // cells are deduped unintentionally here. On the other hand, if we don't dedup them,
+        // HBase will treat them as duplicates.
```
igorbernstein2: I dont understand what this comment means. How we are deduping them. Also where/when will hbase treat them as duplicates?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ProtoRowAdapter.java
igorbernstein2: I dont think this needs to be a standalone class. I think it can just be a private method in the wrapper impl

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ProtoRowAdapter.java
rahulKQL: Yes, we are going to apply whileMatchFilter for rows returned by readRows() operation(i.e. `ResultScanner<Result>`).

But `readModifyWriteRowAsync()` returns the response directly without any filtering logic, Should we apply the filter on this operation as well. something like
```java
@Override
  public ApiFuture<Result> readModifyWriteRowAsync(ReadModifyWriteRow readModifyWriteRow) {
    ListenableFuture<ReadModifyWriteRowResponse> response =
        delegate.readModifyWriteRowAsync(readModifyWriteRow.toProto(requestContext));
    return ApiFutureUtil.transformAndAdapt(
        response,
        new Function<ReadModifyWriteRowResponse, Result>() {
          @Override
          public Result apply(ReadModifyWriteRowResponse response) {
            return Adapters.BIGTABLE_WHILE_MATCH_FILTER_ADAPTER.adapt(
                protoRowToResultUtil(response.getRow()));
          }
        });
  }
```

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ProtoRowAdapter.java
rahulKQL: >How we are deduping them.

If a user starts cells with reverse timestamp then we are deduping those timestamp to HConstants.LATEST_TIMESTAMP.

>Also where/when will hbase treat them as duplicates?

If cells have timestamp differences of only microseconds then HBase would treat 2nd and remaining cells until the next millisecond as duplicate.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ProtoRowAdapter.java
igorbernstein2: I dont whilematchfilter & labels make sense for ReadModifyWrite

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ProtoRowAdapter.java
igorbernstein2: Lets chat offline

PR #2422, opened 2020-03-16
chore: adds BigtableWrapper which with classic implementation

rahulKQL:
This change contains `BigtableWrapper`, The intention with this class is to replace `BigtableSession` references across `bigtable-hbase` & it's child modules.

Note: BigtableInstanceAdmin is not yet exposed through BigtableWrapper(we may add it once migrating of data & admin operations is finished).

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableClassicApi.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableClassicApi.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Google LLC.
+ * Copyright 2020 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableClassicApi.java
```diff
@@ -0,0 +1,74 @@
...
+import java.io.IOException;
+
+/** For internal use only - public for technical reasons. */
+@InternalApi("For internal usage only")
+public class BigtableClassicApi extends BigtableWrapper {
+
+  private final BigtableSession bigtableSession;
+  private final DataClientWrapper dataClientWrapper;
+
+  // Delaying the admin wrapper instantiation until its needed
```
igorbernstein2: Why?

rahulKQL: I was following the BigtableSession [design](https://github.com/googleapis/java-bigtable-hbase/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java#L737-L761) for AdminClient initialization. I'll update it to initialize it in the constructor.

igorbernstein2: But BigtableSession will initialize a connection to bigtableadmin.googleapis.com regardless. I don't think this is buying us anything. Maybe in the future we can optimize this. But for now, lets keep it simple

PR #2427, opened 2020-03-16
chore: part 1 of replacing BigtableOptions to BigtableHBaseSettings

rahulKQL:
This is part-1 of series of pull requests.

This change introduces a getter in the `CommonConnection` interface, which would be used across `bigtable-hbase` and it's child module to fetch bigtable settings.


bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -72,7 +73,7 @@
   private BigtableSession session;
 
   private volatile boolean cleanupPool = false;
-  private final BigtableOptions options;
+  private final BigtableHBaseSettings baseSettings;
```
igorbernstein2: why base settings? I think just settings should be fine

PR #2435, opened 2020-03-18
chore: part 3 of replacing BigtableOptions to BigtableHBaseSettings

rahulKQL:
This change updates `AbstractBigtableRegionLocator` and `AbstractBigtableTable` and it's references.

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableRegionLocator.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableRegionLocator.java
@@ -20,12 +20,13 @@ import com.google.api.core.ApiFuture;
 import com.google.api.core.ApiFutureCallback;
 import com.google.api.core.ApiFutures;
 import com.google.api.core.InternalApi;
-import com.google.cloud.bigtable.config.BigtableOptions;
 import com.google.cloud.bigtable.core.IBigtableDataClient;
+import com.google.cloud.bigtable.data.v2.internal.NameUtil;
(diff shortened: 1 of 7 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableRegionLocator.java
```diff
@@ -56,11 +57,16 @@
...
 
   public AbstractBigtableRegionLocator(
-      TableName tableName, BigtableOptions options, IBigtableDataClient client) {
+      TableName tableName, BigtableHBaseSettings settings, IBigtableDataClient client) {
     this.tableName = tableName;
     this.client = client;
-    this.bigtableTableName = options.getInstanceName().toTableName(tableName.getNameAsString());
-    ServerName serverName = ServerName.valueOf(options.getDataHost(), options.getPort(), 0);
+    this.bigtableTableName =
+        new BigtableTableName(
```
kolea2: how frequently do we need to do this operation? Is it worth adding this as a getter on `BigtableHbaseSettings` (similar to the convenience method `toTableName()`)?

rahulKQL: This is an intermediate state, We would be able to replace `BigtableTableName` with string `tableID` once the veneer client wrappers are in place.

kolea2: sounds good, thanks!

rahulKQL:
rebased and added a unit to fix CodeCov build, which had -0.02% coverage change.

PR #2443, opened 2020-03-19
chore: introducing BigtableApi with classic client implementation(Part 1)

rahulKQL:
This change performs the following tasks:
- Renames BigtableWrapper to BigtableApi.
- Introduced new constructor in BigtableApi to accept BigtableSession.
- Added BigtableApi getter in CommonConnection, This should replace BigtableSession getter in future.
- Added unit tests for AbstractBigtableConnection

This is part 1 of a series of changes, with an end goal to replace `BigtableSession` to `BigtableApi`.

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/BigtableApi.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/BigtableApi.java
@@ -27,11 +27,11 @@ import java.io.IOException;
  * <p>For internal use only - public for technical reasons.
  */
 @InternalApi("For internal usage only")
-public abstract class BigtableWrapper implements AutoCloseable {
+public abstract class BigtableApi implements AutoCloseable {
 
   private final BigtableHBaseSettings hBaseSettings;
 
-  public static BigtableWrapper create(BigtableHBaseSettings settings) throws IOException {
+  public static BigtableApi create(BigtableHBaseSettings settings) throws IOException {
     if (settings instanceof BigtableHBaseVeneerSettings) {
       throw new UnsupportedOperationException("Veneer client is not yet supported.");
     } else {
@@ -39,7 +39,7 @@ public abstract class BigtableWrapper implements AutoCloseable {
     }
   }
 
-  protected BigtableWrapper(BigtableHBaseSettings hbaseSettings) {
+  protected BigtableApi(BigtableHBaseSettings hbaseSettings) {
     this.hBaseSettings = hbaseSettings;
   }
 
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableClassicApi.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableClassicApi.java
@@ -20,32 +20,37 @@ import com.google.cloud.bigtable.data.v2.internal.RequestContext;
 import com.google.cloud.bigtable.grpc.BigtableInstanceName;
 import com.google.cloud.bigtable.grpc.BigtableSession;
 import com.google.cloud.bigtable.hbase.wrappers.AdminClientWrapper;
-import com.google.cloud.bigtable.hbase.wrappers.BigtableWrapper;
+import com.google.cloud.bigtable.hbase.wrappers.BigtableApi;
+import com.google.cloud.bigtable.hbase.wrappers.BigtableHBaseSettings;
 import com.google.cloud.bigtable.hbase.wrappers.DataClientWrapper;
 import java.io.IOException;
 
 /** For internal use only - public for technical reasons. */
 @InternalApi("For internal usage only")
-public class BigtableClassicApi extends BigtableWrapper {
+public class BigtableClassicApi extends BigtableApi {
 
   private final BigtableSession bigtableSession;
   private final DataClientWrapper dataClientWrapper;
   private final AdminClientWrapper adminClientWrapper;
 
   public BigtableClassicApi(BigtableHBaseClassicSettings settings) throws IOException {
+    this(settings, new BigtableSession(settings.getBigtableOptions()));
+  }
+
+  public BigtableClassicApi(BigtableHBaseSettings settings, BigtableSession session)
+      throws IOException {
     super(settings);
-    this.bigtableSession = new BigtableSession(settings.getBigtableOptions());
+    this.bigtableSession = session;
 
     RequestContext requestContext =
         RequestContext.create(
             settings.getProjectId(),
             settings.getInstanceId(),
-            settings.getBigtableOptions().getAppProfileId());
+            session.getOptions().getAppProfileId());
     this.dataClientWrapper = new DataClientClassicApi(bigtableSession, requestContext);
 
     BigtableInstanceName instanceName =
-        new BigtableInstanceName(
-            getBigtableHBaseSettings().getProjectId(), getBigtableHBaseSettings().getInstanceId());
+        new BigtableInstanceName(settings.getProjectId(), settings.getInstanceId());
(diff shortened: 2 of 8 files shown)


rahulKQL:
@igorbernstein2 / @kolea2  Should we introduce unit tests for these classes as the coverage difference is showing `-0.10%`? These classes does not have unit tests because it has been tested through Integration test cases.

kolea2:
> @igorbernstein2 / @kolea2 Should we introduce unit tests for these classes as the coverage difference is showing `-0.10%`? These classes does not have unit tests because it has been tested through Integration test cases.

Yes please, that would be great!

bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -70,7 +73,9 @@
   private volatile ExecutorService batchPool = null;
   private ExecutorService bufferedMutatorExecutorService;
 
-  private BigtableSession session;
+  private final BigtableSession session;
+  // TODO: remove this when introducing BigtableWrapper, once incremental changes are in place
+  private final DataClientWrapper dataClient;
```
igorbernstein2: If I'm not mistaken I think the final goal is to replace BigtableSession with BigtableWrapper?
If that's true, shouldn't we add BigtableWrapper here?  We can add a temporary constructor to it that takes a prebuilt BigtableSession. 

I think we will end up with a lot less noise this way

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/BigtableApi.java
```diff
@@ -27,19 +27,19 @@
  * <p>For internal use only - public for technical reasons.
  */
 @InternalApi("For internal usage only")
-public abstract class BigtableWrapper implements AutoCloseable {
+public abstract class BigtableApi implements AutoCloseable {
```
rahulKQL: @igorbernstein2 
Please let me know if this name sounds ok to you. Somehow I felt this sounds better. If not then I can revert back.

Also `DataClientWrapper` & `AdminClientWrapper` has getters with `BigtableApi#getDataClient` & `BigtableApi#getAdminClient`.

bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
rahulKQL: Thanks for the suggestion! I have updated this PR to follow the same approach. PTAL

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/BigtableApi.java
igorbernstein2: Sounds good either way. I don't have strong feelings

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableClassicApi.java
```diff
@@ -20,32 +20,37 @@
...
 
   private final BigtableSession bigtableSession;
   private final DataClientWrapper dataClientWrapper;
   private final AdminClientWrapper adminClientWrapper;
 
   public BigtableClassicApi(BigtableHBaseClassicSettings settings) throws IOException {
+    this(settings, new BigtableSession(settings.getBigtableOptions()));
+  }
+
+  public BigtableClassicApi(BigtableHBaseSettings settings, BigtableSession session)
```
igorbernstein2: Please add a comment that this is temporary and should be removed after the transition to the new wrappers

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/classic/TestBigtableClassicApi.java
```diff
@@ -69,29 +71,43 @@ public void setUp() throws IOException {
...
+    configuration.set(BigtableOptionsFactory.BIGTABLE_NULL_CREDENTIAL_ENABLE_KEY, "true");
     configuration.set(BigtableOptionsFactory.BIGTABLE_DATA_CHANNEL_COUNT_KEY, "1");
     configuration.set(BigtableOptionsFactory.BIGTABLE_EMULATOR_HOST_KEY, "localhost:" + port);
     bigtableHBaseSettings = BigtableHBaseClassicSettings.create(configuration);
-    bigtableWrapper = BigtableWrapper.create(bigtableHBaseSettings);
+    bigtableApi = BigtableApi.create(bigtableHBaseSettings);
+  }
+
+  @Test
+  public void testWithBigtableSession() throws Exception {
```
igorbernstein2: I don't  really see what this is testing.

bigtable-client-core-parent/bigtable-hbase/src/test/java/org/apache/hadoop/hbase/client/TestAbstractBigtableConnection.java
```diff
@@ -0,0 +1,247 @@
...
+    connection = new TestBigtableConnectionImpl(configuration);
+  }
+
+  @After
+  public void tearDown() throws IOException {
+    connection.close();
+  }
+
+  @Test
+  public void testGetter() throws IOException {
```
igorbernstein2: I dont understand what this is testing.

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/classic/TestBigtableClassicApi.java
rahulKQL: This is to maintain or increase the codeCov percentage. If not these than one of the CodeCov pre-submit fails.

I am not sure if this is an acceptable workaround. Please let me know.

bigtable-client-core-parent/bigtable-hbase/src/test/java/org/apache/hadoop/hbase/client/TestAbstractBigtableConnection.java
rahulKQL: These are for getters present in AbstractBigtableConnection. Again I will try to update it with happy/failure paths.

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/classic/TestBigtableClassicApi.java
igorbernstein2: But this isn't a meaningful test. So the increased coverage isn't real.

I think it would be more meaningful to stand up a grpc server and send a send a request to each of the apis. Maybe a  MutateRow , a GetInstance and a GetTable request? Just to make sure that settings are getting propagated and that the clients are able to connect to a backend

rahulKQL: Thanks for the suggestion, updated with the same test logic. 😄

bigtable-client-core-parent/bigtable-hbase/src/test/java/org/apache/hadoop/hbase/client/TestAbstractBigtableConnection.java
rahulKQL: I have removed this test case as it was not unnecessary.

PR #2447, opened 2020-03-23
chore: add small fixes from various module

rahulKQL:
This change contains following small fixes:
 - Fix for `BigtableConnection#getAllRegionInfos` and unit test case for the same.
 - Small fix for `AbstractBigtableTable`.
 - Moved `TestTableAdapter2x` class to appropriate package.
 - Added dependencies used for performing test.
 - Moved `FlatRowAdapter#adaptToRow` under `TestFlatRowAdapter`.

### DIFF
--- a/bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
+++ b/bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
@@ -105,5 +105,31 @@ limitations under the License.
       <version>${junit.version}</version>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-api</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-stub</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.api.grpc</groupId>
+      <artifactId>proto-google-cloud-bigtable-v2</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-bigtable</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.api.grpc</groupId>
+      <artifactId>grpc-google-cloud-bigtable-v2</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 </project>
--- a/bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
+++ b/bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
@@ -92,6 +92,35 @@ limitations under the License.
       <version>${dropwizard.metrics.version}</version>
     </dependency>
 
+    <!-- These are direct dependencies of TestBigtableConnection.java which is in test scope,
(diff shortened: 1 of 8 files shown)


bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
```diff
@@ -105,5 +105,31 @@ limitations under the License.
       <version>${junit.version}</version>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-api</artifactId>
+      <scope>test</scope>
```
igorbernstein2: Although this module doesn't reference grpc-api directly, it's declared as a compile time dep of bigtable-client-core. So if you override it to test here it will be missing during runtime.

Please mark this as compile

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
```diff
@@ -105,5 +105,31 @@ limitations under the License.
...
+    </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-stub</artifactId>
+      <scope>test</scope>
```
igorbernstein2: compile scope

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
```diff
@@ -105,5 +105,31 @@ limitations under the License.
...
+    </dependency>
+    <dependency>
+      <groupId>com.google.api.grpc</groupId>
+      <artifactId>proto-google-cloud-bigtable-v2</artifactId>
+      <scope>test</scope>
```
igorbernstein2: compile scope

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
```diff
@@ -105,5 +105,31 @@ limitations under the License.
...
+    </dependency>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-bigtable</artifactId>
+      <scope>test</scope>
```
igorbernstein2: compile

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
```diff
@@ -105,5 +105,31 @@ limitations under the License.
...
+    </dependency>
+    <dependency>
+      <groupId>com.google.api.grpc</groupId>
+      <artifactId>grpc-google-cloud-bigtable-v2</artifactId>
+      <scope>test</scope>
```
igorbernstein2: compile

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
rahulKQL: I just wanted to clear before changing the scope. We need these dependencies for `TestBigtableConnection.java` to for the Server for FakeDataService and BigtableGrpc.

igorbernstein2: These are direct dependencies of TestBigtableConnection which is in test scope, but they are transitive dependencies of bigtable-client-core, which needs them during runtime. 
test scope inherits compile scope, which inherits runtime scope.

So runtime scope would probably be the most correct. But I'm not sure how that will play with shading, so I'm suggesting compile scope to be safe. Maybe add a comment explaining this next to each declaration

Issue #2454, opened 2020-03-24
chore: Transition of existing client's references to new wrappers


rahulKQL:
## Background
To remove dependency from `bigtable-core-client` we have moved some of the classes and introduced fresh wrappers in the `bigtable-hbase` module. The new wrappers can be found under [`bigtable-hbase`'s wrapper package](https://github.com/googleapis/java-bigtable-hbase/tree/95883e748f3cf37384f243b68ff0b30c2364b245/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers). 

This issue is to track the transition of BigtableSession, IBigtableDataClient, IBigtableTableAdminClient to BigtbaleApi, DataClientWrapper, AdminClientWrapper respectively.

- [x] Move important classes from `bigtable-core-client` to `bigtable-hbase`.
- [x] introduce Settings to switch between classic and veneer clients.
- [x] Introduce new wrapper interfaces.
- [x] Add Classic client implementation.
- [x] Align defaults between classic and veneer.
- [x] Replace references of Session/Data/Admin client with new wrappers.
- [x] introduce veneer implementation for the wrappers.
- [ ] perform benchmark tests with classic as well as veneer client.

PR #2459, opened 2020-03-24
chore: replacing bulk read with new wrapper (Part-4)

rahulKQL:
Towards #2454 

This change includes the transition of BulkRead to BulkReadWrapper in BatchExecutor.

Note: This is part 4 of the series of PRs to replace references of BigtableSession/Data/Admin clients with new wrappers.

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
@@ -26,7 +26,9 @@ import com.google.cloud.bigtable.grpc.scanner.FlatRow;
 import com.google.cloud.bigtable.hbase.adapters.Adapters;
 import com.google.cloud.bigtable.hbase.adapters.HBaseRequestAdapter;
 import com.google.cloud.bigtable.hbase.util.Logger;
+import com.google.cloud.bigtable.hbase.wrappers.BigtableApi;
 import com.google.cloud.bigtable.hbase.wrappers.BigtableHBaseSettings;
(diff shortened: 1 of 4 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
```diff
@@ -103,6 +105,8 @@ public final void onSuccess(Object message) {
           result =
               Adapters.ROW_ADAPTER.adaptResponse(
                   (com.google.cloud.bigtable.data.v2.models.Row) message);
+        } else if (message instanceof Result) {
+          result = (Result) message;
```
igorbernstein2: Shouldnt this always be a Result? when would we see FlatRow or Row?

rahulKQL: As of this change, we have yet to migrate `BufferedMutatorHelper` with new wrappers, 

https://github.com/googleapis/java-bigtable-hbase/blob/fafa313acbe30374346b8264297edbea4b73d8d1/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java#L191-L193

so it is a possibility that it may call readModifyWriteRow() for mutation.
https://github.com/googleapis/java-bigtable-hbase/blob/fafa313acbe30374346b8264297edbea4b73d8d1/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutatorHelper.java#L205-L210

igorbernstein2: got it, thanks

PR #2460, opened 2020-03-25
chore: replacing bulk read with bulk read wrapper (Part-5)

rahulKQL:
Towards #2454

This PR contains following changes:
  - transition of data client in BigtableBufferedMutator, BufferedMutatorHelper.
  - Also BulkMutation to new wrapper.
  - updated BatchExecutor constructor to remove BigtableSession dependency.
  - remove unneeded FlatRow checks and revisited unit test for BatchExecutor.

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
@@ -92,14 +92,14 @@ public class BatchExecutor {
     @SuppressWarnings("unchecked")
     @Override
     public final void onSuccess(Object message) {
-      Result result = Result.EMPTY_RESULT;
-      try {
-        if (message instanceof Result) {
-          result = (Result) message;
-        }
-      } catch (Throwable throwable) {
-        onFailure(throwable);
-        return;
+      final Result result;
+
+      if (message instanceof Result) {
+        // Handle the completion of Gets
+        result = (Result) message;
+      } else {
+        // Handles the completion of other operations like Deletes & Puts that don't return anything
+        result = Result.EMPTY_RESULT;
       }
 
       resultsArray[index] = result;
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
@@ -131,17 +131,17 @@ public class BatchExecutor {
   /**
    * Constructor for BatchExecutor.
    *
-   * @param bigtableApi a {@link BigtableApi} object.
-   * @param requestAdapter a {@link HBaseRequestAdapter} object.
+   * @param bigtableApi a {@link BigtableApi} object to access bigtable data client.
+   * @param settings a {@link BigtableHBaseSettings} object for bigtable settings.
+   * @param adapter a {@link HBaseRequestAdapter} object to convert HBase object to Bigtable protos.
    */
-  public BatchExecutor(BigtableApi bigtableApi, HBaseRequestAdapter requestAdapter) {
-    this.requestAdapter = requestAdapter;
-    this.settings = bigtableApi.getBigtableHBaseSettings();
+  public BatchExecutor(
+      BigtableApi bigtableApi, BigtableHBaseSettings settings, HBaseRequestAdapter adapter) {
+    this.requestAdapter = adapter;
+    this.settings = settings;
     this.bulkRead =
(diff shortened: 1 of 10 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
```diff
@@ -97,15 +93,8 @@ public RpcResultFutureCallback(
...
-
       try {
-        if (message instanceof FlatRow) {
-          result = Adapters.FLAT_ROW_ADAPTER.adaptResponse((FlatRow) message);
-        } else if (message instanceof com.google.cloud.bigtable.data.v2.models.Row) {
-          result =
-              Adapters.ROW_ADAPTER.adaptResponse(
-                  (com.google.cloud.bigtable.data.v2.models.Row) message);
-        } else if (message instanceof Result) {
+        if (message instanceof Result) {
```
igorbernstein2: What else can this be?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutator.java
```diff
@@ -65,18 +64,16 @@
...
       HBaseRequestAdapter adapter,
-      BigtableHBaseSettings settings,
-      BigtableSession session,
+      BigtableApi bigtableApi,
       BufferedMutator.ExceptionListener listener) {
-    helper = new BigtableBufferedMutatorHelper(adapter, settings, session);
+    helper = new BigtableBufferedMutatorHelper(adapter, bigtableApi);
     this.listener = listener;
-    this.host = settings.getDataHost();
+    this.host = bigtableApi.getBigtableHBaseSettings().getDataHost();
```
igorbernstein2: I'd prefer for the bigtable api not expose it's settings. Please pass the settings in parallel to bigtableApi

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutatorHelper.java
```diff
@@ -65,25 +65,23 @@
...
-   * @param settings For bigtable settings
-   * @param session a {@link BigtableSession} object.
+   * @param bigtableApi For bigtable settings and data client
    */
-  public BigtableBufferedMutatorHelper(
-      HBaseRequestAdapter adapter, BigtableHBaseSettings settings, BigtableSession session) {
+  public BigtableBufferedMutatorHelper(HBaseRequestAdapter adapter, BigtableApi bigtableApi) {
     this.adapter = adapter;
-    this.settings = settings;
+    this.settings = bigtableApi.getBigtableHBaseSettings();
```
igorbernstein2: Same as above

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
rahulKQL: It could be Void as well from BufferedMutatorHelper#mutate:

https://github.com/googleapis/java-bigtable-hbase/blob/792fe728b6feaa67ee3229f2ada22e56a5eb3e69/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutatorHelper.java#L192-L203

igorbernstein2: Looking at this closer, why do we even need the if block or the try block? They are basically no-ops

rahulKQL: We would need these Result for AbstractBigtableTable#get(List) operation which is HBase's Table method. Internally this batch is used for AbstractBigtbaleTable's batch(List), delete(List), get(List), put(List) operation.

igorbernstein2: I see, please update it to make it clearer:

```java
final Result result;

if (message instanceof Result) {
   // Handle the completion of Gets
   result = (Result) message;
} else {
  // Handle the completion of other operations like Deletes & Puts that don't return anything
   result = Rest.EMPTY_RESULT
}
```

igorbernstein2: The try isn't necessary because nothing can fail

Issue #2448, opened 2020-03-23
Clean up tests with unused mockito calls


kolea2:
Example from logs:
```
[MockitoHint] TestRetryingUnaryOperation.testOK (see javadoc for MockitoHint):
[MockitoHint] 1. Unused -> at com.google.cloud.bigtable.grpc.async.TestRetryingUnaryOperation.setup(TestRetryingUnaryOperation.java:74)
[MockitoHint] 2. Unused -> at com.google.cloud.bigtable.grpc.async.OperationClock.initializeMockSchedule(OperationClock.java:81)
```

PR #2461, opened 2020-03-25
chore(fix): refactored mocking in unit test case.

rahulKQL:
Towards #2448 

This change updates `bigtable-core-client` module's unit tests to reduces unused mocking.

This is part 1 of the series of PRs towards reducing unused mocking.

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/TestAppProfile.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/TestAppProfile.java
@@ -51,8 +51,10 @@ import java.net.ServerSocket;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.TimeUnit;
 import org.junit.After;
+import org.junit.AfterClass;
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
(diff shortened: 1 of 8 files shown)


bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/TestAppProfile.java
```diff
@@ -61,27 +63,36 @@
 public class TestAppProfile {
   private static final String TABLE_ID = "fake-table";
 
-  private FakeDataService fakeDataService;
-  private Server server;
+  private static FakeDataService fakeDataService = new FakeDataService();
```
igorbernstein2: Why static?

bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/TestBigtableDataGrpcClient.java
```diff
@@ -79,53 +78,62 @@
...
+                public Void answer(InvocationOnMock invocation) {
+                  checkHeader(invocation.getArgument(1, Metadata.class));
+                  return null;
+                }
+              })
+          .when(mockClientCall)
+          .start(any(ClientCall.Listener.class), any(Metadata.class));
+    }
+
+    return new BigtableDataGrpcClient(mockChannel, mockScheduler, options);
```
igorbernstein2: This makes the actual tests extremely hard to read. 
Reading the test code I have no idea what this does: `createClient(false, false, false);` and I have scroll 

Lets put this on hold until we finish the veneer migration

bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/grpc/TestAppProfile.java
rahulKQL: To instantiate Server only once for all the tests. currently, Server creation is in setUp().

igorbernstein2: Server creation is cheap, in general we should keep tests isolated as much as possible unless there is a large penalty. But as discussed below, I'm going to close this for the time being and we will come back to it in the future

PR #2444, opened 2020-03-20
chore: replacing Data Client with new wrappers (Part-3)

rahulKQL:
Towards #2454 

This change has followings updates:
- Updates Data client references in `AbstractBigtableTable` and `BigtableAsyncTable`.
- Removed label's check from FlatRowAdapter, and moved it to `BigtableWhileMatchResultScannerAdapter`.
   -  `Table#get`, `Table#getScanner`(except when whileMatchFilter is applied) would return cells with labels to user.
- Updates `BigtableWhileMatchResultScannerAdapter` to adapt to hbase's `ResultScanner`.

Note: This is part 3 of the series of PRs to replace references of BigtableSession/Data/Admin clients with new wrappers.

PR #2462, opened 2020-03-26
chore: introduce point read in new Data client wrapper

rahulKQL:
Towards #2454 

This change is suggested in [#2444](https://github.com/googleapis/java-bigtable-hbase/pull/2444#discussion_r397186293). In this PR, the point reads are depending on underlying readFlatRowsAsync(). Once the veneer client is implemented then we can set deadlines separately for this.


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/DataClientWrapper.java
```diff
@@ -56,6 +59,13 @@
    */
   ApiFuture<List<KeyOffset>> sampleRowKeysAsync(String tableId);
 
+  /** Reads a single row, if the row not found then returns an empty {@link Result}. */
+  ApiFuture<Result> readRowAsync(String tableId, ByteString rowKey);
```
igorbernstein2: I dont think we need this. Lets just stick with the nullable filter below. I'd like to keep this abstraction layer narrow

PR #2465, opened 2020-03-27
chore(cleanup): deprecated unused method and removed references

rahulKQL:
Towards #2454

This PR address following changes:
 - marked deprecated unused methods from CommonConnection.
 - removed unused Adapter constant from Adapters.
 - add a note in `BigtableResultScannerAdapter` for removal.
 - replace the missed reference of RowKeyUtil.


bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/CommonConnection.java
```diff
@@ -54,11 +51,8 @@
    */
   Configuration getConfiguration();
 
-  /**
-   * Getter for the field <code>options</code>.
-   *
-   * @return a {@link com.google.cloud.bigtable.config.BigtableOptions} object.
-   */
+  /** @deprecated Please use {@link #getBigtableHBaseSettings()}. */
```
igorbernstein2: nit. Can we rename getBigtableHBaseSettings -> getBigtableSettings? The hbase part is implied since this is an hbase connection

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutatorHelper.java
```diff
@@ -204,10 +204,10 @@ public long getWriteBufferSize() {
         future = bulkMutation.add(adapter.adaptEntry((Delete) mutation));
       } else if (mutation instanceof Increment) {
         future = dataClient.readModifyWriteRowAsync(adapter.adapt((Increment) mutation));
-        operationAccountant.registerOperation(ApiFutureUtil.adapt(future));
+        operationAccountant.registerOperation(new ApiFutureToListenableFuture<>(future));
```
igorbernstein2: Can we avoid the wrapping here altogether? and just have OperationAccountant work with ApiFutures?

rahulKQL: marking as resolved as address this on #2468

bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/CommonConnection.java
rahulKQL: marking as resolved(address this on #2466)

rahulKQL:
Thanks for the feedback. I have updated this PR, Please take a fresh look!

bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/CommonConnection.java
igorbernstein2: Please update the javadoc

PR #2467, opened 2020-03-30
chore: Marking snapshot operations as unsupported

rahulKQL:
This change makes all snapshot related operations as unsupported, Also it removes test cases around snapshot operation from integration test modules.


rahulKQL:
forced commit with rewards to trigger failed Travis CI. 

Failed CI build: https://travis-ci.org/github/googleapis/java-bigtable-hbase/builds/668678551?utm_source=github_status&utm_medium=notification

PR #2481, opened 2020-04-03
chore: adding veneer implementation for DataClientWrapper

rahulKQL:
Towards #2454 

This PR adds veneer specific implementation for `DataClientWrapper`. Also, adding ResultRowAdapter which helps in converting `ReadRowsResponse` to HBase's `Result`.

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ResultRowAdapter.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ResultRowAdapter.java
@@ -1,0 +1,181 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.hbase.adapters.read;
+
+import com.google.api.core.InternalApi;
+import com.google.cloud.bigtable.data.v2.models.RowAdapter;
+import com.google.cloud.bigtable.hbase.util.ByteStringer;
+import com.google.cloud.bigtable.hbase.util.TimestampConverter;
+import com.google.common.collect.ImmutableList;
+import com.google.protobuf.ByteString;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.TreeMap;
+import org.apache.hadoop.hbase.Cell;
+import org.apache.hadoop.hbase.client.Result;
+
+/**
+ * Adapter for {@link RowAdapter} that uses {@link Result} to represent logical rows.
+ *
+ * <p>For internal use only - public for technical reasons.
+ */
+@InternalApi("For internal usage only")
+public class ResultRowAdapter implements RowAdapter<Result> {
+
+  @Override
+  public RowBuilder<Result> createRowBuilder() {
+    return new ResultRowBuilder();
+  }
+
+  @Override
+  public boolean isScanMarkerRow(Result result) {
+    return result.isEmpty();
+  }
+
+  @Override
+  public ByteString getKey(Result result) {
+    return ByteStringer.wrap(result.getRow());
+  }
+
+  public static class ResultRowBuilder implements RowBuilder<Result> {
+    private byte[] currentKey;
+    private String family;
+    private ByteString qualifier;
+    private List<String> labels;
+    private long timestamp;
+    private ByteString value;
+
+    private Map<String, List<RowCell>> cells = new TreeMap<>();
+    private List<RowCell> currentFamilyCells = null;
+    private String previousFamily;
+    private int totalCellCount = 0;
+
+    @Override
+    public void startRow(ByteString rowKey) {
+      this.currentKey = ByteStringer.extract(rowKey);
+    }
+
+    @Override
+    public void startCell(
+        String family, ByteString qualifier, long timestamp, List<String> labels, long size) {
+      this.family = family;
+      this.qualifier = qualifier;
+      this.timestamp = timestamp;
+      this.labels = labels;
+      this.value = ByteString.EMPTY;
+    }
+
+    @Override
+    public void cellValue(ByteString value) {
+      this.value = this.value.concat(value);
+    }
+
+    /**
+     * Adds a Cell to {@link RowCell}'s map which is ordered by family. cells received from {@link
+     * RowBuilder} has ordering as:
+     *
+     * <ul>
+     *   <li>family names clustered, but not sorted
+     *   <li>qualifiers in each family cluster is sorted lexicographically
+     *   <li>then descending by timestamp
+     * </ul>
+     *
+     * The end result will be that {@link RowCell} are ordered as:
+     *
+     * <ul>
+     *   <li>lexicographical by family
+     *   <li>then lexicographical by qualifier
+     *   <li>then descending by timestamp
+     * </ul>
+     *
+     * A flattened version of the {@link RowCell} map will be sorted correctly.
+     */
+    @Override
+    public void finishCell() {
+      if (!Objects.equals(this.family, this.previousFamily)) {
+        previousFamily = this.family;
+        currentFamilyCells = new ArrayList<>();
+        cells.put(this.family, this.currentFamilyCells);
+      }
+
+      RowCell rowCell =
+          new RowCell(
+              this.currentKey,
+              this.family.getBytes(),
+              ByteStringer.extract(this.qualifier),
+              TimestampConverter.bigtable2hbase(this.timestamp),
+              ByteStringer.extract(this.value),
+              this.labels);
+      this.currentFamilyCells.add(rowCell);
+      totalCellCount++;
+    }
+
+    /**
+     * This method flattens the {@code cells} which has a map of Lists keyed by family name. The
+     * {@code cells} treeMap is sorted lexicographically, and each List is sorted by qualifier in
+     * lexicographically ascending order, and timestamp in descending order.
+     *
+     * @return an object of HBase {@link Result}.
+     */
+    @Override
+    public Result finishRow() {
+      if (totalCellCount == 0) {
+        return Result.EMPTY_RESULT;
+      }
+
+      ImmutableList.Builder<Cell> combined = ImmutableList.builder();
+      for (List<RowCell> familyCellList : cells.values()) {
+
+        RowCell previous = null;
+        for (RowCell rowCell : familyCellList) {
+          if (previous == null || !rowCell.getLabels().isEmpty() || !keysMatch(rowCell, previous)) {
+            combined.add(rowCell);
+          }
+          previous = rowCell;
+        }
+      }
+
+      return Result.create(combined.build());
+    }
+
+    private boolean keysMatch(RowCell current, RowCell previous) {
+      return current.getTimestamp() == previous.getTimestamp()
+          && Arrays.equals(current.getQualifierArray(), previous.getQualifierArray());
+    }
+
+    @Override
+    public void reset() {
+      this.currentKey = null;
+      this.family = null;
+      this.qualifier = null;
+      this.labels = null;
+      this.timestamp = 0L;
+      this.value = null;
+      this.cells = new TreeMap<>();
+      this.currentFamilyCells = null;
+      this.previousFamily = null;
+      this.totalCellCount = 0;
+    }
+
+    @Override
+    public Result createScanMarkerRow(ByteString rowKey) {
+      return Result.EMPTY_RESULT;
+    }
+  }
+}
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
@@ -18,6 +18,7 @@ package com.google.cloud.bigtable.hbase.wrappers.veneer;
 import com.google.api.core.ApiFunction;
 import com.google.api.core.ApiFuture;
 import com.google.api.core.ApiFutures;
+import com.google.api.core.InternalApi;
 import com.google.api.gax.rpc.ServerStream;
 import com.google.api.gax.rpc.StateCheckingResponseObserver;
 import com.google.api.gax.rpc.StreamController;
@@ -44,6 +45,8 @@ import org.apache.hadoop.hbase.client.AbstractClientScanner;
 import org.apache.hadoop.hbase.client.Result;
 import org.apache.hadoop.hbase.client.ResultScanner;
 
+/** For internal use only - public for technical reasons. */
+@InternalApi("For internal usage only")
 public class DataClientVeneerApi implements DataClientWrapper {
 
   private final BigtableDataClient delegate;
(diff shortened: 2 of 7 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ResultRowAdapter.java
```diff
@@ -0,0 +1,181 @@
...
+     *   <li>lexicographical by family
+     *   <li>then lexicographical by qualifier
+     *   <li>then descending by timestamp
+     * </ul>
+     *
+     * A flattened version of the {@link RowCell} map will be sorted correctly.
+     */
+    @Override
+    public void finishCell() {
+      if (!Objects.equals(this.family, this.previousFamily)) {
```
igorbernstein2, line 111: Please verify that nextIndex == -1 or nextIndex == value.size()

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ResultRowAdapter.java
```diff
@@ -0,0 +1,181 @@
...
+      this.value = null;
+      this.cells = new TreeMap<>();
+      this.currentFamilyCells = null;
+      this.previousFamily = null;
+      this.totalCellCount = 0;
+    }
+
+    @Override
+    public Result createScanMarkerRow(ByteString rowKey) {
+      return Result.EMPTY_RESULT;
```
igorbernstein2, line 178: I'm a little concerned about re-using the EMPTY_RESULT marker here.

Can we create our own static instance of an empty Result and use referential equality to detect it as a marker?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  @Override
+  public ResultScanner readRows(Query request) {
+    return new RowResultScanner(delegate.readRowsCallable(new ResultRowAdapter()).call(request));
+  }
+
+  @Override
+  public ApiFuture<List<Result>> readRowsAsync(Query request) {
+    return delegate.readRowsCallable(new ResultRowAdapter()).all().futureCall(request);
```
igorbernstein2, line 117: Please create a single instance as a constant and reuse it

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/adapters/read/TestResultRowAdapter.java
```diff
@@ -0,0 +1,160 @@
...
+
+  @Test
+  public void testWhenSplitCell() {
+    ByteString valuePart1 = ByteString.copyFromUtf8("part-1");
+    ByteString valuePart2 = ByteString.copyFromUtf8("part-2");
+    ByteString valuePart3 = ByteString.copyFromUtf8("part-3");
+
+    RowAdapter.RowBuilder<Result> rowBuilder = underTest.createRowBuilder();
+    rowBuilder.startRow(ROW_KEY_ONE);
+    rowBuilder.startCell(COL_FAMILY, QUAL_ONE, 10000L, Collections.<String>emptyList(), 0);
```
igorbernstein2, line 88: This looks wrong, why is size 0 for a split cell?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ResultRowAdapter.java
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  @Override
+  public boolean isScanMarkerRow(Result result) {
+    return result.isEmpty();
+  }
+
+  @Override
+  public ByteString getKey(Result result) {
+    return ByteStringer.wrap(result.getRow());
```
igorbernstein2, line 53: I'm a little concerned about this:
We get a ByteString from a chunk, extract the bytes and then  almost immediately re-wrap it for ReadRowsResumptionStrategy. I need to think about how this can be improved

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ResultRowAdapter.java
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  public static class ResultRowBuilder implements RowBuilder<Result> {
```
igorbernstein2, line 56: I think this should use a custom subclass of Result. The subclass should store the original ByteString alongside the Cells. isMarkerRow and getRowKey can downcast to get the appropriate info

rahulKQL: Thanks a lot for the review and suggestions.
I am not sure where we will be using cells fields from subclass, So kept a boolean field for isMarkerRow(). like this:
```java
static class RowResult extends Result {
    private final ByteString rowKey;
    private final boolean isMarkerRow;

    RowResult(ByteString rowKey, List<Cell> cells) {
      super();
      this.rowKey = rowKey;
      this.isMarkerRow = cells == null || cells.isEmpty();

      // all except default ctor of Result are private, So instantiating cells through copyFrom()
      // because value(), size(), isEmpty() rawCells(), advance(). are directly using Result's cells field.
      this.copyFrom(Result.create(cells));
    }

    ByteString getKey() {
      return rowKey;
    }

    boolean isMarkerRow() {
      return isMarkerRow;
    }
  }
```

Please let me know you thoughts on this.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ResultRowAdapter.java
```diff
@@ -0,0 +1,181 @@
...
+      if (totalCellCount == 0) {
+        return Result.EMPTY_RESULT;
+      }
+
+      ImmutableList.Builder<Cell> combined = ImmutableList.builder();
+      for (List<RowCell> familyCellList : cells.values()) {
+
+        RowCell previous = null;
+        for (RowCell rowCell : familyCellList) {
+          if (previous == null || !rowCell.getLabels().isEmpty() || !keysMatch(rowCell, previous)) {
```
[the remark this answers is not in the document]
rahulKQL: This was originally added to `FlatRowAdapter`, I remember some of the Integration tests were flaky due to duplicated cells.

Flaky test cases were `TestFilter`
  - testWhileMatchFilter_simple
  - testWhileMatchFilter_singleChained
  - testWhileMatchFilter_withUpdate

I have removed these checks as we are allowing cells with labels now.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ResultRowAdapter.java
```diff
@@ -0,0 +1,181 @@
...
+     * {@code cells} treeMap is sorted lexicographically, and each List is sorted by qualifier in
+     * lexicographically ascending order, and timestamp in descending order.
+     *
+     * @return an object of HBase {@link Result}.
+     */
+    @Override
+    public Result finishRow() {
+      if (totalCellCount == 0) {
+        return Result.EMPTY_RESULT;
+      }
```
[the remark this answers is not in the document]
rahulKQL: This could execute only be in case of marker rows. Removing this check now as we can handle it with custom Result.

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/adapters/read/TestResultRowAdapter.java
rahulKQL: This was my overlook because I had not checked against the `size` parameter in this draft.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/RowResultAdapter.java
```diff
@@ -0,0 +1,224 @@
...
+      } else {
+        this.value = EMPTY_VALUE;
+        this.nextValueIndex = -1;
+      }
+    }
+
+    @Override
+    public void cellValue(ByteString newValue) {
+      // TODO: Verify if can value ever be null?
+      if (newValue == null || newValue.isEmpty()) {
```
igorbernstein2, line 127: it should never be null

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/RowResultAdapter.java
```diff
@@ -0,0 +1,224 @@
...
+        return;
+      }
+
+      if (nextValueIndex == -1) {
+        this.value = newValue.toByteArray();
```
igorbernstein2, line 132: Please unwrap using ByteStringer

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/RowResultAdapter.java
```diff
@@ -0,0 +1,224 @@
...
+        nextValueIndex = newValue.size();
+      } else {
+        int newValueSize = newValue.size();
+        Preconditions.checkState(
+            Integer.MAX_VALUE - nextValueIndex > newValueSize,
```
igorbernstein2, line 137: This should check for overflow on the array size not the maximum integer

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/RowResultAdapter.java
```diff
@@ -0,0 +1,224 @@
...
+        this.value = new byte[(int) size];
+        this.nextValueIndex = 0;
+      } else {
+        this.value = EMPTY_VALUE;
+        this.nextValueIndex = -1;
+      }
+    }
+
+    @Override
+    public void cellValue(ByteString newValue) {
```
igorbernstein2, line 125: The current implementation doesn't really make sense to me.

Can I suggest something like this instead?

```java
    public void cellValue(ByteString newValue) {
      // Optimize unsplit cells by avoiding a copy
      if (nextValueIndex == -1) {
        this.value = ByteStringer.extract(newValue);
        return;
      }

      Preconditions.checkState(nextValueIndex + newValue.size() <= value.length,
              "Cell value is larger than expected");

      newValue.copyTo(this.value, this.nextValueIndex);
      nextValueIndex += newValue.size();
    }
```

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/RowResultAdapter.java
```diff
@@ -0,0 +1,224 @@
...
+     *   <li>then lexicographical by qualifier
+     *   <li>then descending by timestamp
+     * </ul>
+     *
+     * A flattened version of the {@link RowCell} map will be sorted correctly.
+     */
+    @Override
+    public void finishCell() {
+      Preconditions.checkState(
+          nextValueIndex == -1 || nextValueIndex == value.length, "Inconsistent value found");
```
igorbernstein2, line 169: nextValueIndex should not be -1. Also the error message is not helpful

reconditions.checkState(nextValueIndex == value.length, "Cell value too short");

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestRowResultAdapter.java
```diff
@@ -0,0 +1,167 @@
...
+    rowBuilder.cellValue(valuePart1);
+    rowBuilder.cellValue(valuePart2);
+    rowBuilder.cellValue(valuePart3);
+    rowBuilder.finishCell();
+
+    Result actualResult = rowBuilder.finishRow();
+    Assert.assertArrayEquals(
+        valuePart1.concat(valuePart2).concat(valuePart3).toByteArray(),
+        actualResult.getValue(COL_FAMILY.getBytes(), QUAL_ONE.toByteArray()));
+  }
```
igorbernstein2, line 102: Since we just identified bugs in the implementation of error handling, can you add tests to make sure it doesnt happen again

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/RowResultAdapter.java
```diff
@@ -0,0 +1,223 @@
...
+    return new RowResultBuilder();
+  }
+
+  @Override
+  public boolean isScanMarkerRow(Result result) {
+    if (result instanceof RowResult) {
+      return ((RowResult) result).isMarkerRow();
+    }
+    // TODO: This may never be executed still is it ok to fallback on this?
+    return result.isEmpty();
```
igorbernstein2, line 55: I would add a Precondition for Result being an instanceof RowResult so that we know if our assumption here is in correct

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/RowResultAdapter.java
```diff
@@ -0,0 +1,223 @@
...
+  }
+
+  @Override
+  public ByteString getKey(Result result) {
+    if (result instanceof RowResult) {
```
igorbernstein2, line 60: same here

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/RowResultAdapter.java
```diff
@@ -0,0 +1,223 @@
...
+      return new RowResult(rowKey, cells);
+    }
+
+    static RowResult createMarker(ByteString rowKey) {
+      return new RowResult(rowKey, ImmutableList.<Cell>of());
+    }
+
+    private RowResult(ByteString rowKey, List<Cell> cells) {
+      this.rowKey = rowKey;
+      this.isMarkerRow = cells == null || cells.isEmpty();
```
igorbernstein2, line 80: Can this be explicitly passed from the factory methods?

rahulKQL: Yes, we could another param to RowResult ctor, 
```java
private RowResult(ByteString rowKey, boolean isMarkerRow, List<Cell> cells) {
      this.rowKey = rowKey;
      this.isMarkerRow = isMarkerRow;
      // ...
    }
```
May I suggest an alternative, Looking at finishRow(), `List<Cell>` would never be `null`, then we can directly use that to check for markerRow, eg:
```java
private RowResult(ByteString rowKey, List<Cell> cells) {
      this.rowKey = rowKey;
      this.isMarkerRow = cells.isEmpty();
      // ...
    }
```
Please let me know your thoughts on this:

rahulKQL: Have updated with `isMarkerRow` as a `ctor` param, this would save us a check after each Row creation(looked up StateMachine.java to understand the usages).

PR #2484, opened 2020-04-06
fix: deadlock scenario from BulkReadVeneerApi and fixed flaky tests

rahulKQL:
Apologies for not adding enough test cases in the original BulkReadVeneerApi PR. Due to that, we are seeing flaky test cases from TestBulkReadVeneerApi in CI jobs.

### Bug Description

The [BulkReadVeneerApi#cleanUp](https://github.com/googleapis/java-bigtable-hbase/blob/4f9e7b65aaf6bf5bd87b9f706d59b54afb49649f/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BulkReadVeneerApi.java#L100) goes into a deadlock caused by the fact that the element future resolves before the Batch's status is updated. More information can be found [here](https://github.com/googleapis/java-bigtable-hbase/pull/2484#issuecomment-612972727):

> The order of events is:
> 
> * Batch#onBatchSuccess()
>   
>   * BatchingDescriptor#splitResponse()
>     
>     * elementFuture#set()
>     * BatcherStats#recordBatchElementsCompletion
>   * Batcher#onBatchCompletion
>     
>     * BatcherImpl#flushLock.notifyAll()
> 
> Which means that the callback in BulkReadVeneerApi is notified before onBatchCompletion is executed, preventing numOfOutstandingBatches from decrementing. In other words we are notifying external callers before finishing updating our own internal state. So I would argue that the bug is actually in gax's BatcherImpl not here.

### Proposed Fix:
 
This PR adds a workaround for this problem by introducing a separate single thread executor, Now we defer the callback which decrement the entry response counter on this executor, which gives a chance to the original grpc thread to call onBatchCompletion.

I confirm this solution by running these unit tests for 100 times each on my local system.

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestBulkReadVeneerApi.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestBulkReadVeneerApi.java
@@ -92,12 +92,12 @@ public class TestBulkReadVeneerApi {
     ApiFuture<Result> secBatchResult1 = bulkReadWrapper.add(ByteString.copyFromUtf8("1"), filter);
     ApiFuture<Result> secBatchResult2 = bulkReadWrapper.add(ByteString.copyFromUtf8("2"), filter);
 
+    bulkReadWrapper.sendOutstanding();
+
     resultFuture1_1.get();
     resultFuture1_2.get();
     secBatchResult1.get();
     secBatchResult2.get();
-
-    bulkReadWrapper.sendOutstanding();
 
     // If one entry of the batch is resolved then another should also be
     assertTrue(resultFuture1_1.isDone());
--- a/bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestBulkReadVeneerApi.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestBulkReadVeneerApi.java
@@ -76,20 +76,17 @@ public class TestBulkReadVeneerApi {
   public void testAdd() throws Exception {
     BulkReadVeneerApi bulkReadWrapper =
         new BulkReadVeneerApi(BigtableDataClient.create(settingsBuilder.build()), TABLE_ID);
-    ApiFuture<Result> resultFuture1 = bulkReadWrapper.add(ByteString.copyFromUtf8("one"), null);
-    assertFalse(resultFuture1.isDone());
+    ApiFuture<Result> resultFuture = bulkReadWrapper.add(ByteString.copyFromUtf8("one"), null);
+    assertFalse(resultFuture.isDone());
+    // Here AutoFlush triggers the batch
+    resultFuture.get();
 
     Filters.Filter filter = Filters.FILTERS.key().regex("cf");
-    ApiFuture<Result> secBatchResult1 = bulkReadWrapper.add(ByteString.copyFromUtf8("1"), filter);
-    ApiFuture<Result> secBatchResult2 = bulkReadWrapper.add(ByteString.copyFromUtf8("2"), filter);
+    ApiFuture<Result> secBatchResult = bulkReadWrapper.add(ByteString.copyFromUtf8("1"), filter);
+    secBatchResult.get();
 
-    // Here AutoFlush triggers the batch
-    resultFuture1.get();
-    secBatchResult1.get();
-
-    // If one entry of the batch is resolved then another should also be
-    assertTrue(resultFuture1.isDone());
-    assertTrue(secBatchResult2.isDone());
+    assertTrue(resultFuture.isDone());
+    assertTrue(secBatchResult.isDone());
 
     assertEquals(2, fakeDataService.getReadRowsCount());
   }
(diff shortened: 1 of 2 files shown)


bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestBulkReadVeneerApi.java
```diff
@@ -76,20 +76,17 @@ public void tearDown() throws InterruptedException {
...
-    ApiFuture<Result> resultFuture1 = bulkReadWrapper.add(ByteString.copyFromUtf8("one"), null);
-    assertFalse(resultFuture1.isDone());
+    ApiFuture<Result> resultFuture = bulkReadWrapper.add(ByteString.copyFromUtf8("one"), null);
+    assertFalse(resultFuture.isDone());
+    // Here AutoFlush triggers the batch
+    resultFuture.get();
 
     Filters.Filter filter = Filters.FILTERS.key().regex("cf");
-    ApiFuture<Result> secBatchResult1 = bulkReadWrapper.add(ByteString.copyFromUtf8("1"), filter);
-    ApiFuture<Result> secBatchResult2 = bulkReadWrapper.add(ByteString.copyFromUtf8("2"), filter);
```
kolea2, line 84: how come we are removing the second batch?

igorbernstein2: I think this test currently flawed. There is a race condition between the response setting the 2nd item's future and us checking it. 

I think this entire thing can be simplified by calling flush instead of relying on the autoflush behavior

rahulKQL:
After spending some time on this today, It seems BulkReadVeneerApi has a bug:  If we try to close all the batchers on the map then this bug makes one of the `batcher` instance in the map to hung until infinite time. 

https://github.com/googleapis/java-bigtable-hbase/blob/7b2d0df5a3ebc7da35c5193bcd697ab5e92c2d00/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BulkReadVeneerApi.java#L100-L109

While debugging it seems even after all the `resultEntry` futures are resolved, `BatcherImpl`'s `close()` -> `flush()`  still await for the same batch to invoke its Batch#onSuccess callback, Which should already be finished until now.
https://github.com/googleapis/gax-java/blob/a8827adc670e99c289993791c52afeed0fc7a6ef/gax/src/main/java/com/google/api/gax/batching/BatcherImpl.java#L166-L187

Beside this bug, another test case from the same class is flaky i.e. TestBulkReadVeneerApi#testSendOutstanding.

I am still investigating this, I will add more information on this tomorrow.

igorbernstein2:
Thanks for the update

rahulKQL:
@igorbernstein2 I have updated this PR with bug description.

Please have a look and let me know your thoughts on the proposed fix.

igorbernstein2:
So I don't think this is a race condition, but a deadlock caused by the fact that the element future resolves before the Batch's status is updated.

The order of events is:
- Batch#onBatchSuccess()
  - BatchingDescriptor#splitResponse()
     - elementFuture#set()
     - BatcherStats#recordBatchElementsCompletion
   - Batcher#onBatchCompletion
     - BatcherImpl#flushLock.notifyAll()

Which means that the callback in BulkReadVeneerApi is notified before onBatchCompletion is executed, preventing numOfOutstandingBatches from decrementing. In other words we are notifying external callers before finishing updating our own internal state. So I would argue that the bug is actually in gax's BatcherImpl not here.

This PR has a bunch of changes, but I believe only one of the them is the actual workaround to the bug, I think other changes are unnecessary (adding additional atomics).

I believe the actual fix (temporary workaround) is to defer the callback onto a separate thread so that the original grpc thread has a chance to call onBatchCompletion. I think the cleanest way to fix this is to replace the directExecutor with a SingleThreadExecutor when adding the rowFutureListener:

```java
private static final Executor CLEANUP_EXECUTOR = Executors.newSingleThreadExecutor(new ThreadFactory() {
    @Override
    public Thread newThread(Runnable r) {
      Thread thread = new Thread(r);
      thread.setDaemon(true);
      thread.setName("bigtable-bulkread-cleanup");
      return thread;
    }
  });
//...
rowFuture.addListener(
        new Runnable() {
          @Override
          public void run() {
            notifyArrival();
          }
        },
            CLEANUP_EXECUTOR);
```


If you agree with this, then I propose the following:
1. In the short term use the above approach
2. In the medium term, add Batcher#closeAsync() to gax and remove the workaround 
3. In the longer term:
  1. add `BatchEntry#setResult(T value)` and `BatchEntry#setException(Throwable)` which will temporarily delegate to the underlying SettableFuture
  2. update gax's BatchEntry interface to hide the underlying future from the BatchingDescriptor
  3. introduce temporary state to BatchEntry to store the THrowable or the result. 
  4. add a `BatchEntry#finalizeResult()` that will resolve the internal SettableFuture
  5. call finalizeResult() after all of the internal bookkeeping in BatcherImpl is done

rahulKQL:
Thanks a lot for taking the time and adding a detailed analysis of this bug.

I agree with your listed approach and will update this PR with the proposed workaround.

igorbernstein2:
Also please update the PR title since this is a bugfix now

PR #2491, opened 2020-04-14
chore(fix): remove duplicated cell when interleave filter is applied

rahulKQL:
Towards #2454 

It seems Bigtable classic client & veneer client behaves differently for duplicated `cells`. A Row can contain duplicated `cells` when the user applies `Interleave` filters. The behavior of both of these clients are:
   - Classic client: Here, we check if a `cell` contains some `labels` or not, If it does then we include that cell in end result if it doesn't then we compare `timestamp` and `qualifier` with previous no label cell(Because a cell with a `label` could have been produced by applying filters).
    https://github.com/googleapis/java-bigtable-hbase/blob/4f9e7b65aaf6bf5bd87b9f706d59b54afb49649f/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RowMerger.java#L302-L316
   - Veneer client: Here, we do not perform these checks. Most of the other client allows duplicate cells, I confirmed with `Go/NodeJs/C#` bigtable client. Not sure why but `python-bigtable` does not allow duplicate cells.

**Assumption:** The labels are applied to determine which filters produced those cells, So we are including all cells where labels are present.

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/RowResultAdapter.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/RowResultAdapter.java
@@ -16,6 +16,8 @@
 package com.google.cloud.bigtable.hbase.wrappers.veneer;
 
 import com.google.api.core.InternalApi;
+import com.google.bigtable.v2.RowFilter.Interleave;
+import com.google.cloud.bigtable.data.v2.models.Filters.InterleaveFilter;
 import com.google.cloud.bigtable.data.v2.models.RowAdapter;
 import com.google.cloud.bigtable.hbase.adapters.read.RowCell;
 import com.google.cloud.bigtable.hbase.util.ByteStringer;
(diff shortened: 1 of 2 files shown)


igorbernstein2:
I don't quite understand this. The only time that labels are relevant for the hbase client is when the WhileMatchFilter is applied. So the wrappers should just pass labels as is and the relevant labels should be stripped there.

rahulKQL:
I added `labels` check for unknown use cases which could have been supported by classic client. 

After your confirmation on previous message, I would remove this check.

igorbernstein2:
Sorry you are correct we need to do this because hbase's MUST_PASS_ONE filter gets converted into an Interleave.  However Interleave will duplicate the cells while MUST_PASS_ONE doesn't. So we need to deduplicate the cells somewhere. However we need to be careful, because the WhileMatchFilter depends on accurate counts. So we should only de-dupe cells without labels.

Please add comments describing the above to the code.

In the future we should move the cell dedupe closer to the Interleave filter implementation to make it easier to maintain. Maybe we can do something similar to WhileMatchFilterScanAdapter. But for now we should go with this approach .

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/RowResultAdapter.java
```diff
@@ -97,6 +99,7 @@ boolean isMarkerRow() {
     private List<String> labels;
```
igorbernstein2, line 99: Please add javadoc to the class:

<p>This adapter is responsible for cell deduplication. bigtable-hbase will convert a MUST_PASS_ONE filter into an Interleave. Unfortunately there is a bit of a mismatch between the 2 filters: MUST_PASS_ONE will not duplicate the cell if it matches multiple branches of the of MUST_PASS_ONE, but Interleave will. This adapter will pave over the difference by removing the duplicate cells while building the Result. However, HBase's WhileMatchFilter depends on duplicate labelled cells for its implementation. So this adapter will not deduplicate labelled cells.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/RowResultAdapter.java
```diff
@@ -179,7 +186,19 @@ public void finishCell() {
               TimestampConverter.bigtable2hbase(this.timestamp),
               this.value,
               this.labels);
-      this.currentFamilyCells.add(rowCell);
+
```
igorbernstein2, line 189: please add a comment referring readers to the class javadoc for explaining this:

`// dedupe user visible cells. Please see class javadoc for details`

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestRowResultAdapter.java
```diff
@@ -119,38 +119,60 @@ public void testWhenRowKeyIsNotSet() {
 
   @Test
   public void testOnlyValueIsDifferent() {
```
igorbernstein2, line 121: Can you add a separate test for dedupe functionality. It'll make it easier to debug if this ever gets messed up

rahulKQL:
Thanks for the review and detailed suggestions! PTAL

PR #2529, opened 2020-05-12
Catch and log exceptions in Watchdog runnable to ensure it never stops running

### LINKED ISSUE #2528: If an exception is thrown in Watchdog#run, future executions will never run
When statically looking at the code and trying to determine if there is ever a case where Watchdog stops running periodically, we noticed this potential error. We can't confirm if this is the root cause for the behavior we have noticed (connections doesn't time out within the Watchdog watch period), but we suspect this is an issue and we should fix it anyways.

justinuang:
See https://stackoverflow.com/questions/6894595/scheduledexecutorservice-exception-handling for details

Fixes #2528 ☕️

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/Watchdog.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/Watchdog.java
@@ -95,20 +95,24 @@ public class Watchdog implements Runnable {
   }
 
   public void run() {
-    Iterator<Entry<WatchedCall<?, ?>, Object>> it = openStreams.entrySet().iterator();
+    try {
+      Iterator<Entry<WatchedCall<?, ?>, Object>> it = openStreams.entrySet().iterator();
 
-    int count = 0;
+      int count = 0;
 
-    while (it.hasNext()) {
-      WatchedCall<?, ?> stream = it.next().getKey();
-      if (stream.cancelIfStale()) {
-        count++;
-        it.remove();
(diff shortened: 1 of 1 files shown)


justinuang:
After this is approved, I will also copy the changes over to https://github.com/googleapis/gax-java/blob/master/gax/src/main/java/com/google/api/gax/rpc/Watchdog.java#L114

justinuang:
@igorbernstein2

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/io/Watchdog.java
```diff
@@ -95,20 +95,24 @@ public Watchdog(Clock clock, long waitTimeoutMs, long idleTimeoutMs) {
...
       }
-    }
 
-    if (count > 0) {
-      LOG.warn("Found %d stale streams and cancelled them", count);
+      if (count > 0) {
+        LOG.warn("Found %d stale streams and cancelled them", count);
+      }
+    } catch (Throwable t) {
+      LOG.error("Caught throwable in periodic Watchdog run. Continuing.", t);
```
igorbernstein2, line 115: I think it would be cleaner to move error handling out of this method, something like:

run() {
  try {
   runUnsafe();
  } catch (...
}

runUnsafe() {
  while(..)
}

justinuang: Done

justinuang:
@googlebot rescan

PR #2535, opened 2020-05-19
test: add new nightly snapshot IT config

kolea2:

### DIFF
--- a/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/pom.xml
+++ b/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/pom.xml
@@ -87,6 +87,70 @@ limitations under the License.
       </build>
     </profile>
     <profile>
+      <id>bigtableIntegrationTestSnapshot</id>
+      <!-- use the snapshot veneer artifact to run the tests-->
+      <properties>
+        <bigtable.version>LATEST</bigtable.version>
+      </properties>
+
+      <repositories>
+        <repository>
+          <id>sonatype-snapshots</id>
+          <url>https://oss.sonatype.org/content/repositories/snapshots</url>
+          <releases>
+            <enabled>false</enabled>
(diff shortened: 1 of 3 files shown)


bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/pom.xml
```diff
@@ -86,6 +86,70 @@ limitations under the License.
         </plugins>
       </build>
     </profile>
+    <profile>
```
igorbernstein2, line 89: Do we need a new profile for this?

Can't we execute:

mvn -P bigtableIntegrationTest -Dbigtable.version=LATEST?

kolea2: Interestingly, `LATEST` from the command line doesn't seem to work for me (but hardcoding a snapshot version, ie 1.12.3-SNAPSHOT, does). Either way, we'd still have to add the `<repositories>` section if we reused the main profile, which I'm not sure I'd prefer to do.

igorbernstein2: In that case can we make the profile additive?

mvn -P bigtableIntegrationTest -PuseVeneerSnapshot?

kolea2: Done. Also moved the profile to the parent to activate for all unit tests (not just integration modules)

pom.xml
```diff
@@ -252,4 +252,26 @@ limitations under the License.
       </plugin>
     </plugins>
   </build>
+  <profiles>
+    <profile>
+      <id>use-bigtable-snapshots</id>
```
igorbernstein2, line 257: Can we keep existing camelCase convention for profile ids?

PR #2583, opened 2020-07-14
fix(build): update return codes for build scripts

kolea2:


bigtable-client-core-parent/bigtable-client-core/src/test/java/com/google/cloud/bigtable/util/ByteStringComparatorTest.java
```diff
@@ -49,5 +49,6 @@ private void compare(ByteString a, ByteString b) {
     Assert.assertTrue(underTest.compare(a, b) < 0);
     Assert.assertTrue(underTest.compare(b, a) > 0);
     Assert.assertTrue(underTest.compare(b, b) == 0);
+    Assert.assertTrue("test", false);
```
igorbernstein2, line 52: Please remove this before committing

kolea2: done, thanks for the quick review!

PR #2570, opened 2020-07-08
fix: RPCs spending 2*timeout before throwing DEADLINE_EXCEEDED.

vermas2012:
Moving the call to exponentialRetryAlgorithm.createFirstAttempt before the first RPC is sent. This allows the ExponentialRetryAlgorithm to capture the correct time of first call which is later used to evaluate if deadline has expired.

Fixes #2567

PR #2591, opened 2020-07-14
fix: ReadRows not counting the rpc as the first attempt

vermas2012:
edit (kolea2): 

This is to fix test failures introduced in #2570. `currentBackoff` was originally null as a performance optimization. This new change will create new TimedAttemptSettings for every message.

Behavioral change: the first rpc counts as the first attempt

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
@@ -301,7 +301,9 @@ public abstract class AbstractRetryingOperation<RequestT, ResponseT, ResultT>
 
   @VisibleForTesting
   public boolean inRetryMode() {
-    return currentBackoff != null;
+    // ResetStatusBasedBackoff will create the first attempt which creates currentBackoff with 0
+    // attempts.
+    return currentBackoff != null && currentBackoff.getAttemptCount() > 0;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
```diff
@@ -301,15 +301,19 @@ protected Long getNextBackoff() {
 
   @VisibleForTesting
   public boolean inRetryMode() {
-    return currentBackoff != null;
+    // ResetStatusBasedBackoff will create the first attempt which creates currentBackoff with 0
+    // attempts.
+    return currentBackoff != null && currentBackoff.getAttemptCount() > 0;
```
igorbernstein2, line 306: Can you add a note as to when this value is null and when attempt count is 0

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
```diff
@@ -301,15 +301,19 @@ protected Long getNextBackoff() {
...
 
   /**
    * Either a response was found, or a timeout event occurred. Reset the information relating to
    * Status oriented exception handling.
    */
   protected void resetStatusBasedBackoff() {
-    this.currentBackoff = null;
+    // Reset the backoff parameters. CreateFirstAttempt will log the current time as the time of
+    // first call and enforce timeout from this timeOfFirstCall.
+    this.currentBackoff = exponentialRetryAlgorithm.createFirstAttempt();
```
igorbernstein2, line 316: After our conversation, I looked at the code and I think the original reason this was null was set to null was a performance optimization. This new change will create new TimedAttemptSettings  for every message.

I dont anticipate this adding too much overhead, but please note this in the PR description for future reference

igorbernstein2: Also please note in the description that there is behavioral change here: the first rpc counts as the first attempt

PR #2628, opened 2020-09-22
fix: ZeroCopyByteStringUtil can return incorrect results in some cases

### LINKED ISSUE #2533: ZeroCopyByteStringUtil can return incorrect results in some cases
If you carefully construct a RopeByteString, such that it contains two "extractable" LiteralByteStrings, ZeroCopyByteStringUtil will only return the second.

here's a simple test that demonstrates it:

```
  test("extracts rope") {
    // make a 128+ byte ByteString
    val bs1 = ByteString.copyFrom(Array.fill(128)(5.toByte))
    // anything else
    val bs2 = ByteString.copyFromUtf8("part 2")
    val rope = bs1.concat(bs2)
    val bytes = ZeroCopyByteStringUtil.get(rope)
    bytes should be (rope.toByteArray)
  }
```

I think the fix is to make sure that `writeLazy` in `ZeroCopyByteOutput` isn't called more than once.

dmitry-fa:
Fixes #2533


igorbernstein2 (approved):
lgtm

PR #2631, opened 2020-09-28
feat: implement veneer metrics part 1 - implement MetricsApiTracerAdapter

kolea2:


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapterFactory.java
```diff
@@ -0,0 +1,62 @@
...
+
+import com.google.api.gax.tracing.ApiTracer;
+import com.google.api.gax.tracing.ApiTracerFactory;
+import com.google.api.gax.tracing.SpanName;
+import com.google.bigtable.v2.BigtableGrpc;
+import com.google.cloud.bigtable.metrics.RpcMetrics;
+import java.util.HashMap;
+
+public class MetricsApiTracerAdapterFactory implements ApiTracerFactory {
+  private final RpcMetrics readRowsMetrics =
```
mutianf, line 26: Why are readRows and mutateRow created differently? Maybe add a comment here?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapter.java
```diff
@@ -0,0 +1,117 @@
...
+import org.threeten.bp.Duration;
+
+public class MetricsApiTracerAdapter implements ApiTracer {
+
+  private final RpcMetrics rpcMetrics;
+  private Context operationTimer;
+  private Context rpcTimer;
+
+  private int attemptCount = 0;
+  private Stopwatch attemptTimer;
```
mutianf, line 31: Looks like attemptTimer is not used and can be removed here?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapter.java
```diff
@@ -0,0 +1,117 @@
...
+    };
+  }
+
+  @Override
+  public void operationSucceeded() {
+    operationTimer.close();
+  }
+
+  @Override
+  public void operationCancelled() {}
```
mutianf, line 53: Should we call operationTimer.close() here too?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapter.java
```diff
@@ -0,0 +1,117 @@
...
+
+  @Override
+  public void operationFailed(Throwable error) {
+    rpcMetrics.markFailure();
```
mutianf, line 57: Should we call operationTimer.close() here too?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapter.java
```diff
@@ -0,0 +1,117 @@
...
+
+  @Override
+  public void attemptFailedRetriesExhausted(Throwable error) {
+    rpcTimer.close();
+    rpcMetrics.markRetriesExhausted();
+  }
+
+  @Override
+  public void attemptPermanentFailure(Throwable error) {
+    rpcTimer.close();
```
mutianf, line 94: Do wee need to call rpcMetrics.markFailure() here?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapter.java
igorbernstein2: Also, I dont think markFailure belongs here. Instead, it should be in `attemptPermanentFailure`

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapter.java
```diff
@@ -0,0 +1,117 @@
...
+
+  @Override
+  public void attemptCancelled() {
+    rpcTimer.close();
+  }
+
+  @Override
+  public void attemptFailed(Throwable error, Duration delay) {
+    rpcTimer.close();
+    rpcMetrics.markFailure();
```
igorbernstein2, line 83: I dont markFailure belongs here. It should count operation failures

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapter.java
igorbernstein2: +1

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapter.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  @Override
+  public void connectionSelected(String id) {}
+
+  @Override
+  public void attemptStarted(int attemptNumber) {
+    rpcTimer = rpcMetrics.timeRpc();
+    attemptCount++;
+    attemptResponseCount = 0;
```
igorbernstein2, line 67: if (attemptCount > 0) { markRetry() }

bigtable-client-core-parent/bigtable-metrics-api/src/main/java/com/google/cloud/bigtable/metrics/RpcMetrics.java
```diff
@@ -0,0 +1,81 @@
...
+
+@InternalApi
+public class RpcMetrics {
+  private final String methodName;
+  private final Timer operationTimer;
+  private final Timer rpcTimer;
+  private final Meter retryMeter;
+  private final Meter failureMeter;
+  private final Meter retriesExhaustedMeter;
+  private final Timer readRowsFirstResponseTimer;
```
igorbernstein2, line 29: This looks new and unused. I'm guessing it leaked from a larger changeset?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapterFactory.java
```diff
@@ -0,0 +1,62 @@
...
+    RpcMetrics rpcMetrics = getRpcMetrics(spanName);
+    return new MetricsApiTracerAdapter(rpcMetrics);
+  }
+
+  private RpcMetrics getRpcMetrics(SpanName spanName) {
+    String key = spanName.getMethodName();
+
+    switch (key) {
+      case "ReadRows":
+        return readRowsMetrics;
```
igorbernstein2, line 43: I think veneer has a method for readRow (singular) that should be mapped back to readRowsMetrics

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapterFactory.java
```diff
@@ -0,0 +1,62 @@
...
+      case "MutateRow":
+        return mutateRowMetrics;
+      default:
+        RpcMetrics metrics = otherMetrics.get(key);
+        if (metrics != null) {
+          return metrics;
+        }
+        synchronized (this) {
```
igorbernstein2, line 51: Why does this need to be synchronized? won't this be called on the thread that instantiates the client?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
```diff
@@ -60,7 +60,7 @@ limitations under the License.
     </dependency>
     <dependency>
       <groupId>com.google.cloud.bigtable</groupId>
-      <artifactId>bigtable-client-core</artifactId>
+      <artifactId>bigtable-metrics-api</artifactId>
```
igorbernstein2, line 63: why does metrics replace client core?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
```diff
@@ -137,6 +137,14 @@ limitations under the License.
       <scope>compile</scope>
     </dependency>
 
+    <!--  TODO: remove this dependency when upgraded through transitive dependency (http-client -> org.apache.httpcomponents:httpclient)
```
igorbernstein2, line 140: Why is this coming up now? Should this be part of a different PR?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapter.java
```diff
@@ -0,0 +1,117 @@
...
+import com.google.api.gax.tracing.ApiTracer;
+import com.google.cloud.bigtable.metrics.RpcMetrics;
+import com.google.cloud.bigtable.metrics.Timer.Context;
+import com.google.common.base.Stopwatch;
+import org.threeten.bp.Duration;
+
+public class MetricsApiTracerAdapter implements ApiTracer {
+
+  private final RpcMetrics rpcMetrics;
+  private Context operationTimer;
```
igorbernstein2, line 27: I think operationTimer can be final as well

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapter.java
igorbernstein2: +1

igorbernstein2 (changes requested):

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
kolea2: I think this needs to be rebased

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
kolea2: I think this is a different change that also needs to be rebased.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
kolea2: Nevermind: I think the classes that this module depended on were all moved from core into the metrics module

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
kolea2: Nevermind: I think the classes that this module depended on were all moved from core into the metrics module

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
kolea2: Took a closer look - this was added to fix the deps build: https://github.com/googleapis/java-bigtable-hbase/pull/2631/commits/bf9a9222fb99bb7d342cdc5e7d045062fabf3ba4 when switching the dependency from bigtable-client-core to metrics

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapter.java
kolea2: will rename to `markOperationFailure` for clarity

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapter.java
```diff
@@ -0,0 +1,122 @@
...
+  @Override
+  public void attemptFailedRetriesExhausted(Throwable error) {
+    rpcTimer.close();
+    rpcMetrics.markRetriesExhausted();
+  }
+
+  @Override
+  public void attemptPermanentFailure(Throwable error) {
+    rpcTimer.close();
+    rpcMetrics.markFailure();
```
igorbernstein2, line 99: This is a little tricky: if retries are disabled, attemptPermanentFailure will never be called which means we will never mark failures for operations. So markFailure needs to move to operationFailed. However from what I understand markFailure() must be exclusive of markRetriesExhausted(). In other words I think that the current behavior is to count retry exhaustion separately from permanent failures.
So I think you need have attemptPermanentFailure and attemptFailedRetriesExhausted set a volatile flag that operationFailed reads to call the appropriate method

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapterFactory.java
igorbernstein2: +1 (the answer is that this is small perf optimization to avoid a hashmap lookup for the 2 most used ops)

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapterFactory.java
```diff
@@ -0,0 +1,61 @@
...
+package com.google.cloud.bigtable.hbase.wrappers.veneer.metrics;
+
+import com.google.api.gax.tracing.ApiTracer;
+import com.google.api.gax.tracing.ApiTracerFactory;
+import com.google.api.gax.tracing.SpanName;
+import com.google.bigtable.v2.BigtableGrpc;
+import com.google.cloud.bigtable.metrics.RpcMetrics;
+import java.util.HashMap;
+
+public class MetricsApiTracerAdapterFactory implements ApiTracerFactory {
```
igorbernstein2, line 25: javadoc & visibility restrictions

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapterFactory.java
igorbernstein2: please add a case for this ie
case "ReadRows":
case "ReadRow":

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
igorbernstein2: ?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
igorbernstein2: please rebase to minimize the noise

igorbernstein2 (changes requested):
We need some tests for this:

- Create a fake bigtable service, that will return a predictable set of results
- Implement a fake MetricsRegistry that records all of the measurements
- connect the client to the emulator and exercise a meaningful subset of hbase methods
- verify that your fake metrics registry recorded all of the predicted measurements

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
kolea2: Sorry, what I mean was this change is necessary since we moved code from core to metrics modules in this change.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-shaded/pom.xml
```diff
@@ -186,6 +186,7 @@ limitations under the License.
                     -->
                     <exclude>com.google.cloud.bigtable.hbase*.**</exclude>
                     <exclude>com.google.cloud.bigtable.metrics.**</exclude>
+                    <exclude>com.google.cloud.bigtable.hbase.wrappers.veneer.metrics.**</exclude>
```
igorbernstein2, line 189: here and all of the other shaded poms: make sure to exclude the artifact as well

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapter.java
```diff
@@ -0,0 +1,138 @@
...
+public class MetricsApiTracerAdapter implements ApiTracer {
+
+  private final RpcMetrics rpcMetrics;
+  private Context operationTimer;
+  private Context rpcTimer;
+
+  private int attemptCount = 0;
+  private Stopwatch attemptTimer;
+  private long attemptResponseCount = 0;
+  private volatile RetryStatus call;
```
igorbernstein2, line 33: I think something like `lastRetryStatus` is would be more appropriate. Also this needs some a comment explaining why it exists

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapter.java
```diff
@@ -0,0 +1,138 @@
...
+
+  @Override
+  public void operationCancelled() {}
+
+  @Override
+  public void operationFailed(Throwable error) {
+    operationFailed(error, call);
+  }
+
+  public void operationFailed(Throwable error, RetryStatus retryStatus) {
```
igorbernstein2, line 61: this seems to be called only by `operationFailed(Throwable error)`. Can this be inlined?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapter.java
igorbernstein2: Also I think you need to initialize this to PERMANENT_FAILURE for when retries are disabled

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/src/test/java/com/google/cloud/bigtable/hbase1_x/TestMetrics.java
```diff
@@ -0,0 +1,243 @@
...
+  public static void tearDownServer() throws InterruptedException {
+    if (server != null) {
+      server.shutdownNow();
+      server.awaitTermination();
+    }
+  }
+
+  @After
+  public void tearDown() throws IOException {
+    connection.close();
```
igorbernstein2, line 107: make sure to reset BigtableClientMetrics as well

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/src/test/java/com/google/cloud/bigtable/hbase1_x/TestMetrics.java
```diff
@@ -0,0 +1,243 @@
...
+    assertEquals(FULL_TABLE_NAME, request.getTableName());
+
+    BigtableHBaseVeneerSettings bigtableSettings =
+        (BigtableHBaseVeneerSettings) connection.getBigtableSettings();
+    MetricsApiTracerAdapterFactory tracerFactory =
+        (MetricsApiTracerAdapterFactory)
+            bigtableSettings.getDataSettings().getStubSettings().getTracerFactory();
+
+    RpcMetrics readRow = tracerFactory.getMetrics().get("ReadRow");
+    Assert.assertNotNull(readRow);
```
igorbernstein2, line 133: I dont think this check is necessary.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/src/test/java/com/google/cloud/bigtable/hbase1_x/TestMetrics.java
```diff
@@ -0,0 +1,243 @@
...
+        fakeMetricRegistry.results.get("google-cloud-bigtable.grpc.method.ReadRow.failure");
+    AtomicLong operationLatency =
+        fakeMetricRegistry.results.get(
+            "google-cloud-bigtable.grpc.method.ReadRow.operation.latency");
+    AtomicLong tableGetLatency =
+        fakeMetricRegistry.results.get("google-cloud-bigtable.table.get.latency");
+
+    Assert.assertEquals(3, readRowFailure.get());
+    Assert.assertTrue(
+        operationLatency.get() <= readRowsTime + 50 || operationLatency.get() >= readRowsTime - 50);
```
igorbernstein2, line 145: I dont think you need to extend the upper bound since it will always be bounded by your stopwatch

igorbernstein2: Also it would be nice to split these assertions and add an assertion message so that when it fails we get more info than just expected true, but got false

igorbernstein2 (changes requested):
This is looking good. I think there is one major issue left and the rest are nits.
The main left is the packaging. You need make sure that the metrics artifact is excluded from the shaded jars and it manually promoted

bigtable-client-core-parent/bigtable-metrics-api/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.dropwizard.metrics</groupId>
+      <artifactId>metrics-core</artifactId>
+      <version>${dropwizard.metrics.version}</version>
```
igorbernstein2, line 35: please rebase & update to the new variable name

bigtable-dataflow-parent/bigtable-hbase-beam/pom.xml
```diff
@@ -249,6 +249,13 @@ limitations under the License.
           </usedDependencies>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
+        </configuration>
+      </plugin>
```
igorbernstein2, line 258: please remove this

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-hadoop/pom.xml
```diff
@@ -77,6 +77,11 @@ limitations under the License.
       <artifactId>metrics-core</artifactId>
       <version>${dropwizard.metrics.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.cloud.bigtable</groupId>
+      <artifactId>bigtable-metrics-api</artifactId>
+      <version>${project.version}</version>
+    </dependency>
```
igorbernstein2, line 84: nit. can you mose this to line 65. I dont like it to be a bit more prominent

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-hadoop/pom.xml
```diff
@@ -164,6 +169,7 @@ limitations under the License.
             <usedDependency>com.google.code.findbugs:jsr305</usedDependency>
             <usedDependency>commons-logging:commons-logging</usedDependency>
             <usedDependency>io.dropwizard.metrics:metrics-core</usedDependency>
+            <usedDependency>com.google.cloud.bigtable:bigtable-metrics-api</usedDependency>
```
igorbernstein2, line 172: nit. please bump this update higher in the list

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/pom.xml
```diff
@@ -98,6 +98,11 @@ limitations under the License.
       <artifactId>hbase-server</artifactId>
       <version>${hbase.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.cloud.bigtable</groupId>
+      <artifactId>bigtable-metrics-api</artifactId>
+      <version>${project.version}</version> <!-- {x-version-update:bigtable-client-core-parent:current} -->
+    </dependency>
```
igorbernstein2, line 105: please rebase and remove this dep:
mapreduce is meant to be a leaf artifact and shades everything (or everything except hbase & hadoop after $2895). So it doesnt need to declare this dep. Please remove it from the dep plugin below as well

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/pom.xml
igorbernstein2: please rebase, this souldnt be necessary anymore

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/src/test/java/com/google/cloud/bigtable/hbase1_x/TestMetrics.java
```diff
@@ -0,0 +1,231 @@
...
+  @Test
+  public void readRows() throws IOException, InterruptedException {
+    FakeMetricRegistry fakeMetricRegistry = new FakeMetricRegistry();
+
+    BigtableClientMetrics.setMetricRegistry(fakeMetricRegistry);
+    Table table = connection.getTable(TABLE_NAME);
+
+    long readRowsStart = System.currentTimeMillis();
+    Result result = table.get(new Get(new byte[2]));
+    long readRowsTime = System.currentTimeMillis() - readRowsStart;
```
igorbernstein2, line 116: Please use a Stopwatch

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/src/test/java/com/google/cloud/bigtable/hbase1_x/TestMetrics.java
```diff
@@ -0,0 +1,231 @@
...
+    }
+  }
+
+  @After
+  public void tearDown() throws IOException {
+    connection.close();
+  }
+
+  @Test
+  public void readRows() throws IOException, InterruptedException {
```
igorbernstein2, line 108: Please add a small description of what this does.
"This tests metric instrumentation by using a fake service to inject failures. The fake service will fail the first 3 readrows requests, causing the client to start exponential retries".

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/src/test/java/com/google/cloud/bigtable/hbase1_x/TestMetrics.java
```diff
@@ -0,0 +1,231 @@
...
+    AtomicLong tableGetLatency =
+        fakeMetricRegistry.results.get("google-cloud-bigtable.table.get.latency");
+
+    Assert.assertEquals(3, readRowFailure.get());
+    Assert.assertTrue(
+        "operation latency for ReadRow took longer than expected",
+        operationLatency.get() <= readRowsTime + 50);
+    Assert.assertTrue(
+        "operation latency for table.get took longer than expected",
+        tableGetLatency.get() <= readRowsTime + 50);
```
igorbernstein2, line 135: this can be a lot tighter of a check. I would put a 5ms sleep in the fake service and bound this to 20ms

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/src/test/java/com/google/cloud/bigtable/hbase1_x/TestMetrics.java
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  private static class FakeDataService extends BigtableGrpc.BigtableImplBase {
+
+    final BlockingQueue<Object> requests = Queues.newLinkedBlockingDeque();
```
igorbernstein2, line 140: this should be a concurrent queue

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/src/test/java/com/google/cloud/bigtable/hbase1_x/TestMetrics.java
```diff
@@ -0,0 +1,231 @@
...
+
+    @SuppressWarnings("unchecked")
+    <T> T popLastRequest() throws InterruptedException {
+      return (T) requests.poll(1, TimeUnit.SECONDS);
+    }
+
+    @Override
+    public void readRows(
+        ReadRowsRequest request, StreamObserver<ReadRowsResponse> responseObserver) {
+      requests.add(request);
```
igorbernstein2, line 150: I would add a 1 or 2 ms sleep here so that we have something to measure

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/src/test/java/com/google/cloud/bigtable/hbase1_x/TestMetrics.java
igorbernstein2: Also, please add checks for the min bound for both measurements

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/src/test/java/com/google/cloud/bigtable/hbase1_x/TestMetrics.java
```diff
@@ -0,0 +1,231 @@
...
+      } else {
+        responseObserver.onNext(ReadRowsResponse.newBuilder().build());
+        responseObserver.onCompleted();
+      }
+    }
+  }
+
+  private static class FakeMetricRegistry implements MetricRegistry {
+
+    private final Map<String, AtomicLong> results = new HashMap<>();
```
igorbernstein2, line 162: I think this will cause flakes because the registry will be accessed from multiple threads.  I think you might be better off with using synchronization here:

```java
private final Object lock = new Object();
Map<String, Long> stats = new HashMap();

inc() {
  synchronized(lock) {
    long count = Objects.firstNonNull(stats.get(name), 0);
    count++;
    stats.put(name, count);
  }
}
```

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/src/test/java/com/google/cloud/bigtable/hbase1_x/TestMetrics.java
```diff
@@ -0,0 +1,231 @@
...
+          }
+          results.get(name).getAndDecrement();
+        }
+      };
+    }
+
+    @Override
+    public Timer timer(final String name) {
+      return new Timer() {
+        final long start = System.currentTimeMillis();
```
igorbernstein2, line 192: Please use a StopWatch

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/src/test/java/com/google/cloud/bigtable/hbase1_x/TestMetrics.java
```diff
@@ -0,0 +1,231 @@
...
+      if (callCount.getAndIncrement() < 4) {
+        responseObserver.onError(new StatusRuntimeException(Status.UNAVAILABLE));
+      } else {
+        responseObserver.onNext(ReadRowsResponse.newBuilder().build());
+        responseObserver.onCompleted();
+      }
+    }
+  }
+
+  private static class FakeMetricRegistry implements MetricRegistry {
```
igorbernstein2, line 160: Please add a bit of doc explaining how this registry deals with multiple values...The implementation is a bit surprising: count() is persistent, but time() records the last value , meter is persistent

bigtable-hbase-2.x-parent/bigtable-hbase-2.x-hadoop/pom.xml
```diff
@@ -81,6 +81,11 @@ limitations under the License.
       <artifactId>metrics-core</artifactId>
       <version>${dropwizard.metrics.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.cloud.bigtable</groupId>
+      <artifactId>bigtable-metrics-api</artifactId>
+      <version>${project.version}</version>
+    </dependency>
```
igorbernstein2, line 88: please move this up to line 69 so its more prominent

bigtable-hbase-2.x-parent/bigtable-hbase-2.x-hadoop/pom.xml
```diff
@@ -197,6 +202,7 @@ limitations under the License.
             <usedDependency>com.google.code.findbugs:jsr305</usedDependency>
             <usedDependency>commons-logging:commons-logging</usedDependency>
             <usedDependency>io.dropwizard.metrics:metrics-core</usedDependency>
+            <usedDependency>com.google.cloud.bigtable:bigtable-metrics-api</usedDependency>
```
igorbernstein2, line 205: Please move this up

bigtable-hbase-2.x-parent/bigtable-hbase-2.x-shaded/pom.xml
```diff
@@ -109,6 +109,11 @@ limitations under the License.
       <artifactId>metrics-core</artifactId>
       <version>${dropwizard.metrics.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.cloud.bigtable</groupId>
+      <artifactId>bigtable-metrics-api</artifactId>
+      <version>${project.version}</version>
+    </dependency>
```
igorbernstein2, line 116: please move this up to just before jsr305 (it should be the first in the manually promoted list)
but hbase-shaded-client is out of place. This is fixed in master

bigtable-hbase-2.x-parent/bigtable-hbase-2.x-shaded/pom.xml
```diff
@@ -179,6 +184,7 @@ limitations under the License.
                   <exclude>org.apache.htrace:htrace-core4</exclude>
                   <exclude>org.apache.yetus:audience-annotations</exclude>
                   <exclude>javax.annotation:javax.annotation-api</exclude>
+                  <exclude>com.google.cloud.bigtable:bigtable-metrics-api</exclude>
```
igorbernstein2, line 187: move up please

bigtable-hbase-2.x-parent/bigtable-hbase-2.x-shaded/pom.xml
```diff
@@ -343,6 +350,7 @@ limitations under the License.
               io.opencensus:opencensus-exporter-stats-stackdriver
             </usedDependency>
             <usedDependency>io.grpc:grpc-census</usedDependency>
+            <usedDependency>com.google.cloud.bigtable:bigtable-metrics-api</usedDependency>
```
igorbernstein2, line 353: move up please

bigtable-hbase-2.x-parent/bigtable-hbase-2.x-hadoop/pom.xml
kolea2: this plugin is now skipped

bigtable-client-core-parent/bigtable-metrics-api/pom.xml
```diff
@@ -32,7 +32,7 @@
     <dependency>
       <groupId>io.dropwizard.metrics</groupId>
       <artifactId>metrics-core</artifactId>
-      <version>${dropwizard.metrics.version}</version>
+      <version>${hbase1-metrics.version}</version>
```
kolea2, line 35: @igorbernstein2 is hbase1 the right version to use?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapter.java
igorbernstein2: I think you want to just update `lastRetryStatus`

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapterFactory.java
```diff
@@ -0,0 +1,62 @@
...
+      return metrics;
+    }
+
+    metrics = this.metrics.get(key);
+    if (metrics != null) {
+      return metrics;
+    }
+    metrics = RpcMetrics.createRpcMetrics(key);
+    this.metrics.put(key, metrics);
+    return metrics;
```
igorbernstein2, line 60: this should be synchronized...multiple threads can call newTracer concurrently

bigtable-client-core-parent/bigtable-metrics-api/src/main/java/com/google/cloud/bigtable/metrics/DropwizardMetricRegistry.java
```diff
@@ -107,7 +107,7 @@ public Timer timer(String name) {
     return new Timer() {
 
       @Override
-      public Timer.Context time() {
+      public Context time() {
```
igorbernstein2, line 110: Please keep this as Timer.Context ... Context is a bit generic of a name

bigtable-dataflow-parent/bigtable-hbase-beam/pom.xml
```diff
@@ -33,7 +33,7 @@ limitations under the License.
       <dependency>
         <groupId>com.google.errorprone</groupId>
         <artifactId>error_prone_annotations</artifactId>
-        <version>2.3.3</version>
+        <version>2.3.4</version>
```
igorbernstein2, line 36: is this intentional? I'm a bit surprised that this would change

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-hadoop/pom.xml
```diff
@@ -71,6 +71,12 @@ limitations under the License.
       <version>${hbase1.version}</version>
     </dependency>
 
+    <dependency>
+      <groupId>com.google.cloud.bigtable</groupId>
+      <artifactId>bigtable-metrics-api</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+
```
igorbernstein2, line 79: I think this is belongs in the manually promoted section below : we aren't shading it, and the reason its here is so that it ends up in the deps

bigtable-dataflow-parent/bigtable-hbase-beam/pom.xml
kolea2: The enforcer fails with:
```
+-com.google.cloud.bigtable:bigtable-hbase-beam:1.14.1-SNAPSHOT
  +-com.google.cloud.bigtable:bigtable-hbase-1.x-shaded:1.14.1-SNAPSHOT
    +-com.google.cloud.bigtable:bigtable-metrics-api:1.14.1-SNAPSHOT
      +-com.google.api:api-common:1.10.1
        +-com.google.guava:guava:29.0-android
          +-com.google.errorprone:error_prone_annotations:2.3.3 (managed) <-- com.google.errorprone:error_prone_annotations:2.3.4
```

bigtable-hbase-1.x-parent/bigtable-hbase-1.x/src/test/java/com/google/cloud/bigtable/hbase1_x/TestMetrics.java
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    @Override
+    public void readRows(
+        ReadRowsRequest request, StreamObserver<ReadRowsResponse> responseObserver) {
+      requests.add(request);
+      try {
+        Thread.sleep(5);
+      } catch (InterruptedException e) {
+        e.printStackTrace();
```
igorbernstein2, line 167: shouldn't this bubble the error to responseObserver?

bigtable-hbase-2.x-parent/bigtable-hbase-2.x-shaded/pom.xml
```diff
@@ -192,6 +199,7 @@ limitations under the License.
                     -->
                     <exclude>com.google.cloud.bigtable.hbase*.**</exclude>
                     <exclude>com.google.cloud.bigtable.metrics.**</exclude>
+                    <exclude>com.google.cloud.bigtable.hbase.wrappers.veneer.metrics.**</exclude>
```
igorbernstein2, line 202: this should already be covered by `com.google.cloud.bigtable.hbase*.**`

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/metrics/MetricsApiTracerAdapter.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+    operationTimer.close();
+  }
+
+  @Override
+  public void connectionSelected(String id) {}
+
+  @Override
+  public void attemptStarted(int attemptNumber) {
+    rpcTimer = rpcMetrics.timeRpc();
```
igorbernstein2, line 77: I dont think it makes a difference, but just in case code shifts around, can you reset lastRetryStatus here?

igorbernstein2 (approved):
lgtm after nit

PR #2642, opened 2020-10-08
fix: Bigtable emulator rejects fuzzy key filters using integer values

dmitry-fa:
Fixes https://github.com/googleapis/java-bigtable/issues/378

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ReaderExpressionHelper.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ReaderExpressionHelper.java
@@ -108,6 +108,13 @@ public class ReaderExpressionHelper {
         writeNullCharacterBytes();
         return;
       }
+      if (unquoted < 0) {
+        // -1 -> \xFF
+        for (byte b : Bytes.toBytes(String.format("\\x%02X", (byte) unquoted))) {
+          delegate.write(b);
+        }
+        return;
+      }
       if ((unquoted < 'a' || unquoted > 'z')
           && (unquoted < 'A' || unquoted > 'Z')
           && (unquoted < '0' || unquoted > '9')
--- a/bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/adapters/filters/TestFuzzyRowFilterAdapter.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/adapters/filters/TestFuzzyRowFilterAdapter.java
@@ -56,4 +56,88 @@ public class TestFuzzyRowFilterAdapter {
 
     Assert.assertEquals(expected.toProto(), adaptedFilter.toProto());
   }
+
+  @Test
+  public void singleByteKey() throws IOException {
+    Pair<Byte, String>[] testData =
+        new Pair[] {
+          Pair.newPair((byte) -128, "\\x80"),
+          Pair.newPair((byte) -127, "\\x81"),
+          Pair.newPair((byte) -42, "\\xD6"),
+          Pair.newPair((byte) -2, "\\xFE"),
+          Pair.newPair((byte) -1, "\\xFF"),
+          Pair.newPair((byte) 0, "\\x00"),
+          Pair.newPair((byte) 1, "\\\001"),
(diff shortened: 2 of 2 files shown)


bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/adapters/filters/TestFuzzyRowFilterAdapter.java
```diff
@@ -56,4 +56,88 @@ public void fuzzyKeysAreTranslatedToRegularExpressions() throws IOException {
...
+
+      FuzzyRowFilter fuzzyNo = new FuzzyRowFilter(ImmutableList.of(Pair.newPair(key, no)));
+      Filters.Filter filterNo = FilterAdapter.buildAdapter().adaptFilter(null, fuzzyNo).get();
+      Assert.assertEquals(expectedNo.toProto(), filterNo.toProto());
+    }
+  }
+
+  @Test
+  public void keysAreProducedFromIntegers() throws IOException {
+    Pair<Integer, String>[] testData =
```
elharo, line 102: go/nopair

dmitry-fa: What do you mean?

igorbernstein2: I think a map would be more appropriate

dmitry-fa: agree, fixed.

Naveg:
Looking forward to this fix! Let me know if I can help somehow

kolea2:
@dmitry-fa this looks good, we will also need to add this to the `bigtable-1.x` branch for the next release. @igorbernstein2 any additional comments?

igorbernstein2 (changes requested):
I think this also needs an integration test that shows that this works with the emulator

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ReaderExpressionHelper.java
```diff
@@ -108,6 +108,13 @@ public void write(int unquoted) throws IOException {
         writeNullCharacterBytes();
         return;
       }
+      if (unquoted < 0) {
```
igorbernstein2, line 111: Shouldn't this be `> 127` ? unquoted is supposed to represent an unsigned byte here

igorbernstein2:
I think this should be fixed in the emulator instead, otherwise every client will have to workaround this issue:
https://github.com/googleapis/google-cloud-go/issues/1373

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/read/ReaderExpressionHelper.java
dmitry-fa: @igorbernstein2 No, this method accepts int but is invoked with a byte passed as the parameter:
https://github.com/googleapis/java-bigtable-hbase/blob/534288cfebf4732b7998c369a6e278581a64f758/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/adapters/filters/FuzzyRowFilterAdapter.java#L75

dmitry-fa:
@igorbernstein2  I think this change is still required, without it the pattern not correct:

Quoting Stack trace from the issue: `Bad pattern "\\x00\\x00\\x00\x80\\C\\C\\C\\C\\C\\C\\C\\C\\C\\C\\C\\C\\C\\C\\C\\C\\C*"`
\x80 instead of \\x80

dmitry-fa:
@igorbernstein2 

> 
> I think this also needs an integration test that shows that this works with the emulator

the problem is not in the emulator, I believe the provided unit test is enough to cover the problem: it fails with the previous version and passes with the updated one.

Naveg:
Any update on this one? We're eagerly looking forward to the fix

PR #2650, opened 2020-10-13
deps: update slf4j to 1.7.30

kolea2:
Opening in favor of #2577

### DIFF
--- a/bigtable-hbase-2.x-parent/bigtable-hbase-2.x-integration-tests/pom.xml
+++ b/bigtable-hbase-2.x-parent/bigtable-hbase-2.x-integration-tests/pom.xml
@@ -395,6 +395,7 @@ limitations under the License.
       <scope>test</scope>
     </dependency>
 
+    <!-- adding slf4j due to older versions in transitive deps-->
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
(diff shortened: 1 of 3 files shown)


elharo (approved):

bigtable-hbase-2.x-parent/bigtable-hbase-2.x-integration-tests/pom.xml
```diff
@@ -395,6 +395,12 @@ limitations under the License.
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+      <version>${slf4j.version}</version>
+      <scope>test</scope>
+    </dependency>
```
igorbernstein2, line 403: I'm a bit confused why this had to change. Can you add a comment?

kolea2: The deps build was failing as we were still pulling in some older versions of slf4j: https://source.cloud.google.com/results/invocations/12b8712b-605a-4315-86c3-0c50dadbd210/targets/cloud-devrel%2Fclient-libraries%2Fjava%2Fjava-bigtable-hbase%2Fpresubmit%2Fdependencies/log. Because this is a test module, I figure the easiest solution was to declare it explicitly. Will add a comment

igorbernstein2: It seems like hbase-shaded-client uses slf4j 1.7.25 and since we are trying to be compatible with hbase, we should use that version

4oVUA2zfQRQU7oB

kolea2: ok, fair enough. I'll close this

PR #2673, opened 2020-11-02
chore: configure autorelease for 1.x branch

kolea2:


chingor13 (changes requested):
There should be a secret manager script in `.kokoro/populate-secrets.sh` which is invoked from `.kokoro/trampoline.sh`

kolea2:
> There should be a secret manager script in `.kokoro/populate-secrets.sh` which is invoked from `.kokoro/trampoline.sh`

We already have these as integration tests are already enabled: https://github.com/googleapis/java-bigtable-hbase/blob/bigtable-1.x/.kokoro/populate-secrets.sh. Let me know if I'm missing something!

chingor13 (approved):

PR #2680, opened 2020-11-04
feat: Enabling OpenCensus publishing for HBase shaded clients

shantstepanian:
Including the OpenCensus libraries into the shaded jars so that clients can easily enable it. Needed to help diagnose some user issues that are still on the HBase client


kolea2 (approved):

igorbernstein2 (approved):
lgtm

PR #2683, opened 2020-11-06
feat: feat: Add keepalive in grpc channelbuilder #2682 in bigtable-1.x

### LINKED ISSUE #2681: Add keepalive in grpc channel for Bigtable client.
Keepalive is recommended to use when your peer dies before it is able to notify you or does not notify you. The reasons can be kernel panic, termination of the process handling that pee

Hence, if a server does not send GOAWAY to the client, the grpc client will send packets to the TCP peer but the TCP peer will not send any packets. This can happen when a server nodes goes down due to machine failure, release/rollouts.

Adding keepalive will prevent requests being blackholed/client being hung.

sushanb:
Sets KeepAlive timer to 30s (Google Frontends are configured limits keepalive calls at 30s by default. If used below 30s, grpc automatically increases keepalive time by 2x after too_many_ping commands).

Sets KeepAliveTimeout to 10s (conservative)

Sets keepAliveWithoutCalls to false as it can potentially increase load.

Fixes #2681


igorbernstein2 (approved):
lgtm

PR #2687, opened 2020-11-10
fix: Extract out trace management from ResultScanners

### LINKED ISSUE #2458: Extract out trace management from ResultScanners
Related to [#2444](https://github.com/googleapis/java-bigtable-hbase/pull/2444#discussion_r397181648) 

While transitioning the `BigtableWhileMatchResultScannerAdapter` with new DataClientWrappers the `span`s are now are left open, It needs to be closed/ends explicitly when the read stream ends.

dmitry-fa:
Fixes #2458

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
@@ -55,6 +55,7 @@ import org.apache.hadoop.hbase.DoNotRetryIOException;
 import org.apache.hadoop.hbase.HTableDescriptor;
 import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.client.AbstractBigtableConnection;
+import org.apache.hadoop.hbase.client.AbstractClientScanner;
 import org.apache.hadoop.hbase.client.Admin;
 import org.apache.hadoop.hbase.client.Append;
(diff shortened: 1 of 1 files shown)


kolea2:
@shantstepanian mind taking a look?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
```diff
@@ -276,15 +277,40 @@ public Result get(Get get) throws IOException {
...
+        }
+
+        @Override
+        public void close() {
+          try {
+            scanner.close();
+          } finally {
+            span.end();
+          }
+        }
```
igorbernstein2, line 307: This is a bit problematic: under normal usage this will end up calling span.end() twice. Also, when an exception is thrown, it needs to end the span with the appropriate status.

In general I actually think that trying to wrap a span around the lifetime of an iterator is problematic to begin with. I think we should probably remove this and ensure that the underlying ReadRowsOperation has a proper span

igorbernstein2 (changes requested):

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
dmitry-fa: refixed, see #2780

PR #2702, opened 2020-11-18
fix: bigtable-hbase-2.x-hadoop incompatible with hbase-shaded-client 2.x

### LINKED ISSUE #2601: bigtable-hbase-2.x-hadoop incompatible with hbase-shaded-client 2.x
The bigtable-hbase-2.x-hadoop dependency expects hbase-shaded-client 2.x, while the code is compiled against HBase 1.x.

#### Steps to reproduce

  1. Setup a Java project and include the dependency com.google.cloud.bigtable:bigtable-hbase-2.x-hadoop:1.15.0. This dependency transitively depends on org.apache.hbase:hbase-shaded-client:2.2.3
  2. call getClusterMetrics() on a AbstractBigtableAdmin

#### Code example

```java
org.apache.hadoop.hbase.client.Connection connection = createConnection();
org.apache.hadoop.hbase.client.AbstractBigtableAdmin admin = connection.getAdmin().
int size = admin.getClusterMetrics().getLiveServerMetrics().size();
```

#### Stack trace
```
Caused by: java.lang.NoSuchMethodError: org.apache.hadoop.hbase.ClusterStatus: method <init>()V not found
	at org.apache.hadoop.hbase.client.AbstractBigtableAdmin$2.<init>(AbstractBigtableAdmin.java:705)
	at org.apache.hadoop.hbase.client.AbstractBigtableAdmin.getClusterStatus(AbstractBigtableAdmin.java:705)
	at com.google.cloud.bigtable.hbase2_x.BigtableAdmin.getClusterMetrics(BigtableAdmin.java:787)
	at org.apache.hadoop.hbase.client.Admin.getClusterMetrics(Admin.java:1564)
	at com.privitar.agrotera.adapters.cluster.hadoop.hbase.apache.admin.ApacheHBaseAdmin.getNumberOfRegionServersInCluster(ApacheHBaseAdmin.java:43)
```

dmitry-fa:
Fixes #2601

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
@@ -40,9 +40,11 @@ import com.google.common.util.concurrent.SettableFuture;
 import com.google.protobuf.ByteString;
 import io.grpc.Status;
 import java.io.IOException;
+import java.lang.reflect.Constructor;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
```diff
@@ -688,13 +690,50 @@ public void deleteColumn(final String tableName, final String columnName) throws
...
+          // TODO(sduskis): Point the server name to options.getServerName()
+          return Collections.emptyList();
+        }
+      };
+    } catch (NoSuchMethodError e) {
+      return getClusterStatusHbase2x();
+    }
+  }
+
+  private ClusterStatus getClusterStatusHbase2x() {
```
kolea2, line 706: is it possible to instead pull down and override the logic to the 2x implementation of BigtableAdmin?

dmitry-fa: I don't quite understand what do you mean by saying "to pull down the logic".
The problem here that ClusterStatus 1x is dramatically different from ClusterStatus 2x. They don't even have the same constructor. So, I can't construct ClusterStatus 2x using API from 1x, the code won't compile.

igorbernstein2: I agree with @kolea2, the there should be 2 separate implementations of this in bigtable-hbase1x and in bigtable-hbase2x.

igorbernstein2 (changes requested):

bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableAdmin.java
dmitry-fa: I got it, thanks for the explanation! Fixed as you suggested in a separate PR, so I close this one.

PR #2717, opened 2020-11-25
feat: Adding timeouts for RPC Retry Attempts for HBase client (on bigtable-1.x branch)

shantstepanian:
We'd like to give user the ability to set a shorter timeout for individual RPC attempts within an overall timeout window (e.g. to allow multiple retries during the timeout window)


bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestRpcRetryBehavior1x.java
```diff
@@ -0,0 +1,167 @@
+package com.google.cloud.bigtable.hbase;
```
shantstepanian, line 1: This test is the main work in progress:

* Still looking to test different methods
* Will see how to add this for the HBase 2.x client too

shantstepanian (commented):
The key change is in AbstractRetryingOperation to leverage the new retry settings, BigtableOptionsFactory to provide the options, and BigtableDataGrpcClient to pass the options from the user to AbstractRetryingOperation.

There are a lot of classes in the middle that needed changing to facilitate this (e.g. a bunch of builders)

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
```diff
@@ -191,7 +253,9 @@ private CallOptionsConfig(Builder builder) {
       readRowsTimeout = longRpcTimeoutMs;
     }
     this.mutateRpcTimeoutMs = mutateTimeout;
-    this.readStreamRpcTimeoutMs = readRowsTimeout;
+    this.mutateRpcAttemptTimeoutMs = builder.mutateRpcAttemptTimeoutMs;
+    this.readStreamRpcTimeoutMs = readRowsTimeout;  // TODO make the names consistent?
```
shantstepanian, line 257: Highlighting this naming inconsistency for @igorbernstein2 

Based on the usages of CallOptionsConfig.getReadStreamRpcTimeoutMs, it seems like the ReadRows name is preferred over ReadStream, so I'd prefer to rename to readRows. LMKWYT

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/CallOptionsFactory.java
```diff
@@ -39,7 +39,23 @@
    * contain state, specifically start time with an expiration is set; in cases when timeouts are
    * used, implementations should create a new CallOptions each time this method is called.
    */
-  <RequestT> CallOptions create(MethodDescriptor<RequestT, ?> descriptor, RequestT request);
+  <RequestT> CallOptions createForRpc(RequestT request);
```
shantstepanian, line 42: Note the renaming of the original create() method - now named as createForRpc and createForRpcAttempt to be more explicit

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/CallOptionsFactory.java
```diff
@@ -39,7 +39,23 @@
...
+
+  /**
+   * Provide a {@link CallOptions} object to be used in a single RPC attempt, where an RPC operation
+   * may have multiple RPC attempts. {@link CallOptions} can
+   * contain state, specifically start time with an expiration is set; in cases when timeouts are
+   * used, implementations should create a new CallOptions each time this method is called.
+   */
+  <RequestT> CallOptions createForRpcAttempt(RequestT request);
+
+  /**
```
shantstepanian, line 52: Note this javadoc and the corresponding implementation for why we use Optional<Integer> in the other classes above

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
```diff
@@ -104,6 +112,19 @@ public Builder setShortRpcTimeoutMs(int shortRpcTimeoutMs) {
       return this;
     }
 
+    /**
+     * The amount of milliseconds to wait before issuing a client side timeout for short RPCs.
```
igorbernstein2, line 116: Please expand this comment to state that how this differs from shortRpcTimeout. ie something along the lines of "Unlike shortRpcTimeout, this limits how long each attempt will take, whereas shortRpcTimeout will limit the total amount across all of the attempts".

Also please mention what short means here. ie "This setting affects point reads and writes like checkAndMutate, mutateRow and readRows with single keys"

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
```diff
@@ -135,6 +156,21 @@ public Builder setMutateRpcTimeoutMs(int mutateRpcTimeoutMs) {
       return this;
     }
 
+    /**
+     * The amount of time in milliseconds to wait before issuing a client side timeout for row
+     * mutation RPCs.
```
igorbernstein2, line 161: Please clarify that this only affects bulk mutations. And similar to above explain the difference between mutateRpcTimeout

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
```diff
@@ -149,16 +185,34 @@ public Builder setReadRowsRpcTimeoutMs(int readStreamRpcTimeoutMs) {
       return this;
     }
 
+    /**
+     * The amount of time in milliseconds to wait before issuing a client side timeout for readRows
+     * streaming RPCs.
+     *
+     * @param readRowsRpcAttemptTimeoutMs timeout value in milliseconds.
+     * @return a {@link Builder} object, for chaining
```
igorbernstein2, line 193: Similar to above, explain that this used for scans and how it differs from the existing non-attempt setting

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
```diff
@@ -223,6 +283,15 @@ public int getShortRpcTimeoutMs() {
     return shortRpcTimeoutMs;
   }
 
+  /**
+   * Getter for the field <code>shortRpcAttemptTimeoutMs</code>.
+   *
```
igorbernstein2, line 288: also add that if null, then the attempt rpc will be bounded by the remaining operation timeout.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
```diff
@@ -104,6 +112,19 @@ public Builder setShortRpcTimeoutMs(int shortRpcTimeoutMs) {
       return this;
     }
 
+    /**
+     * The amount of milliseconds to wait before issuing a client side timeout for short RPCs.
+     *
+     * @param shortRpcAttemptTimeoutMs timeout value in milliseconds.
+     * @return a {@link Builder} object, for chaining
+     */
+    public Builder setShortRpcAttemptTimeoutMs(int shortRpcAttemptTimeoutMs) {
```
igorbernstein2, line 121: Can you annotate all of the new methods with `@BetaApi`?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
```diff
@@ -253,6 +331,15 @@ public int getReadStreamRpcTimeoutMs() {
...
+   *
+   * @return an Optional int, populated if the property was present.
+   */
+  public Optional<Integer> getReadStreamRpcAttemptTimeoutMs() {
+    return readStreamRpcAttemptTimeoutMs;
+  }
+
   /** {@inheritDoc} */
   @Override
   public boolean equals(Object obj) {
```
igorbernstein2, line 345: Please update equals & hashcode & toString

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/CallOptionsFactory.java
igorbernstein2: Can you also update the javadoc that despite the name being RPC, its actually used for operations?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/CallOptionsFactory.java
```diff
@@ -48,17 +63,29 @@
    * <p>For internal use only - public for technical reasons.
    */
   @InternalApi("For internal usage only")
-  public static class Default implements CallOptionsFactory {
+  class Default implements CallOptionsFactory {
```
mutianf, line 66: Not sure why this was a class before. Maybe we don't need to have separate classes for default and configured behaviors? CallOptionFactory could return default values if CallOptionsConfig is not provided?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/CallOptionsFactory.java
```diff
@@ -98,6 +124,34 @@ public ConfiguredCallOptionsFactory(CallOptionsConfig config) {
       }
     }
 
+    @Override
+    public <RequestT> CallOptions createForRpcAttempt(RequestT request, Deadline currentDeadline) {
```
igorbernstein2: please rename currentDeadline to operationDeadline

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/CallOptionsFactory.java
```diff
@@ -98,6 +124,34 @@ public ConfiguredCallOptionsFactory(CallOptionsConfig config) {
...
+      if (config.isUseTimeout() && request != null) {
```
igorbernstein2: The logic is a bit hard to follow here. Can you reorg with an early return?
```java
// grpc timeouts are disabled so don't set a deadline
if (!config.isUseTimeout()) {
  return CallOptions.DEFAULT;
}
// TODO: investigate if this is ever happens
if (request == null) {
  return CallOptions.DEFAULT;
}

Deadline deadline = getRequestAttemptDeadline(request);
// shorten the attempt deadline if the remaining operation deadline is earlier
if (operationDeadline != null) {
  deadline = deadline.minimum(operationDeadline);
}
return CallOptions.DEFAULT.withDeadline(deadline);
```

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/CallOptionsFactory.java
```diff
@@ -98,6 +124,34 @@ public ConfiguredCallOptionsFactory(CallOptionsConfig config) {
...
+        } else {
+          chosenTimeoutMs = timeoutMs;
+        }
+        return CallOptions.DEFAULT.withDeadline(Deadline.after(chosenTimeoutMs, TimeUnit.MILLISECONDS));
+      } else {
+        return CallOptions.DEFAULT;
+      }
+    }
+
+    @Override
```
igorbernstein2, line 139: If you use the logic above, you no longer need to check this as it will fallback correctly

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/CallOptionsFactory.java
```diff
@@ -112,6 +166,24 @@ private int getRequestTimeout(Object request) {
       }
     }
 
+    /**
```
igorbernstein2, line 164: I think we should centralize the deadline merging logic to above and make this a simple getter for attempt deadlines

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/CallOptionsFactory.java
```diff
@@ -98,6 +124,29 @@ public ConfiguredCallOptionsFactory(CallOptionsConfig config) {
       }
     }
 
+    @Override
+    public <RequestT> CallOptions createForRpcAttempt(RequestT request, Deadline operationDeadline) {
+      if (!config.isUseTimeout() || request == null) {
```
igorbernstein2, line 129: Please split these in to 2 if statements and comment why we skip:

```java
public <RequestT> CallOptions createForRpcAttempt(RequestT request, CallOptions operationCallOpts) {

// grpc deadlines are disabled in the BigtableOptions, so skip setting any deadlines
if (!useTimeout)  { return operationCallOpts); }

// TODO: check why this exists and remove if request is never null
 if(request == null)  { return operationCallOpts); }

Also add an early return for isStreamingRead to retain previous behavior for no deadlines on scans
// If this is a streaming read, then the Watchdog will take affect and ensure that hanging does not occur.
if (isStreamingRead()  { return operationCallOpts); }

Optional<Deadline> attemptDeadline = getMaxRequestDeadline(request);
if (attemptDeadline.isPresent()) {
   if (operationCallOpts.getDeadline() == null || operationCallOpts.getDeadline.isAfter(attemptDeadline.get()) {
    return operationCallOpts.withDeadline(attemptDeadline.get());
  }
}
```

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
```diff
@@ -410,15 +418,19 @@ protected CallOptions getOperationCallOptions() {
    *     reads.
    * @return a {@link CallOptions}
    */
-  protected CallOptions getRpcCallOptions() {
-    if (callOptions.getDeadline() != null || isStreamingRead()) {
+  protected CallOptions getRpcAttemptCallOptions() {
+    if (callOptionsFactory.isRpcAttemptTimeoutEnabled(request)) {
+      return callOptionsFactory.createForRpcAttempt(request, operationCallOptions.getDeadline());
+    }
```
igorbernstein2, line 424: I think entire method can be just be simplified to `return callOptionsFactory.createForRpcAttempt(request, operationCallOptions)`

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
```diff
@@ -264,6 +265,15 @@
...
   public static final String BIGTABLE_RPC_TIMEOUT_MS_KEY = "google.bigtable.rpc.timeout.ms";
 
+  /**
+   * If timeouts are set, how many milliseconds should pass before a DEADLINE_EXCEEDED for an
+   * individual RPC attempt? Note that multiple attempts may happen within an overall operation,
+   * whose timeout is governed by {@link #BIGTABLE_RPC_TIMEOUT_MS_KEY}. Currently, this feature is
+   * experimental.
+   */
+  public static final String BIGTABLE_RPC_ATTEMPT_TIMEOUT_MS_KEY =
+      "google.bigtable.rpc.attempt_timeout.ms";
```
mutianf, line 275: To follow the naming convention, should it be attempt.timeout?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/CallOptionsFactory.java
```diff
@@ -98,6 +124,29 @@ public ConfiguredCallOptionsFactory(CallOptionsConfig config) {
       }
     }
 
+    @Override
+    public <RequestT> CallOptions createForRpcAttempt(RequestT request, Deadline operationDeadline) {
```
igorbernstein2, line 128: I think it would cleaner to pass operationCallOptions instead of just the Deadline here. This way if we start putting other attributes into the CallOptions they would get preserved (see sketch below)

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
```diff
@@ -282,13 +292,31 @@
...
    * If timeouts are set, how many milliseconds should pass before a DEADLINE_EXCEEDED for a long
    * read. Currently, this feature is experimental.
    */
   public static final String BIGTABLE_READ_RPC_TIMEOUT_MS_KEY =
       "google.bigtable.read.rpc.timeout.ms";
 
+  /**
+   * If timeouts are set, how many milliseconds should pass before a DEADLINE_EXCEEDED for an RPC
+   * attempt within a long mutation. Note that multiple attempts may happen within an overall
+   * operation, whose timeout is governed by {@link #BIGTABLE_MUTATE_RPC_TIMEOUT_MS_KEY}. Currently,
```
mutianf, line 314: Link here should be READ_RPC instead of MUTATE_RPC

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
```diff
@@ -410,15 +418,19 @@ protected CallOptions getOperationCallOptions() {
...
+
+    if (operationCallOptions.getDeadline() != null || isStreamingRead()) {
       // If the user set a deadline, honor it.
       // If this is a streaming read, then the Watchdog will take affect and ensure that hanging
       // does not occur.
       return getOperationCallOptions();
     } else {
       // Unary calls should fail after 6 minutes, if there isn't any response from the server.
-      return callOptions.withDeadlineAfter(UNARY_DEADLINE_MINUTES, TimeUnit.MINUTES);
+      return operationCallOptions.withDeadlineAfter(UNARY_DEADLINE_MINUTES, TimeUnit.MINUTES);
```
igorbernstein2, line 433: Lets move this into CallOptionsFactory to centralize all of the deadline computations

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
```diff
@@ -264,6 +265,15 @@
    */
   public static final String BIGTABLE_RPC_TIMEOUT_MS_KEY = "google.bigtable.rpc.timeout.ms";
 
+  /**
+   * If timeouts are set, how many milliseconds should pass before a DEADLINE_EXCEEDED for an
+   * individual RPC attempt? Note that multiple attempts may happen within an overall operation,
+   * whose timeout is governed by {@link #BIGTABLE_RPC_TIMEOUT_MS_KEY}. Currently, this feature is
+   * experimental.
+   */
+  public static final String BIGTABLE_RPC_ATTEMPT_TIMEOUT_MS_KEY =
```
igorbernstein2, line 274: Please mark all new public symbols  with BetaApi

pom.xml
```diff
@@ -148,8 +148,8 @@ limitations under the License.
         </plugin>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-surefire-plugin</artifactId>
-          <version>2.22.2</version>
+          <artifactId>maven-dependency-plugin</artifactId>
+          <version>3.1.2</version>
```
igorbernstein2, line 152: Can this be split into a separate pr? In case we have to rollback timeout changes

igorbernstein2 (changes requested):

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
igorbernstein2: +1

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/CallOptionsFactory.java
shantstepanian: I had already renamed the method to createForOperation, though I may still do the refactorings we discussed offline.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
shantstepanian: Done

pom.xml
shantstepanian: I will split this into a separate commit. I assume that should be fine? e.g. if we merge the PR and don't squash, then both commits will be there, and we can just revert the other commit if needed

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
shantstepanian: Done; I assume the same should be done for CallOptionsConfig?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
shantstepanian: Done

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
shantstepanian: Fixed

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
shantstepanian: Done, and updated the other new methods too

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
shantstepanian: Done

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
shantstepanian: Done

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
shantstepanian: Refactor is done

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
shantstepanian: Refactor is done

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/config/CallOptionsConfig.java
shantstepanian: done

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/CallOptionsFactory.java
shantstepanian: I see both sides of this; if the main implementation is simple, we may not need a default one. In this case, I can somewhat see why a default impl was used, since it didn't use the configurations.

Having said that - I refactored this whole thing quite a bit, as we wanted the CallOptionsFactory to handle more of the RPC and Deadline logic that was originally in AbstractRetryOperations. So I now do definitely have a separate default and main implementation.

(And they are now in separate files; this file was already too big as it was :) )

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/ConfiguredDeadlineGeneratorFactory.java
```diff
@@ -0,0 +1,166 @@
...
+     */
+    private Optional<Deadline> createOperationDeadline() {
+      Deadline contextDeadline = Context.current().getDeadline();
+      if (contextDeadline != null) {
+        return Optional.of(contextDeadline);
+      } else if (useTimeout) {
+        return Optional.of(Deadline.after(requestTimeoutMs, TimeUnit.MILLISECONDS));
+      } else {
+        return Optional.absent();
+      }
```
igorbernstein2, line 126: minor code style improvement:
```java
Deadline deadline = Optional.ofNullable(Context.current().getDeadline());
if (useTimeout) {
  deadline = deadline.or(Deadline.after(requestTimeoutMs, TimeUnit.MILLISECONDS));
}
return deadline;
```

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/ConfiguredDeadlineGeneratorFactory.java
```diff
@@ -0,0 +1,166 @@
...
+    public Optional<Deadline> getRpcAttemptDeadline() {
+      // Check for RPC Attempt Timeouts first for backward-compatibility, so that
+      // the existing logic (based mostly on the operation timeout) is preserved.
+      if (useTimeout && requestAttemptTimeoutMs.isPresent()) {
+        Deadline attemptDeadline =
+            Deadline.after(requestAttemptTimeoutMs.get(), TimeUnit.MILLISECONDS);
+        // Ensure we don't exceed the absolute operation deadline by taking the min of the attempt
+        // and
+        // operation deadlines.
+        return Optional.of(attemptDeadline.minimum(operationDeadline.get()));
```
igorbernstein2, line 154: I'm not sure that operationDeadline is always non-null here. If its guaranteed that if useTimeout is true then operationDeadline is set, can you add a precondition in the constructor with a comment explaining why.

And add a note here referencing it

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/ConfiguredDeadlineGeneratorFactory.java
shantstepanian: Fixed, though a few more Optionals were needed

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/ConfiguredDeadlineGeneratorFactory.java
shantstepanian: It is guaranteed to be non-null due to the useTimeout condition you mentioned.
I've added notes to clarify.
Not sure how strict of a precondition you want; let's talk offline.

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/ConfiguredDeadlineGeneratorFactory.java
```diff
@@ -0,0 +1,169 @@
...
+      return new DeadlineGeneratorImpl(
+          config.isUseTimeout(),
+          config.getShortRpcTimeoutMs(),
+          config.getShortRpcAttemptTimeoutMs(),
+          // Unary calls should fail after 6 minutes, if there isn't any response from the server.
+          Optional.of(TimeUnit.MINUTES.toMillis(UNARY_DEADLINE_MINUTES)));
+    }
+  }
+
+  // TODO(sduskis): This is only required because BigtableDataGrpcClient doesn't always use
```
igorbernstein2, line 73: I think its safe to remove this comment

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/DeadlineGenerator.java
```diff
@@ -0,0 +1,67 @@
...
+import io.grpc.CallOptions;
+import io.grpc.Context;
+import io.grpc.Deadline;
+
+/**
+ * A provider for RPC timeouts and deadlines for a given Bigtable client operation, governing both
+ * the full operation and retry attempts within an operation. A single {@link DeadlineGenerator}
+ * object should be used for a client operation instance and reused across retry attempts.
+ */
+public interface DeadlineGenerator {
```
igorbernstein2, line 28: I think this should be InternalApi

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
```diff
@@ -323,12 +324,9 @@ protected void resetStatusBasedBackoff() {
...
    */
   private ExponentialRetryAlgorithm createRetryAlgorithm(ApiClock clock) {
-    long timeoutMs = retryOptions.getMaxElapsedBackoffMillis();
+    Optional<Long> operationTimeoutMs = deadlineGenerator.getOperationTimeoutMs();
 
-    Deadline deadline = getOperationCallOptions().getDeadline();
-    if (deadline != null) {
-      timeoutMs = deadline.timeRemaining(TimeUnit.MILLISECONDS);
-    }
+    long timeoutMs = operationTimeoutMs.or((long) retryOptions.getMaxElapsedBackoffMillis());
```
igorbernstein2, line 329: Can you add a todo to try to move this logic into your deadline generator impl?

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/ConfiguredDeadlineGeneratorFactory.java
```diff
@@ -0,0 +1,169 @@
...
+          // Timeouts aren't specified if we have come this far.  If this is a streaming read, then
+          // no need to specify a fallback deadline, since the Watchdog will take affect and ensure
+          // that hanging does not occur.
+          Optional.<Long>absent());
+    } else if (request instanceof MutateRowsRequest) {
+      return new DeadlineGeneratorImpl(
+          config.isUseTimeout(),
+          config.getMutateRpcTimeoutMs(),
+          config.getMutateRpcAttemptTimeoutMs(),
+          // Unary calls should fail after 6 minutes, if there isn't any response from the server.
```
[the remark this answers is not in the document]
shantstepanian: Fixed up the comments in a few places around the fallback timeout (note the comment I added on line 52 for the streaming read and watchdog, which I had moved from elsewhere)

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/ConfiguredDeadlineGeneratorFactory.java
shantstepanian: Done

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/DeadlineGenerator.java
shantstepanian: Done

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestRpcRetryBehavior.java
```diff
@@ -0,0 +1,285 @@
...
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.Connection;
+import org.apache.hadoop.hbase.client.ConnectionFactory;
+import org.apache.hadoop.hbase.client.Table;
+import org.hamcrest.Matchers;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public abstract class TestRpcRetryBehavior {
```
igorbernstein2, line 54: I think you have to mark these as `@KnownHBaseGap` so that they don't run during our hbase compatibility tests

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/AbstractRetryingOperation.java
shantstepanian: Added a TODO comment. I did consider that option (basically having the backoffmillis as a parameter to getOperationTimeoutMs). I didn't as it ended up just being a fallback option that Optional easily represented. I'm not strongly in favor either way though, so I guess a TODO is fine for now

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/TestRpcRetryBehavior.java
shantstepanian: Done

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/CallOptionsFactory.java
igorbernstein2: I agree with Mattie, the Defaults arent ever actually used in production code. So I think we should remove them. However, I think that will create a lot more noise in PR and I dont want to block this feature. So I think we should just ship it as is

igorbernstein2 (approved):
LGTM!

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/CallOptionsFactory.java
shantstepanian: FYI, we do use this in BigtableTableAdminGrpcClient. There are no CallOptionsConfig defined for that class.
While it wouldn't be hard to inject it, I don't know what the implications would be to have it.
(If we trust our existing test suites to handle this change, then it would be quick to do this)

PR #2755, opened 2020-12-11
feat: Ability to import HBase Snapshot data into Cloud Bigtable using Dataflow

lichng:
A Dataflow pipeline template that could import HBase data from an HBase snapshot into Cloud Bigtable.


bigtable-dataflow-parent/bigtable-beam-import/pom.xml
```diff
@@ -249,6 +276,16 @@ limitations under the License.
           <transformers>
             <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer" />
           </transformers>
+	  <filters>
+	      <filter>
+		  <artifact>*:*</artifact>
+		  <excludes>
+		      <exclude>META-INF/*.SF</exclude>
```
vermas2012, line 283: What does this do?

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfiguration.java
```diff
@@ -0,0 +1,95 @@
...
+ */
+class HBaseSnapshotInputConfiguration {
+
+  private static final Log LOG = LogFactory.getLog(HBaseSnapshotInputConfiguration.class);
+  private static final int BATCH_SIZE = 1000;
+
+  private final Configuration hbaseConf;
+
+  /**
+   * Constructs a new top level source.
```
vermas2012, line 53: I am not sure, what source are we creating here?

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfiguration.java
```diff
@@ -0,0 +1,95 @@
...
+      ValueProvider<String> restoreDir) {
+
+    Preconditions.checkArgument(
+        snapshotDir.toString().startsWith("gs://"),
+        "snapshot folder must be hosted in a GCS bucket ");
+
+    Configuration conf = HBaseConfiguration.create();
+    try {
+      conf.set("hbase.rootdir", snapshotDir.toString());
+      conf.set("fs.AbstractFileSystem.gs.impl", "com.google.cloud.hadoop.fs.gcs.GoogleHadoopFS");
```
vermas2012, line 70: There must be a class somewhere that defines these config keys as constants?

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfiguration.java
```diff
@@ -0,0 +1,95 @@
...
+
+      this.LOG.debug(conf);
+      Job job = Job.getInstance(conf); // creates internal clone of hbaseConf
+      TableSnapshotInputFormat.setInput(
+          job, snapshotName.toString(), new Path(restoreDir.toString()));
+      conf = job.getConfiguration(); // extract the modified clone
+    } catch (Exception e) {
+      this.LOG.fatal(e);
+    }
+    this.hbaseConf = new SerializableConfiguration(conf).get();
```
vermas2012, line 89: To make this class serializable hbaseConf should be a SerializableConfiguration and the getHbaseConf should do hbaseConf.get(). 

SerializableConfiguration is a wrapper around conf, so all you are doing is wrapping the conf and unwrapping it in the same statement.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
```diff
@@ -0,0 +1,152 @@
...
+ * a metadata file that contains information about the runtime parameters that can be used for
+ * parameter validation purpose and more. A sample metadata file can be found at
+ * "src/main/resources/ImportJob_metadata".
+ *
+ * <p>An example using gcloud command line:
+ *
+ * <pre>
+ * gcloud beta dataflow jobs run $JOB_NAME \
+ *   --gcs-location gs://$TEMPLATE_PATH \
+ *   --parameters bigtableProject=$PROJECT,bigtableInstanceId=$INSTANCE,bigtableTableId=$TABLE,hbaseRootDir=gs://$HBASE_ROOT_PATH,snapshotName=$SNAPSHOT_NAME,restoreDir=gs://$RESTORE_PATH
```
vermas2012, line 84: "gs://$HBASE_ROOT_PATH

Should we expect the command line argument to be a path containing "gs://"?

bigtable-dataflow-parent/bigtable-beam-import/src/main/resources/log4j.properties
```diff
@@ -21,4 +21,6 @@ log4j.appender.stdout.Target=System.out
 log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
 log4j.appender.stdout.layout.ConversionPattern=%d{yyyy-MM-dd HH:mm:ss} %-5p %c{1}:%L - %m%n
 
-log4j.category.org.apache.hadoop.io.compress.CodecPool=WARN
+log4j.category.org.apache.hadoop.io.compress.CodecPool=TRACE
```
vermas2012, line 24: Is this intentional?  Can the log volume in TRACE mode flood the logs?

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
```diff
@@ -0,0 +1,191 @@
...
+  public void teardown() throws IOException {
+    final List<GcsPath> paths = gcsUtil.expand(GcsPath.fromUri(restoreDir + "/*"));
+
+    if (!paths.isEmpty()) {
+      final List<String> pathStrs = new ArrayList<>();
+
+      for (GcsPath path : paths) {
+        pathStrs.add(path.toString());
+      }
+      this.gcsUtil.remove(pathStrs);
```
vermas2012, line 134: Can't we just delete the whole restore folder?

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
```diff
@@ -0,0 +1,191 @@
...
+
+      // run pipeline
+      State state = ImportJobFromHbaseSnapshot.buildPipeline(importOpts).run().waitUntilFinish();
+      Assert.assertEquals(State.DONE, state);
+
+      // check data in bigtable
+      BigtableTableUtils destTable = new BigtableTableUtils(connection, tableId, CF);
+      Assert.assertEquals(
+          100 /* There are 100 rows in test snapshot*/,
+          destTable.readAllCellsFromTable().toArray().length);
```
vermas2012, line 188: Please add a TODO for me, we will add the sync-table after this. That way the data in this table can be verified.

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
lichng: It is for avoiding errors like "User class threw exception: java.lang.SecurityException: Invalid signature file digest for Manifest main attributes" when creating a shaded jar file.

references: 
https://stackoverflow.com/questions/34855649/invalid-signature-file-digest-for-manifest-main-attributes-exception-while-tryin
https://maven.apache.org/plugins/maven-shade-plugin/examples/includes-excludes.html

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfiguration.java
lichng: I think I missed this one when updating the code comments, will update this. Thanks for catching it.

bigtable-dataflow-parent/bigtable-beam-import/src/main/resources/log4j.properties
lichng: No it is not, will revert this.

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
```diff
@@ -25,7 +25,8 @@ limitations under the License.
   <artifactId>bigtable-beam-import</artifactId>
 
   <properties>
-    <mainClass>com.google.cloud.bigtable.beam.sequencefiles.Main</mainClass>
+    <mainClass>com.google.cloud.bigtable.beam.Main</mainClass>
+    <skipITs>false</skipITs>
```
kolea2, line 29: I don't think we should be changing this, ITs should be activated via profile.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * <pre>
+ * mvn compile exec:java \
+ *   -DmainClass=com.google.cloud.bigtable.beam.hbasesnapshots.ImportJobFromHbaseSnapshot \
+ *   -Dexec.args="--runner=DataflowRunner \
+ *                --stagingLocation=gs://$STAGING_PATH \
+ *                --project=$PROJECT \
+ *                --bigtableInstanceId=$INSTANCE \
+ *                --bigtableTableId=$TABLE \
+ *                --hbaseRootDir=gs://$HBASE_ROOT_PATH \
```
kolea2, line 56: the naming here is a bit confusing. is the bucket a temp or staging bucket, or permanent? `gs://$HBASE_ROOT_PATH` isn't clear to me.

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
lichng: Sure, will revert this.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
lichng: I believe so, will update the document.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
lichng: This is the path to the exported HBase snapshots(the data source). it would serve as the HBase root path in the job configuration.

Would $HBASE_EXPORT_ROOT_PATH be a better name?

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
lichng: This is following the prior practice in the sequencefiles import tests, the reason being we didn't create the folder in the setup, so it should not be deleted during teardown.

If we use setup to create this folder, then the setup and teardown will also need to deal with more cases like:
1, if the folder already exists?
2. the folder created but the setup it not successful?
3. the folder didn't got created so the teardown will be deleting an non-existing folder.

So for simplicity and conformance with the prior practice, I would suggest that we keep current practice.

lichng:
Hi @kolea2 @vermas2012 ,

This is good for another round of review. Please take another look.

We plan to keep tuning the performance, but it would be great if we can get this merge so we can start collaborating on the downstream processes(validation, syncTable).

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
kolea2: sounds good to me!

lichng: Done. Thanks!

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
vermas2012: this is confusing. gs://$RESTORE tells me that gs will be there and I only need to add "shitanshu-bucket/snapshot"

If we want to show expectations, we can have the sample variables declarations in the comment.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfiguration.java
```diff
@@ -0,0 +1,95 @@
...
+import org.apache.hadoop.io.Writable;
+import org.apache.hadoop.mapreduce.InputFormat;
+import org.apache.hadoop.mapreduce.Job;
+
+/**
+ * A {@link Configuration} that could be used in {@link HadoopFormatIO} for reading HBase snapshot
+ * hosted in Google Cloud Storage(GCS) bucket via GCS connector. It uses {@link
+ * TableSnapshotInputFormat} for reading HBase snapshots.
+ */
+class HBaseSnapshotInputConfiguration {
```
igorbernstein2, line 45: I dont quite understand why this need to be its own class. I think you can improve in one of 2 directions:
1. Make it a static method in a utility class:
   ```java
    class Utils {
      static Configuration createSnapshotReaderConfig(String projectId, ...)
    }
```
2. Or you can take in the other direction of creating a builder
   class HBaseSnapshotInputConfigBuilder {
        private String projectId;
        ...
        HBaseSnapshotInputConfigBuilder setProjectId(String projectId) {
          this.projectId = projectId;
           return this;
        }
       Configuration build() { 
          Precondition.checkNotNull(projectId);
          Configuration config = HBaseConfiguration.create();
          config.set("fs.gs.project.id", projectId);
          //...
         return config;
     }
```

But in the current approach of `ctor(args).getThing()` I dont see any benefit of.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfiguration.java
```diff
@@ -0,0 +1,95 @@
...
+  /**
+   * Constructs a HBase Configuration that could read HBase snapshot files from GCS Bucket.
+   *
+   * @param snapshotDir The path or pattern of the file(s) to read.
+   */
+  HBaseSnapshotInputConfiguration(
+      ValueProvider<String> gcsProjectId,
+      ValueProvider<String> snapshotDir,
+      ValueProvider<String> snapshotName,
+      ValueProvider<String> restoreDir) {
```
igorbernstein2, line 61: These should not be wrapped in ValueProviders. ValueProviders are intended for templated jobs where the parameters can be resolved during pipeline execution time. This job requires that the values are available at job construction time.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfiguration.java
igorbernstein2: I think you can get a few from here:
https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcs/src/main/java/com/google/cloud/hadoop/fs/gcs/GoogleHadoopFileSystemConfiguration.java#L52

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfiguration.java
```diff
@@ -0,0 +1,95 @@
...
+      conf.set("fs.defaultFS", snapshotDir.toString());
+      conf.set("google.cloud.auth.service.account.enable", "true");
+      conf.setClass(
+          "mapreduce.job.inputformat.class", TableSnapshotInputFormat.class, InputFormat.class);
+      conf.setClass("key.class", ImmutableBytesWritable.class, Writable.class);
+      conf.setClass("value.class", Result.class, Object.class);
+      ClientProtos.Scan proto = ProtobufUtil.toScan(new Scan().setBatch(BATCH_SIZE));
+      conf.set(TableInputFormat.SCAN, Base64.encodeBytes(proto.toByteArray()));
+
+      this.LOG.debug(conf);
```
igorbernstein2, line 81: I think `this` is a typo

pom.xml
```diff
@@ -76,10 +76,10 @@ limitations under the License.
     <hamcrest.version>1.3</hamcrest.version>
     <mockito.version>3.6.28</mockito.version>
     <!-- TODO: check if commons-codec was transitively updated to 1.13 and okhttp was updated to 2.7.5 when upgrading-->
-    <beam.version>2.19.0</beam.version>
```
igorbernstein2, line 79: Please move the version upgrades to a separate PR

pom.xml
```diff
@@ -76,10 +76,10 @@ limitations under the License.
...
+    <beam.version>2.24.0</beam.version>
     <!-- referred from bigtable-beam-import and bigtable-emulator -->
     <guava.version>30.0-android</guava.version>
-    <beam-guava.version>20.0</beam-guava.version>
```
igorbernstein2, line 82: It seems like beam started vendoring their guava. So I think we should update this prop name to just guava.version

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/Main.java
```diff
@@ -13,10 +13,14 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.google.cloud.bigtable.beam.sequencefiles;
+package com.google.cloud.bigtable.beam;
```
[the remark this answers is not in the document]
lichng: Sure, I kept the old Main class in the sequencefiles package.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
lichng: My understanding is that people will likely set RESTORE as env variable and just copy the full commandline (including "gs://") for execution.

I prefer this form since:
1. it provide more information for what the URI schema should be look like, and $*_PATH give intuitive guidance for what the value should be look like.
2. it is consistent with what has been published in https://github.com/googleapis/java-bigtable-hbase/blob/534288cfebf4732b7998c369a6e278581a64f758/bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/sequencefiles/ImportJob.java#L51

So unless there is another reviewer find this confusing, I am don't think we should change it to something like `--gcs-location $RESTORE_PATH`

Is there any format you believe could be more intuitive that we are not aware of?

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfiguration.java
lichng: Thanks for the detailed advice! 

Will refactor this into a builder class for better readability.

pom.xml
lichng: https://github.com/googleapis/java-bigtable-hbase/pull/2775 created for upgrading beam version

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfiguration.java
lichng: Done

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
```diff
@@ -78,6 +76,21 @@ limitations under the License.
       <version>${hbase.version}</version>
     </dependency>
 
+    <!-- For HBase 2.x, this should be hbase-mapreduce
```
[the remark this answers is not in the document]
lichng: Thanks @igorbernstein2 ,

Could you elaborate on how can I find out about this my own? "hbase-shaded-server includes hbase-shaded-client"

I checked `hbase-shaded-server` in maven repo page and didn't see a indirect dependency.
Also didn't find anything from the output of effective-pom.

Is it possible that the hbase-shaded-client is brought in by the two dependencies you mentioned?

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfiguration.java
lichng: I did some search in hadoop-connector repo as well as hadoop repo unfortunately didn't find any constant definition that we can use. 

Most of the code I found are using these immediate string values, would we prefer using constants(that we could define within this package)?

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
```diff
@@ -25,7 +25,7 @@ limitations under the License.
   <artifactId>bigtable-beam-import</artifactId>
 
   <properties>
-    <mainClass>com.google.cloud.bigtable.beam.sequencefiles.Main</mainClass>
+    <mainClass>com.google.cloud.bigtable.beam.Main</mainClass>
```
kolea2, line 28: what are our current instructions to users on how to use this? Will this break existing implementations?

lichng: This class path is only specified here, so it should be transparent to the outside users.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
```diff
@@ -0,0 +1,138 @@
...
+
+    return pipeline;
+  }
+
+  static PTransform<PCollection<Mutation>, PDone> createSink(ImportOptions opts) {
+    CloudBigtableTableConfiguration config = TemplateUtils.BuildImportConfig(opts);
+    config =
+        config
+            .toBuilder()
+            .withConfiguration(BigtableOptionsFactory.BIGTABLE_BULK_AUTOFLUSH_MS_KEY, "0")
```
vermas2012, line 134: I think the Autoflush bug is fixed. Please check with Mattie and remove this line.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/sequencefiles/HBaseResultToMutationFn.java
```diff
@@ -43,7 +44,8 @@
  * A {@link DoFn} function that converts a {@link Result} in the pipeline input to a {@link
  * Mutation} for output.
  */
-class HBaseResultToMutationFn extends DoFn<KV<ImmutableBytesWritable, Result>, Mutation> {
+@InternalApi
+public class HBaseResultToMutationFn extends DoFn<KV<ImmutableBytesWritable, Result>, Mutation> {
```
vermas2012, line 48: Should we move this class up a level?

vermas2012 (approved):

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
mutianf: Yes, it is fixed and merged into master: https://github.com/googleapis/java-bigtable-hbase/pull/2772

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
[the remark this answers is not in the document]
igorbernstein2: hbase-shaded-server is the shaded version of hbase-server. hbase-server has a direct dependency on hbase-client. Since hbase-shaded-server doesnt declare it, then you can guess that hbase-shaded-server absorbed the class files.  You can verify it by looking at the pom as well:
https://github.com/apache/hbase/blob/rel/1.4.13/hbase-shaded/hbase-shaded-server/pom.xml#L53

hbase-shaded-server depends on hbase-server, which depends on hbase-client
hbase-shaded-server doesnt exclude it
hbase-shaded-server uses the shaded plugin w/o any exclusions

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/sequencefiles/HBaseResultToMutationFn.java
lichng: It could be in the beam package as it is now shared.

However I think we have some refactoring expected in near future for sequencefiles import and for this PR I was trying to avoid modifying too many sequencefiles related files so to keep the review simple.

Can we leave this to the refactoring when we are unifying the styles of these two pipeline?

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
lichng: Thanks for confirming, will remove this line

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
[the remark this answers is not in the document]
lichng: Thanks @igorbernstein2 

I updated the pom file accordingly, PTAL.

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
kolea2: Yes makes sense, my larger question is do we expect users to already be importing `com.google.cloud.bigtable.beam.sequencefiles.Main` in source files that would break them by moving its package?

lichng: I since that is unlikely but should be fine as well since we left the original file in the old location, too.

kolea2: Oooh ok perfect. Thanks!

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfigBuilder.java
```diff
@@ -0,0 +1,105 @@
...
+  public Configuration build() {
+    Preconditions.checkNotNull(projectId);
+    Preconditions.checkNotNull(exportedSnapshotDir);
+    Preconditions.checkNotNull(snapshotName);
+    Preconditions.checkArgument(
+        exportedSnapshotDir.startsWith("gs://"), "snapshot folder must be hosted in a GCS bucket ");
+
+    Configuration conf = HBaseConfiguration.create();
+    try {
+      conf.set("hbase.rootdir", exportedSnapshotDir);
```
igorbernstein2, line 84: Please group these into logical groups of settings and add a comment as to what they do. ie. its not obvious what `hbase.rootdir` is 

Something like:
```
// Enable gcs are a filesystem
 conf.set("fs.AbstractFileSystem.gs.impl", "com.google.cloud.hadoop.fs.gcs.GoogleHadoopFS");
conf.set("fs.gs.project.id", projectId);
conf.set("google.cloud.auth.service.account.enable", "true");

// Configure the job input path
conf.set("hbase.rootdir", exportedSnapshotDir);
conf.set("fs.defaultFS", exportedSnapshotDir);

// Configure the inputformat for the snapshot files
conf.ssetClass ...

```

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfigBuilder.java
```diff
@@ -0,0 +1,105 @@
...
+      conf.set("fs.AbstractFileSystem.gs.impl", "com.google.cloud.hadoop.fs.gcs.GoogleHadoopFS");
+      conf.set("fs.gs.project.id", projectId);
+      conf.set("fs.defaultFS", exportedSnapshotDir);
+      conf.set("google.cloud.auth.service.account.enable", "true");
```
igorbernstein2, line 88: conf.setBoolean

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfigBuilder.java
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * A {@link Configuration} that could be used in {@link HadoopFormatIO} for reading HBase snapshot
+ * hosted in Google Cloud Storage(GCS) bucket via GCS connector. It uses {@link
+ * TableSnapshotInputFormat} for reading HBase snapshots.
+ */
+class HBaseSnapshotInputConfigBuilder {
+
+  private static final Log LOG = LogFactory.getLog(HBaseSnapshotInputConfigBuilder.class);
+  private static final int BATCH_SIZE = 1000;
```
igorbernstein2, line 46: Please comment to give context as to what batch this is

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfigBuilder.java
```diff
@@ -0,0 +1,105 @@
...
+      conf.set("fs.defaultFS", exportedSnapshotDir);
+      conf.set("google.cloud.auth.service.account.enable", "true");
+      conf.setClass(
+          "mapreduce.job.inputformat.class", TableSnapshotInputFormat.class, InputFormat.class);
+      conf.setClass("key.class", ImmutableBytesWritable.class, Writable.class);
+      conf.setClass("value.class", Result.class, Object.class);
+      ClientProtos.Scan proto = ProtobufUtil.toScan(new Scan().setBatch(BATCH_SIZE));
+      conf.set(TableInputFormat.SCAN, Base64.encodeBytes(proto.toByteArray()));
+
+      // LOG.debug(conf);
```
igorbernstein2, line 96: please remove commented out debug code

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfigBuilder.java
```diff
@@ -0,0 +1,105 @@
...
+      Job job = Job.getInstance(conf); // creates internal clone of hbaseConf
+      TableSnapshotInputFormat.setInput(job, snapshotName, new Path(restoreDir));
+      return job.getConfiguration(); // extract the modified clone
+    } catch (Exception e) {
+      LOG.fatal(e);
```
igorbernstein2, line 101: Why are you swallowing this exception? how can the job have any chance of continuing if this failed?

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
```diff
@@ -0,0 +1,132 @@
...
+import org.apache.beam.sdk.values.PDone;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.hadoop.hbase.client.Mutation;
+import org.apache.hadoop.hbase.client.Result;
+import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
+
+/**
+ * A job that imports data from HBase snapshot exports hosted in Cloud Storage bucket into Cloud
+ * Bigtable. This job can be run directly or as a Dataflow template.
```
igorbernstein2, line 43: I dont think the template part is try anymore

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
```diff
@@ -0,0 +1,132 @@
...
+    void setGcsProject(String gcsProjectId);
+
+    @Description("The HBase root dir where HBase snapshot files resides.")
+    String getHbaseRootDir();
+
+    @SuppressWarnings("unused")
+    void setHbaseRootDir(String hbaseRootDir);
+
+    @Description("Temp location for restoring snapshots")
+    String getRestoreDir();
```
igorbernstein2, line 79: Can you clarify what a temp location for restoring snapshots means? if its just a scratch dir, why not use a subdir in the job's temp directory?

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
```diff
@@ -0,0 +1,132 @@
...
+
+  public interface ImportOptions extends ImportJob.ImportOptions {
+    @Description("The GCP project id for the GCS bucket")
+    String getGcsProject();
+
+    @SuppressWarnings("unused")
+    void setGcsProject(String gcsProjectId);
+
+    @Description("The HBase root dir where HBase snapshot files resides.")
+    String getHbaseRootDir();
```
igorbernstein2, line 73: I think hbaseSnapshotSourcePath might be a better name

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
```diff
@@ -0,0 +1,132 @@
...
+ *                --restoreDir=gs://$RESTORE_PATH
+ * </pre>
+ */
+@InternalExtensionOnly
+public class ImportJobFromHbaseSnapshot {
+  private static final Log LOG = LogFactory.getLog(ImportJobFromHbaseSnapshot.class);
+
+  public interface ImportOptions extends ImportJob.ImportOptions {
+    @Description("The GCP project id for the GCS bucket")
+    String getGcsProject();
```
igorbernstein2, line 67: I think it would be nice to default this to the job project

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
```diff
@@ -0,0 +1,132 @@
...
+    void setHbaseRootDir(String hbaseRootDir);
+
+    @Description("Temp location for restoring snapshots")
+    String getRestoreDir();
+
+    @SuppressWarnings("unused")
+    void setRestoreDir(String restoreDir);
+
+    @Description("Snapshot name")
+    String getSnapshotName();
```
igorbernstein2, line 85: What is a snapshot name? how does it relate to hbaseRootDir? please document

bigtable-dataflow-parent/bigtable-beam-import/src/main/resources/log4j.properties
```diff
@@ -22,3 +22,5 @@ log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
 log4j.appender.stdout.layout.ConversionPattern=%d{yyyy-MM-dd HH:mm:ss} %-5p %c{1}:%L - %m%n
 
 log4j.category.org.apache.hadoop.io.compress.CodecPool=WARN
+log4j.logger.org.apache.hadoop.io.compress.Compression=TRACE
+log4j.logger.org.apache.hadoop.util.NativeCodeLoader=DEBUG
```
igorbernstein2, line 26: Why do these need such fine grained levels?

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
```diff
@@ -0,0 +1,198 @@
...
+  @Before
+  public void setup() throws Exception {
+    projectId = getTestProperty(BigtableOptionsFactory.PROJECT_ID_KEY);
+    instanceId = getTestProperty(BigtableOptionsFactory.INSTANCE_ID_KEY);
+    dataflowStagingLocation = getTestProperty(GOOGLE_DATAFLOW_STAGING_LOCATION);
+    region = getTestProperty(DATAFLOW_REGION);
+    String cloudTestDataFolder = getTestProperty(CLOUD_TEST_DATA_FOLDER);
+    if (!cloudTestDataFolder.endsWith(File.separator)) {
+      cloudTestDataFolder = cloudTestDataFolder + File.separator;
+    }
```
igorbernstein2, line 99: For tests, I think it would be better to enforce a single format

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
```diff
@@ -0,0 +1,198 @@
...
+    // Cloud Storage config
+    GcpOptions gcpOptions = PipelineOptionsFactory.create().as(GcpOptions.class);
+    gcpOptions.setProject(projectId);
+    gcsUtil = new GcsUtil.GcsUtilFactory().create(gcpOptions);
+
+    // Bigtable config
+    connection = BigtableConfiguration.connect(projectId, instanceId);
+    tableId = "test_" + UUID.randomUUID().toString();
+
+    System.out.println("Setting up integration tests");
```
igorbernstein2, line 114: Here and everywhere else, please use log statements instead of sysout

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
```diff
@@ -0,0 +1,198 @@
...
+  public void teardown() throws IOException {
+    final List<GcsPath> paths = gcsUtil.expand(GcsPath.fromUri(restoreDir + "/*"));
+
+    if (!paths.isEmpty()) {
+      final List<String> pathStrs = new ArrayList<>();
+
+      for (GcsPath path : paths) {
+        pathStrs.add(path.toString());
+      }
+      this.gcsUtil.remove(pathStrs);
```
igorbernstein2, line 137: can you avoid all of this by just setting a ttl on the files when creating them? It will make clean up more robust

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
```diff
@@ -0,0 +1,198 @@
...
+    // delete test table
+    BigtableConfiguration.connect(projectId, instanceId)
+        .getAdmin()
+        .deleteTable(TableName.valueOf(tableId));
+  }
+
+  @Test
+  public void testHBaseSnapshotImport() throws Exception {
+
+    try (Connection connection = BigtableConfiguration.connect(projectId, instanceId)) {
```
igorbernstein2, line 151: why not use the connection you already setup?

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
```diff
@@ -0,0 +1,198 @@
...
+      importPipelineOpts.setProject(projectId);
+      importPipelineOpts.setRegion(region);
+
+      ImportJobFromHbaseSnapshot.ImportOptions importOpts =
+          importPipelineOpts.as(ImportJobFromHbaseSnapshot.ImportOptions.class);
+      // setup GCP and bigtable
+      importOpts.setBigtableProject(StaticValueProvider.of(projectId));
+      importOpts.setBigtableInstanceId(StaticValueProvider.of(instanceId));
+      importOpts.setBigtableTableId(StaticValueProvider.of(tableId));
+      importOpts.setBigtableAppProfileId(null);
```
igorbernstein2, line 177: why is this necessary?

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfigBuilder.java
```diff
@@ -0,0 +1,105 @@
...
+  public HBaseSnapshotInputConfigBuilder setRestoreDir(String restoreDir) {
+    this.restoreDir = restoreDir;
+    return this;
+  }
+
+  public Configuration build() {
+    Preconditions.checkNotNull(projectId);
+    Preconditions.checkNotNull(exportedSnapshotDir);
+    Preconditions.checkNotNull(snapshotName);
+    Preconditions.checkArgument(
```
igorbernstein2, line 79: exportedSnapshotDir isnt an argument, I think that checkState is more appropriate

bigtable-dataflow-parent/bigtable-beam-import/src/test/generate_test_data.txt
```diff
@@ -0,0 +1,107 @@
+create 'test', 'cf', {SPLITS => ["1", "2", "3", "4", "5", "6", "7", "8", "9"]}
```
igorbernstein2, line 1: Please add some docs as to how someone can use this

igorbernstein2 (changes requested):

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfigBuilder.java
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.beam.hbasesnapshots;
+
+import static java.lang.System.*;
```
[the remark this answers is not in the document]
lichng: removed.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfigBuilder.java
```diff
@@ -0,0 +1,105 @@
...
+  private static final int BATCH_SIZE = 1000;
+
+  private String projectId;
+  private String exportedSnapshotDir;
+  private String snapshotName;
+  private String restoreDir;
+
+  public HBaseSnapshotInputConfigBuilder() {}
+
+  public HBaseSnapshotInputConfigBuilder setProjectId(String projectId) {
```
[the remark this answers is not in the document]
lichng: Done

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
lichng: You are right, I think this is messed up in the process of copy-pasting... fixed.

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
lichng: It is not, removed, thanks!

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfigBuilder.java
lichng: Agreed, this was missed in the refactoring for this builder class, thanks for catching it!

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
lichng: This was copied from https://github.com/googleapis/java-bigtable-hbase/blob/486589b9361bf0e62b466b332bbc23964ffe4d0c/bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/sequencefiles/EndToEndIT.java#L198 with some confusing comment so I kept it.

After some testing, it is not needed, will remove this.

bigtable-dataflow-parent/bigtable-beam-import/src/test/generate_test_data.txt
lichng: Adding a README.md

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
lichng: This was the practice I found and followed in https://github.com/googleapis/java-bigtable-hbase/blob/486589b9361bf0e62b466b332bbc23964ffe4d0c/bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/sequencefiles/EndToEndIT.java#L105

If we use a TTL, since this folder could be reused across multiple consecutive test runs, it could introduce flakiness caused by timing, no?

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/Main.java
```diff
@@ -0,0 +1,79 @@
...
+import com.google.cloud.bigtable.beam.sequencefiles.ExportJob;
+import com.google.cloud.bigtable.beam.sequencefiles.ImportJob;
+import java.io.File;
+import java.net.URISyntaxException;
+import java.util.Arrays;
+
+/** Entry point for create-table/import/export job submission. */
+@InternalExtensionOnly
+public class Main {
+  /** For internal use only - public for technical reasons. */
```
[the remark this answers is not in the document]
lichng: This is a direct copy from https://github.com/googleapis/java-bigtable-hbase/blob/c7ae163b167881dacab45d6474f6e636c4f0a478/bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/sequencefiles/Main.java#L29

It was introduced in https://github.com/googleapis/java-bigtable-hbase/commit/c7ae163b167881dacab45d6474f6e636c4f0a478

TBH I am not sure why this is even needed.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/Main.java
```diff
@@ -0,0 +1,79 @@
...
+import com.google.cloud.bigtable.beam.hbasesnapshots.ImportJobFromHbaseSnapshot;
+import com.google.cloud.bigtable.beam.sequencefiles.CreateTableHelper;
+import com.google.cloud.bigtable.beam.sequencefiles.ExportJob;
+import com.google.cloud.bigtable.beam.sequencefiles.ImportJob;
+import java.io.File;
+import java.net.URISyntaxException;
+import java.util.Arrays;
+
+/** Entry point for create-table/import/export job submission. */
+@InternalExtensionOnly
```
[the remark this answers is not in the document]
lichng: Done

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/Main.java
```diff
@@ -0,0 +1,79 @@
...
+        CreateTableHelper.main(subArgs);
+        break;
+      default:
+        usage();
+        System.exit(1);
+    }
+  }
+
+  private static void usage() {
+    String jarName;
```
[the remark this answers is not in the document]
lichng: Since the initialization could through an exception, hence in the code this is actually initialized twice, making it final will not compile, are you suggesting using a temp variable in the try block? IMO that would make the logic more complex.
Did I miss anything?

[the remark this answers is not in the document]
igorbernstein2: I dont quite understand what you mean here. The variable will only be initialized once: either in the try or the catch. Marking it final makes sure that in the future someone doesnt accidentally remove the assignment in the catch

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfigBuilder.java
```diff
@@ -0,0 +1,115 @@
...
+  /*
+   * Set the GCS path where the HBase snapshot data is located
+   */
+  public HBaseSnapshotInputConfigBuilder setExportedSnapshotDir(String exportedSnapshotDir) {
+    this.exportedSnapshotDir = exportedSnapshotDir;
+    return this;
+  }
+
+  /*
+   * Set the name of the snapshot to be imported
```
igorbernstein2, line 70: Please add a bit more doc: how does the name of the snapshot relate to the ExportedSnapshotDir?

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfigBuilder.java
```diff
@@ -0,0 +1,115 @@
...
+   */
+  public HBaseSnapshotInputConfigBuilder setSnapshotName(String snapshotName) {
+    this.snapshotName = snapshotName;
+    return this;
+  }
+
+  /*
+   * Set the temporal GCS path used by TableSnapshotInputFormat while reading the HBase snapshot
```
igorbernstein2, line 78: s/temporal/temporary

why does the user want to set this? can't you just use a subdir in the gcs temp dir?

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
igorbernstein2: the workdir should always be unique due to the uuid. I think the practice in the sequencefile test can be improved and has caused issues in the past. When I wrote it I didnt think to use gcs ttls

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfigBuilder.java
lichng: TableSnapshotInputFormat perform a restore operation when reading from snapshot, the restoreDir is the  targeted restore folder. so the resulting file size could be huge, hence we don't want to do this under the hood.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfigBuilder.java
lichng: Done. Added a sample

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/Main.java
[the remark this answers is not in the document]
lichng: Since we initialize `jarName`  within a try/catch statement, the compiler think this variable could get assigned twice and will throw a `variable jarName might already have been assigned` error.

I did some google search, the best workaround is to use a second variable before assignment or use a function like:

```
    final String jarName = getJarName();

    String getJarName() {
        String temp;
        try {
            temp = ...
        } catch (Exception e) {
            temp = ...
        }
        return temp
    }
```

Which I find hardly any better...

Please advise if I have misunderstood this.

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
lichng: After some investigation, there are one more concern for using TTL:
TTL(lifecycle management for GCS) could only apply to Bucket level, to leverage that, we will need to create a separate bucket just for the restoreDir every time since current bucket also include test input files which shouldn't be deleted.

we can set up the bucket to delete its content after 1 day but the bucket will be there, so overtime the project we use for testing will accumulate a lot of empty buckets.

Current solution also has this issue(generates a lot of empty dirs instead), but this could be solved by adding:

`gcsUtil.remove(Arrays.asList(restoreDir));` after removing all files in the directory.

lichng: Also, per this weird check: https://github.com/apache/hbase/blob/3cc24683001998f995d22591ce8d1c74f67fb9f7/hbase-server/src/main/java/org/apache/hadoop/hbase/snapshot/RestoreSnapshotHelper.java#L837

HBase treats the bucket name as part of file system so we could not possibly use another bucket for the restore...

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfigBuilder.java
lichng: Will throw this and exit the pipeline building

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
lichng: Done

bigtable-dataflow-parent/bigtable-beam-import/src/main/resources/log4j.properties
lichng: It was for debugging, removed.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
lichng: Added more document and example in the class comment

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
lichng: Added more document in the Class document

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
lichng: Done

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
lichng: Noted, Will leave this one for another PR for re-structuring the test data and bucket lifecycle management.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
igorbernstein2: why not use the beam's temp directory? why does a user need to specify a temp dir?

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
```diff
@@ -0,0 +1,139 @@
...
+    void setSnapshotName(String snapshotName);
+  }
+
+  public static void main(String[] args) {
+    PipelineOptionsFactory.register(ImportOptions.class);
+
+    ImportOptions opts =
+        PipelineOptionsFactory.fromArgs(args).withValidation().as(ImportOptions.class);
+
+    LOG.info("DEBUG===> Building Pipeline");
```
igorbernstein2, line 96: `LOG.info("DEBUG`, this looks a bit odd. Why is DEBUG==> there?

igorbernstein2: Same below

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
```diff
@@ -0,0 +1,139 @@
...
+              "Read from HBase Snapshot",
+              HadoopFormatIO.<ImmutableBytesWritable, Result>read()
+                  .withConfiguration(configuration))
+          .apply("Create Mutations", ParDo.of(new HBaseResultToMutationFn()))
+          .apply("Write to Bigtable", createSink(opts));
+
+    } catch (Exception e) {
+      LOG.fatal("Failed to create HBaseConfiguration for HadoopFormatIO");
+      LOG.fatal(e);
+      System.exit(-1);
```
igorbernstein2, line 130: Why not just throw an exception? having a helper function kill the jvm is really unexpected

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+    return pipeline;
+  }
+
+  static PTransform<PCollection<Mutation>, PDone> createSink(ImportOptions opts) {
+    CloudBigtableTableConfiguration config = TemplateUtils.BuildImportConfig(opts);
+    return CloudBigtableIO.writeToTable(config);
+  }
```
igorbernstein2, line 138: I'm not sure if this helper is necessary...why not just inline it?

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
lichng: Removed.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
lichng: Done. 

There was an extra configuration we used to workaround the bulk flush issue, now it is not needed.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
lichng: Done.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
lichng: Technically we could use a default temp dir, however, the restoring process will save some of the user data into the restoreDir, e.g, snapshot metadata, regioninfo, which could be sensitive.
Hence I figured we might want to make this explicit and let the users decide how to clean it up.
Otherwise I am ok to use `opts.getTempLocation()` for that path
What do you think?

igorbernstein2: Take a look at how BigQueryIO handles it and see if its applicable:
https://github.com/xsm110/Apache-Beam/blob/master/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIO.java#L503-L533

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfigBuilder.java
lichng: Done

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfigBuilder.java
lichng: Done

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
lichng: Created a DoFn to cleanup temp folders similar to https://github.com/apache/beam/blob/7c43ab6a8df9b23caa7321fddff9a032a71908f6/sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java#L582 as it is simpler.

Had to use GcsUtil directly instead of beam FileSystem abstraction since the wrapping GcsFileSystem filterout directories intetionally while listing buckets.

PTAL

igorbernstein2 (changes requested):

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
```diff
@@ -0,0 +1,133 @@
...
+  @VisibleForTesting
+  static Pipeline buildPipeline(ImportOptions opts) throws Exception {
+
+    Pipeline pipeline = Pipeline.create(Utils.tweakOptions(opts));
+    Configuration configuration =
+        new HBaseSnapshotInputConfigBuilder()
+            .setProjectId(opts.getProject())
+            .setHbaseSnapshotSourceDir(opts.getHbaseSnapshotSourceDir())
+            .setSnapshotName(opts.getSnapshotName())
+            .setRestoreDirSuffix(opts.getJobName())
```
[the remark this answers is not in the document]
lichng: It should be, from dataflow's document, if you specify --jobName to an existing job, it should considered a replacement job: https://cloud.google.com/dataflow/docs/guides/updating-a-pipeline

Add uuid would need us to passing this new suffix between steps which could introduce some more complexity hence I reused jobName

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
```diff
@@ -0,0 +1,133 @@
...
+            "Write to Bigtable",
+            CloudBigtableIO.writeToTable(TemplateUtils.BuildImportConfig(opts)));
+
+    final List<KV<String, String>> tempFiles =
+        Arrays.asList(
+            KV.of(
+                opts.getHbaseSnapshotSourceDir(),
+                new HBaseSnapshotInputConfigBuilder()
+                    .setRestoreDirSuffix(opts.getJobName())
+                    .getRestoreDir()));
```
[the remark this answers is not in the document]
lichng: Done

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/CleanupHBaseSnapshotRestoreFilesFn.java
```diff
@@ -0,0 +1,85 @@
...
+    List<String> results = new ArrayList<>();
+    do {
+      Objects objects = gcsUtil.listObjects(bucketName, prefix, pageToken);
+      if (objects.getItems() == null) {
+        break;
+      }
+      results.addAll(
+          objects.getItems().stream()
+              .map(storageObject -> GcsPath.fromObject(storageObject).toString())
+              .collect(Collectors.toList()));
```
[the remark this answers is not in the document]
lichng: Done

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/HBaseSnapshotInputConfigBuilder.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  public String getRestoreDir() {
+    return RESTORE_DIR + this.restoreDirSuffix;
+  }
+
+  public Configuration build() throws Exception {
+    Preconditions.checkNotNull(projectId);
+    Preconditions.checkNotNull(hbaseSnapshotSourceDir);
+    Preconditions.checkNotNull(snapshotName);
```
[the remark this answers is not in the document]
lichng: Done

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/hbasesnapshots/ImportJobFromHbaseSnapshot.java
```diff
@@ -0,0 +1,133 @@
...
+    readResult
+        .apply("Create Mutations", ParDo.of(new HBaseResultToMutationFn()))
+        .apply(
+            "Write to Bigtable",
+            CloudBigtableIO.writeToTable(TemplateUtils.BuildImportConfig(opts)));
+
+    final List<KV<String, String>> tempFiles =
+        Arrays.asList(
+            KV.of(
+                opts.getHbaseSnapshotSourceDir(),
```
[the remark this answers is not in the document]
lichng: No, the source file will not be deleted, the hbaseSnapshoSourceDir is used to get the bucket name(which the temp folder reside in). 

Actually the reason I passed `hbaseSnapshotSourceDir` in instead of just `GcsPath.fromUri(hbaseSnapshotDir).getBucket()` is because we need to make sure restoreDir does not include hbaseSnapshotSourceDir and I believe that is some check belongs to the DoFn itself instead of in the buildPipeline() call.

To avoid confustion, I renamed the variable from `tempFiles` to `sourceAndRestoreFolders`, hopefully together with the name `CleanupHBaseSnapshotRestoreFilesFn`, it will cause less confusion.

As for the failure handling, I can see 2 options:
1. ask the user to fix the pipeline and launch a replacement pipeline(so the same restore folder is used and will be cleanup)
2. ask the user to manually delete the restore folder

I am not sure there is a way to guarantee cleanup will be executed if the pipeline fails, since the restore folder will be generated even start actual reading. in my experiment with customized worker images, sometimes the job will be killed before even start reading from snapshots, so in that case there is nothing the pipeline can do to clean it up.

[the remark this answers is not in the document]
igorbernstein2: Ah, sorry, I misread the code I thought you were making a list of that include the source dir. 

I dont think you need to figure out how to cleanup on failure right now. Just add it to the docs and we can figure it out later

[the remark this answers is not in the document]
lichng: Done.

I added a paragraph in the header comments for `ImportJobFromHbaseSnapshot`, eventually when all pieces are together, we would like to also provide a README.md for this job. But I would avoid that right now since the validation part isn't checked in yet.

igorbernstein2 (approved):
lgtm!

PR #2756, opened 2020-12-11
fix: Non-retriable operations should ignore attempt timeout and default to RPC timeout instead

shantstepanian:


igorbernstein2 (approved):
LGTM. But please change the title to `fix:`

PR #2780, opened 2020-12-24
fix: Extract out trace management from ResultScanners

### LINKED ISSUE #2458: Extract out trace management from ResultScanners

dmitry-fa:
Fixes #2458

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
@@ -276,25 +276,22 @@ public abstract class AbstractBigtableTable implements Table {
   @Override
   public ResultScanner getScanner(final Scan scan) throws IOException {
     LOG.trace("getScanner(Scan)");
-    Span span = TRACER.spanBuilder("BigtableTable.scan").startSpan();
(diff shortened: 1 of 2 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
```diff
@@ -276,25 +276,22 @@ public Result get(Get get) throws IOException {
   @Override
   public ResultScanner getScanner(final Scan scan) throws IOException {
     LOG.trace("getScanner(Scan)");
-    Span span = TRACER.spanBuilder("BigtableTable.scan").startSpan();
+    Span span = TRACER.spanBuilder("BigtableTable.readRows").startSpan();
```
kolea2, line 279: why is this changing from `scan` to `readRows`?

dmitry-fa: Originally we traced the scope of  readRows() and iterator over rows, the scope is limited only to readRows() method.

suztomo (dismissed):

PR #2781, opened 2020-12-24
fix: Bigtable emulator rejects fuzzy key filters using integer values over 127

dmitry-fa:
Backport: #2642


dmitry-fa:
@kolea2 backport PR as you asked

suztomo (dismissed):

PR #2783, opened 2020-12-29
fix: add documentation for scan/read timeouts

### LINKED ISSUE #2375: add documentation for scan/read timeouts
These are generated when a row is available, but hasn't been consumed by the reader of a readrows stream. We should document this behavior and point users to the appropriate setting that affects it

dmitry-fa:
Fixes #2375


igorbernstein2:
@shantstepanian

PR #2789, opened 2021-01-05
deps: fix #2788 Using bom files to avoid possible version discrepancy among grpc components

### LINKED ISSUE #2788: Switch to using grpc-bom instead of overriding individual artifact versions
#2775 added some dep overrides for piecemeal grpc components. It would be more consistent to import the grpc bom to ensure that we dont mismatch grpc components.

lichng:


igorbernstein2 (approved):
lgtm

PR #2791, opened 2021-01-05
chore(deps): pin specific deps to be ignored by renovate-bot

kolea2:
To avoid getting PRs like https://github.com/googleapis/java-bigtable-hbase/pull/2790 where we are instead pinning to a specific version.

The 'com.google.cloud.bigtable' is temporary as we currently get renovate updates when we do a release from bigtable-1.x.


igorbernstein2 (approved):
lgtm

PR #2828, opened 2021-02-04
feat: Add a new pipeline to validate data imported from HBase

vermas2012:


bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
```diff
@@ -185,6 +245,81 @@ public void testHBaseSnapshotImport() throws Exception {
...
+    PipelineResult result = SyncTableJob.buildPipeline(syncOpts).run();
+    state = result.waitUntilFinish();
+    Assert.assertEquals(State.DONE, state);
+
+    List<GcsPath> outputs = gcsUtil.expand(GcsPath.fromUri(syncTableOutputDir + "*"));
+
+    System.out.println("OUTPUTS: " + outputs);
+    // FileSink will shard the outputs and will created >1 files.
+    Assert.assertTrue(outputs.size() > 1);
+    // TODO read the files and validate that the ranges are there instead of size check.
```
vermas2012, line 315: Pending output format discussion on this PR.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/TemplateUtils.java
```diff
@@ -72,6 +73,19 @@ public static CloudBigtableTableConfiguration BuildImportConfig(ImportOptions op
     return builder.build();
   }
 
+  /** Builds CloudBigtableTableConfiguration from input runtime parameters for import job. */
+  public static CloudBigtableTableConfiguration BuildSyncTableConfig(SyncTableOptions opts) {
```
aimiebell, line 77: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFn.java
```diff
@@ -0,0 +1,232 @@
...
+
+  private final TableHashWrapperFactory tableHashWrapperFactory;
+
+  // Counter for reporting matching and mismatching ranges. Names are similar to HBase sync-table
+  // job.
+  private final Counter matches = Metrics.counter("cbt-dataflow-validate", "ranges_matched");
+  private final Counter mismatches = Metrics.counter("cbt-dataflow-validate", "ranges_not_matched");
+
+  public ComputeAndValidateHashFromBigtableDoFn(SyncTableOptions options) {
+    super(TemplateUtils.BuildSyncTableConfig(options));
```
aimiebell, line 65: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/SyncTableJob.java
```diff
@@ -0,0 +1,199 @@
...
+  public interface SyncTableOptions extends GcpOptions {
+
+    @Description("This Bigtable App Profile id.")
+    ValueProvider<String> getBigtableAppProfileId();
+
+    @SuppressWarnings("unused")
+    void setBigtableAppProfileId(ValueProvider<String> appProfileId);
+
+    @Description("The project that contains the table to export. Defaults to --project.")
+    @Default.InstanceFactory(Utils.DefaultBigtableProjectFactory.class)
```
aimiebell, line 102: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/BufferedHadoopHashTableSource.java
```diff
@@ -0,0 +1,212 @@
...
+
+    // Reads from hashReader and buffers the RangeHashes.
+    // Returns true if any RangeHashes were read from hashReader.
+    private boolean bufferRangeHashes() throws IOException {
+      boolean readRangeHashes = false;
+      while (buffer.size() < source.maxBufferSize && hashReader.advance()) {
+        readRangeHashes = true;
+        buffer.add(hashReader.getCurrent());
+      }
+      return readRangeHashes;
```
igorbernstein2, line 179: nit. return !buffer.isEmpty()

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/BufferedHadoopHashTableSource.java
```diff
@@ -0,0 +1,212 @@
...
+      Preconditions.checkArgument(!buffer.isEmpty(), "Can not get current on empty buffer.");
+      List<RangeHash> hashes = buffer;
+      // Reset the buffer for next batch.
+      buffer = new ArrayList<>(source.maxBufferSize);
+      // GroupBy key is a string and not ImmutableBytesWritable because the WritableCoder is not
+      // deterministic. The outputted PCollection is grouped by the K and needs a deterministic
+      // coder. Having a String K leads to an unfortunate double encoding, ImmutableBytesWritable->
+      // HEX string -> UTF8 encoded string. The number of batches are significantly smaller than
+      // data fetched from Bigtable and should not have meaningful impact on the job performance.
+      return KV.of(Bytes.toStringBinary(hashes.get(0).startInclusive.copyBytes()), hashes);
```
igorbernstein2, line 199: consider using `hashes.get(0).startInclusive.hashCode()` for the batch?

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFn.java
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  @ProcessElement
+  public void processElement(ProcessContext context) throws Exception {
+    // BufferedHadoopHashTableSource generates only 1 item per groupby key, but iterate just in
+    // case.
+    for (List<RangeHash> rangeHashes : context.element().getValue()) {
+      if (rangeHashes.isEmpty()) {
+        // No rows ranges found, return;
+        return;
```
igorbernstein2, line 94: continue? or better yet, just flatten the value

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFn.java
```diff
@@ -0,0 +1,232 @@
...
+          rangeHashes.get(rangeHashes.size() - 1).stopExclusive;
+
+      BigtableResultHasher resultHasher = new BigtableResultHasher();
+      resultHasher.startBatch(rangeStartInclusive);
+
+      // Since all the row-ranges are sorted in HashTable's data files, 1 big scan can be used
+      // to read all the row ranges. Parallelism is achieved by splitting the HashTable's data
+      // files into smaller bundle of row-ranges in GroupBy.
+      ResultScanner scanner =
+          createBigtableScan(rangeStartInclusive.copyBytes(), rangeEndExclusive.copyBytes());
```
igorbernstein2, line 108: why copyBytes instead of get()?

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFn.java
```diff
@@ -0,0 +1,232 @@
...
+      long numRows = 0;
+
+      RangeHash currentRangeHash = rangeHashIterator.next();
+
+      // Process each row and validate hashes
+      for (Result result : scanner) {
+        numRows++;
+        if (numRows % 10_000 == 0) {
+          // Heartbeat in logs in case a large scan gets hung.
+          DOFN_LOG.debug("Processed " + numRows + " rows ");
```
igorbernstein2, line 120: would a counter be a better choice?

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFn.java
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        ImmutableBytesWritable rowKey = new ImmutableBytesWritable(result.getRow());
+
+        // Check if the rowKey belongs to current range, if not keep iterating through the
+        // rangeHashes until rowKey's range is found.
+        while (!isWithinUpperBound(currentRangeHash.stopExclusive, rowKey)) {
```
igorbernstein2, line 127: consider using `RowEndKeyComparator`

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFn.java
```diff
@@ -0,0 +1,232 @@
...
+
+      // Process each row and validate hashes
+      for (Result result : scanner) {
+        numRows++;
+        if (numRows % 10_000 == 0) {
+          // Heartbeat in logs in case a large scan gets hung.
+          DOFN_LOG.debug("Processed " + numRows + " rows ");
+        }
+
+        ImmutableBytesWritable rowKey = new ImmutableBytesWritable(result.getRow());
```
igorbernstein2, line 123: you can avoid a copy here by comparing result.getRow() to stopExclusive.get()

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFn.java
```diff
@@ -0,0 +1,232 @@
...
+
+        // Check if the rowKey belongs to current range, if not keep iterating through the
+        // rangeHashes until rowKey's range is found.
+        while (!isWithinUpperBound(currentRangeHash.stopExclusive, rowKey)) {
+          validateBatchHash(context, resultHasher, currentRangeHash);
+          if (!rangeHashIterator.hasNext()) {
```
igorbernstein2, line 129: nit. replace with Preconditions.checkState

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/SyncTableJob.java
```diff
@@ -0,0 +1,199 @@
...
+    return pipeline;
+  }
+
+  static class RangeHashToString extends SimpleFunction<RangeHash, String> {
+    // TODO maybe explore a sequenceFile sink for RangeHash. Hadoop jobs using this output may be
+    // easier to write for sequence file.
+
+    // GSON is not serializable, keep it transient. Member variable to avoid creating a Gson object
+    // per apply call.
+    private transient Gson gson = null;
```
igorbernstein2, line 189: why not make it static?

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/SyncTableUtils.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+package com.google.cloud.bigtable.beam.validation;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.HBaseConfiguration;
+import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
+import org.apache.hadoop.hbase.util.Bytes;
+
+/** Utility class for SyncTable job. */
+public class SyncTableUtils {
```
igorbernstein2, line 24: InternalApi?

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/TableHashWrapperImpl.java
```diff
@@ -0,0 +1,119 @@
...
+
+class TableHashWrapperImpl implements TableHashWrapper {
+
+  static TableHashWrapper create(Configuration conf, String hashTableOutputDir) throws IOException {
+    TableHash tableHash = TableHash.read(conf, new Path(hashTableOutputDir));
+
+    TableHashWrapper tableHashWrapper = new TableHashWrapperImpl(tableHash);
+    Preconditions.checkArgument(
+        tableHashWrapper.getNumHashFiles() == (tableHashWrapper.getPartitions().size() + 1),
+        String.format(
```
igorbernstein2, line 39: checkArgument accepts format strings already, you dont need String.format

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/org/apache/hadoop/hbase/mapreduce/BigtableTableHashAccessor.java
```diff
@@ -0,0 +1,77 @@
...
+import com.google.common.collect.ImmutableList;
+import java.io.IOException;
+import org.apache.hadoop.hbase.client.Result;
+import org.apache.hadoop.hbase.client.Scan;
+import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
+import org.apache.hadoop.hbase.mapreduce.HashTable.ResultHasher;
+import org.apache.hadoop.hbase.mapreduce.HashTable.TableHash;
+
+/** A helper class to access package private fields of HashTable.TableHash. */
+public class BigtableTableHashAccessor {
```
igorbernstein2, line 27: package private or InternalApi

Also, this could really use a unit test to alert us when the internal impl details change in hbase

igorbernstein2 (changes requested):

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/BufferedHadoopHashTableSource.java
```diff
@@ -0,0 +1,212 @@
...
+ * Buffers the RangeHashes generated by {@link HadoopHashTableSource}. This is an optimization that
+ * allows {@link ComputeAndValidateHashFromBigtableDoFn} to issue fewer ReadRow APIs with larger row
+ * ranges.
+ *
+ * <p>Hadoop HashTable output is sorted by row-key and contains a row-range and hash. Beam
+ * Pcollection do not guarantee any ordering. To fetch a batch of ranges in 1 ReadRows operation,
+ * this source buffers then and outputs a List<RangeHash> guaranteeing the sorted order of ranges.
+ */
+@InternalApi
+class BufferedHadoopHashTableSource extends BoundedSource<KV<String, List<RangeHash>>> {
```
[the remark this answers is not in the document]
vermas2012: done.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/HadoopHashTableSource.java
```diff
@@ -0,0 +1,464 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.beam.validation;
+
+import static com.google.cloud.bigtable.beam.validation.SyncTableUtils.createConfiguration;
+import static com.google.cloud.bigtable.beam.validation.SyncTableUtils.immutableBytesToString;
+
+import autovalue.shaded.com.google$.common.annotations.$VisibleForTesting;
```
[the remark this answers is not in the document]
vermas2012: fixed

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
```diff
@@ -26,6 +26,7 @@ limitations under the License.
 
   <properties>
     <mainClass>com.google.cloud.bigtable.beam.Main</mainClass>
+    <skipITs>false</skipITs>
```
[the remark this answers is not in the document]
vermas2012: Just before merging. I need to run the integ test as part of the changes.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/HadoopHashTableSource.java
```diff
@@ -0,0 +1,464 @@
...
+@InternalApi
+class HadoopHashTableSource extends BoundedSource<RangeHash> implements Serializable {
+
+  private static final long serialVersionUID = 2383724L;
+
+  /**
+   * A simple POJO encapsulating a row range and the corresponding hash generated by HashTable job.
+   */
+  @DefaultCoder(RangeHashCoder.class)
+  public static class RangeHash {
```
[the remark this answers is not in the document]
vermas2012: What are pros and cons of custom coder vs Serializable coder? We agreed on a custom coder in our earlier conversations.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/HadoopHashTableSource.java
```diff
@@ -0,0 +1,464 @@
...
+
+    @Override
+    public int hashCode() {
+      return Objects.hashCode(startInclusive, stopExclusive, hash);
+    }
+  }
+
+  public static final Log LOG = LogFactory.getLog(HadoopHashTableSource.class);
+
+  private ValueProvider<String> projectId;
```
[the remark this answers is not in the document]
vermas2012: WriteObject needs to set them and hence, they can't be final.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/HadoopHashTableSource.java
```diff
@@ -0,0 +1,464 @@
...
+    int numPartitions = partitions.size();
+
+    List<HadoopHashTableSource> splitSources = new ArrayList<>(numPartitions + 1);
+    if (numPartitions == 0) {
+      // There are 0 partitions and 1 hashfile, return single source with full key range.
+      splitSources.add(
+          new HadoopHashTableSource(
+              projectId,
+              sourceHashDir,
+              new ImmutableBytesWritable(hash.getStartRow()),
```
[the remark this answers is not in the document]
vermas2012: done

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/HadoopHashTableSource.java
```diff
@@ -0,0 +1,464 @@
...
+            + immutableBytesToString(startRow.get())
+            + ", "
+            + immutableBytesToString(stopRow.get())
+            + "]");
+    // Add the last range for [lastPartition, stopRow).
+    splitSources.add(
+        new HadoopHashTableSource(
+            projectId,
+            sourceHashDir,
+            partitions.get(numPartitions - 1),
```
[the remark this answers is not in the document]
vermas2012: this is the last range, so we need to bound it on stopRow.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/HadoopHashTableSource.java
```diff
@@ -0,0 +1,464 @@
...
+
+    public HashBasedReader(
+        HadoopHashTableSource source,
+        ImmutableBytesWritable startRowInclusive,
+        ImmutableBytesWritable stopRowExclusive,
+        TableHashReader reader) {
+      this.reader = reader;
+      this.source = source;
+      this.startRowInclusive = startRowInclusive;
+      this.stopRowExclusive = stopRowExclusive;
```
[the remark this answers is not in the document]
vermas2012: done

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/HadoopHashTableSource.java
```diff
@@ -0,0 +1,464 @@
...
+    }
+
+    @Override
+    public boolean start() throws IOException {
+      // NO CHECKED EXCEPTIONS HERE.
```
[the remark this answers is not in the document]
vermas2012: not applicable now. Deleted.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/HadoopHashTableSource.java
```diff
@@ -0,0 +1,464 @@
...
+      ImmutableBytesWritable startKey = this.currentRangeStartKey;
+      ImmutableBytesWritable hash = this.currentHash;
+
+      if (!readNextKey()) {
+        this.currentRangeHash = RangeHash.of(startKey, stopRowExclusive, hash);
+        // return true since we have lastBatchStartKey to emit. Set isDone=true to prevent reading
+        // from a potentially exhausted reader.
+        isDone = true;
+      } else {
+        this.currentRangeHash = RangeHash.of(startKey, reader.getCurrentKey(), hash);
```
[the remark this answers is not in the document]
vermas2012: currentRangeStartKey represents the current range's start key and we are using it as the start key. The end of the range is the reader.currentKey().

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/HadoopHashTableSource.java
```diff
@@ -0,0 +1,464 @@
...
+        this.currentRangeStartKey = reader.getCurrentKey();
+        if ( // StopRow is not set, everything is in bounds.
+        (stopRowExclusive.equals(HConstants.EMPTY_END_ROW)
+            || currentRangeStartKey.compareTo(stopRowExclusive) < 0)) { // currentKey < stopKey
+          // There is a key to read and the key is within the bounds of this workitem. Return true.
+          this.currentHash = reader.getCurrentHash();
+          return true;
+        } else {
+          // There is a key to read but its outside of the bounds of this workitem.
+          this.currentHash = null;
```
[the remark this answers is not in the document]
vermas2012: I didn't get the subrange part? We need to interpret the ranges.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/org/apache/hadoop/hbase/mapreduce/BigtableTableHashAccessor.java
vermas2012: This could not be package private as its whole purpose it to expose the package private methods/fields. I added internal API annotation.

vermas2012: I am not sure we can create a unit test for this as we will need HBase with files in HDFS. Maybe integ test can cover this, but there is no logic here, if any field changes, it will likely lead to compiler issues.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/TableHashWrapperImpl.java
vermas2012: done.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/SyncTableUtils.java
vermas2012: Done.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/SyncTableJob.java
vermas2012: Actually, we can do that. Fixed.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFn.java
vermas2012: done.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/HadoopHashTableSource.java
```diff
@@ -0,0 +1,464 @@
...
+  public String toString() {
+    return "HadoopHashTableSource ["
+        + immutableBytesToString(startRowInclusive)
+        + ", "
+        + immutableBytesToString(stopRowExclusive)
+        + ')';
+  }
+
+  private void writeObject(ObjectOutputStream s) throws IOException {
+    // s.defaultWriteObject();
```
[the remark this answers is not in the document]
vermas2012: If we mark them transient, we convey to the reader that they are not going to be serialized. I think it gives wrong message. 

I will delete the commented reference to default(read|write)Object, as it leads to confusion.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/BufferedHadoopHashTableSource.java
```diff
@@ -0,0 +1,212 @@
...
+
+  private static final long serialVersionUID = 39842743L;
+
+  public static final Log LOG = LogFactory.getLog(BufferedHadoopHashTableSource.class);
+  private static final int DEFAULT_BATCH_SIZE = 50;
+
+  // Max number of RangeHashes to buffer.
+  private int maxBufferSize;
```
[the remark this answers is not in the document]
vermas2012: Serialization methods needed it to be not final. I have removed the custom serialization and marked them final.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/BufferedHadoopHashTableSource.java
```diff
@@ -0,0 +1,212 @@
...
+
+  @Override
+  public Coder<KV<String, List<RangeHash>>> getOutputCoder() {
+    return coder;
+  }
+
+  @Override
+  public long getEstimatedSizeBytes(PipelineOptions options) throws Exception {
+    // HashTable data files don't expose a method to estimate size or lineCount.
+    return 0;
```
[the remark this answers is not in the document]
vermas2012: done

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/BufferedHadoopHashTableSource.java
```diff
@@ -0,0 +1,212 @@
...
+  public String toString() {
+    return "BufferedHadoopHashTableSource ["
+        + immutableBytesToString(hashTableSource.startRowInclusive)
+        + ", "
+        + immutableBytesToString(hashTableSource.stopRowExclusive)
+        + "), maxBufferSize="
+        + maxBufferSize;
+  }
+
+  private void writeObject(ObjectOutputStream s) throws IOException {
```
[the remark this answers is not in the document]
vermas2012: Makes sense. done.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/BufferedHadoopHashTableSource.java
```diff
@@ -0,0 +1,212 @@
...
+
+  private void readObject(ObjectInputStream s) throws IOException, ClassNotFoundException {
+    this.hashTableSource = (HadoopHashTableSource) s.readObject();
+    this.coder = KvCoder.of(StringUtf8Coder.of(), ListCoder.of(RangeHashCoder.of()));
+    this.maxBufferSize = s.readInt();
+  }
+
+  private static class BufferedHashBasedReader extends BoundedReader<KV<String, List<RangeHash>>> {
+
+    private BoundedReader<RangeHash> hashReader;
```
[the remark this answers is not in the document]
vermas2012: done

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/BufferedHadoopHashTableSource.java
vermas2012: The current return value indicates if anything new was buffered, there is a case for start when the buffer will be non-empty to start with. 

We can change it to return if the buffer is not empty, but that is not what it is today.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFn.java
vermas2012: get() returns the full byte[] and the caller is responsible to prune it to the bounds covered by this object. 

```
 public byte[] get() {
    if (this.bytes == null) {
      throw new IllegalStateException("Uninitialiized. Null constructor called w/o accompaying readFields invocation");
    } else {
      return this.bytes;
    }
  }


  public byte[] copyBytes() {
    return Arrays.copyOfRange(this.bytes, this.offset, this.offset + this.length);
  }
```

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFn.java
vermas2012: I am not sure if counter will be better. We already have a Success/fail counter that can be monitored. This is for debugging if one worker had an issue, logs were filtered on a worker and you see these logs.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFn.java
vermas2012: I am not sure i get it? It will be true only once per batch, and will make code little complex with an extra branch?

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
```diff
@@ -126,6 +158,17 @@ private static String getTestProperty(String name) {
...
+
+    if (!paths.isEmpty()) {
+      final List<String> pathStrs = new ArrayList<>();
+
+      for (GcsPath path : paths) {
+        pathStrs.add(path.toString());
+      }
+      this.gcsUtil.remove(pathStrs);
+    }
+
```
igorbernstein2, line 171: This doesnt have to happen in this PR, so maybe just add a TODO. But we've had issues in the past of failing to cleanup resources during teardown. For example the test timing out and teardown never running. A effective way to avoid such issues is to embed a timestamp in your resource names and have setup check for orphans from previous runs and clean them up:
https://github.com/googleapis/java-bigtable/blob/35588d89b9b243eb691a29d3aff16b9f5a08fbb8/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/AbstractTestEnv.java#L108-L119

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFn.java
vermas2012: RowKeyComparator will require us to copy the byte[] again. Not sure if its worth the extra object and GC/memory overheads.

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
```diff
@@ -185,6 +245,81 @@ public void testHBaseSnapshotImport() throws Exception {
...
+    // Run SyncTable job and expect 4 mismatches.
+    SyncTableOptions syncOpts = createSyncTableOptions();
+    PipelineResult result = SyncTableJob.buildPipeline(syncOpts).run();
+    state = result.waitUntilFinish();
+    Assert.assertEquals(State.DONE, state);
+
+    List<GcsPath> outputs = gcsUtil.expand(GcsPath.fromUri(syncTableOutputDir + "*"));
+
+    System.out.println("OUTPUTS: " + outputs);
```
igorbernstein2, line 312: Please avoid using sys.out. Please use logging instead or better yet assert that the outputs match expected output

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFnTest.java
```diff
@@ -0,0 +1,441 @@
...
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@RunWith(JUnit4.class)
+public class ComputeAndValidateHashFromBigtableDoFnTest {
```
igorbernstein2, line 59: I dont see tests for the counters/metrics. Would be to good to verify them in each scenario

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/HadoopHashTableSource.java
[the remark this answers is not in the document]
igorbernstein2: I think the benefit here is that its a lot less boilerplate. But I dont feel strongly about this, so its up to you

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/HadoopHashTableSource.java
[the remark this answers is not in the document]
igorbernstein2: ValueProvider is serializable, readObject doesnt need to touch it

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/HadoopHashTableSource.java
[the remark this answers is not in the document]
igorbernstein2: after the last iteration of the for loop, startRow == partitions.get(numPartitions - 1)

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/HadoopHashTableSource.java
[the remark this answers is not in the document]
igorbernstein2: I disagree: marking things as transient means that they are hydrated in a different way. I would prefer to mark things as transient and leaving a comment that they use custom serialization

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFnTest.java
vermas2012: done.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/HadoopHashTableSource.java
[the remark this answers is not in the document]
igorbernstein2: stopRowExclusive should never land in the middle of chunk, it should always equal the start of the next chunk (unless its the very last chunk). If doesnt, then that means that there is a bug somewhere else which will lead to under reading data in bigtable. I'm proposing to to proactively sanity check this invariant

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
vermas2012: Yes, we will assert them. Need to have a conversation with you on the output format.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/HadoopHashTableSource.java
```diff
@@ -0,0 +1,464 @@
...
+          this.currentHash = reader.getCurrentHash();
+          return true;
+        } else {
+          // There is a key to read but its outside of the bounds of this workitem.
+          this.currentHash = null;
+          return false;
+        }
+      }
+
+      // Nothing left to read for this workitem.
```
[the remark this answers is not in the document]
igorbernstein2: you have 2 end states: 1. you find the start of the next chunk. 2. you hit eof 

In case#1:
- reader.next() returns true 
- `numKeys++;` regardless if you hit the terminating row key.
- advance() makes a synthetic tail range

In case#2
- reader.next() returns false
- numKeys is unchanged
- advance() makes a synthetic tail range

This is inconsistent: you need to decide if you are counting the synthetic tail range or not

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/BufferedHadoopHashTableSource.java
igorbernstein2: practically I dont think it makes a difference, just saves you a couple of lines of code

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/BufferedHadoopHashTableSource.java
igorbernstein2: Yes, the K of KV.

I thought that the key is just used to split the bundles, where does it appear in error messages?
Also, I dont think you really care if there is hash collision, because you are pre-grouping

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFn.java
igorbernstein2: Thats a bit annoying. How would you feel about adding a helper to avoid copies in most cases:

```java
byte[] unwrapBytes(ImmutableBytesWriteable w) {
  byte[] innerBytes = w.getBytes();
  if (w.getOffset() == 0 && w.getLength() == innerBytes.length) {
    return innerBytes;
  } else {
     return w.copyBytes();
  }
}
```

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFn.java
igorbernstein2: fair enough

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFn.java
igorbernstein2: Sorry that was a brain fart, ImmutableBytesWritable(byte[]) doesnt create copies

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/org/apache/hadoop/hbase/mapreduce/BigtableTableHashAccessor.java
igorbernstein2: ok

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
vermas2012: done.

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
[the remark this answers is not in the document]
vermas2012: done.

bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/hbasesnapshots/EndToEndIT.java
vermas2012: done

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/HadoopHashTableSource.java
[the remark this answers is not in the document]
vermas2012: Ack! I am going to leave it as is. I have had hard time debugging with auto-gen code.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/HadoopHashTableSource.java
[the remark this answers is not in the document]
vermas2012: done.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/HadoopHashTableSource.java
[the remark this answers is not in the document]
vermas2012: But we are always reading just one line. So are we checking for a bug in our own code? How does unit tests not do it? If we were reading a <start,end> tuple from HBase then it would have made sense that we are validating the end[i] == start[i+1] but we are just reading starts and creating the [start,end] ourselves, which is tested via unit/integ tests.

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFn.java
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    return table.getScanner(scan);
+  }
+
+  /**
+   * Determines if row >= stopExclusive for a row range (start, stopExclusive). Empty stopExclusive
+   * represents a range with no upper bound.
+   */
+  private boolean isWithinUpperBound(
```
igorbernstein2, line 190: static

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/TableHashWrapperFactory.java
```diff
@@ -0,0 +1,33 @@
...
+
+package com.google.cloud.bigtable.beam.validation;
+
+import static com.google.cloud.bigtable.beam.validation.SyncTableUtils.createConfiguration;
+
+import java.io.IOException;
+import java.io.Serializable;
+
+/** Factory to create a TableHashWrapper. */
+public class TableHashWrapperFactory implements Serializable {
```
igorbernstein2, line 25: InternalApi or package private

igorbernstein2 (approved):
lgtm after 2 nits

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFn.java
vermas2012: done

PR #2837, opened 2021-02-18
chore(deps): remove dependency on google-http-client

igorbernstein2:
We unintentionally started depending on google-http-client for for minimal functionality. This will either inline or pull the functionality from elsewhere and drop the dep

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
@@ -56,7 +56,6 @@ import static org.threeten.bp.Duration.ofMillis;
 import static org.threeten.bp.Duration.ofMinutes;
 import static org.threeten.bp.Duration.ofSeconds;
 
-import com.google.api.client.util.SecurityUtils;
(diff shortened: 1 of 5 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -399,14 +399,15 @@ private void configureCredentialProvider(StubSettings.Builder stubSettings) thro
...
       String serviceAccountEmail, String privateKeyFile) throws IOException {
+
     try {
+      KeyStore keyStore = KeyStore.getInstance("PKCS12");
+
+      try (FileInputStream fin = new FileInputStream(privateKeyFile)) {
+        keyStore.load(fin, "notasecret".toCharArray());
+      }
       PrivateKey privateKey =
-          SecurityUtils.loadPrivateKeyFromKeyStore(
```
kolea2, line 404: I think we should continue to use this dependency here. I don't want to maintain this code in two places.

igorbernstein2: As discussed offline, I think taking on an external dependency for 3 lines of code is too high of a cost to maintain.

kolea2 (approved):

PR #2851, opened 2021-02-26
feat: mapreduce snapshot import to cloud bigtable

### LINKED ISSUE #2850: MapReduce snapshot import to Bigtable
Tool for importing an HBase snapshot into Bigtable. An initial version supports:

* simple tool for reading an HBase snapshot and loading the data to a new/existing table in cbt
* portability - job can be run on Hadoop clusters (tested on Hadoop 2.10.1 Dataproc 1.5.x)
* the table being imported to does not require the same name, but requires the same column families as the source snapshot
* initial version does not provide source filters
* ignores deletes as the snapshot inputformat will not return tombstones or deleted cells
* utilizes HBase libraries for import framework and relies on underlying java-bigtable-hbase libraries to write to Bigtable

jhambleton:
Tool for importing an HBase snapshot into Bigtable. Testing includes mr-unit for asserting mapper output and relies on java-bigtable-hbase library through TableOutputFormat Hadoop class for writing to Bigtable. 

ie. snip from a single task log
```
...
2021-02-25 23:14:33,751 INFO [main] org.apache.hadoop.mapred.MapTask: Processing split: org.apache.hadoop.hbase.mapreduce.TableSnapshotInputFormat$TableSnapshotRegionSplit@73ba6fe6
2021-02-25 23:14:34,209 INFO [main] com.google.bigtable.repackaged.com.google.cloud.bigtable.grpc.BigtableSession: Opening session for projectId jh-data-sandbox, instanceId abc12345, on data host bigtable.googleapis.com, admin host bigtableadmin.googleapis.com.
2021-02-25 23:14:34,220 INFO [main] com.google.bigtable.repackaged.com.google.cloud.bigtable.grpc.BigtableSession: Bigtable options: BigtableOptions{dataHost=bigtable.googleapis.com, adminHost=bigtableadmin.googleapis.com, port=443, projectId=jh-data-sandbox, instanceId=abc12345, appProfileId=, userAgent=hbase-1.4.12, credentialType=DefaultCredentials, dataChannelCount=16, retryOptions=RetryOptions{retriesEnabled=true, allowRetriesWithoutTimestamp=false, statusToRetryOn=[UNAUTHENTICATED, DEADLINE_EXCEEDED, UNAVAILABLE, ABORTED], initialBackoffMillis=10, maxElapsedBackoffMillis=600000, backoffMultiplier=2.0, streamingBufferSize=60, readPartialRowTimeoutMillis=300000, maxScanTimeoutRetries=10}, bulkOptions=BulkOptions{asyncMutatorCount=2, useBulkApi=true, bulkMaxKeyCount=100, bulkMaxRequestSize=20971520, autoflushMs=1000, maxInflightRpcs=160, maxMemory=104857600, enableBulkMutationThrottling=false, bulkMutationRpcTargetMs=100}, callOptionsConfig=CallOptionsConfig{useTimeout=true, shortRpcTimeoutMs=20000, longRpcTimeoutMs=600000, mutateRpcTimeoutMs=600000, readStreamRpcTimeoutMs=43200000}, usePlaintextNegotiation=false, useCachedDataPool=false, useBatch=false, useGCJClient=false}.
2021-02-25 23:14:34,450 INFO [main] com.google.bigtable.repackaged.com.google.cloud.bigtable.grpc.BigtableSession: Creating new channel for bigtable.googleapis.com
2021-02-25 23:14:38,042 INFO [main] com.google.bigtable.repackaged.com.google.cloud.bigtable.grpc.BigtableSession: Creating new channel for bigtable.googleapis.com
...
2021-02-25 23:14:49,296 INFO [main] org.apache.hadoop.hbase.regionserver.HRegion: Closed validation_test,e6666661,1594940279992.23e8284620a8e0db6bb65248f2d36737.
2021-02-25 23:14:49,355 INFO [main] org.apache.hadoop.mapred.Task: Task:attempt_1614033965977_0035_m_000002_0 is done. And is in the process of committing
2021-02-25 23:14:49,523 INFO [main] org.apache.hadoop.mapred.Task: Task 'attempt_1614033965977_0035_m_000002_0' done.
2021-02-25 23:14:49,538 INFO [main] org.apache.hadoop.mapred.Task: Final Counters for attempt_1614033965977_0035_m_000002_0: Counters: 43
	File System Counters
		FILE: Number of bytes read=0
		FILE: Number of bytes written=249505
		FILE: Number of read operations=0
		FILE: Number of large read operations=0
		FILE: Number of write operations=0
		GS: Number of bytes read=9464204
		GS: Number of bytes written=0
		GS: Number of read operations=170
		GS: Number of large read operations=0
		GS: Number of write operations=0
		HDFS: Number of bytes read=0
		HDFS: Number of bytes written=0
		HDFS: Number of read operations=0
		HDFS: Number of large read operations=0
		HDFS: Number of write operations=0
	Map-Reduce Framework
		Map input records=10106
		Map output records=10106
		Input split bytes=771
		Spilled Records=0
		Failed Shuffles=0
		Merged Map outputs=0
		GC time elapsed (ms)=721
		CPU time spent (ms)=18750
		Physical memory (bytes) snapshot=638541824
		Virtual memory (bytes) snapshot=4532940800
		Total committed heap usage (bytes)=478674944
	HBase Counters
		BYTES_IN_REMOTE_RESULTS=0
		BYTES_IN_RESULTS=9699718
		MILLIS_BETWEEN_NEXTS=0
		NOT_SERVING_REGION_EXCEPTION=0
		NUM_SCANNER_RESTARTS=0
		NUM_SCAN_RESULTS_STALE=0
		REGIONS_SCANNED=0
		REMOTE_RPC_CALLS=0
		REMOTE_RPC_RETRIES=0
		ROWS_FILTERED=0
		ROWS_SCANNED=10106
		RPC_CALLS=0
		RPC_RETRIES=0
	com.google.cloud.bigtable.mapreduce.hbasesnapshots.ImportHBaseSnapshotJob$ScanCounter
		NUM_CELLS=101144
		NUM_ROWS=10106
	File Input Format Counters 
		Bytes Read=0
	File Output Format Counters 
		Bytes Written=0
```

Fixes #2850 ☕️


igorbernstein2:
This looks like a great feature, thanks for submitting this. I took a really quick glance and was wondering if this can be merged with existing mapreduce based sequencefile importer?
https://github.com/googleapis/java-bigtable-hbase/tree/master/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce

I think it would be nice to follow the same pattern as the beam importer: having a single driver with multiple subjobs. What do you think?

Also it would be good to try to make the invocation for the jobs similar: standardizing on sys props vs env vars vs positional args etc

jhambleton:
@igorbernstein2 - thanks for the suggestions. updated per changes below.

> This looks like a great feature, thanks for submitting this. I took a really quick glance and was wondering if this can be merged with existing mapreduce based sequencefile importer?
> https://github.com/googleapis/java-bigtable-hbase/tree/master/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce

Agreed. I integrated the code with the existing mapreduce libs in bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce.
> 
> I think it would be nice to follow the same pattern as the beam importer: having a single driver with multiple subjobs. What do you think?

I agree. The existing MapReduce tools had a [Driver](https://github.com/googleapis/java-bigtable-hbase/blob/master/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/Driver.java) for the existing Import/Export tools that I added this tool to. 
> 
> Also it would be good to try to make the invocation for the jobs similar: standardizing on sys props vs env vars vs positional args etc

The existing Import/Export MR jobs use system properties for Hadoop/Bigtable configurations (as this is how they would be loaded from a config file such as /etc/hbase/conf/hbase-site.xml through the Hadoop framework) and then use positional arguments for tool configurations. Instead of refactoring the existing tools so they're all standardized I've done the following:

* followed the same conventions for system properties and positional args as the existing Import/Export 
* added a readme with details on these conventions

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
```diff
@@ -0,0 +1,286 @@
...
+        new Scan().setMaxVersions(),
+        SnapshotMapper.class,
+        ImmutableBytesWritable.class,
+        Writable.class,
+        job,
+        true,
+        new Path(conf.get(SNAPSHOT_RESTOREDIR_KEY)));
+
+    job.setNumReduceTasks(0);
+    job.setOutputFormatClass(TableOutputFormat.class);
```
igorbernstein2, line 206: Something needs to randomize the splits otherwise you'll sequentially hotspot each tablet

igorbernstein2 (commented):
This is looking good. I added a couple nits.

@vermas2012, wanna take a look as well?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
```diff
@@ -0,0 +1,286 @@
...
+            + conf.get(TABLENAME_KEY));
+
+    // implicit bigtable configs
+    conf.setIfUnset(
+        ClusterConnection.HBASE_CLIENT_CONNECTION_IMPL,
+        BigtableConfiguration.getConnectionClass().getName());
+    conf.setBooleanIfUnset(BigtableOptionsFactory.BIGTABLE_USE_TIMEOUTS_KEY, true);
+    conf.setIfUnset(
+        BigtableOptionsFactory.BIGTABLE_MUTATE_RPC_TIMEOUT_MS_KEY,
+        String.valueOf(BIGTABLE_MUTATE_RPC_TIMEOUT_MS_DEFAULT));
```
vermas2012, line 174: Is there a way to initialize Bigtable configuration that creates all the default values? If we were to add another config later, how can we ensure that its initialized?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
```diff
@@ -0,0 +1,286 @@
...
+
+    // implicit table outputformat configs
+    conf.set(TableOutputFormat.OUTPUT_TABLE, conf.get(TABLENAME_KEY));
```
vermas2012, line 177: Please add a little more descriptive comment. I am not sure what OUTPUT_FORMAT is, the TABLE_INPUT_FORMAT needs to be a snapshot, what does output format mean here?

vermas2012 (commented):
Overall looks good.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/README.md
```diff
@@ -0,0 +1,192 @@
...
+export DATAPROC_CLUSTER=<DATAPROC_CLUSTER_NAME>
+
+# bigtable table properties
+export CBT_INSTANCE=<BIGTABLE_INSTANCE>
+export CBT_CLUSTER=<BIGTABLE_CLUSTER>
+export CBT_TABLENAME=<TABLENAME>
+export CBT_COLUMN_FAMILY=<CF1[,CF]>
+
+# import job configurations
+export DATAPROC_JOB_JAR=bigtable-mapreduce-import-1.14.1-SNAPSHOT-shaded.jar
```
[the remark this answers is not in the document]
jhambleton: was added with rebase

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/pom.xml
```diff
@@ -52,6 +52,14 @@ limitations under the License.
...
           <artifactId>bigtable-hbase-1.x-shaded</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>org.apache.hadoop</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.apache.hbase</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
```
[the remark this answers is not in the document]
jhambleton: removed in #2895

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/pom.xml
```diff
@@ -127,6 +148,8 @@ limitations under the License.
                     <exclude>commons-codec:commons-codec</exclude>
                     <exclude>org.apache.zookeeper:zookeeper</exclude>
                     <exclude>org.apache.hadoop:hadoop-yarn-api</exclude>
+                    <exclude>org.objenesis:objenesis</exclude>
+                    <excude>com.google.cloud.bigdataoss:gcs-connector</excude>
```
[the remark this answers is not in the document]
jhambleton: removed with update from #2895

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/Driver.java
```diff
@@ -34,7 +35,11 @@ public static void main(String[] args) {
       programDriver.addClass(
           "export-table", Export.class, "A map/reduce program that exports a table to a file.");
       programDriver.addClass(
-          "import-table", Import.class, "A map/reduce program that imports a table to a file.");
+          "import-table", Import.class, "A map/reduce program that imports a file to a table.");
```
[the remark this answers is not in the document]
jhambleton: updated as suggested

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/README.md
```diff
@@ -0,0 +1,192 @@
...
+export DATAPROC_JOB_IMPORT_CLASS=com.google.cloud.bigtable.mapreduce.Driver
+
+# import job configurations
+export IMPORT_PROPS="-Dgoogle.bigtable.project.id=${PROJECT_ID} \
+ -Dgoogle.bigtable.instance.id=${CBT_INSTANCE}"
+
+export ARG_SNAPSHOT_NAME=<SNAPSHOT_NAME>
+export ARG_SNAPSHOT_ROOT_DIR=<SNAPSHOT_ROOT_DIR>
+export ARG_RESTORE_BASE_DIR=<SNAPSHOT_RESTORE_BASE_DIR>
```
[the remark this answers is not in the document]
jhambleton: updated restore dir to temp dir as this is scratch space when running the job that is cleaned up after.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
```diff
@@ -0,0 +1,286 @@
...
+    conf.setIfUnset(
+        IMPORT_SNAPSHOT_JOBNAME_KEY,
+        ImportHBaseSnapshotJob.class.getName()
+            + " - from snapshot "
+            + conf.get(SNAPSHOTNAME_KEY)
+            + " to "
+            + conf.get(TABLENAME_KEY));
+
+    // implicit bigtable configs
+    conf.setIfUnset(
```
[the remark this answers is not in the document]
jhambleton: updated as suggested

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
jhambleton: hotspotting is targeted for a subsequent release due to complexity. however, in this release, we could add `BIGTABLE_BUFFERED_MUTATOR_ENABLE_THROTTLING` with a default low target that has helped reduce effects of hotspotting in our production use. thoughts in adding it as a default here with a 100ms target?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/README.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Main Driver
+
+The main driver used to launch MapReduce tools.  
+
+```bash
+Usage: hadoop jar <JAR> com.google.cloud.bigtable.mapreduce.Driver <PROGRAM> <TOOL_ARGS>
+
+Valid program names are:
+  export-table: A map/reduce program that exports a table to a file.
```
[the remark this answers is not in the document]
jhambleton: updated as suggested

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/README.md
```diff
@@ -0,0 +1,192 @@
...
+  import-snapshot: A map/reduce program that imports a table snapshot.
+  import-table: A map/reduce program that imports a file to a table.
```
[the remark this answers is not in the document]
jhambleton: updated as suggested

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/README.md
```diff
@@ -0,0 +1,192 @@
...
+  -Dgoogle.bigtable.project.id=<bigtable-project-id>
+  -Dgoogle.bigtable.instance.id=<bigtable-instance-id>
+
+Required arguments:
+  snapshot-name     - name of hbase snapshot to read
+  tablename         - bigtable table to import into
+  snapshot-root-dir - directory of snapshot,
+         e.g., gs://hbase-migration-table1-bucket/export/table1-snapshot
+  restore-base-dir  - a temporary base directory to restore hlinks of the 
+         snapshot into for read,
```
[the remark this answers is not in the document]
jhambleton: updated as suggested

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/README.md
```diff
@@ -0,0 +1,192 @@
...
+```bash
+hbase org.apache.hadoop.hbase.snapshot.ExportSnapshot \
+ -snapshot <snapshotName> \
+ -copy-to gs://<bucket/<snapshot-dir> \
+ -mappers <num-mappers>
+```
+
+&nbsp;3. [Create a Dataproc Cluster](https://cloud.google.com/dataproc/docs/guides/create-cluster) for executing the Import Job
+
+&nbsp;4. Provision [Bigtable instance and cluster](https://cloud.google.com/bigtable/docs/instances-clusters-nodes) and update the below environment variables for the project configuration. **Important note:** the below is an example with default 1 node with HDD type. A production deployment should size the Bigtable cluster and use the appropriate [storage type](https://cloud.google.com/bigtable/docs/choosing-ssd-hdd) that meets the workload requirements.
```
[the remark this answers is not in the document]
jhambleton: updated with a few notes around scaling and link in step 8

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
jhambleton: most default values should be initialized in `BigtableOptionsFactory` and a few other classes, but this is a simple client override intended to show users how to explicitly set the timeout for batch mutations (and increase it if they timeout). i would assume that other configs added later should be verified by the user, but these shouldn't have a unit test as that would fail if these are adjusted.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
jhambleton: updated as suggested

jhambleton:
thanks for the reviews. PTAL at the updates.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
```diff
@@ -0,0 +1,290 @@
...
+    if (conf.get(BigtableOptionsFactory.INSTANCE_ID_KEY) == null) {
+      usage("Must specify the property " + BigtableOptionsFactory.INSTANCE_ID_KEY);
+      return -1;
+    }
+
+    // set arguments in configuration for ease of use
+    conf.set(SNAPSHOTNAME_KEY, args[0]);
+    conf.set(HConstants.HBASE_DIR, args[1]);
+    conf.set(TABLENAME_KEY, args[2]);
+    conf.set(SNAPSHOT_RESTOREDIR_KEY, args[3]);
```
igorbernstein2, line 153: Do these not need precondition checks as well?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
```diff
@@ -0,0 +1,290 @@
...
+    // implicit bigtable configs
+    BigtableConfiguration.configure(
+        conf,
+        conf.get(BigtableOptionsFactory.PROJECT_ID_KEY),
+        conf.get(BigtableOptionsFactory.INSTANCE_ID_KEY),
+        conf.get(BigtableOptionsFactory.APP_PROFILE_ID_KEY, ""));
+    conf.setBooleanIfUnset(BigtableOptionsFactory.BIGTABLE_USE_TIMEOUTS_KEY, true);
+    conf.setIfUnset(
+        BigtableOptionsFactory.BIGTABLE_MUTATE_RPC_TIMEOUT_MS_KEY,
+        String.valueOf(BIGTABLE_MUTATE_RPC_TIMEOUT_MS_DEFAULT));
```
igorbernstein2, line 176: why is this necessary?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
igorbernstein2: I think you can expose latency based throttling as an option, but I dont think its relevant for my concern. Not randomizing the chunks will not allow this job to scale to multiple nodes.

Can you clarify where the complexity is coming from? Can't we use the same approach as ShuffledSequenceFileInputFormat?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
```diff
@@ -0,0 +1,290 @@
...
+        return;
+      }
+
+      ctx.getCounter(ScanCounter.NUM_ROWS).increment(1);
+      ctx.getCounter(ScanCounter.NUM_CELLS).increment(cells.size());
+
+      // Split the row into multiple puts if it exceeds the maximum mutation limit
+      Iterator<Cell> cellIt = cells.iterator();
+
+      while (cellIt.hasNext()) {
```
igorbernstein2, line 249: I dont have the history as to why this was necessary, but the Import for SequenceFiles seems to be doing a lot more work in convertings Results to Puts. Can you take a look and comment as to the difference?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
```diff
@@ -0,0 +1,290 @@
...
+        Put put = new Put(key.get());
+
+        for (int i = 0; i < MAX_CELLS && cellIt.hasNext(); i++) {
+          put.add(cellIt.next());
+        }
```
igorbernstein2, line 254: This looks like silent data loss. I think it needs a bit more thought

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportJobCommon.java
```diff
@@ -0,0 +1,30 @@
...
+  public static final String SNAPSHOTNAME_KEY = "google.bigtable.import.snapshot.name";
+
+  public static final String TABLENAME_KEY = "google.bigtable.import.bigtable.tablename";
+
+  public static final String SNAPSHOT_RESTOREDIR_KEY =
+      "google.bigtable.import.snapshot.restore.dir";
+
+  public static final String IMPORT_SNAPSHOT_JOBNAME_KEY = "google.bigtable.import.jobname";
+
+  public static final Integer BIGTABLE_MUTATE_RPC_TIMEOUT_MS_DEFAULT = 600_000;
```
igorbernstein2, line 29: Why does this need special timeout?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/TestImportHBaseSnapshotJob.java
```diff
@@ -0,0 +1,69 @@
...
+  private Configuration conf;
+
+  @Before
+  public void setup() {
+    this.conf = HBaseConfiguration.create();
+  }
+
+  @Test
+  public void testSetConfFromArgs() {
+    Configuration conf = HBaseConfiguration.create();
```
igorbernstein2, line 45: shouldnt this be create(false) to avoid accidentally pulling in an hbase-site.xml?

igorbernstein2 (changes requested):
How hard do you think it its to write an end to end integration test here?
1. create a couple of rows to an hbase minicluster
2. snapshot it
3. export it using minicluster
4. import it into bigtable
5. verify the data is correct?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
jhambleton: shuffled splits has been added and note in readme for indicating how to enable latency based throttling

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/README.md
```diff
@@ -75,4 +100,106 @@ hadoop jars. However we encourage our users to migrate to
...
+The following is an example for backing up an HBase table in Google Cloud Storage 
+as an HBase snapshot and then importing the HBase snapshot to Bigtable.
+
+#### On-prem hbase cluster
+&nbsp;1. Take a snapshot of the HBase table.
+```bash
+hbase> snapshot ‘<tableName>’, ‘<snapshotName>’
+```
+
+&nbsp;2. Export the HBase snapshot to a GCS bucket (Note: the [GCS connector is required](https://cloud.google.com/dataproc/docs/concepts/connectors/install-storage-connector)
```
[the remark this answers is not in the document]
jhambleton: reformatted readme to remove these

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
jhambleton: we check the number of args as a precondition for basic verification of args such as the other mr jobs do.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
jhambleton: instead of relying on a default request timeout, the approach here is to increase this for batch mutations explicitly to indicate to users how to adjust this threshold. it may not be evident to end users that TableOutputFormat uses a bufferedmutator with this specific timeout config. what are the specs for the default request timeout if we remove this?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
jhambleton: Sure, the main differences between this & HBaseResultToMutationFn for SequenceFile Beam Import are:
* handling deletes - this isn't required as scan does read through deleted cells/rows. deletes are only returned when ["raw" mode for Scan is enabled](https://hbase.apache.org/1.4/apidocs/org/apache/hadoop/hbase/client/Scan.html#setRaw(boolean)). 
* splitting 100k - implemented to avoid [100k mutation limit](https://cloud.google.com/bigtable/docs/reference/data/rpc/google.bigtable.v2#mutaterowrequest)

lmk if there are other inputs needed here.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
jhambleton: this mimics the behavior in sequence file import [HBaseResultToMutationFn.java#L100](https://github.com/googleapis/java-bigtable-hbase/blob/master/bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/sequencefiles/HBaseResultToMutationFn.java#L100) and we test this condition in [TestSnapshotMapper.testRowExceedingMaxCells()](https://github.com/jhambleton/java-bigtable-hbase/blob/master/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/TestSnapshotMapper.java#L111) and assert if 2 rows are emitted and on number of cells matching. 

i've reviewed this again and don't see conditions where the iterator would escape the loop and silently lose data. lmk if you had anything in mind and i'd be happy to incorporate it

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportJobCommon.java
jhambleton: Instead of using a CallOptionsConfig.LONG_TIMEOUT_MS_DEFAULT as the default that is set in BigtableOptionsFactory for mutate rows timeout, we set our own default to make sure that any changes to CallOptionsConfig.LONG_TIMEOUT_MS_DEFAULT aren't impacting this value. If this timeout is deemed unnecessary from the other comment, i'll remove this.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/TestImportHBaseSnapshotJob.java
jhambleton: done

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/README.md
```diff
@@ -75,4 +100,106 @@ hadoop jars. However we encourage our users to migrate to
 `bigtable-hbase-1.x-mapreduce-1.14.0-shaded-byo-hadoop.jar` to avoid dependency
 conflicts with the existing classpath on Hadoop workers.
 
+## Examples
```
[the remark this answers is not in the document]
jhambleton: done

jhambleton:
> How hard do you think it its to write an end to end integration test here?

great suggestion and overall comments which all should be addressed. ptal

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
igorbernstein2: it should follow java-bigtable's configs. Which are defined here:
https://github.com/googleapis/googleapis/blob/master/google/bigtable/v2/bigtable_grpc_service_config.json#L44

and augmented here:
https://github.com/googleapis/java-bigtable/blob/77e6f9ef11ad7cfa474eed1826fe5ad1e32f16a2/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java#L571-L598

If the intention is to give users an example how to change this, I think it should be moved to docs?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
igorbernstein2: can you just add a code comment with this info?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
igorbernstein2: Sorry should've been more clear. If there are more cells than MAX_CELLS then they will be silently dropped.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/pom.xml
```diff
@@ -52,6 +55,18 @@ limitations under the License.
...
+    <dependencies>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud-bom</artifactId>
+        <version>${google-cloud-bom.version}</version>
+        <type>pom</type>
+        <scope>test</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
```
igorbernstein2, line 68: This looks a bit off:
1. boms should be imported with scope import
2. we shouldnt be using the google-cloud-bom in hadoop jobs

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/pom.xml
```diff
@@ -90,6 +105,45 @@ limitations under the License.
...
+      <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>com.google.protobuf</groupId>
+          <artifactId>protobuf-java</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>com.google.protobuf</groupId>
+          <artifactId>protobuf-java-util</artifactId>
+        </exclusion>
```
igorbernstein2, line 128: I'm a bit surprised to see these exclusions. emulator shouldnt pull in an deps other then junit & api-common

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/pom.xml
```diff
@@ -90,6 +105,45 @@ limitations under the License.
...
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.hbase</groupId>
+      <artifactId>hbase-testing-util</artifactId>
+      <version>${hbase1.version}</version>
+      <scope>test</scope>
+      <exclusions>
```
igorbernstein2, line 136: can you add a comment explaining the exclusions?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
```diff
@@ -0,0 +1,317 @@
...
+
+    Job job = createSubmittableJob(conf);
+    int successStatus = (job.waitForCompletion(true) ? 0 : 1);
+
+    Counters counters = job.getCounters();
+    long numRows = counters.findCounter(ScanCounter.NUM_ROWS).getValue();
+    long numCells = counters.findCounter(ScanCounter.NUM_CELLS).getValue();
+
+    System.out.println(("############# Num of ROWS imported= " + numRows));
+    System.out.println(("############# Num of CELLS imported= " + numCells));
```
igorbernstein2, line 138: should these be logs? it seems inconsistent to log the start of the job but sout the end

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/TestImportHBaseSnapshotJob.java
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.CommonConfigurationKeysPublic;
+import org.apache.hadoop.hbase.HConstants;
+import org.apache.hadoop.hbase.client.ClusterConnection;
+import org.apache.hadoop.hbase.mapreduce.TableOutputFormat;
+import org.junit.Assert;
+import org.junit.Test;
+
+/** test driver function */
+public class TestImportHBaseSnapshotJob {
```
igorbernstein2, line 32: This looks awesome!
Can you add a TODO to parameterize this to run against prod in the future?
same of the other e2e test

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
igorbernstein2: my mistake

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
jhambleton: done

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
jhambleton: added code comment

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/pom.xml
jhambleton: good catch. removed.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/pom.xml
jhambleton: good catch, these weren't necessary and likely something that slipped in.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/pom.xml
jhambleton: added comment in pom

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
jhambleton: done

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/TestImportHBaseSnapshotJob.java
jhambleton: done

jhambleton:
updated to address comments. ptal, thanks!

igorbernstein2 (approved):

PR #2852, opened 2021-02-26
test: enable snapshots integration tests

lichng:
Now we can run integration tests for HBase snapshot data import via:

```
mvn verify -PhbasesnapshotsIntegrationTest -Dgoogle.bigtable.project.id=<project_id> \
-Dgoogle.bigtable.instance.id=<instance_id> \
-Dgoogle.dataflow.stagingLocation=gs://<staging-path> \
-Dcloud.test.data.folder=gs://<test-data-bucket-path> \
-Dregion=<region>
```

Newer version of Beam SDK requires region so added that to every test pipeline.


bigtable-dataflow-parent/bigtable-beam-import/pom.xml
```diff
@@ -26,6 +26,9 @@ limitations under the License.
```
kolea2, line 26: this profile should be added to a new kokoro config corresponding to a dataflow presubmit check so the tests will run as part of this PR

kolea2 (changes requested):

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
lichng: Sure, will do.

lichng:
Verified the integration tests are running.

Next step is to optimize the test setup so user does not need to upload test data themselves.

kolea2 (approved):
lgtm, would like @igorbernstein2 to look as well

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
```diff
@@ -28,6 +28,9 @@ limitations under the License.
     <mainClass>com.google.cloud.bigtable.beam.Main</mainClass>
     <beam-import-netty.version>4.1.48.Final</beam-import-netty.version>
     <beam-import-google-http-client.version>1.34.1</beam-import-google-http-client.version>
+    <!-- Needed to enable integration tests before we figure out -->
+    <!-- possible misconfiguration of Failsafe plugin -->
+    <skipITs>false</skipITs>
```
igorbernstein2, line 33: where is this set to true?

bigtable-dataflow-parent/bigtable-beam-import/src/test/resources/log4j.properties
```diff
@@ -22,3 +22,8 @@ log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
 log4j.appender.stdout.layout.ConversionPattern=%d{yyyy-MM-dd HH:mm:ss} %-5p %c{1}:%L - %m%n
 log4j.category.org.apache.beam.sdk.io.FileBasedSource=WARN
 log4j.category.com.google.cloud.bigtable.beam.sequencefiles.SequenceFileSource=WARN
+# make hbase snapshot import integration tests output less verbose.
+log4j.category.org.apache.hadoop.hbase.snapshot.RestoreSnapshotHelper=WARN
```
igorbernstein2, line 26: can the entirety of org.apache.hadoop be set to WARN?

.kokoro/nightly/integration-beam.cfg
```diff
@@ -0,0 +1,38 @@
...
+# Configure the docker image for kokoro-trampoline.
+env_vars: {
+  key: "TRAMPOLINE_IMAGE"
+  value: "gcr.io/cloud-devrel-kokoro-resources/java8"
+}
+
+env_vars: {
+    key: "INTEGRATION_TEST_ARGS"
+    value: "-PhbasesnapshotsIntegrationTest -Dgoogle.bigtable.project.id=gcloud-devel -Dgoogle.bigtable.instance.id=hbase-snapshot-import-test -Dgoogle.dataflow.stagingLocation=gs://java-bigtable-hbase-testing/staging -Dcloud.test.data.folder=gs://java-bigtable-hbase-testing/hbase-snapshot-import-integration-tests -Dregion=us-central1"
+}
```
igorbernstein2, line 12: The profiles are a bit of mess:
`sequencefileIntegrationTest` will run all of the ITs but doesnt seem like it has the necessary region prop

`beamIntegrationTest` will run `CloudBigtableBeamITTest`
`hbasesnapshotsIntegrationTest` will run just the snapshots

I dont quite understand why all of these are separate profiles? I think the original reason for having a separate profile is that cloud resource names are available so we want to avoid running these during the regular verify phase.

I dont think we should be defining separate kokoro jobs for each test case. So I think all of the profiles should be merged. If the other ones are currently broken, feel free to exclude them in the failsafe config.

igorbernstein2 (changes requested):

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
lichng: I don't believe we have set it explicitly but it is current set to true in effective pom, if we run the following in master: 

```
$ mvn org.apache.maven.plugins:maven-help-plugin:3.2.0:effective-pom|grep skipITs
    <skipITs>true</skipITs>
```

I suspect this is caused by some misconfiguration of Failsafe but does not have enough expertise to debug at this moment.

bigtable-dataflow-parent/bigtable-beam-import/src/test/resources/log4j.properties
lichng: Done

.kokoro/nightly/integration-beam.cfg
lichng: Agree it make sense to let CI to run all integration tests in one profile.

However, I believe it would also be beneficial to keep these individual profiles as `hbasesnapshotsIntegrationTest` would launch multiple dataflow jobs hence it quite time consuming to run, hence if we would want to only test part of the package and get result faster, we can use those individual profiles.

I think `sequencefileIntegrationTest` is not currently running in any of our presubmit tasks, I can update the test and add the `region` argument.

lichng: Ok. now all integration tests under `bigtable-beam-import` should be running and passing in this kokoro task.

PTAL

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
igorbernstein2: Until we figure out whats going on, can we move it to the profile?

igorbernstein2 (approved):
LGTM after the property move

PR #2872, opened 2021-03-05
chore: remove eclipse plugin

igorbernstein2:
Its not currently used and most likely not doing what we want


kolea2 (approved):
plugin seems to be retired, https://maven.apache.org/plugins/maven-eclipse-plugin/, LGTM

PR #2875, opened 2021-03-08
chore: stop shading test artifacts

igorbernstein2:
I think this was added awhile back when our integration tests used shaded artifacts. This is no longer the case and adds a lot of noise

### DIFF
--- a/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-hadoop/pom.xml
+++ b/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-hadoop/pom.xml
@@ -114,7 +114,6 @@ limitations under the License.
               <goal>shade</goal>
             </goals>
             <configuration>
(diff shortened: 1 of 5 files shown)


bigtable-hbase-1.x-parent/bigtable-hbase-1.x-hadoop/pom.xml
```diff
@@ -118,7 +118,6 @@ limitations under the License.
               <goal>shade</goal>
             </goals>
             <configuration>
```
kolea2, line 120: does this also stop creating an artifact/publishing the test jar? If no, could we add it as part of this?

igorbernstein2: not sure, I'll come back to this after the dependencies have settled

kolea2 (approved):

PR #2874, opened 2021-03-08
chore(deps): remove dependency analysis for packaging artifacts and non-compile deps

igorbernstein2:
This should pull in less unnecessary deps as hbase-client explicitly excludes a bunch of transitive deps from hadoop-common.  Also this should simplify maintenance.

Also remove dependency analysis from bigtable-hbase-1.x-mapreduce. The artifact is primarily a copy of the mapreduce jobs that exist in hbase-server with the bigtable-hbase-1.x-hadoop inserted. This means that module lives within the confines of hbase-server dependencies and direct dep analysis doesn't buy much value.

PR #2877, opened 2021-03-09
chore: ensure that multi module dependencies align on the same version

igorbernstein2:
Note: this depends on #2874 getting merged, it resolves an issue where depending on hadoop-common directly caused slf4j modules to get misaligned.

The check showed that netty & google-http-client modules were misaligned. I fixed them using the boms for each project. This is a temporary solution in an upcoming PR I will replace both of the boms with the new beam bom


bigtable-hbase-1.x-parent/bigtable-hbase-1.x-hadoop/pom.xml
```diff
@@ -76,6 +77,23 @@ limitations under the License.
...
+    </dependency>
+
+    <!-- unlike hbase-shaded-client, hbase-client depends on slf4j 1.6.1.
+      Which is older than the version requested by metrics-core. However,
+      metrics-core will work fine with the older version and we want to stay
+      compatible with hbase-client deps -->
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+      <version>1.6.1</version>
```
kolea2, line 96: please double check this is in renovate config to skip

bigtable-client-core-parent/bigtable-client-core/pom.xml
```diff
@@ -224,7 +224,7 @@ this is not used directly, but upgrading due to transitive vulnerabilities in ol
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-databind</artifactId>
       <!-- align version with veneer's flattened dep -->
-      <version>2.11.1</version>
+      <version>2.12.1</version>
```
kolea2, line 227: I think this will break as per comment above

igorbernstein2: veneer uses 2.12.1 now....this is just catching up to it

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-hadoop/pom.xml
igorbernstein2: I updated the renovate config

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-hadoop/pom.xml
```diff
@@ -30,12 +30,18 @@ limitations under the License.
...
+    <hbase1-hadoop-slf4j.version>1.6.1</hbase1-hadoop-slf4j.version>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>bigtable-hbase-1.x-shaded</artifactId>
       <version>${project.version}</version>
       <exclusions>
+        <!-- hbase-shaded-client will be replaced with hbase-client -->
```
kolea2, line 44: nit: add a TODO infront of this comment

kolea2 (approved):
lgtm with one nit

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-hadoop/pom.xml
igorbernstein2: This just explaining whats happening in this pom: a few lines below we are adding hbase-client which was removed on this line

PR #2879, opened 2021-03-09
chore: stop shading conscrypt

igorbernstein2:
Until https://github.com/google/conscrypt/issues/811 is fixed, we can't shade it reliably.

A follow up PR will add build time checks to ensure that grpc-conscrypt property is properly kept in sync

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,6 +58,7 @@ limitations under the License.
     <!-- core dependency versions -->
     <bigtable.version>1.21.0</bigtable.version>
     <dropwizard.metrics.version>3.2.6</dropwizard.metrics.version>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -58,6 +58,7 @@ limitations under the License.
     <!-- core dependency versions -->
     <bigtable.version>1.21.0</bigtable.version>
     <dropwizard.metrics.version>3.2.6</dropwizard.metrics.version>
+    <grpc-conscrypt.version>2.5.1</grpc-conscrypt.version>
```
kolea2, line 61: probably want to add to renovate config?

igorbernstein2: good call, done!

kolea2 (approved):

PR #2884, opened 2021-03-10
chore: remove stale dependency override

igorbernstein2:
This was originally added to force upgrade a transitive dep that had security vulnerabilities. This dep is no longer used, so the override is not necessary


kolea2 (commented):
to confirm, this isn't showing up in `mvn dependency:tree`?

kolea2 (approved):

PR #2885, opened 2021-03-10
chore(test): introduce a build checks to help maintain proper dep & shading configs

igorbernstein2:
note: this PR depends on #2882 and currently includes it's commit. This will be removed once #2882 gets merged

This PR employs the maven plugins merged in #2847 to ensure that when upgrading dependencies in the future doesnt break. Currently there are a couple of issues that this explicitly ignores like conscrypt, error_prone, etc

Future PRs will address these issues

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/pom.xml
+++ b/bigtable-client-core-parent/bigtable-hbase/pom.xml
@@ -98,6 +98,13 @@ limitations under the License.
       <artifactId>grpc-api</artifactId>
     </dependency>
     <dependency>
+      <!-- See notes in verify-conscrypt-version execution below -->
+      <groupId>org.conscrypt</groupId>
+      <artifactId>conscrypt-openjdk-uber</artifactId>
+      <version>${grpc-conscrypt.version}</version>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
       <groupId>io.grpc</groupId>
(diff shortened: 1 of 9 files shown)


bigtable-client-core-parent/bigtable-hbase/pom.xml
```diff
@@ -203,6 +210,40 @@ limitations under the License.
           </systemProperties>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>com.google.cloud.bigtable.test</groupId>
+        <artifactId>bigtable-build-helper</artifactId>
+        <version>1.14.1-SNAPSHOT</version> <!-- {x-version-update:bigtable-test:current} -->
```
kolea2, line 216: can we use the project.version variable instead?

kolea2 (changes requested):

bigtable-client-core-parent/bigtable-hbase/pom.xml
igorbernstein2: its different artifacts build-helper uses bigtable-test version while project.version follows bigtable-client-parent version. While we have the split in versions.txt, we should keep this line as is. In the future we can collapse everything to a single identifier in versions.txt and then update this

kolea2 (approved):

PR #2892, opened 2021-03-10
chore(deps): upgrade beam 2.28.0

igorbernstein2:
New version of beam requires a region now. So to avoid breaking our existing users I added code that will infer the region from the zone.
I also removed the docs for beam specific options to avoid these kinds of issues in the future

### DIFF
--- a/bigtable-dataflow-parent/bigtable-beam-import/pom.xml
+++ b/bigtable-dataflow-parent/bigtable-beam-import/pom.xml
@@ -26,41 +26,25 @@ limitations under the License.
 
   <properties>
     <mainClass>com.google.cloud.bigtable.beam.Main</mainClass>
-    <beam-import-netty.version>4.1.48.Final</beam-import-netty.version>
-    <beam-import-google-http-client.version>1.34.1</beam-import-google-http-client.version>
   </properties>
 
   <!-- Adding this to resolve version conflict within beam sdk-->
   <dependencyManagement>
     <dependencies>
       <dependency>
-        <groupId>io.grpc</groupId>
-        <artifactId>grpc-bom</artifactId>
-        <version>${beam-grpc.version}</version>
+        <groupId>org.apache.beam</groupId>
+        <artifactId>beam-sdks-java-bom</artifactId>
+        <version>${beam.version}</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
-
-      <!-- BEGIN temporary workaround before beam upgrade -->
-      <!-- Until we upgrade to a version of beam that includes a bom,
-      temporarily align transitive dep modules -->
-      <!-- TODO: replace this with beam's bom -->
       <dependency>
-        <groupId>io.netty</groupId>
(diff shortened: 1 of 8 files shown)


bigtable-dataflow-parent/bigtable-beam-import/pom.xml
```diff
@@ -140,21 +120,18 @@ limitations under the License.
...
-      <version>${beam-auto-value.version}</version>
+      <version>1.7.4</version>
       <scope>provided</scope>
     </dependency>
 
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>${gcs-guava.version}</version>
+      <version>30.1-jre</version>
```
kolea2, line 130: to confirm, this a 1.8 module?

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/sequencefiles/Utils.java
```diff
@@ -46,6 +47,21 @@ public static PipelineOptions tweakOptions(PipelineOptions opts) {
     }
     DataflowPipelineOptions dataflowOpts = opts.as(DataflowPipelineOptions.class);
 
+    // Region is a newly added requirement in newer versions of dataflow runner.
+    // Make the pipeline backwards compatible by inferring the region from the zone.
+    // This is done by chopping off the last dash
+    if (Strings.isNullOrEmpty(dataflowOpts.getRegion())) {
```
kolea2, line 53: please add a test for this

kolea2 (changes requested):

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
igorbernstein2: yep, we can use java8 in beam

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/sequencefiles/Utils.java
igorbernstein2: added, it actually caught a typo :)

kolea2 (approved):

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
```diff
@@ -140,21 +120,18 @@ limitations under the License.
     <dependency>
       <groupId>com.google.auto.value</groupId>
       <artifactId>auto-value</artifactId>
-      <version>${beam-auto-value.version}</version>
+      <version>1.7.4</version>
```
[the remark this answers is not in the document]
igorbernstein2: nope, we beam now uses vendored versions of these, so we are free to upgrade all we want

PR #2894, opened 2021-03-11
chore(deps): cleanup forced transitive dep upgrades

igorbernstein2:
Awhile back we added direct deps on indirect dependencies to exclude deps that have security issues. The security issues are not relevant to bigtable, but we decided it was good practice. This PR revisits that decision:
1. it add build checks that will by default fail the build if we accidentally depend on the problematic versions
2. it specifically ignores the new checks in scenarios where are not in control of our deps

This depends on beam upgrade (#2892  and needs to be rebased once that PR merges


kolea2 (changes requested):
lgtm, please highlight changes in readme

kolea2 (approved):

PR #2897, opened 2021-03-15
chore(deps): reorganize dependency ordering

igorbernstein2:
This provides 2 benefits:
1. easier for me to reason about
2. gives precedence to more relevant deps: in a beam job we should prefer beam transitive deps to hbase, since our client is shaded the hbase deps dont matter, but the deps in the beam worker are actively used

### DIFF
--- a/bigtable-dataflow-parent/bigtable-beam-import/pom.xml
+++ b/bigtable-dataflow-parent/bigtable-beam-import/pom.xml
@@ -49,6 +49,41 @@ limitations under the License.
   </dependencyManagement>
 
   <dependencies>
+    <!-- Beam Group: should come first since we will be running in the beam ecosystem -->
+    <dependency>
+      <groupId>org.apache.beam</groupId>
+      <artifactId>beam-sdks-java-core</artifactId>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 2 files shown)


bigtable-dataflow-parent/bigtable-beam-import/pom.xml
```diff
@@ -49,6 +49,41 @@ limitations under the License.
...
+    <dependency>
+      <groupId>org.apache.beam</groupId>
+      <artifactId>
+        beam-sdks-java-extensions-google-cloud-platform-core
+      </artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.beam</groupId>
+      <artifactId>beam-runners-google-cloud-dataflow-java</artifactId>
+      <exclusions>
```
kolea2, line 66: mind adding a small comment about exclusions? At first glance I thought we were adding these as deps

kolea2 (approved):
lgtm one small request

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
igorbernstein2: done. Thanks for the quick review

PR #2911, opened 2021-03-30
fix: enabling the integration test on the correct kokoro target

lichng:
When moving the <skipITs> to individual profile, I accidentally moved it to a incorrect target instead of `beamIntegrationTest`.

This PR is to fix this and run Beam integration tests in presubmit and nightly.

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -51,7 +51,7 @@ javadoc)
     ;;
 integration)
 # clean needed when running more than one IT profile
-    mvn clean verify -B ${INTEGRATION_TEST_ARGS} -DtrimStackTrace=false -Dclirr.skip=true -fae
+    mvn clean verify -B ${INTEGRATION_TEST_ARGS} -Penable-integration-tests -DtrimStackTrace=false -Dclirr.skip=true -fae
     RETURN_CODE=$?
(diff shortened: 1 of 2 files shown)


igorbernstein2:
I dont think this is the correct way to go about this.

java-shared-config disables IT tests by default:
https://github.com/googleapis/java-shared-config/blob/393ba5669675afb44740fdc45c8110feccf0b3b6/pom.xml#L65

This is done because we want to be able to build the project immediately after git clone and integration tests require the developer to set up cloud resources and creds.

To actually run the integration tests, java-shared-config provides this profile:
https://github.com/googleapis/java-shared-config/blob/393ba5669675afb44740fdc45c8110feccf0b3b6/pom.xml#L651

Which can be activated by -Penable-integration-tests. I believe the intention is to enable that profile during build similar to how java-bigtable does it:
https://github.com/googleapis/java-bigtable/blob/2821902b34ae04596771a9fc5e2b62d5c24a7253/.kokoro/build.sh#L64

Would you mind updating build.sh instead?

igorbernstein2 (changes requested):
please update build.sh instead

igorbernstein2 (approved):
lgtm

PR #2917, opened 2021-04-02
fix: dont use channel pools for admin api for veneer

igorbernstein2:


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -366,9 +366,11 @@ public ManagedChannelBuilder apply(ManagedChannelBuilder channelBuilder) {
           });
     }
 
-    String channelCount = configuration.get(BIGTABLE_DATA_CHANNEL_COUNT_KEY);
-    if (!Strings.isNullOrEmpty(channelCount)) {
-      channelProvider.setPoolSize(Integer.parseInt(channelCount));
+    if (endpointKey.equals(BIGTABLE_HOST_KEY)) {
```
kolea2, line 369: i'd flip this assertion

kolea2 (approved):
lgtm with one nit

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
igorbernstein2: I think I should keep it as is...BIGTABLE_DATA_CHANNEL_COUNT_KEY only applies to BIGTABLE_HOST_KEY. If we ever need to add a third host, I wouldnt want to update this logic

PR #2918, opened 2021-04-02
fix: remove duplicate classic setting parsing

igorbernstein2:
BigtableHBaseClassicSettings should just delegate to BigtableOptionsFactory.fromConfig(). There is no reason to duplicate the logic


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BigtableHBaseClassicSettings.java
```diff
@@ -15,101 +15,22 @@
...
-      bigtableOptionsBuilder.enableEmulator(emulatorHost);
-    }
-
-    String useBatchStr = configuration.get(BIGTABLE_USE_BATCH);
-    if (!isNullOrEmpty(useBatchStr)) {
-      bigtableOptionsBuilder.setUseBatch(Boolean.parseBoolean(useBatchStr));
-    }
-
-    this.bigtableOptions = bigtableOptionsBuilder.build();
+    this.bigtableOptions = BigtableOptionsFactory.fromConfiguration(configuration);
```
kolea2, line 33: confirming this is already tested somewhere?

igorbernstein2: There is `TestBigtableOptionsFactory`, which tests some of this stuff and it could be improved. But expanding that test is out of scope for this PR especially since we are deleting BIgtableHBaseClassicSettings shortly after the alpha

kolea2 (approved):

PR #2922, opened 2021-04-14
chore: stop publishing test artifacts & fix maven-jar-plugin failures on test artifacts

### LINKED ISSUE #2880: stop publishing test artifacts

igorbernstein2:
- This updates all test artifacts to disable release related plugins: deploy, nexus-staging, site, source, javadoc, gpg, & clirr.
- The integration test modules were failing build because they don't have any prod classes, so maven-jar-plugin would fail because it couldnt find the primary jar. This fixes it by adding placeholder classes
- Finally this pre-emptively applies changes from https://github.com/googleapis/synthtool/pull/1043

fixes #2880

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/pom.xml
+++ b/bigtable-client-core-parent/bigtable-hbase-integration-tests-common/pom.xml
@@ -67,19 +67,75 @@ limitations under the License.
   </dependencies>
 
   <build>
-    <plugins>
-      <plugin>
-        <artifactId>maven-surefire-plugin</artifactId>
-        <configuration>
-          <skip>true</skip>
-        </configuration>
-      </plugin>
-      <plugin>
-        <artifactId>maven-failsafe-plugin</artifactId>
-        <configuration>
-          <skip>true</skip>
-        </configuration>
-      </plugin>
-    </plugins>
+    <pluginManagement>
+      <plugins>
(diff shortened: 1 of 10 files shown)


bigtable-client-core-parent/bigtable-hbase-integration-tests-common/pom.xml
```diff
@@ -67,19 +67,75 @@ limitations under the License.
...
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-javadoc-plugin</artifactId>
+          <configuration>
+            <skip>true</skip>
+          </configuration>
+        </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-gpg-plugin</artifactId>
```
kolea2, line 125: do you need to add this? I think the first two are sufficient; see the benchmarks package as an example https://github.com/googleapis/java-bigtable-hbase/blob/bigtable-1.x/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-benchmarks/pom.xml#L84-L99

igorbernstein2: there is no point in signing artifacts if we wont publish them

kolea2 (approved):

PR #2925, opened 2021-04-15
fix: fix retry so it won't fail when rows read == rows limit

mutianf:


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ReadRowsRequestManager.java
```diff
@@ -64,6 +60,19 @@ void incrementRowCount(int count) {
     rowCount += count;
   }
 
+  boolean isConsumed() {
+    if (this.lastFoundKey == null) {
+      return false;
+    }
+    if (filterRows().equals(RowSet.getDefaultInstance())) {
+      return true;
+    }
```
igorbernstein2, line 69: Can you merge these if statements and add a comment along the lines of:
A non-null lastFoundKey implies that we had at least one row that was returned and the filtered request should be truncated. In this case an empty filtered row set implies that all rows have been seen and this stream is complete.

igorbernstein2 (approved):
lgtm! thanks for fixing this

PR #2930, opened 2021-04-19
fix: multithreaded batch operations

igorbernstein2:
Fix thread safety issues with batch operations on BigtableTable


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BulkMutationVeneerApi.java
```diff
@@ -42,13 +42,13 @@
 
   /** {@inheritDoc} */
   @Override
-  public void sendUnsent() {
+  public synchronized void sendUnsent() {
```
mutianf, line 45: Does this need to be synchronized? The methods in Batcher should be thread safe? https://github.com/googleapis/gax-java/blob/master/gax/src/main/java/com/google/api/gax/batching/BatcherImpl.java#L221

mutianf (approved):

PR #2953, opened 2021-04-29
fix: Update metric logging to make sure metrics always show up.

vermas2012:
Updating the counters to be logged everytime. This will result in counters being 0 on UI rather than being absent.

### DIFF
--- a/bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFn.java
+++ b/bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFn.java
@@ -90,9 +90,14 @@ class ComputeAndValidateHashFromBigtableDoFn
     // BufferedHadoopHashTableSource generates only 1 item per groupby key, key is startKey for the
     // Sorted ranges.
     Preconditions.checkState(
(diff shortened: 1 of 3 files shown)


bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/validation/ComputeAndValidateHashFromBigtableDoFn.java
```diff
@@ -90,10 +90,15 @@ public void processElement(ProcessContext context) throws Exception {
...
     // Sorted ranges.
     Preconditions.checkState(
-        wrapperdRangeHashes.size() == 1, "Can not have muiple entries for a key");
+        wrapperdRangeHashes.size() == 1, "Can not have multiple entries for a key");
     List<RangeHash> rangeHashes = wrapperdRangeHashes.get(0);
     Preconditions.checkState(!rangeHashes.isEmpty(), "Can not have empty ranges in DO_FN");
 
+    // If a metric is not logged, it is absent from all the metrics (as opposed to being
+    // 0). By logging a 0 value for the metrics we guarantee that they shows up on Dataflow UIs.
+    mismatches.inc(0);
```
vermas2012, line 99: @igorbernstein2 I checked the beam docs and not sure if logging these metrics in constructor will work. We can move it into a StartBundle method but I didn't think this big enough change to introduce the methods in this class.

igorbernstein2 (approved):
lgtm

PR #2959, opened 2021-05-05
chore: reset veneer executor to be gRPC default instead of gax

igorbernstein2:
This aligns veneer impl to use the same executor as bigtable-client-core


mutianf (approved):
lgtm

PR #2987, opened 2021-05-24
build: only publish bigtable-hbase-beam and its dep bigtable-hbase-1x-shaded for sp branch

kolea2:
This will only publish bigtable-hbase-beam, its dependency bigtable-hbase-1x-shaded, and the parent poms


igorbernstein2 (approved):
lgtm

PR #2988, opened 2021-05-24
chore(test): use same the same resource name pattern as java-bigtable tests

igorbernstein2:
Use the pattern temp-${secs since epoch}-${random}-${process counter} to create new tables. This should guarantee test uniqueness when running multiple tests concurrently


mutianf (approved):
lgtm

PR #2994, opened 2021-05-24
build: only build bigtable-hbase-beam and its dependencies for releases

kolea2:


igorbernstein2 (approved):
lgtm, but please:
- adds comments to the .sh files
- exclude sh files from being overwritten

PR #3007, opened 2021-06-09
fix: keep only failed actions in List<Delete>

### LINKED ISSUE #2998: Table#delete(List) should mutate the list in place and remove successful actions
According to
https://hbase.apache.org/1.4/apidocs/org/apache/hadoop/hbase/client/Table.html#delete(java.util.List)

When an error happens, the list should get modified to only include failed deletes

mutianf:

Fixes #2998 ☕️

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
@@ -267,7 +267,7 @@ public class BatchExecutor {
    * @throws java.lang.InterruptedException if any.
    * @param <R> a R object.
    */
-  public <R> void batchCallback(
+  <R> void batchCallback(
       List<? extends Row> actions,
       Object[] results,
       Batch.Callback<R> callback,
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
@@ -222,17 +222,23 @@ public class BatchExecutor {
     }
   }
 
+  public Result[] batch(List<? extends Row> actions) throws IOException {
+    return batch(actions, false);
+  }
+
   /**
    * batch.
    *
    * @param actions a {@link java.util.List} object.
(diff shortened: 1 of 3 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
```diff
@@ -222,17 +222,23 @@ public void batch(List<? extends Row> actions, @Nullable Object[] results)
...
   /**
    * batch.
    *
    * @param actions a {@link java.util.List} object.
+   * @param removeSucceededActions if true, remove succeeded actions from {@code actions}
    * @return an array of {@link org.apache.hadoop.hbase.client.Result} objects.
    * @throws java.io.IOException if any.
    */
-  public Result[] batch(List<? extends Row> actions) throws IOException {
+  public Result[] batch(List<? extends Row> actions, boolean removeSucceededActions)
```
igorbernstein2, line 237: Should this be package private?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java
```diff
@@ -284,6 +300,14 @@ public void batch(List<? extends Row> actions, @Nullable Object[] results)
         hosts.add(options.getDataHost());
       }
     }
+    if (removeSucceededActions) {
+      for (int i = results.length - 1; i >= 0; i--) {
+        // if result is not null, it succeeded
+        if (results[i] instanceof Result) {
```
igorbernstein2, line 306: I'm a bit confused by this check. Isnt the result type for a delete a Void? and Result is for Get?

mutianf: Hmm, I thought it would be an empty result? https://github.com/googleapis/java-bigtable-hbase/blob/master/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BatchExecutor.java#L107

I could also check if it's not a throwable

igorbernstein2: I think the later would be better

igorbernstein2 (approved):
lgtm

PR #3020, opened 2021-06-18
fix: keep only failed actions in List<Delete> in master

### LINKED ISSUE #2998: Table#delete(List) should mutate the list in place and remove successful actions

mutianf:
Fixing in master

Fixes #2998 ☕️


igorbernstein2 (approved):
lgtm

PR #3031, opened 2021-06-23
feat: allow DirectPath by default + update integration tests

igorbernstein2:
* DirectPath is now opt-out by setting google.bigtable.direct.path.allowed to false
* Integration tests can now verify that they are running over the desired mode by specifying "bigtable.connection-mode" system property
* Integration tests update their user agent to signal membership in allow/deny lists
* DirectPathFallbackIT now supports IPv4


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -794,7 +711,7 @@ public BigtableInstanceClient getInstanceAdminClient() throws IOException {
   public static BigtableInstanceClient createInstanceClient(BigtableOptions options)
       throws IOException, GeneralSecurityException {
 
-    ManagedChannel rawAdminChannel = createNettyChannel(options.getAdminHost(), options);
+    ManagedChannel rawAdminChannel = createNettyChannel(options.getAdminHost(), options, false);
```
kolea2, line 714: I think the parameters are backwards here (options should go last)?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/util/IpVerificationInterceptor.java
```diff
@@ -0,0 +1,61 @@
+package com.google.cloud.bigtable.hbase.util;
```
kolea2, line 1: missing header

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOptionsFactory.java
```diff
@@ -396,6 +410,24 @@ public static BigtableOptions fromConfiguration(final Configuration configuratio
...
+    if (s != null) {
+      final ChannelConfigurator oldConfigurator = bigtableOptionsBuilder.getChannelConfigurator();
+      bigtableOptionsBuilder.setChannelConfigurator(new ChannelConfigurator() {
+        @Override
+        public ManagedChannelBuilder configureChannel(ManagedChannelBuilder builder, String host) {
+          if (oldConfigurator != null) {
+            builder = oldConfigurator.configureChannel(builder, host);
+          }
+          if (!host.contains("admin")) {
+            builder = builder.intercept(new IpVerificationInterceptor(Pattern.compile(s)));
```
kolea2, line 424: I'm a little worried about adding this code into the prod path when it's really just meant for testing. Is there another way we could do this in the test itself?

igorbernstein2: The only other way is to inject classes into the Configuration, which quickly becomes a rube goldberg machine:

```
Class configuratorCls = config.getClass("google.bigtable.connection.configurator.class", ChannelConfigurator.class, NoopClass.class);
ChannelConfigurator instance = configuratorCls.newInstance();
instance.configureChannel(builder, host);

class DirectPathIpv4InterceptorConfigurator implements ChannelConfigurator {
// filtering for ipv4
}

class DirectPathIpv6InterceptorConfigurator implements ChannelConfigurator {
// filtering for ipv6
}
class DirectPathAnyInterceptorConfigurator implements ChannelConfigurator {
// filtering for ipv4 or ipv6
}
class CfeInterceptorConfigurator implements ChannelConfigurator {
// ! Direct path filtering
}
```
Which makes this a lot more difficult to deal with. Veneer already has this test, so once we transition to veneer, this can go away. It will be removed in 2.0 beta along with bigtable-client-core

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/util/IpVerificationInterceptor.java
igorbernstein2: fixed

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableSession.java
```diff
@@ -535,24 +538,29 @@ public static ManagedChannel createNettyChannel(
...
+          .keepAliveTime(CHANNEL_KEEP_ALIVE_TIME_SECONDS, TimeUnit.SECONDS)
+          .keepAliveTimeout(CHANNEL_KEEP_ALIVE_TIMEOUT_SECONDS, TimeUnit.SECONDS);
     }
 
     if (options.getChannelConfigurator() != null) {
       builder = options.getChannelConfigurator().configureChannel(builder, host);
     }
 
     return builder
+        .disableServiceConfigLookUp()
```
kolea2, line 561: why was this added? Do we need to test something about it?

igorbernstein2: Service configs arent currently published for bigtable and the client doesnt expect any service configs. If at some point in the future we want to start using them then we need to ensure that the service configs in the client arent overridden by things that it doesnt expect. Unfortunately trying to test config fetching requires a dns server and very hard to achieve

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/test_env/BigtableEnv.java
```diff
@@ -81,6 +81,29 @@ protected void setup() throws IOException {
...
 
+    // Configure DirectPath settings:
+    // - when a required mode is specified, ensure that ip address match the mode
+    // - specify a user agent that will trigger deny/allow lists on the serverside
+    switch (getConnectionMode()) {
+      case DEFAULT:
+        // nothing special
+        break;
+      case REQUIRE_CFE:
+        configuration.set("google.bigtable.test.data.ip.regex", "^(?!2001:4860:8040|34\\.126).*");
```
mutianf, line 92: nit: Should we use `BigtableOptionsFactory.BIGTABLE_TEST_DATA_IP_REGEX` and `CUSTOM_USER_AGENT_KEY ` here instead of "google.bigtable.test.data.ip.regex" and "google.bigtable.custom.user.agent"?

igorbernstein2: Common doesn’t know about bigtable-hbase…it uses hbase apis only

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/util/DirectPathUtil.java
```diff
@@ -0,0 +1,66 @@
...
+  private static final String GCE_PRODUCTION_NAME_AFTER_2016 = "Google Compute Engine";
+
+  private static final Supplier<Boolean> isOnComputeEngine =
+      Suppliers.memoize(
+          new Supplier<Boolean>() {
+            @Override
+            public Boolean get() {
+              String osName = System.getProperty("os.name");
+              if ("Linux".equals(osName)) {
+                String cmd = "cat /sys/class/dmi/id/product_name";
```
coryan, line 40: drive-by: 

why `cat` and then read the output?  That file (if it exists) `/sys/class/dmi/id/produc_name` can be opened like any other file.

Also: I think this won't work on GKE, the containers in that case run without the necessary permissions to read that file.

igorbernstein2: These are excellent points. I was trying to keep consistent with all of the other java clients:
https://github.com/googleapis/gax-java/pull/1250

It seems like that file is accessible in gke:
https://github.com/googleapis/gax-java/pull/1250/files/cf494cef60f4d9ef265e5b4690ebe3cb49537893#r589861385

I'll update both libraries to simply read the file w/o exec

coryan: Ooops, my recollection was wrong regarding GKE, it was Cloud Run and gVisor, but that may have changed again:

https://github.com/googleapis/google-cloud-cpp/pull/3959

For these purposes it might not matter because using Direct Path with Cloud Run might not be that useful or important.  If you are ever interested in supporting Windows, I wrote the code a long time ago for that:

https://github.com/grpc/grpc/blob/master/src/core/lib/security/credentials/alts/check_gcp_environment_windows.cc

igorbernstein2: Thanks for the pointer. Though registry access from java is a tough pill to swallow.

mutianf (approved):

PR #3040, opened 2021-06-29
fix: use binary search to get region location

mutianf:
Use binary search in `getRegionLocation` instead of iterating through all the regions because the keys returned by SampleRowKeys are sorted. When used in `org.apache.hadoop.hbase.mapreduce.TableInputFormatBase` (https://github.com/apache/hbase/blob/master/hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/TableInputFormatBase.java#L316) to get splits, this can improve the performance from N^2 to Nlog(N).

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableRegionLocator.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableRegionLocator.java
@@ -75,10 +75,11 @@ public abstract class BigtableRegionLocator extends AbstractBigtableRegionLocato
       HRegionLocation regionLocation = regions.get(mid);
       HRegionInfo regionInfo = regionLocation.getRegionInfo();
 
-      // This isn't the last region (endKey != "") and row key is fall outside of it
+      // This isn't the last region (endKey != "") and row key is greater than the current bound
       if (regionInfo.getEndKey().length > 0 && Bytes.compareTo(row, regionInfo.getEndKey()) >= 0) {
         low = mid + 1;
       } else if (Bytes.compareTo(row, regionInfo.getStartKey()) < 0) {
+        // no need to check empty key because it will compare naturally
         high = mid - 1;
       } else {
         return regionLocation;
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableRegionLocator.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableRegionLocator.java
@@ -22,6 +22,7 @@ import com.google.cloud.bigtable.core.IBigtableDataClient;
 import java.io.IOException;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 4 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableRegionLocator.java
```diff
@@ -61,11 +62,29 @@ public HRegionLocation getRegionLocation(byte[] row) throws IOException {
...
+  private HRegionLocation findRegion(List<HRegionLocation> regions, byte[] row) throws IOException {
+    int low = 0;
+    int high = regions.size() - 1;
+
+    while (low <= high) {
+      int mid = (low + high) >>> 1;
+      HRegionLocation regionLocation = regions.get(mid);
+      HRegionInfo regionInfo = regionLocation.getRegionInfo();
+
+      // This isn't the last region (endKey != "") and row key is fall outside of it
```
igorbernstein2, line 78: // This isn't the last region (endKey != "") and row key is greater than the current bound

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableRegionLocator.java
```diff
@@ -61,11 +62,29 @@ public HRegionLocation getRegionLocation(byte[] row) throws IOException {
...
+      if (regionInfo.getEndKey().length > 0 && Bytes.compareTo(row, regionInfo.getEndKey()) >= 0) {
+        low = mid + 1;
+      } else if (Bytes.compareTo(row, regionInfo.getStartKey()) < 0) {
```
igorbernstein2, line 81: // no need to check empty key because it will compare naturally

bigtable-hbase-2.x-parent/bigtable-hbase-2.x/pom.xml
```diff
@@ -149,6 +149,16 @@ limitations under the License.
...
     </dependency>
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.api.grpc</groupId>
+      <artifactId>grpc-google-cloud-bigtable-v2</artifactId>
+      <scope>test</scope>
```
igorbernstein2: Both of these should be `runtime`:
They are build deps of a dep (bigtable-hbase-2x -> bigtable-hbase). However they werent listed here because until now they werent directly referenced in bigtable-hbase-2x. Now you are referencing them in a test.
Maven treats your declaration here as an override, which means that during runtime in production, the end user will get a missing class error. You have 2 options here:
1. change scope to runtime and add a comment (runtime is available in both test and runtime)
2. exclude the dep from dependency plugin and add comment

mutianf: hmm I think scope needs to be compile

igorbernstein2: ok

igorbernstein2 (approved):
lgtm

PR #3041, opened 2021-06-29
chore: Wrapping local test Server starts within a retry to reduce test flakiness

shantstepanian:
Code areas to review:

1) TestServerBuilder (though the final location (which module, test vs main) are TBD
2) All the test changes that remove the "new ServerSocket(0)" call and move to TestServerBuilder
3) One test calls "new ServerSocket(0)" unnecessarily (see TestBigtableHBaseVeneerSettings). I just specify a number there now.

### DIFF
--- a/bigtable-test/bigtable-internal-test-helper/pom.xml
+++ b/bigtable-test/bigtable-internal-test-helper/pom.xml
@@ -1,0 +1,107 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>bigtable-test</artifactId>
+        <groupId>com.google.cloud.bigtable</groupId>
+        <version>2.0.0-alpha-1-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>bigtable-internal-test-helper</artifactId>
+    <description>
+        Module with helper code for Bigtable unit and integration tests.
+        Not intended to be shared with external users.
+    </description>
+
+    <properties>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
+    </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>com.google.cloud</groupId>
+                <artifactId>google-cloud-bigtable-bom</artifactId>
+                <version>${bigtable.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+            <dependency>
+                <groupId>com.google.cloud</groupId>
+                <artifactId>google-cloud-bigtable-deps-bom</artifactId>
+                <version>${bigtable.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-api</artifactId>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <pluginManagement>
+            <plugins>
+                <!-- Start Skip publishing -->
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-deploy-plugin</artifactId>
+                    <version>3.0.0-M1</version>
+                    <configuration>
+                        <skip>true</skip>
+                    </configuration>
+                </plugin>
+                <plugin>
+                    <groupId>org.sonatype.plugins</groupId>
+                    <artifactId>nexus-staging-maven-plugin</artifactId>
+                    <configuration>
+                        <skipNexusStagingDeployMojo>true</skipNexusStagingDeployMojo>
+                    </configuration>
+                </plugin>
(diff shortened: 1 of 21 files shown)


bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/util/TestUtil.java
```diff
@@ -0,0 +1,37 @@
+package com.google.cloud.bigtable.util;
+
+import com.google.cloud.bigtable.config.Logger;
+import io.grpc.Server;
+import io.grpc.ServerBuilder;
+
+import java.io.IOException;
+import java.net.ServerSocket;
+
+public class TestUtil {
```
shantstepanian, line 10: I haven't thought about the right package/ module to put this in yet. Putting it in this location for now just to show the code

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/util/TestUtil.java
```diff
@@ -0,0 +1,37 @@
...
+      } catch (IOException e) {
+        LOG.trace("Could not setup port; retrying ...", e);
+      }
+    }
+
+    // Exhausted retries.
+    throw new IllegalArgumentException("Could not create server after " + maxRetries + " retries; unable to find port");
+  }
+
+  public interface ServerBuilderFactory {
```
shantstepanian, line 34: We're on Java 7, so I guess it's a functional interface and abstract classes on the calling side for now ... I'm open to better suggestions

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/util/TestUtil.java
```diff
@@ -0,0 +1,37 @@
...
+
+  public static Server buildServer(ServerBuilderFactory serverBuilderFactory) throws IOException {
+    int maxRetries = 3;
+
+    for (int retry = 0; retry < maxRetries; retry++) {
+      int port;
+      try (ServerSocket s = new ServerSocket(0)) {
+        port = s.getLocalPort();
+      }
+      Server server = serverBuilderFactory.buildServer(port).build();
```
shantstepanian, line 21: I wanted the interface to return the server builder so that we can call the start in here and simplify the clien code.
I'm open to having the ServerBuilder.forPort(port) call be in here too

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/util/TestUtil.java
```diff
@@ -0,0 +1,37 @@
...
+      try {
+        server.start();
+        return server;
+      } catch (IOException e) {
+        LOG.trace("Could not setup port; retrying ...", e);
```
igorbernstein2, line 26: I dont think transient failure is necessary here

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/util/TestUtil.java
kolea2: FWIW, we have something similar in the veneer client: https://github.com/googleapis/java-bigtable/blob/master/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/FakeServiceHelper.java

shantstepanian: Cool, good to have a similar reference. I'd suggest changing that to have retries too, but I don't _think_ that codebase is used in our integration tests so the flakes are less prominent (though I think it would be fine to add still)

bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/util/TestUtil.java
shantstepanian: Fixed on the new update

bigtable-test/bigtable-internal-test-helper/pom.xml
```diff
@@ -0,0 +1,107 @@
...
+
+    <artifactId>bigtable-internal-test-helper</artifactId>
+    <description>
+        Module with helper code for Bigtable unit and integration tests.
+        Not intended to be shared with external users.
+    </description>
+
+    <properties>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
```
igorbernstein2, line 20: unfortunately bigtable-client-core has to support java7, and since this is a dep of it, you are stuck with java7

bigtable-test/bigtable-internal-test-helper/src/main/java/com/google/cloud/bigtable/test/helper/TestServerBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+        server.start();
+        return server;
+      } catch (IOException e) {
+        server.shutdown();
+        lastError = e;
+      }
+    }
+    throw lastError;
+  }
+}
```
igorbernstein2, line 128: newline

igorbernstein2 (changes requested):

bigtable-test/bigtable-internal-test-helper/src/main/java/com/google/cloud/bigtable/test/helper/TestServerBuilder.java
shantstepanian: Fixed

igorbernstein2 (approved):
lgtm

PR #3053, opened 2021-07-02
fix: veneer adapter batching

igorbernstein2:
There was a race condition of the BulkReadWrapper being cleaned up before the batch executor was done with it. This PR fixes the issue by making lifecycle management more explicit  by scoping batch executors to a single  method(List) invocation and making close explicit. This will be further cleaned up once  https://github.com/googleapis/gax-java/pull/1423 lands

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
@@ -112,7 +112,6 @@ public abstract class AbstractBigtableTable implements Table {
   protected final HBaseRequestAdapter hbaseAdapter;
 
   protected final DataClientWrapper clientWrapper;
-  private BatchExecutor batchExecutor;
   protected final AbstractBigtableConnection bigtableConnection;
   private TableMetrics metrics = new TableMetrics();
 
@@ -183,7 +182,9 @@ public abstract class AbstractBigtableTable implements Table {
       for (Get get : gets) {
         existGets.add(GetAdapter.setCheckExistenceOnly(get));
       }
-      return getBatchExecutor().exists(existGets);
+      try (BatchExecutor executor = createBatchExecutor()) {
+        return executor.exists(existGets);
+      }
     }
   }
 
@@ -194,7 +195,9 @@ public abstract class AbstractBigtableTable implements Table {
     LOG.trace("batch(List<>, Object[])");
     try (Scope scope = TRACER.spanBuilder("BigtableTable.batch").startScopedSpan()) {
       addBatchSizeAnnotation(actions);
-      getBatchExecutor().batch(actions, results);
+      try (BatchExecutor executor = createBatchExecutor()) {
+        executor.batch(actions, results);
+      }
     }
   }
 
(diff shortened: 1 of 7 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BulkReadClassicApi.java
```diff
@@ -44,6 +44,12 @@
     this.tableId = tableId;
   }
 
+  @Override
+  public void close() {
+    isClosed = true;
+    sendOutstanding();
```
mutianf, line 50: Do we need to check
`if (isClosed) return` ? 

Should the order be:

sendOutstanding();
isClosed = true; ?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BulkReadVeneerApi.java
```diff
@@ -77,13 +79,22 @@ public Thread newThread(Runnable r) {
...
+    // TODO: use closeAsync after https://github.com/googleapis/gax-java/pull/1423 is available
+    if (!isClosed.compareAndSet(false, true)) {
+      return;
+    }
+    notifyArrival();
   }
 
   @Override
   public ApiFuture<Result> add(ByteString rowKey, @Nullable Filters.Filter filter) {
+    Preconditions.checkState(!isClosed.get());
```
mutianf, line 96: Maybe also add the error message here? "can't add request when the bulk read is closed."

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
```diff
@@ -183,7 +182,9 @@ public boolean exists(Get get) throws IOException {
       for (Get get : gets) {
         existGets.add(GetAdapter.setCheckExistenceOnly(get));
       }
-      return getBatchExecutor().exists(existGets);
+      try (BatchExecutor executor = createBatchExecutor()) {
```
mutianf, line 185: Will it be expensive if everytime a method is called it'll create and close a batcher?

mutianf (changes requested):

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/classic/BulkReadClassicApi.java
igorbernstein2: Practically I dont think it matters in this case because add() & close() can be called only from a single thread. However, logically I would want to prevent addition of elements first and then flush the buffered ones

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BulkReadVeneerApi.java
igorbernstein2: done

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
igorbernstein2: I dont think so. It just allocates a handful of objects. I dont think it will be noticeable

mutianf (approved):

PR #3061, opened 2021-07-12
chore: use a single version for all modules

igorbernstein2:


kolea2 (commented):
As part of this, could we remove any references to `${project.version}`? Would help align and be more consistent IMO

kolea2 (approved):

PR #3151, opened 2021-07-13
feat: backport features from main branch

igorbernstein2:
Api changes:
* expose subsettings on BigtableOptionsBuilder

Default changes:
* increase batch mutations to 20 MiB
* decrease batch mutation element count 125 -> 100
* enable 1s autoflush by default
* hardcode flow control of outstanding bytesize to 100 MiB (instead of 10% of memory)
* decrease short timeout: 60s -> 20s
* increase partial row timeout 1 min -> 5 mins
* increase initial exponential backoff 5ms -> 10ms
* increase exponential backoff multiplier 1.5 -> 2
* increase long timeout to 10 mins
* increase # of attempts: 3 -> 10

Test changes:
* fix random tests
* remove some stale gcj tests
* migrates tests from jackson to gson


bigtable-client-core-parent/bigtable-client-core/pom.xml
```diff
@@ -273,30 +266,24 @@ this is not used directly, but upgrading due to transitive vulnerabilities in ol
     </resources>
     <plugins>
       <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-surefire-plugin</artifactId>
+        <!-- skip for bigtable-client-core to allow for upcoming refactoring-->
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>clirr-maven-plugin</artifactId>
```
kolea2, line 271: was anything actually failing here? Or are we just doing this to be consistent?

igorbernstein2: just consistency

kolea2 (approved):

bigtable-client-core-parent/bigtable-client-core/pom.xml
kolea2: should we remove it? Feels like a good check to keep

PR #2999, opened 2021-05-26
chore(test): optimize beam import tests

igorbernstein2:
This should significantly reduce test execution time. However it does have a negative repercussion of the log being buffered until all of the tests finish. This cuts the runtime by 50%

* run all tests in parallel threads
* pre-stage files to ensure that parallel threads aren't clobbering each other
* fix report paths
* extract all system props into a common helper
* stop using surefire for integration tests
* rename staging dir prop to be a workdir prop which can contain other ephemeral files
* allow dataflow runner to log so that job urls are logged
* update rad/write test to run in parallel and lower # of rows

PR #3177, opened 2021-08-03
chore(test): optimize beam import tests (#2999)

kolea2:
* chore(test): optimize beam import tests

This should significantly reduce test execution time. However it does have a negative repercussion of the log being buffered until all of the tests finish

* run all tests in parallel threads
* pre-stage files to ensure that parallel threads aren't clobbering each other
* fix report paths
* extract all system props into a common helper
* stop using surefire for integration tests
* rename staging dir prop to be a workdir prop which can contain other ephemeral files
* allow dataflow runner to log so that job urls are logged

* fix deps

* format

* npe

* optimize CloudBigtableBeamIT

- allow read & write parts to run in parallel
- lower row count

* codestyle

* address feedback

(cherry picked from commit 6c3b151bc69349ff30deeb7a6c47770157940b58)


igorbernstein2 (approved):
thank you soooooo much for back porting this!

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
```diff
@@ -184,6 +184,19 @@ limitations under the License.
       <version>v1-rev171-1.25.0</version>
     </dependency>
 
+    <!-- Used transitively by beam, and directly in tests -->
+    <dependency>
+      <groupId>org.apache.beam</groupId>
+      <artifactId>beam-runners-core-construction-java</artifactId>
+      <version>2.24.0</version>
```
igorbernstein2, line 191: ${beam.version}

PR #3189, opened 2021-08-17
feat!: migrate to java 8

kolea2:
Removed all compiler plugins as it should now reference the shared config (updated to 1.0.0).

Updated bigtable version to JDK 8 version

Misc changes for enforcer


igorbernstein2 (approved):
lgtm

PR #3217, opened 2021-09-16
chore: clean up classic wrapper for beta release

mutianf:


kolea2 (commented):
is there any pom cleanup (deps on bigtable-client-core)?

mutianf:
> is there any pom cleanup (deps on bigtable-client-core)?

Yeah, I think this: https://github.com/googleapis/java-bigtable-hbase/blob/master/bigtable-client-core-parent/bigtable-hbase/pom.xml#L55-L59 needs to be cleaned up, but there is still code in VeneerWrapper depending on it, I'm looking into it..

igorbernstein2 (approved):
This looks like a great start!
We will need to follow up by cutting the dependency on bigtable-client-core as well, which will bring up a couple of issues that we'll need to address

PR #3220, opened 2021-09-20
Skeleton for the mirroring client

dopiera:
This PR introduces configuration, a `MirroringTable` with a simplistic implementation of some requests and an `AsyncTableWrapper`, which enables that implementation.

### DIFF
--- a/bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/pom.xml
+++ b/bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/pom.xml
@@ -62,5 +62,11 @@ limitations under the License.
       <version>1.10.4</version>
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>3.8.0</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 </project>
--- a/bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/pom.xml
+++ b/bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/pom.xml
@@ -1,0 +1,66 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Copyright 2021 Google LCC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>com.google.cloud.bigtable</groupId>
+    <artifactId>bigtable-hbase-mirroring-client-1.x-parent</artifactId>
+    <version>0.0.1-alpha1-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>bigtable-hbase-mirroring-client-1.x</artifactId>
+  <packaging>jar</packaging>
+  <name>${project.groupId}:${project.artifactId}</name>
+  <version>0.0.1-alpha1-SNAPSHOT</version>
+  <description>
+  </description>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.hbase</groupId>
+      <artifactId>hbase-shaded-client</artifactId>
+      <version>${hbase1.version}</version>
+    </dependency>
+
+    <!-- Test dependencies-->
+    <dependency>
+      <groupId>io.dropwizard.metrics</groupId>
+      <artifactId>metrics-graphite</artifactId>
+      <version>${hbase1-metrics.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>${junit.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.truth</groupId>
+      <artifactId>truth</artifactId>
+      <version>1.1.2</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.api</groupId>
+      <artifactId>api-common</artifactId>
+      <version>1.10.4</version>
+      <scope>compile</scope>
+    </dependency>
+  </dependencies>
+</project>
--- a/bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/src/main/java/com/google/cloud/bigtable/mirroring/hbase1_x/AsyncTableWrapper.java
+++ b/bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/src/main/java/com/google/cloud/bigtable/mirroring/hbase1_x/AsyncTableWrapper.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.mirroring.hbase1_x;
+
+import com.google.api.core.InternalApi;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListeningExecutorService;
+import java.util.List;
+import java.util.concurrent.Callable;
+import org.apache.hadoop.hbase.client.Get;
+import org.apache.hadoop.hbase.client.Result;
+import org.apache.hadoop.hbase.client.Table;
+
+/**
+ * MirroringClient verifies consistency between two databases asynchronously - after the results are
+ * delivered to the user. HBase Table object does not have an synchronous API, so we simulate it by
+ * wrapping the regular Table into AsyncTableWrapper.
+ *
+ * <p>Table instances are not thread-safe, every operation is synchronized to prevent concurrent
+ * accesses to the table from different threads in the executor.
+ */
+@InternalApi("For internal usage only")
+public class AsyncTableWrapper {
+  private final Table table;
+  private final ListeningExecutorService executorService;
+
+  public AsyncTableWrapper(Table table, ListeningExecutorService executorService) {
+    this.table = table;
+    this.executorService = executorService;
+  }
+
+  public ListenableFuture<Result> get(final Get gets) {
+    return this.executorService.submit(
+        new Callable<Result>() {
+          @Override
+          public Result call() throws Exception {
+            synchronized (table) {
+              return table.get(gets);
+            }
+          }
+        });
+  }
+
+  public ListenableFuture<Result[]> get(final List<Get> gets) {
+    return this.executorService.submit(
+        new Callable<Result[]>() {
+          @Override
+          public Result[] call() throws Exception {
+            synchronized (table) {
+              return table.get(gets);
+            }
+          }
+        });
+  }
+
+  public ListenableFuture<Boolean> exists(final Get get) {
+    return this.executorService.submit(
+        new Callable<Boolean>() {
+          @Override
+          public Boolean call() throws Exception {
+            synchronized (table) {
+              return table.exists(get);
+            }
+          }
+        });
+  }
+
+  public ListenableFuture<boolean[]> existsAll(final List<Get> gets) {
+    return this.executorService.submit(
+        new Callable<boolean[]>() {
+          @Override
+          public boolean[] call() throws Exception {
+            synchronized (table) {
+              return table.existsAll(gets);
+            }
+          }
+        });
+  }
+}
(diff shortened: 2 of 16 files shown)


dopiera:
@googlebot I fixed it.

dopiera:
@googlebot I fixed it.

mwalkiewicz:
@googlebot I consent.

bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>com.google.cloud.bigtable</groupId>
+    <artifactId>bigtable-hbase-mirroring-client-1.x-parent</artifactId>
+    <version>0.0.1-alpha1-SNAPSHOT</version>
```
vermas2012, line 23: Lets keep the versioning consistent with the other packages. Typically, we have the same version number for all the artifacts in this repo.

bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/src/main/java/com/google/cloud/bigtable/mirroring/hbase1_x/AsyncTableWrapper.java
```diff
@@ -0,0 +1,92 @@
...
+import com.google.common.util.concurrent.ListeningExecutorService;
+import java.util.List;
+import java.util.concurrent.Callable;
+import org.apache.hadoop.hbase.client.Get;
+import org.apache.hadoop.hbase.client.Result;
+import org.apache.hadoop.hbase.client.Table;
+
+/**
+ * MirroringClient verifies consistency between two databases asynchronously - after the results are
+ * delivered to the user. HBase Table object does not have an synchronous API, so we simulate it by
```
vermas2012, line 29: you mean async?

bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/src/main/java/com/google/cloud/bigtable/mirroring/hbase1_x/MirroringConfiguration.java
```diff
@@ -0,0 +1,157 @@
...
+   * <p>Prefixes should not contain dot at the end.
+   */
+  public static final String MIRRORING_PRIMARY_CONFIG_PREFIX_KEY =
+      "google.bigtable.mirroring.primary-client.prefix";
+
+  /**
+   * If this key is set, then only parameters that start with given prefix are passed to secondary
+   * Connection.
+   */
+  public static final String MIRRORING_SECONDARY_CONFIG_PREFIX_KEY =
```
vermas2012, line 60: I am not sure I understand the user experience here. Lets say we have an HBase primary and a Bigtable secondary (most common setup), then all the timeouts etc will be common for both, HBase will require zookeeper config while Bigtable will require projectId/InstanceId.  In such scenario you will set all the configs without any prefix (primary or secondary) set.  Each connection impl will take what it wants from the config and everything will work. 

However, if say, we had to setup different timeout values, then we need to specify a prefix (either on primary or secondary), Does that mean that we have to copy all the configurations on that database? A full copy of un-prefixed configs and then a prefixed version with all the configs (including some that will differ)?

bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/src/main/java/com/google/cloud/bigtable/mirroring/hbase1_x/MirroringConfiguration.java
```diff
@@ -0,0 +1,157 @@
...
+
+  private Configuration constructSecondaryConfiguration(Configuration conf) {
+    return constructConnectionConfiguration(
+        conf, MIRRORING_SECONDARY_CONNECTION_CLASS_KEY, MIRRORING_SECONDARY_CONFIG_PREFIX_KEY);
+  }
+
+  private Configuration constructConnectionConfiguration(
+      Configuration conf, String connectionClassKey, String prefixKey) {
+    String connectionClassName = conf.get(connectionClassKey);
+    String prefix = conf.get(prefixKey, "");
```
vermas2012, line 103: Why do this instead of hardcoding primary/secondary as the prefix? 

Are you hoping this will help with swapping primary and secondary? That this will remain hbase/bigtable and you will only change the prefix keys to swap the primary database?

bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/src/main/java/com/google/cloud/bigtable/mirroring/hbase1_x/MirroringConfiguration.java
```diff
@@ -0,0 +1,157 @@
...
+        throw new IllegalArgumentException(
+            String.format(
+                "Mirroring connections using the same client class requires a separate "
+                    + "configuration for one of them. Specify either %s or %s and use its value "
+                    + "as a prefix for configuration options.",
+                MIRRORING_PRIMARY_CONFIG_PREFIX_KEY, MIRRORING_SECONDARY_CONFIG_PREFIX_KEY));
+      } else {
+        throw new IllegalArgumentException(
+            String.format(
+                "Values of %s and %s should be different.",
```
vermas2012, line 134: Can we use more descriptive message here and explain why we are demanding it to be different?

bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/src/main/java/com/google/cloud/bigtable/mirroring/hbase1_x/MirroringConnection.java
```diff
@@ -0,0 +1,109 @@
...
+   * The constructor called from {@link
+   * org.apache.hadoop.hbase.client.ConnectionFactory#createConnection(Configuration)} and in its
+   * many forms via reflection with this specific signature.
+   *
+   * <p>Parameters are passed down to ConnectionFactory#createConnection method, connection errors
+   * are passed back to the user.
+   */
+  public MirroringConnection(Configuration conf, boolean managed, ExecutorService pool, User user)
+      throws IOException {
+    assert !managed; // This is always-false legacy hbase parameter.
```
vermas2012, line 46: When will it be true, or rather what is the behavior when this is true that we are asserting here? 

Also, should we stick with Preconditions for all validation?

bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/src/main/java/com/google/cloud/bigtable/mirroring/hbase1_x/utils/Comparators.java
```diff
@@ -0,0 +1,54 @@
...
+    Cell[] cells2 = result2.rawCells();
+    if (cells1.length != cells2.length) {
+      return false;
+    }
+    for (int i = 0; i < cells1.length; i++) {
+      int cellResult = CellComparator.compare(cells1[i], cells2[i], true);
+      if (cellResult != 0) {
+        return false;
+      }
+      if (!CellUtil.matchingValue(cells1[i], cells2[i])) {
```
vermas2012, line 48: Does CellComparator.compare not compare values?

vermas2012 (commented):
Reviewed the code first.  Will review the tests later tomorrow.

bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/pom.xml
mwalkiewicz: Done in next PRs.

bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/src/main/java/com/google/cloud/bigtable/mirroring/hbase1_x/AsyncTableWrapper.java
mwalkiewicz: Yes, will fix in future PR.

bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/src/main/java/com/google/cloud/bigtable/mirroring/hbase1_x/AsyncTableWrapper.java
```diff
@@ -0,0 +1,92 @@
...
+    this.executorService = executorService;
+  }
+
+  public ListenableFuture<Result> get(final Get gets) {
+    return this.executorService.submit(
+        new Callable<Result>() {
+          @Override
+          public Result call() throws Exception {
+            synchronized (table) {
+              return table.get(gets);
```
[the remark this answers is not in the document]
mwalkiewicz: The `table` object is not thread safe, and we issue a single asynchronous operation that is synchronized on this `table` object per single blocking client operation. If the secondary database has the same performance as the primary, then we think that is shouldn't really matter, because this synchronized operation will more-or-less overlap with next synchronous operation issued by the user thread. If it is not the case we will hit flow controller limit anyways.

bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/src/main/java/com/google/cloud/bigtable/mirroring/hbase1_x/MirroringConfiguration.java
mwalkiewicz: Yes, as long as all the keys that are common for both clients are the same then no prefix is needed. If The user would want to specify separate configuration for both the databases, we want them to make a copy and keep a prefix. We wanted the user to be explicit about which database is configured to prevent accidental misconfiguration.

bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/src/main/java/com/google/cloud/bigtable/mirroring/hbase1_x/MirroringConfiguration.java
```diff
@@ -0,0 +1,157 @@
...
+      "google.bigtable.mirroring.secondary-client.prefix";
+
+  public MirroringConfiguration(
+      Configuration primaryConfiguration,
+      Configuration secondaryConfiguration,
+      Configuration mirroringConfiguration) {
+    super.set("hbase.client.connection.impl", MirroringConnection.class.getCanonicalName());
+    this.primaryConfiguration = primaryConfiguration;
+    this.secondaryConfiguration = secondaryConfiguration;
+    this.mirroringOptions = new MirroringOptions(mirroringConfiguration);
```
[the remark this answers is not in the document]
mwalkiewicz: This constructor is only for tests, should be package-private and marked as `@VisibleForTesting` or maybe even removed in future PRs.

bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/src/main/java/com/google/cloud/bigtable/mirroring/hbase1_x/MirroringConfiguration.java
mwalkiewicz: This is exactly the reason why we have done it. We could have used `bigtable.` and `hbase.` prefixes if we hardcoded them, but we wanted to remain agnostic to underlying implementations and not to make any assumptions that only those two client will be used.

bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/src/main/java/com/google/cloud/bigtable/mirroring/hbase1_x/MirroringConfiguration.java
mwalkiewicz: OK, I've created an issue for this.

bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/src/main/java/com/google/cloud/bigtable/mirroring/hbase1_x/MirroringConnection.java
mwalkiewicz: In HBase > 1.0 it won't be true afaik, because managed connections were removed from the API. Bigtable HBase client also asserts that managed is false.

> Also, should we stick with Preconditions for all validation?

ok, I'll do in in next PRs.

bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/src/main/java/com/google/cloud/bigtable/mirroring/hbase1_x/utils/Comparators.java
mwalkiewicz: According to https://hbase.apache.org/devapidocs/org/apache/hadoop/hbase/CellComparatorImpl.html#compare-org.apache.hadoop.hbase.Cell-org.apache.hadoop.hbase.Cell- it does compare `The key part of the cell is taken for comparison which includes row, family, qualifier, timestamp and type`.

bigtable-hbase-mirroring-client-1.x-parent/pom.xml
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Copyright 2021 Google LCC
```
kolea2, line 3: please double check headers - looks like a typo?

kolea2 (changes requested):

bigtable-hbase-mirroring-client-1.x-parent/pom.xml
kolea2: https://github.com/GoogleCloudPlatform/java-docs-samples#source-code-headers

bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x/pom.xml
```diff
@@ -0,0 +1,72 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Copyright 2021 Google LCC
```
kolea2, line 3: same comment here - please check all headers. https://github.com/GoogleCloudPlatform/java-docs-samples#source-code-headers

kolea2 (changes requested):

PR #3267, opened 2021-09-28
fix: Clean up RowResultAdapter

mutianf:
`RowResult` class will cause problems when mapping `Result` elements. Remove `RowResult` class and create / check for marker rows directly.

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/RowResultAdapter.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/RowResultAdapter.java
@@ -46,6 +46,9 @@ import org.apache.hadoop.hbase.filter.FilterList.Operator;
  * on duplicate labelled cells for its implementation. So this adapter will not deduplicate labelled
  * cells.
  *
+ * <p>This adapter will also return and check for scan marker rows, which will be an empty row with
+ * the scan marker row label.
+ *
  * <p>For internal use only - public for technical reasons.
  */
 @InternalApi("For internal usage only")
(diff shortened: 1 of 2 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/RowResultAdapter.java
```diff
@@ -61,45 +61,23 @@
 
   @Override
   public boolean isScanMarkerRow(Result result) {
-    Preconditions.checkState(result instanceof RowResult, "Should be an instance of RowResult");
-    return ((RowResult) result).isMarkerRow();
+    if (result.listCells().size() != 1) {
+      return false;
+    }
+    Preconditions.checkState(
```
igorbernstein2, line 67: Should this just return false?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/RowResultAdapter.java
```diff
@@ -61,45 +61,23 @@
...
+        result.listCells().get(0) instanceof RowCell, "Should be instance of RowCell");
+
+    RowCell cell = (RowCell) result.listCells().get(0);
```
igorbernstein2, line 70: I think you can avoid an unneeded allocation by using rawCells()

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/RowResultAdapter.java
```diff
@@ -61,45 +61,23 @@
...
-    }
-
-    ByteString getKey() {
-      return rowKey;
-    }
-
-    boolean isMarkerRow() {
-      return isMarkerRow;
-    }
+    return ByteStringer.wrap(result.getRow());
```
igorbernstein2, line 80: getRow will cause an unnecessary copy, instead just grab it from the first cell

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/RowResultAdapter.java
```diff
@@ -53,6 +52,7 @@
 public class RowResultAdapter implements RowAdapter<Result> {
```
igorbernstein2, line 52: Please update the javadoc to include a description of marker rows

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/RowResultAdapter.java
```diff
@@ -47,59 +46,43 @@
...
-    }
-
-    ByteString getKey() {
-      return rowKey;
-    }
-
-    boolean isMarkerRow() {
-      return isMarkerRow;
-    }
+    return ByteStringer.wrap(result.rawCells()[0].getRowArray());
```
igorbernstein2, line 85: I think this makes an assumption that is not in line with the contract of a Cell. You are assuming that getRowOffset() == 0 and getRowLength() == getRowArray().length. The RowCell implementation aligns with this, but there is nothing in this method to ensure that implementation. I think it would be a lot safer to this instead:
```
Cell c = result.rawCells()[0];
ByteString.wrap(c.getRowArray(), c.getRowOffset(), c.getRowLength()
```

igorbernstein2 (approved):
lgtm, but please update the key extraction

PR #3270, opened 2021-09-30
deps: upgrade hbase1 to 1.7.1

kolea2:


pom.xml
```diff
@@ -221,22 +221,6 @@ limitations under the License.
       </plugins>
     </pluginManagement>
     <plugins>
-      <plugin>
```
kolea2, line 224: unrelated to this PR but removed to reduce noise - we have a separate lint check now, don't need to include this in the main build path

mutianf (approved):

PR #3271, opened 2021-09-30
chore: deprecate asyncConfigure

mutianf:

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConfiguration.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConfiguration.java
@@ -48,13 +48,19 @@ public class BigtableConfiguration {
   public static final String BIGTABLE_HBASE_CLIENT_ASYNC_REGISTRY_CLASS =
       "org.apache.hadoop.hbase.client.BigtableAsyncRegistry";
 
+  private static final String HBASE_ASYNC_CONNECTION_CLASS =
+      "org.apache.hadoop.hbase.client.AsyncConnection";
+
   private static final String[] CONNECTION_CLASS_NAMES = {
     "com.google.cloud.bigtable.hbase1_x.BigtableConnection",
     "com.google.cloud.bigtable.hbase2_x.BigtableConnection",
   };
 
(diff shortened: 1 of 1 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConfiguration.java
```diff
@@ -81,6 +84,25 @@
...
 
+  /**
+   * Set up connection impl classes. If org.apache.hadoop.hbase.client.AsyncConnection exists, set
+   * up async connection impl class as well.
+   */
+  private static Configuration configureConnectionClass(Configuration configuration) {
+    configuration.set(HBASE_CLIENT_CONNECTION_IMPL, getConnectionClass().getCanonicalName());
+    try {
+      // Set up HBase async registry impl if AsyncConnection class exists
+      Class.forName(HBASE_ASYNC_CONNECTION_CLASS);
```
igorbernstein2, line 95: I think you should check for the classes that you actually want to use (BIGTABLE_HBASE_CLIENT_ASYNC_CONNECTION_CLASS & BIGTABLE_HBASE_CLIENT_ASYNC_REGISTRY_CLASS) instead of which hbase classes exist.

Or maybe check for all 3?

We can end up in a situation where the application has a mismatch of hbase and bigtable-hbase versions. ie the classpath can look like: hbase-client 1.x, bigtable-hbase-1x, hbase-client 2.x.
Since the classpath is searched in order, this might have worked before this change, but now it will crash.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableConfiguration.java
```diff
@@ -81,6 +84,25 @@
     return CONNECTION_CLASS;
   }
 
+  /**
+   * Set up connection impl classes. If org.apache.hadoop.hbase.client.AsyncConnection exists, set
+   * up async connection impl class as well.
+   */
+  private static Configuration configureConnectionClass(Configuration configuration) {
```
igorbernstein2, line 91: I think `injectBigtableImpls()` might be more precise

igorbernstein2 (approved):

PR #3326, opened 2021-10-28
deps: migrate to log4j-core

kolea2:


vermas2012 (commented):
How have you tested this change?

vermas2012 (approved):
Overall changes look good. Approved, provided we have tested the logging for the tool.

kolea2:
> Overall changes look good. Approved, provided we have tested the logging for the tool.

Thanks @vermas2012 for taking a look! I tried some testing and realized I had to make a few additional fixes. I was able to successfully get this to work by running the unit tests with the log4j2.xml file uncommented. I did change this to STDOUT as the output - is this ok?

vermas2012:
STDout is fine. I wonder if we should test logging by running the actual tool?

vermas2012:
> STDout is fine. I wonder if we should test logging by running the actual tool?

You don't have to setup HBase, you can use this json file as a source for schema
[schema-2.json.zip](https://github.com/googleapis/java-bigtable-hbase/files/7549676/schema-2.json.zip)
.

PR #3338, opened 2021-11-18
fix: fix flow controller setting and option javadocs

mutianf:
- Fix flow controller setting
- Update javadocs in `BigtableOptionsFactory` to use batch mutation / read instead of long mutation / read


igorbernstein2 (approved):
lgtm

PR #3380, opened 2021-12-13
deps(fix): update to 2.16.0 log4j version and ban all 2.x versions which are <= 2.15.0

mutianf:

### DIFF
--- a/bigtable-dataflow-parent/bigtable-beam-import/pom.xml
+++ b/bigtable-dataflow-parent/bigtable-beam-import/pom.xml
@@ -26,6 +26,7 @@ limitations under the License.
 
   <properties>
     <mainClass>com.google.cloud.bigtable.beam.Main</mainClass>
+    <log4j2.version>2.15.0</log4j2.version>
   </properties>
 
   <!-- Adding this to resolve version conflict within beam sdk-->
@@ -37,6 +38,11 @@ limitations under the License.
         <version>${beam-grpc.version}</version>
         <type>pom</type>
         <scope>import</scope>
+      </dependency>
+      <dependency>
+        <groupId>org.apache.logging.log4j</groupId>
+        <artifactId>log4j-api</artifactId>
+        <version>${log4j2.version}</version>
       </dependency>
     </dependencies>
   </dependencyManagement>
--- a/bigtable-dataflow-parent/bigtable-beam-import/pom.xml
+++ b/bigtable-dataflow-parent/bigtable-beam-import/pom.xml
@@ -26,7 +26,6 @@ limitations under the License.
 
   <properties>
     <mainClass>com.google.cloud.bigtable.beam.Main</mainClass>
-    <log4j2.version>2.15.0</log4j2.version>
   </properties>
 
   <!-- Adding this to resolve version conflict within beam sdk-->
@@ -38,11 +37,6 @@ limitations under the License.
         <version>${beam-grpc.version}</version>
         <type>pom</type>
         <scope>import</scope>
-      </dependency>
-      <dependency>
-        <groupId>org.apache.logging.log4j</groupId>
-        <artifactId>log4j-api</artifactId>
-        <version>${log4j2.version}</version>
       </dependency>
     </dependencies>
   </dependencyManagement>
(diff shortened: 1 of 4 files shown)


igorbernstein2 (approved):
lgtm

pom.xml
```diff
@@ -266,6 +266,28 @@ limitations under the License.
...
+            <id>enforce-banned-deps</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
+              <rules>
+                <!-- ban all log4j 2.x deps with CVEs -->
+                <bannedDependencies>
+                  <excludes>
+                    <exclude>org.apache.logging.log4j:*:[2.0-beta9, 2.14.1]</exclude>
```
kolea2, line 283: does this also include pre beta9?

kolea2: should this be log4j-api, not log4j?

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
```diff
@@ -38,6 +39,11 @@ limitations under the License.
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
+        <groupId>org.apache.logging.log4j</groupId>
+        <artifactId>log4j-api</artifactId>
```
kolea2, line 44: what about log4j-core?

pom.xml
```diff
@@ -266,6 +266,28 @@ limitations under the License.
           <ignoreNonCompile>true</ignoreNonCompile>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
```
kolea2, line 271: is this overriding the configuration in the parent pom?

mutianf: The checks in parent pom should still be propagated unless we specifically disable it according to https://maven.apache.org/guides/mini/guide-configuring-plugins.html#using-the-inherited-tag-in-build-plugins

I also tested by  removing https://github.com/googleapis/java-bigtable-hbase/blob/bigtable-1.x/bigtable-dataflow-parent/bigtable-beam-import/pom.xml#L259 and got the dependency errors

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
mutianf: I think it only has log4j-api dependency here, log4j-api should be included in the banned dependencies

pom.xml
mutianf: it's excluding `org.apache.logging.log4j:*` so should include all the artifacts?

pom.xml
kolea2: excellent, thank you :)

pom.xml
kolea2: should it be `log4j*` rather than `log4j:*`? Sorry I'm not sure how this field works, just want to make sure!

mutianf: I think log4j:* should work, example here: https://maven.apache.org/enforcer/enforcer-rules/bannedDependencies.html

I also tested by not updating the version to 2.15 and the build failed

bigtable-dataflow-parent/bigtable-hbase-beam/pom.xml
```diff
@@ -27,6 +27,20 @@ limitations under the License.
...
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.apache.logging.log4j</groupId>
+        <artifactId>log4j-api</artifactId>
+        <version>${log4j2.version}</version>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
```
igorbernstein2, line 42: This needs to be listed as a direct dependency. DependencyManagement doesn't work for libraries

pom.xml
```diff
@@ -63,6 +63,8 @@ limitations under the License.
     <slf4j.version>1.7.25</slf4j.version>
     <commons-logging.version>1.2</commons-logging.version>
     <jsr305.version>3.0.2</jsr305.version>
+    <log4j2.version>2.15.0</log4j2.version>
```
kolea2, line 66: you can probably clean up this same property in the tools module and reference this instead.

kolea2 (changes requested):

igorbernstein2 (approved):

pom.xml
mutianf: done!

kolea2 (approved):

pom.xml
```diff
@@ -266,6 +267,28 @@ limitations under the License.
...
+            <id>enforce-banned-deps</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
+              <rules>
+                <!-- ban all log4j 2.x deps with CVEs -->
+                <bannedDependencies>
+                  <excludes>
+                    <exclude>org.apache.logging.log4j:*:[2.0-alpha1,2.14.1]</exclude>
```
kolea2, line 284: I think we should now increase this to 2.15.0 and only allow 2.16.0 +

kolea2 (changes requested):

kolea2 (approved):

PR #3394, opened 2021-12-15
deps: remove explicit log4j2 deps on beam artifacts as they are no longer brought in transitively

kolea2:


igorbernstein2 (approved):
lgtm

PR #3413, opened 2021-12-22
deps: remove explicit auto value dependency in bigtable-beam-import

kolea2:
Opening in favor of #3393


igorbernstein2:
I dont think you need to pin this

igorbernstein2 (approved):
lgtm

PR #3429, opened 2022-01-06
chore: upgrade hbase2 to 2.3.6

mutianf:


pom.xml
```diff
@@ -66,7 +66,7 @@ limitations under the License.
     <!-- hbase dependency versions -->
     <hbase1.version>1.7.1</hbase1.version>
     <hbase1-metrics.version>3.1.2</hbase1-metrics.version>
-    <hbase2.version>2.2.3</hbase2.version>
+    <hbase2.version>2.2.6</hbase2.version>
```
kolea2, line 69: is this the right version? PR title says 2.3.6?

mutianf: Ah yeah I made a mistake, should be 2.3.6

bigtable-client-core-parent/bigtable-hbase/pom.xml
```diff
@@ -168,6 +168,12 @@ limitations under the License.
       </exclusions>
     </dependency>
 
+    <dependency>
+      <groupId>net.bytebuddy</groupId>
+      <artifactId>byte-buddy</artifactId>
+      <version>1.12.6</version>
```
kolea2, line 174: nit - I think this is used now in more than one place? Can we extract the version number to a property?

bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/BigtableAsyncRegistry.java
```diff
@@ -0,0 +1,112 @@
+/*
+ * Copyright 2017 Google LLC
```
kolea2, line 2: 2022

bigtable-hbase-2.x-parent/bigtable-hbase-2.x-integration-tests/pom.xml
```diff
@@ -160,9 +160,9 @@ limitations under the License.
                   </includes>
                   <excludedGroups>KnownEmulatorGap,KnownGap</excludedGroups>
                   <systemPropertyVariables>
-                    <google.bigtable.test_env>bigtable</google.bigtable.test_env>
                     <google.bigtable.project.id>fake-project</google.bigtable.project.id>
                     <google.bigtable.instance.id>fake-instance</google.bigtable.instance.id>
+                    <google.bigtable.test_env>bigtable</google.bigtable.test_env>
```
kolea2, line 165: nit - can revert this?

bigtable-hbase-2.x-parent/bigtable-hbase-2.x-shaded/pom.xml
```diff
@@ -48,6 +48,16 @@ limitations under the License.
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
+        <groupId>org.slf4j</groupId>
+        <artifactId>slf4j-api</artifactId>
+        <version>1.7.30</version>
```
kolea2, line 54: can we use the variable here?

bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/BigtableAsyncRegistry.java
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * <p>This class is injected via the system property: "hbase.client.registry.impl". For further
+ * details See HConstants#CLIENT_CONNECTION_REGISTRY_IMPL_CONF_KEY,
+ * AsyncRegistry#createAsyncConnection() for hbase < 2.3 and
+ * ConnectionFactory#createAsyncConnection() for hbase >= 2.3.
+ *
+ * <p>For internal use only - public for technical reasons.
+ */
+@InternalApi("For internal usage only")
+public class BigtableAsyncRegistry {
```
kolea2, line 46: to confirm my understanding - this will work as a drop in replacement here? https://github.com/googleapis/java-bigtable-hbase/blob/main/bigtable-hbase-2.x-parent/bigtable-hbase-2.x-integration-tests/pom.xml#L38

kolea2 (approved):
LGTM after nits and questions addressed

bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/BigtableAsyncRegistry.java
mutianf: This is updated in this pr: https://github.com/googleapis/java-bigtable-hbase/pull/3452 , BigtableEnv will call BigtableConnection() to create the connection for integration tests, so we don't need to define `google.bigtable.registry.impl` anymore, and BigtableConnection will handle generating the class. I'll also clean up this property in the pom file.

bigtable-hbase-2.x-parent/bigtable-hbase-2.x-integration-tests/pom.xml
mutianf: This was updated in https://github.com/googleapis/java-bigtable-hbase/pull/3452/files to use bigtable/minicluster instead of connection.impl.. I think I might have missed this pom file

PR #3500, opened 2022-02-10
Remove hbase-server dependency from core replication library

sushanb:
Added an interface in core library and each hbase1.x and hbase2.x extend that interface.

Couple of question
1. Is there a way to remove the code duplication across hbase1.x and hbase2.x as the signatures for MetricsSource is same.

### DIFF
--- a/bigtable-hbase-1.x-replication/pom.xml
+++ b/bigtable-hbase-1.x-replication/pom.xml
@@ -147,39 +147,17 @@
     <dependencies>
         <!-- bigtable libs for org.apache.hadoop.hbase -->
         <dependency>
-            <groupId>com.google.cloud.bigtable</groupId>
-            <artifactId>bigtable-hbase-replication</artifactId>
-            <version>2.0.0-beta6</version>
-        </dependency>
-        <dependency>
-            <groupId>com.google.cloud.bigtable</groupId>
-            <artifactId>bigtable-hbase-1.x-hadoop</artifactId>
-            <version>${project.version}</version>
-            <scope>${scope}</scope>
-            <exclusions>
-                <!-- exclude libraries if providing as separate dependency-->
-                <exclusion>
-                    <groupId>org.apache.hadoop</groupId>
-                    <artifactId>*</artifactId>
-                </exclusion>
-                <exclusion>
-                    <groupId>org.apache.hbase</groupId>
-                    <artifactId>*</artifactId>
-                </exclusion>
-            </exclusions>
-        </dependency>
-        <!--  <dependency>-->
-        <!--    <groupId>com.google.cloud.bigtable</groupId>-->
-        <!--    <artifactId>bigtable-hbase-1.x-hadoop</artifactId>-->
-        <!--    <version>${project.version}</version>-->
-        <!--  </dependency>-->
-        <dependency>
             <groupId>org.apache.hbase</groupId>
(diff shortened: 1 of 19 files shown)


bigtable-hbase-1.x-replication/pom.xml
```diff
@@ -181,6 +154,11 @@
             <!--       use the HBase server from the HBase environment-->
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>com.google.cloud.bigtable</groupId>
+            <artifactId>bigtable-hbase-replication</artifactId>
+            <version>2.0.0-beta6</version>
```
vermas2012, line 160: we should not hardcode the version, it should be the repo version, use maven tags and speical comment to auto update this version with every release. Refer to other POMs

bigtable-hbase-1.x-replication/pom.xml
```diff
@@ -181,6 +154,11 @@
...
+        </dependency>
         <dependency>
             <groupId>com.google.protobuf</groupId>
```
vermas2012, line 163: Is it required to bring our own protobuf? If so, please shade it, we don't want our protobuf to conflict with hbase's version.

bigtable-hbase-1.x-replication/src/main/java/com/google/cloud/bigtable/hbase/replication/metrics/HBaseMetricsExporter.java
```diff
@@ -0,0 +1,20 @@
+package com.google.cloud.bigtable.hbase.replication.metrics;
+
+import org.apache.hadoop.hbase.replication.regionserver.MetricsSource;
+
+/**
+ * HBaseMetricsExporter implements MetricExporter which bridges with MetricsSource.
+ */
+public class HBaseMetricsExporter implements MetricsExporter {
+    // set this as static
+    private static MetricsSource metricsSource;
```
vermas2012, line 10: Why does this need to be static? I am not sure this is a good idea?  If hbase context's metric source is not static, please keep this as non-static. 

If you decide to keep it static, make sure its threadsafe. There can be more than one threads accessing this field.

bigtable-hbase-2.x-replication/src/main/java/com/google/cloud/bigtable/hbase/replication/HBaseMetricsExporter.java
```diff
@@ -0,0 +1,20 @@
+package com.google.cloud.bigtable.hbase.replication;
+
+import com.google.cloud.bigtable.hbase.replication.metrics.MetricsExporter;
+import org.apache.hadoop.hbase.replication.regionserver.MetricsSource;
+/**
+ * HBaseMetricsExporter implements MetricExporter which bridges with MetricsSource.
+ */
+public class HBaseMetricsExporter implements MetricsExporter {
+    // set this as static
```
vermas2012, line 9: Usually, its much more valuable to comment on "why" you kept it static. Describing the code usually not helpful.

vermas2012 (commented):
Please add tableName in BigtableWALEntry and uncomment 2.x tests. I will rather have tests with deprecated tools than now have them. This prevents us from working on buggy code, in the absence of 2.x tests. 

There were a few things that we left TODO for in previous review, have you covered them all?

PR #3503, opened 2022-02-10
Remove extraneous mvn deps from replication library

sushanb:


sushanb:
FYI @vermas2012

bigtable-hbase-replication/pom.xml
```diff
@@ -153,11 +153,6 @@
       <version>${project.version}</version>
```
vermas2012, line 153: Similary, use the project.version for hbase-replication-common in 1.x and 2.x package

PR #3521, opened 2022-02-24
feat: add future delete handling

sushanb:
Add incompatible mutation/dropped mutations for deletes at future timestamp


hbase-migration-tools/bigtable-hbase-replication/bigtable-hbase-replication-core/src/main/java/com/google/cloud/bigtable/hbase/replication/adapters/ApproximatingIncompatibleMutationAdapter.java
```diff
@@ -58,6 +59,20 @@ public ApproximatingIncompatibleMutationAdapter(
   protected List<Cell> adaptIncompatibleMutation(BigtableWALEntry walEntry, int index) {
     long walWriteTime = walEntry.getWalWriteTime();
     Cell cell = walEntry.getCells().get(index);
+    long currTime = EnvironmentEdgeManager.currentTime();
+
+    // test if delete is in future timestamp
+    if (CellUtil.isDelete(cell) && cell.getTimestamp() > currTime) {
```
vermas2012, line 65: This does not sound right, we should check with WalWriteTime. In case of large replication delay, currentTime and WAL write time will differ, we want to make sure that delete was issued in future.

vermas2012 (changes requested):

PR #3546, opened 2022-03-08
build: update renovate config to group veneer dependencies

kolea2:
Currently, when veneer has a release we're getting two renovate PRs for both the emulator and google-cloud-bigtable. This cleans up our renovate config to group google-cloud-bigtable and google-cloud-bigtable-emulator together.


igorbernstein2 (changes requested):
This will cause all of our artifacts to use protobuf/guava/etc versions from veneer. Which will break the -hadoop artifacts as they expose hadoop compatible versions of those libraries

kolea2:
> This will cause all of our artifacts to use protobuf/guava/etc versions from veneer. Which will break the -hadoop artifacts as they expose hadoop compatible versions of those libraries

@igorbernstein2 changed this to have renovate manage groups of dependencies together, PTAL

PR #3582, opened 2022-04-19
deps: upgrade guava to 31.1-jre and Bigtable to 2.6.2

prash-mi:
Updated guava to `31.1-jre` to avoid Maven's `RequireUpperBoundDeps` error
Updated Bigtable version to '2.6.2' (On behalf of https://github.com/googleapis/java-bigtable-hbase/pull/3541)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,7 +55,7 @@ limitations under the License.
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
     <!-- core dependency versions -->
-    <bigtable.version>2.5.3</bigtable.version>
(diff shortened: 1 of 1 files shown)


kolea2 (changes requested):
Thanks @prash-mi! Can we also upgrade the bigtable.version as part of this PR to make it clear why we're upgrading guava as well?

prash-mi:
> Thanks @prash-mi! Can we also upgrade the bigtable.version as part of this PR to make it clear why we're upgrading guava as well?

Sure updated the PR's desc and the Bigtable version

kolea2 (approved):

PR #3598, opened 2022-05-02
build: temporarily make clirr optional check, skip clirr on dependency check

kolea2:
To unblock https://github.com/googleapis/java-bigtable-hbase/pull/3592 which needs further investigation


igorbernstein2 (approved):
lgtm

PR #3600, opened 2022-05-02
fix: published dependencies of bigtable-beam-import

igorbernstein2:
Due to a bug in maven shade plugin 3.3.0,  compile dependencies got stripped from bigtable-beam-import. This bug only affects artifacts where shadedArtifactAttached=true.

Reference: https://issues.apache.org/jira/projects/MSHADE/issues/MSHADE-419

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -162,7 +162,8 @@ limitations under the License.
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-shade-plugin</artifactId>
-          <version>3.3.0</version>
+          <!-- MSHADE-419: 3.3.0 Shade plugin causes pom to be created without compile dependencies -->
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -162,7 +162,8 @@ limitations under the License.
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-shade-plugin</artifactId>
-          <version>3.3.0</version>
+          <!-- MSHADE-419: 3.3.0 Shade plugin causes pom to be created without compile dependencies -->
+          <version>3.2.4</version>
```
kolea2, line 166: please update the renovate config to ignore this so we don't accidentally bump it again

kolea2 (changes requested):

kolea2 (approved):

PR #3601, opened 2022-05-02
feat: log version and settings on startup

igorbernstein2:
Add some informative log messages about the client on startup. A similar feature existed in 1x, but got lost in the upgrade 2.0. Will print the artifact name, version and settings for each new connection


moo/pom.xml
```diff
@@ -0,0 +1,19 @@
...
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>bigtable-client-parent</artifactId>
+    <groupId>com.google.cloud.bigtable</groupId>
+    <version>2.1.1</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>moo</artifactId>
```
mutianf, line 12: moo? lol

mutianf (commented):
Please fix the pom file :P

moo/pom.xml
igorbernstein2: 😳

mutianf (approved):
lgtm

PR #3615, opened 2022-05-10
fix: updating google-cloud-bigtable-emulator to latest version for bigtable-beam-import tests

vermas2012:


igorbernstein2 (changes requested):
I'm not sure this is the correct way to do this.

The java-bigtable version is getting pulled via a chain of boms:
bigtable-import -> beam-sdks-java-google-cloud-platform-bom -> libraries-bom -> google-cloud-bom -> google-cloud-bigtable-bom

I think we want to stay compatible with beam's classpath (which is why we use bigtable-hbase-1x-shaded artifact). So I'm not sure if we want to piecemeal override parts of the classpath.

Can this be dealt with by upgrade beam?

PR #3626, opened 2022-05-18
WIP: Mirroring client wip

mwalkiewicz:


migration-tools/mirroring-client/bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x-hadoop/pom.xml
```diff
@@ -182,7 +182,7 @@ limitations under the License.
               <allowedJarClassEntries>
                 <classEntry>${shading-prefix-path}</classEntry>
                 <classEntry>com/google/cloud/bigtable/mirroring</classEntry>
-<!--                <classEntry>org/apache/htrace</classEntry>-->
+                <classEntry>org/apache/htrace</classEntry>
```
vermas2012, line 185: I think we should disable the htrace from being shaded, right? If its a dependency of hbase-shaded, why do we need to shade it?

mwalkiewicz: Hi, @vermas2012 currently I'm trying to run build on kokoro to verify if it works and build successfully at all and I'll try to run integration tests. Changes on this branch might or might not make any sense yet, I'll fix those issues locally.

And back to your question - yes, htrace shouldn't be shaded, and it is not shaded now - this execution verifies that output jar contains classes starting with one of `shading-prefix`, `com/google/.../mirroring` or `org/apache/htrace`. It means that `org/apache/htrace` is not shaded (original path is preserved) but is included in the jar.

I'm not sure now if this is correct or not, but we can either remove `htrace` classes from the jar or leave it as is - I'll verify which of those options matches original `bigtable-hbase-1.x-hadoop` later.

mwalkiewicz: New hbase-client version added `org.apache.htrace:htrace-core4` as a dependency, I've excluded it from `shade` goal.

migration-tools/mirroring-client/bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x-hadoop/pom.xml
```diff
@@ -0,0 +1,240 @@
...
+    <!-- define a property that can be ignored by renovate -->
+    <hbase1-hadoop-slf4j.version>1.6.1</hbase1-hadoop-slf4j.version>
+  </properties>
+
+  <dependencies>
+    <!-- Internal dependencies that will be shaded along with their transitive dependencies. -->
+    <dependency>
+      <groupId>com.google.cloud.bigtable</groupId>
+      <artifactId>bigtable-hbase-mirroring-client-1.x-shaded</artifactId>
+      <version>2.2.1-SNAPSHOT</version> <!-- {x-version-update:bigtable-client-parent:current} -->
```
kolea2, line 45: this (and everywhere else) should be 2.3.1-SNAPSHOT (I believe this will solve your current build issues)

igorbernstein2:
superceded by https://github.com/googleapis/java-bigtable-hbase/pull/3698

PR #3688, opened 2022-07-14
feat: mapreduce hashtable/synctable jobs

### LINKED ISSUE #3687: MapReduce hashtable/synctable jobs
Add MapReduce tools for validating data from source to target table between HBase and Bigtable and vice-versa.

* `hash-table` tool uses HBase HashTable
* `sync-table` tool uses HBase SyncTable
    * counters emitted from the job with no differences
      ```sh
      org.apache.hadoop.hbase.mapreduce.SyncTable$SyncMapper$Counter
         BATCHES=1
         HASHES_MATCHED=1
      ```

    * counters emitted from the job with differences
      ```sh
      org.apache.hadoop.hbase.mapreduce.SyncTable$SyncMapper$Counter
         BATCHES=1
         HASHES_NOT_MATCHED=1
         MATCHINGCELLS=3
         MATCHINGROWS=3
         RANGESNOTMATCHED=1
         ROWSWITHDIFFS=1
         SOURCEMISSINGCELLS=1
         TARGETMISSINGCELLS=1
      ```
    * this job expects connectivity to the source, for diverging hashes, SyncTable compares rows/cells and will emit details when run with debug level logging. i.e.
      ```sh
      022-07-12 23:48:13,561 DEBUG [main] org.apache.hadoop.hbase.mapreduce.SyncTable: Hash mismatch.  Key range:  to  sourceHash: 06c51394f67a083137b0cd0a2700ba4f targetHash: ef6383f1c69e7d296065c1f5e26e8e20
      2022-07-12 23:48:13,598 DEBUG [main] org.apache.hadoop.hbase.mapreduce.SyncTable: Target missing cell: row1#12346/cf1:dump/1657569926224/Put/vlen=9/seqid=0
      2022-07-12 23:48:13,598 DEBUG [main] org.apache.hadoop.hbase.mapreduce.SyncTable: Source missing cell: row1#12346/cf1:dump/1657229802265/Put/vlen=9/seqid=0
      ```

jhambleton:
add mapreduce `hash-table` and `sync-table` tools to use with validation of data in a source and a target. Tools inherit existing tool functionality and are extended to operate with Bigtable as a source or target. 

Fixes #3687 ☕️


jhambleton:
@vermas2012 - can you take a pass on this?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/README.md
```diff
@@ -79,17 +120,46 @@ from/to Cloud Bigtable and import HBase snapshots using a Map Reduce cluster (ie
...
+        --jar bigtable-hbase-1.x-mapreduce-2.5.0-shaded-byo-hadoop.jar \
+        -- \
+        sync-table \
+        --dryrun=true \
+        --sourcezkcluster=<source-zk-quorum> \
+        --targetbigtableproject=<project-id> \
+        --targetbigtableinstance=<instance-id> \
+        <hash-outputdir-hbase> \
+        <source-table-id> \
+        <target-table-id>
```
vermas2012, line 162: Please validate all the commands, I am not very familiar with Hadoop commands and have not run them.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncMapper.java
```diff
@@ -0,0 +1,71 @@
...
+      BigtableSyncTableAdapter.setTargetZkClusterConf(conf, sourceZkClusterConf);
+      LOG.debug("target conn init as source: " + sourceZkClusterConf);
+    }
+
+    if (null == sourceZkClusterConf && null != targetZkClusterConf) {
+      BigtableSyncTableAdapter.setSourceZkClusterConf(conf, targetZkClusterConf);
+      LOG.debug("source conn init as target: " + targetZkClusterConf);
+    }
+
+    // default setup
```
vermas2012, line 58: pl add a detailed comment here. We should clarify what we are doing here vs using super.setup() for?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncMapper.java
```diff
@@ -0,0 +1,71 @@
...
+    super.setup(context);
+
+    // establish target connection to bigtable
+    if (!conf.onlyKeyExists(BigtableSyncTableJob.TARGET_BT_PROJECTID_CONF_KEY)) {
+      BigtableSyncTableAdapter.createBigtableTargetConnection(this, conf);
+    }
+
+    // establish source connection to bigtable
+    if (!conf.onlyKeyExists(BigtableSyncTableJob.SOURCE_BT_PROJECTID_CONF_KEY)) {
+      BigtableSyncTableAdapter.createBigtableSourceConnection(this, conf);
```
vermas2012, line 68: where are we validating the input params? Is there a validation somewhere that checks project/instance/table for CBT?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncTableJob.java
```diff
@@ -0,0 +1,247 @@
...
+    if (!BigtableSyncTableAdapter.parseCommandLine(this, NUM_ARGS, toolDefaultArgs)) return false;
+
+    // return false if options and args are not properly set
+    if (!BigtableSyncTableAdapter.verifyRequiredArgsSet(
+        this,
+        args,
+        sourceBigtableProjectId,
+        sourceBigtableInstance,
+        targetBigtableProjectId,
+        targetBigtableInstance)) return false;
```
vermas2012, line 114: Pl run the formatter.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncTableJob.java
```diff
@@ -0,0 +1,247 @@
...
+
+      final String targetBigtableInstanceKey = "--targetbigtableinstance=";
+      if (cmd.startsWith(targetBigtableInstanceKey)) {
+        targetBigtableInstance = cmd.substring(targetBigtableInstanceKey.length());
+        continue;
+      }
+
+      final String targetBigtableAppProfileKey = "--targetbigtableprofile=";
+      if (cmd.startsWith(targetBigtableAppProfileKey)) {
+        targetBigtableAppProfile = cmd.substring(targetBigtableAppProfileKey.length());
```
vermas2012, line 162: There should be a better way to parse the command line in the package. have you checked other jobs?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncTableJob.java
```diff
@@ -0,0 +1,247 @@
...
+
+  /**
+   * SyncTable command line parsing utility
+   *
+   * @param args
+   * @return
+   */
+  private boolean doCommandLine(final String[] args) {
+    String[] toolDefaultArgs = parseBigtableCommandLineArgs(args);
+    if (!BigtableSyncTableAdapter.parseCommandLine(this, NUM_ARGS, toolDefaultArgs)) return false;
```
vermas2012, line 105: why do we have the argument parsing in 2 functions? can't we wrap both calls in a single function call?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/BigtableSyncTableAdapter.java
```diff
@@ -0,0 +1,384 @@
...
+    // set any defaults & override with any options/args that are set on cli
+    setDefaultConfigs(syncTable);
+
+    if (args.length < NUM_ARGS) {
+      printUsage(null, args);
+      return false;
+    }
+    try {
+      syncTable.sourceHashDir = new Path(args[args.length - 3]);
+      syncTable.sourceTableName = args[args.length - 2];
```
vermas2012, line 158: Where is the args coming from? Where do we enforce strict ordering of the arguments?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/BigtableSyncTableAdapter.java
```diff
@@ -0,0 +1,384 @@
...
+            + ", "
+            + bigtableConf.get(TARGET_BT_INSTANCE_CONF_KEY)
+            + ", "
+            + bigtableConf.get(TARGET_BT_APP_PROFILE_CONF_KEY, ""));
+    try {
+      if (mapper.targetConnection != null) {
+        mapper.targetConnection.close();
+      }
+    } catch (IOException ioe) {
+      LOG.warn("closed target connection, " + ioe.getMessage());
```
vermas2012, line 351: Failed to close target connection? This is very confusing, because you are actually establishing "real" target connection later.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/BigtableSyncTableAdapter.java
```diff
@@ -0,0 +1,384 @@
...
+import org.apache.hadoop.hbase.client.Connection;
+import org.apache.hadoop.hbase.client.ConnectionFactory;
+import org.apache.hadoop.hbase.client.Scan;
+import org.apache.hadoop.hbase.client.Table;
+import org.apache.hadoop.hbase.mapreduce.SyncTable.SyncMapper;
+import org.apache.hadoop.hbase.zookeeper.ZKConfig;
+import org.apache.hadoop.mapreduce.Job;
+
+/** A helper class to access package private fields of SyncTable. */
+public class BigtableSyncTableAdapter {
```
vermas2012, line 45: I would very much prefer for any file in org/apache/hbase directory to be very simple accessor without any real logic. Is it possible to make this file like the HashTableAccessor and move the logic to the BigtableSyncTableJob (or Mapper)?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/validation/EndToEndIT.java
```diff
@@ -0,0 +1,429 @@
...
+
+  @BeforeClass
+  public static void setUpBeforeClass() throws Exception {
+    HB_TEST_UTIL.startMiniCluster(1);
+    HB_TEST_UTIL.setJobWithoutMRCluster();
+  }
+
+  @AfterClass
+  public static void tearDownAfterClass() throws Exception {
+    HB_TEST_UTIL.shutdownMiniCluster();
```
vermas2012, line 80: I think we have a rule for HBASE_TEST_UTIL as well.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/validation/EndToEndIT.java
```diff
@@ -0,0 +1,429 @@
...
+        bigtableSyncTableJobSourceHBaseTargetBigtable(
+            HB_TEST_SOURCE_TABLE, BT_TEST_TARGET_TABLE, testHashOutDir);
+
+    assertVerifySyncMatches(syncCounters);
+  }
+
+  /** assert match/mismatch results from counters and output records */
+  private void assertVerifySyncMatches(Counters syncCounters) {
+    // verify counters
+    assertEquals(60, syncCounters.findCounter(Counter.ROWSWITHDIFFS).getValue());
```
vermas2012, line 192: There counters don't match? 

RowsWithDiff = Sum (all other diffs)??

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/validation/EndToEndIT.java
```diff
@@ -0,0 +1,429 @@
...
+    assertEquals(10, syncCounters.findCounter(Counter.SOURCEMISSINGROWS).getValue());
+    assertEquals(10, syncCounters.findCounter(Counter.TARGETMISSINGROWS).getValue());
+    assertEquals(50, syncCounters.findCounter(Counter.SOURCEMISSINGCELLS).getValue());
+    assertEquals(50, syncCounters.findCounter(Counter.TARGETMISSINGCELLS).getValue());
+    assertEquals(20, syncCounters.findCounter(Counter.DIFFERENTCELLVALUES).getValue());
```
vermas2012, line 197: Where are we checking on the matches?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/validation/EndToEndIT.java
```diff
@@ -0,0 +1,429 @@
...
+    conf.set(
+        BigtableOptionsFactory.BIGTABLE_EMULATOR_HOST_KEY,
+        "localhost:" + bigtableEmulator.getPort());
+    conf.set(BigtableOptionsFactory.BIGTABLE_NULL_CREDENTIAL_ENABLE_KEY, "true");
+    conf.set(BigtableOptionsFactory.BIGTABLE_USE_SERVICE_ACCOUNTS_KEY, "false");
+
+    BigtableSyncTableJob syncTable = new BigtableSyncTableJob(conf);
+    String[] args = Arrays.copyOf(options, options.length + 3);
+    args[options.length] = testHashOutDir.toString(); // sourcehashdir
+    args[options.length + 1] = sourceTableName.getNameAsString(); // sourcetablename
```
vermas2012, line 244: why is this not in the form of `--sourceTable=test-table`

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/validation/EndToEndIT.java
```diff
@@ -0,0 +1,429 @@
...
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Rule;
+import org.junit.Test;
+
+/**
+ * end to end integration test for HashTable and SyncTableValidationOnlyJob from Bigtable to HBase
+ * and HBase to Bigtable
+ */
+public class EndToEndIT {
```
vermas2012, line 55: We should also have a unit tests that validates all the inputs and their vaidation.

vermas2012 (changes requested):
Lets chat offline, I didn't understand the responsibilities of each files and why certain logic belonged in certain class.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncTableJob.java
jhambleton: i confirmed, `mvn com.coveo:fmt-maven-plugin:format` emits this formatting

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/validation/EndToEndIT.java
jhambleton: we use HB_TEST_UTIL within [EndToEndIt for import snapshot job](https://github.com/googleapis/java-bigtable-hbase/blob/main/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/EndToEndIT.java#L61) as well. let me know if this is needed, we may need to address both tests.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/validation/EndToEndIT.java
jhambleton: added two additional junit tests to verify matches `testSyncTableBigtabletoHBaseValidationMatches` and `testSyncTableHBaseToBigtableValidationMatches`

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/README.md
jhambleton: updated test class to validate key-value arguments. other args may be set as strings and we perform basic validation on number of args

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncMapper.java
jhambleton: updated, added a comment

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncMapper.java
jhambleton: we provide a verification in the main `BigtableSyncTableJob` with `verifyRequiredArgsSet` method. ptal

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncTableJob.java
jhambleton: good suggestion, i merged these

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncTableJob.java
jhambleton: we're following the semantics from SyncTable. let me know if changes are required here.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/BigtableSyncTableAdapter.java
jhambleton: great suggestion, refactored as discussed.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/BigtableSyncTableAdapter.java
jhambleton: we discussed the input args, i don't believe any changes are required here, but let me know.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/BigtableSyncTableAdapter.java
jhambleton: updated this to debug level logging and updated msg, ptal

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncMapper.java
```diff
@@ -0,0 +1,71 @@
...
+    String targetZkClusterConf = BigtableSyncTableAdapter.getTargetZkClusterConf(conf);
+    String sourceZkClusterConf = BigtableSyncTableAdapter.getSourceZkClusterConf(conf);
+    if (null == targetZkClusterConf && null != sourceZkClusterConf) {
+      BigtableSyncTableAdapter.setTargetZkClusterConf(conf, sourceZkClusterConf);
+      LOG.debug("target conn init as source: " + sourceZkClusterConf);
+    }
+
+    if (null == sourceZkClusterConf && null != targetZkClusterConf) {
+      BigtableSyncTableAdapter.setSourceZkClusterConf(conf, targetZkClusterConf);
+      LOG.debug("source conn init as target: " + targetZkClusterConf);
```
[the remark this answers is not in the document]
jhambleton: this may be a common configuration issue with the job. i've increased the visibility of the msg to info level and improved the detailed msg to clarify this is a temporary override and for initialization only. map task logs are only viewed when troubleshooting issues for hadoop jobs. ptal

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/validation/EndToEndIT.java
jhambleton: these are basic arguments and not key-value pairs. `SyncTable [options] <sourcehashdir> <sourcetable> <targettable>`

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/validation/EndToEndIT.java
jhambleton: great suggestion, added `TestBigtableSyncTableJob`, ptal

jhambleton:
@vermas2012 - great suggestions with the review. updated the PR, ptal

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncTableJob.java
```diff
@@ -0,0 +1,483 @@
...
+                + sourceBigtableInstance
+                + ")");
+
+      jobConf.set(SOURCE_BT_PROJECTID_CONF_KEY, sourceBigtableProjectId);
+      jobConf.set(SOURCE_BT_INSTANCE_CONF_KEY, sourceBigtableInstance);
+
+      if (sourceBigtableAppProfile != null)
+        jobConf.set(SOURCE_BT_APP_PROFILE_CONF_KEY, sourceBigtableAppProfile);
+    }
+    if (targetBigtableProjectId != null && targetBigtableInstance != null) {
```
vermas2012, line 412: Is there a reason to keep it away from the targetZkCluster setup in the beginning of the method? Can you move this?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncTableJob.java
```diff
@@ -0,0 +1,483 @@
...
+      if (targetZkCluster != null)
+        LOG.warn(
```
vermas2012, line 414: I am not sure if this is before or after the "set fake connections for initializations" step. But if the customer supplies both targetZkCluster and targetBtProject, we should fail the job.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/BigtableSyncTableAccessor.java
```diff
@@ -0,0 +1,148 @@
...
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.Connection;
+import org.apache.hadoop.hbase.client.ConnectionFactory;
+import org.apache.hadoop.hbase.client.Table;
+import org.apache.hadoop.hbase.mapreduce.SyncTable.SyncMapper;
+
+/** A helper class to access package private fields of SyncTable. */
+public class BigtableSyncTableAccessor {
```
vermas2012, line 28: InternalAPI

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/validation/TestBigtableSyncTableJob.java
```diff
@@ -0,0 +1,78 @@
...
+    Configuration conf = new Configuration(false);
+    BigtableSyncTableJob bigtableSyncTable = new BigtableSyncTableJob(conf);
+    boolean isSuccess = bigtableSyncTable.doCommandLine(bigtableSyncTable, args);
+
+    Assert.assertTrue(isSuccess);
+    Assert.assertEquals(
+        parseKeyValueArg(args[0]), BigtableSyncTableAccessor.getSourceZkCluster(bigtableSyncTable));
+    Assert.assertEquals(parseKeyValueArg(args[1]), bigtableSyncTable.getTargetBigtableProjectId());
+    Assert.assertEquals(parseKeyValueArg(args[2]), bigtableSyncTable.getTargetBigtableInstance());
+    Assert.assertEquals(parseKeyValueArg(args[3]), bigtableSyncTable.getTargetBigtableAppProfile());
```
vermas2012, line 48: does it make sense to validate the source/targetTables and hashDir? Do we parse them or just keep the array as is?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/validation/TestBigtableSyncTableJob.java
```diff
@@ -0,0 +1,78 @@
...
+    boolean isSuccess = bigtableSyncTable.doCommandLine(bigtableSyncTable, args);
+
+    Assert.assertTrue(isSuccess);
+    Assert.assertEquals(parseKeyValueArg(args[0]), bigtableSyncTable.getSourceBigtableProjectId());
+    Assert.assertEquals(parseKeyValueArg(args[1]), bigtableSyncTable.getSourceBigtableInstance());
+    Assert.assertEquals(parseKeyValueArg(args[2]), bigtableSyncTable.getSourceBigtableAppProfile());
+    Assert.assertEquals(
+        parseKeyValueArg(args[3]), BigtableSyncTableAccessor.getTargetZkCluster(bigtableSyncTable));
+  }
+
```
vermas2012, line 74: Can you add some tests where validations fail? Something like when both source zkCluster and CBT cluster are specified? or when hashDir is not present (or not a valid path?)

vermas2012 (changes requested):

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncTableJob.java
```diff
@@ -0,0 +1,483 @@
...
+    if (BigtableSyncTableAccessor.getTargetZkCluster(syncTable) == null
+        && (targetBigtableProjectId == null || targetBigtableInstance == null)) {
+      printUsage(
+          "--targetzkcluster or -targetbigtableproject and --targetbigtableinstance required.",
+          args);
+      return false;
+    }
+
+    if (BigtableSyncTableAccessor.getSourceHashDir(syncTable) == null) {
+      printUsage("--sourcehashdir is required.", args);
```
vermas2012, line 285: I don't think the input we require has the --sourceHashDir prefix. The 3 required params are not prefixed by their names. We should remove this "language" in the log and add unit tests where these are not passed

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/validation/EndToEndIT.java
vermas2012: NIT: The whole setup is hidden away somewhere, the readability of this test can be better. Its hard to comprehend the magic numbers that you are validating.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
```diff
@@ -184,6 +184,9 @@ protected static int setConfFromArgs(Configuration conf, String[] args) {
         conf.get(BigtableOptionsFactory.INSTANCE_ID_KEY),
         conf.get(BigtableOptionsFactory.APP_PROFILE_ID_KEY, ""));
 
+    // Set user agent
+    conf.set(BigtableOptionsFactory.CUSTOM_USER_AGENT_KEY, "HBaseMRImport");
```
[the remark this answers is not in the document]
jhambleton: good catch, updated this.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncMapper.java
```diff
@@ -0,0 +1,185 @@
...
+      String appProfileKey,
+      String sourceOrTarget) {
+    Configuration bigtableConf = new Configuration(conf);
+
+    bigtableConf.unset(ClusterConnection.HBASE_CLIENT_CONNECTION_IMPL);
+    BigtableConfiguration.configure(
+        bigtableConf,
+        bigtableConf.get(projectIdKey),
+        bigtableConf.get(instanceKey),
+        bigtableConf.get(appProfileKey, ""));
```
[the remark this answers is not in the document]
jhambleton: added, ptal

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncTableJob.java
```diff
@@ -0,0 +1,483 @@
...
+    return true;
+  }
+
+  /**
+   * set default sync table configs explicitly
+   *
+   * @param syncTable
+   */
+  private void setDefaultConfigs(SyncTable syncTable) {
+    // dry run is enabled by default
```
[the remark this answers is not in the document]
jhambleton: updated, ptal

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncTableJob.java
jhambleton: good suggestion, moved the logic up after setting targetZkCluster

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncTableJob.java
jhambleton: with the existing behavior, if both are set, targetZkCluster and targetBtProject, then we default to Bigtable. The same is true for sourceZkCluster and sourceBigtableProjectId. If both are set, we default to sourceBigtableProjectId for establishing a connection. I agree that we should fail if both are explicitly set as that means the user isn't properly configuring the job. will update this.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/BigtableSyncTableAccessor.java
jhambleton: good catch, updated

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/validation/TestBigtableSyncTableJob.java
jhambleton: we keep these as array items and store them as strings. it doesn't make a lot of sense to verify table names or paths with filesystem schemes as they may be arbitrarily set (ie. hdfs, gs, s3, abfs).

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/validation/TestBigtableSyncTableJob.java
jhambleton: great suggestion, updated junit to test for several other common misconfigs

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncTableJob.java
jhambleton: great catch, this inherits behavior from HBase so we haven't changed the input args vs. prefixed `--` config params.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/validation/TestBigtableSyncTableJob.java
```diff
@@ -24,6 +24,24 @@
 // TODO - parameterize this to run against prod in future
 public class TestBigtableSyncTableJob {
 
+  @Test
+  public void testMissingArgs() {
+    String[] args = {
+      "--sourcezkcluster=hbase-m:2181:/hbase",
+      "--targetbigtableproject=prod-app-bigtable-svcs",
+      "--targetbigtableinstance=prod-app-events",
+      "--targetbigtableprofile=default",
```
vermas2012, line 33: what is this? Is this application profile, if so we should name it that way.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncTableJob.java
```diff
@@ -384,6 +433,31 @@ private Job setupJobWithBigtable(String[] otherArgs) throws IOException {
...
+                + targetBigtableProjectId
+                + "), targetBigtableInstance("
+                + targetBigtableInstance
+                + ")");
+
+      jobConf.set(TARGET_BT_PROJECTID_CONF_KEY, targetBigtableProjectId);
+      jobConf.set(TARGET_BT_INSTANCE_CONF_KEY, targetBigtableInstance);
+
+      if (targetBigtableAppProfile != null)
+        jobConf.set(TARGET_BT_APP_PROFILE_CONF_KEY, targetBigtableAppProfile);
```
vermas2012, line 452: please add curly brackets and scope the if statement, everywhere.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncTableJob.java
```diff
@@ -265,34 +279,69 @@ public boolean verifyRequiredArgsSet(
...
+    // source bigtable project or source instance not set
+    if (BigtableSyncTableAccessor.getSourceZkCluster(syncTable) == null
+        && ((sourceBigtableProjectId == null || sourceBigtableInstance == null))) {
+      printUsage("--sourcebigtableproject and --sourcebigtableinstance required.", args);
+      return false;
+    }
+
+    // neither target zk or target bigtable config set
     if (BigtableSyncTableAccessor.getTargetZkCluster(syncTable) == null
-        && (targetBigtableProjectId == null || targetBigtableInstance == null)) {
```
vermas2012: I think this was correct. if any of instance/project is null, its basically malformed.

vermas2012 (approved):

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/validation/TestBigtableSyncTableJob.java
vermas2012: is this consistent with other jobs?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncTableJob.java
jhambleton: updated

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncTableJob.java
jhambleton: i perform 2 separate validations for hbase or bigtable connection configurations to provide better usage instructions. 

First check if one or the other is set (zk or bigtable) and if neither are set, print usage that zk or bigtable configs are required. 

Later, check if zk is not set, and check bigtable project/instance configs are both set. if not, we print usage of all bigtable configs that are are required.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/test/java/com/google/cloud/bigtable/mapreduce/validation/TestBigtableSyncTableJob.java
jhambleton: good suggestions. it's not the prettiest, but updated app profile options to be more precise - sourcebigtableappprofile and targetbigtableappprofile.

jhambleton: the naming is consistent with upstream [SyncTable.java#L808-L842](https://github.com/apache/hbase/blob/master/hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/SyncTable.java#L808-L842)

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/README.md
```diff
@@ -2,51 +2,89 @@
...
-          -Dgoogle.bigtable.instance.id=<instance-id> \
-          <snapshot-name> \
-          <snapshot-dir> \
-          <table-id> \
-          <tmp-dir>
+    # SyncTable on Bigtable (dryrun enabled by default)
+    GOOGLE_APPLICATION_CREDENTIALS=path/to/service-account.json \
+    hadoop jar bigtable-hbase-1.x-mapreduce-2.5.0-shaded-byo-hadoop.jar \
+        sync-table \
+        --sourcezkcluster=<source-zk-quorum> \
```
igorbernstein2, line 77: why change styles of system properties already in use?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncMapper.java
```diff
@@ -0,0 +1,187 @@
...
+    if (null == targetZkClusterConf && null != sourceZkClusterConf) {
+      BigtableSyncTableAccessor.setTargetZkClusterConf(conf, sourceZkClusterConf);
+      LOG.info(
+          "target conn temporarily set as source for initialization only: " + sourceZkClusterConf);
+    }
+
+    if (null == sourceZkClusterConf && null != targetZkClusterConf) {
+      BigtableSyncTableAccessor.setSourceZkClusterConf(conf, targetZkClusterConf);
+      LOG.info(
+          "source conn temporarily set as target for initialization only: " + targetZkClusterConf);
```
igorbernstein2, line 67: please avoid abbreviating things in logs, it'll make it easier for users to understand

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncMapper.java
```diff
@@ -0,0 +1,187 @@
...
+  public void setup(
+      Mapper<ImmutableBytesWritable, Result, ImmutableBytesWritable, Mutation>.Context context)
+      throws IOException {
+
+    // modify configuration to establish valid connection
+    Configuration conf = context.getConfiguration();
+    conf.unset(ClusterConnection.HBASE_CLIENT_CONNECTION_IMPL);
+
+    // since both source/target require establishing a connection, set the source as the target or
+    // target as source if not set and connecting with bigtable
```
igorbernstein2, line 55: This needs a bit more explanation. If I understand correctly, this is necessary when this used bigtable is on one side of the sync. In that case super.setup() needs a fake config to initialize?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/BigtableSyncTableAccessor.java
```diff
@@ -0,0 +1,150 @@
...
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.Connection;
+import org.apache.hadoop.hbase.client.ConnectionFactory;
+import org.apache.hadoop.hbase.client.Table;
+import org.apache.hadoop.hbase.mapreduce.SyncTable.SyncMapper;
+
+/** A helper class to access package private fields of SyncTable. */
+@InternalApi
+public class BigtableSyncTableAccessor {
```
igorbernstein2, line 30: Please add some doc as to why this class exists. It's not obvious that it works as a bridge to access package protected variables

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/BigtableSyncTableAccessor.java
```diff
@@ -0,0 +1,150 @@
...
+      throws IOException {
+    mapper.targetConnection = ConnectionFactory.createConnection(conf);
+    return mapper.targetConnection;
+  }
+
+  public static void setTargetTable(SyncMapper mapper, Connection connection, Configuration conf)
+      throws IOException {
+    mapper.targetTable =
+        connection.getTable(TableName.valueOf(conf.get(SyncTable.TARGET_TABLE_CONF_KEY)));
+  }
```
igorbernstein2, line 149: I think you should simplify this to just give you the conf keys and use those keys to access the configuration elsewhere.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncMapper.java
```diff
@@ -0,0 +1,187 @@
...
+   * @param sourceOrTarget
+   */
+  private void closeConnection(Connection conn, String sourceOrTarget) {
+    try {
+      if (conn != null) {
+        conn.close();
+      }
+    } catch (IOException ioe) {
+      LOG.debug("error closing temporary " + sourceOrTarget + " connection, " + ioe.getMessage());
+      ioe.printStackTrace();
```
igorbernstein2, line 145: why not bubble this up?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncMapper.java
```diff
@@ -0,0 +1,187 @@
...
+   */
+  private Configuration getBigtableConfiguration(
+      Configuration conf,
+      String projectIdKey,
+      String instanceKey,
+      String appProfileKey,
+      String sourceOrTarget) {
+    Configuration bigtableConf = new Configuration(conf);
+
+    bigtableConf.unset(ClusterConnection.HBASE_CLIENT_CONNECTION_IMPL);
```
igorbernstein2, line 167: is unset necessary? won't .configure() overwrite it?

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/README.md
jhambleton: good question, these are non-conventional, but we're following HBase upstream [HashTable.java#L616-L758](https://github.com/apache/hbase/blob/master/hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/HashTable.java#L616-L758) and  [SyncTable.java#L753-L854](https://github.com/apache/hbase/blob/master/hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/SyncTable.java#L753-L854) conventions with options and args.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncMapper.java
jhambleton: thanks for the suggestion, updated.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncMapper.java
jhambleton: good suggestion, updated and moved this logic this to its own method. 

your understanding is correct. we intend to use SyncMapper super.setup() to initialize the map task, but extend it to establish a connection with Bigtable if it's defined as a source or target. 

in parent [SyncTable.SyncMapper.setup(Context context)](https://github.com/apache/hbase/blob/master/hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/SyncTable.java#L222-L252) a connection is established for creating an HBase connection to both source and target. if source zk or target zk are not set, which is the case when Bigtable is configured as source or target, the HBase connection defaults to localhost:2181 and will fail with establishing a connection.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/BigtableSyncTableAccessor.java
jhambleton: updated the class comment

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/BigtableSyncTableAccessor.java
jhambleton: great suggestion, this cleaned things up a lot. updated, ptal

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncMapper.java
jhambleton: the thought was that we want to log it, but since this is a connection we're throwing away, we want to proceed with the job even if we observe a failure here.

jhambleton: let me know if it would be better if we bubbled this up.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncMapper.java
jhambleton: good suggestion, removed this and verified that it's overwritten as mentioned.

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/README.md
igorbernstein2: its a shame that I didn't follow these to begin with :(

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/README.md
```diff
@@ -226,12 +295,72 @@ setting the properties for the job. For example:
...
+1. Set the following additional environment variables for running the validation steps.
+    ```bash
+    # hash-table validation job
+    export HBASE_TABLENAME=<HBASE_TABLENAME>
+    # hbase zookeeper quorum (ie. zk1.example.com:2181)
+    export HBASE_ZK_QUORUM=<ZK_QUORUM>
+    export HASH_OUTPUTDIR=<HASH_OUTPUTDIR>
+
+    # sync-table validation job
+    export HBASE_ZK_QUORUM_FULL=${HBASE_ZK_QUORUM}:/hbase
```
igorbernstein2, line 309: nit. I dont think these need to be exported

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncMapper.java
igorbernstein2: I think it should be

igorbernstein2 (approved):
a couple of nits, but lgtm

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncMapper.java
jhambleton: thanks for confirming, updated to throw the exception up

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/validation/BigtableSyncMapper.java
jhambleton: this should be closed as we've removed the unset for bigtableConf

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/README.md
jhambleton: thanks, updated these to shell variables

PR #3745, opened 2022-08-19
build: skip clirr for protobuf-java-format-shaded

kolea2:


hbase-migration-tools/mirroring-client/bigtable-hbase-mirroring-client-core-parent/protobuf-java-format-shaded/pom.xml
```diff
@@ -81,6 +81,14 @@
         </configuration>
       </plugin>
 
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
```
igorbernstein2, line 85: please add a comment that this is an internal artifact and we dont need to prevent abi breakage

igorbernstein2 (approved):
lgtm, after comment

hbase-migration-tools/mirroring-client/bigtable-hbase-mirroring-client-core-parent/protobuf-java-format-shaded/pom.xml
kolea2: done!

PR #3805, opened 2022-09-29
feat: add ability to compute number of splits per region in the snapshot importer

igorbernstein2:


bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
```diff
@@ -198,6 +204,40 @@ protected static int setConfFromArgs(Configuration conf, String[] args) {
...
+          regionSize += storeFile.getFileSize();
+        }
+      }
+      maxRegionSize = Math.max(maxRegionSize, regionSize);
+    }
+
+    numSplitsPerRegion = (int) Math.ceil(maxRegionSize / (float) targetSplitSize);
+    if (numSplitsPerRegion < 1) {
+      numSplitsPerRegion = 1;
+    }
```
jhambleton, line 236: can we provide an upper bound of splits and set a max if the total splits for the job (regions * splits per region) exceeds this maximum. It would be safe to use 1 million splits for an upper bound. For example: 

```
max splits = 1million 
total splits = regions * splits per region

if (total splits > max splits ) 
 splits per region = max splits  / num of region
```

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-mapreduce/src/main/java/com/google/cloud/bigtable/mapreduce/hbasesnapshots/ImportHBaseSnapshotJob.java
```diff
@@ -198,6 +204,40 @@ protected static int setConfFromArgs(Configuration conf, String[] args) {
...
+    ;
```
jhambleton, line 237: extra semi

jhambleton (approved):
my tests with the dynamically calculated splits per region were successful. verified multiple cfs. a few minor nits, but overall lgtm.

PR #3854, opened 2022-11-14
deps: update bigtable.version to v2.16.0 and auto-value-annotations to 1.10

prash-mi:
update bigtable.version to v2.16.0 and auto-value-annotations to 1.10


kolea2:
closing in favor of #3855

PR #3877, opened 2022-12-03
chore: pretty print the effective client settings

kongweihan:


kongweihan:
@igorbernstein2

bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -353,4 +352,74 @@ public Set<TableName> getDisabledTables() {
   public BigtableApi getBigtableApi() {
     return bigtableApi;
   }
+
+  // A helper function to parse the pretty print the string from ToStringHelper.
+  public static class ToStringHelperPrettyPrinter {
```
mutianf, line 357: I don't think this needs to be public

bigtable-client-core-parent/bigtable-hbase/src/main/java/org/apache/hadoop/hbase/client/AbstractBigtableConnection.java
```diff
@@ -353,4 +352,74 @@ public Set<TableName> getDisabledTables() {
...
+      IN_SET,  // Print verbatim for sets, including commas and spaces, e.g. "[foo, bar]".
+    }
+
+    private static void appendIndent(StringBuilder sb, int level) {
+      for (int i = 0; i < level; i++) {
+        sb.append("  ");
+      }
+    }
+
+    public static String print(String str) {
```
mutianf, line 371: Same here, I don't think this needs to be public

mutianf (changes requested):
Can you also fix the pr name to follow the convention? something like `chore: pretty print the settings`

mutianf (approved):

PR #3900, opened 2022-12-20
fix: Making outer main public for export/templates

diegomez17:


kongweihan:
Do all the commits in this PR get appended to the main branch if we merge as is or are they squashed into one? If the former,  we should probably improve the commit messages.

igorbernstein2 (approved):

diegomez17:
> Do all the commits in this PR get appended to the main branch if we merge as is or are they squashed into one? If the former, we should probably improve the commit messages.

The title of the commit would be the only thing listed inside of the release. The commits will be shown if someone opens up the PR. Since this is such a small change and the commits don't really mean much, I think we can submit without any worry about naming.

PR #3930, opened 2023-01-19
deps: replace all transitive log4j deps with reload4j

igorbernstein2:
note: this is currently rebased on top of #3929

This breaks the previous assumption that bigtable-hbase doesn't change hbase's classpath. Fortunately due to the severity of the issue, all log4j deps have been removed from hbase1 and replaced with reload4j. The immediate impact is that if the enduser's application depended on log4j, there will now be duplicate log4j classes (from reload4j and the enduser's log4j)


kolea2 (commented):
lgtm, can we add to the enforcer rules to detect any transitive log4j versions (like how we do already with log4j2)

kolea2 (approved):

PR #3940, opened 2023-01-30
fix: fix compatibility of CloudBigtableIO with the emulator env var

igorbernstein2:
There are 2 issues that prevented CloudBigtableIO from working with the emulator:
1. useBulk macro setting would override the endpoint to batch-bigtable.googleapis.com regardless if the BIGTABLE_EMULATOR_ENV was set
2. the emulator doesnt return sample row key sizes when when the data is tiny, so split calculations broke

### DIFF
--- a/bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableIO.java
+++ b/bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableIO.java
@@ -335,7 +335,7 @@ public class CloudBigtableIO {
     protected List<SourceWithKeys> split(
         long regionSize, long desiredBundleSizeBytes, byte[] startKey, byte[] stopKey)
         throws IOException {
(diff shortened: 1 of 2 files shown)


bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableIO.java
```diff
@@ -477,7 +477,7 @@ protected SourceWithKeys(CloudBigtableScanConfiguration configuration, long esti
                   "Source keys not in order: [%s, %s]",
                   Bytes.toStringBinary(startRow), Bytes.toStringBinary(stopRow)));
         }
-        Preconditions.checkState(estimatedSize > 0, "Source size must be positive", estimatedSize);
+        Preconditions.checkState(estimatedSize >= 0, "Source size must be positive", estimatedSize);
```
mutianf, line 480: Should we also update the error message? "Source size can't be negative"?

mutianf (approved):
LGTM after nit

PR #3942, opened 2023-02-01
chore: check compatibility with shared-dependencies upgrade to 2.10.3

prash-mi:
Please do not merge

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -116,6 +116,17 @@ limitations under the License.
     <url>https://github.com/googleapis/java-bigtable-hbase/issues</url>
   </issueManagement>
 
+  <dependencyManagement>
+  <dependencies>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -243,11 +254,11 @@ limitations under the License.
             </goals>
             <configuration>
               <rules>
-                <requireSameVersions>
+              <!--  <requireSameVersions>
```
kolea2, line 257: please undo this, we should be pulling in the same version from the grpc-bom

kolea2 (changes requested):

pom.xml
prash-mi: my bad, wasn't intending to push this change. Updated!

PR #4040, opened 2023-03-28
chore: tweaks to beam

igorbernstein2:
- remove unnecessary dependency plugin disablement
- force upgrade commons-compress
- move plugin config overrides to pluginManagement

### DIFF
--- a/bigtable-dataflow-parent/bigtable-beam-import/pom.xml
+++ b/bigtable-dataflow-parent/bigtable-beam-import/pom.xml
@@ -421,6 +421,8 @@ limitations under the License.
         <filtering>false</filtering>
       </testResource>
     </testResources>
+
+
     <plugins>
       <plugin>
         <artifactId>maven-jar-plugin</artifactId>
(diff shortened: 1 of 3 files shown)


bigtable-dataflow-parent/bigtable-beam-import/pom.xml
```diff
@@ -509,33 +511,6 @@ limitations under the License.
           </execution>
         </executions>
       </plugin>
-
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-failsafe-plugin</artifactId>
-        <configuration>
-          <!-- disable integration tests by default, opt-in via profile -->
-          <skip>true</skip>
```
kolea2, line 518: can you confirm tests are running as expected?

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
```diff
@@ -421,6 +421,8 @@ limitations under the License.
         <filtering>false</filtering>
       </testResource>
     </testResources>
+
```
kolea2, line 424: nit - revert

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
igorbernstein2: yep

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
igorbernstein2: done

kolea2 (approved):

PR #4100, opened 2023-06-29
feat: Add a new configuration key to inject custom credentials class for authentication 

vermas2012:


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOAuthCredentials.java
```diff
@@ -0,0 +1,69 @@
...
+   */
+  public abstract CompletableFuture<Map<String, List<String>>> getRequestMetadata(
+      URI uri, Executor executor) throws IOException;
+
+  /** Returns the HBase configuration used to create this object. */
+  public Configuration getConfiguration() {
+    return configuration;
+  }
+
+  protected Configuration configuration;
```
igorbernstein2, line 68: why is this necessary?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
```diff
@@ -548,6 +550,12 @@ private void configureCredentialProvider(StubSettings.Builder<?, ?> stubSettings
       stubSettings.setCredentialsProvider(
           FixedCredentialsProvider.create(
               buildCredentialFromPrivateKey(serviceAccount, keyFileLocation)));
+    } else if (!Strings.isNullOrEmpty(configuration.get(BIGTABLE_CUSTOM_CREDENTIALS_CLASS_KEY))) {
+      String customAuthClassName = configuration.get(BIGTABLE_CUSTOM_CREDENTIALS_CLASS_KEY);
```
igorbernstein2, line 554: configuration.getClass(key, BigtableOAuthCredentials.class) and then update the method signatures appropriately

igorbernstein2: Also why does this have the least precedence in the auth priority? I would think it belongs right after  BigtableExtendedConfiguration

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/util/CustomCredentialsProvider.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    if (!BigtableOAuthCredentials.class.isAssignableFrom(authClass)) {
+      throw new IllegalArgumentException(
+          "Custom credentials class ["
+              + customAuthClassName
+              + "] must be a child of "
+              + BigtableOAuthCredentials.class.getName()
+              + ".");
+    }
```
igorbernstein2, line 45: this should be replaced with hadoop's Configuration#getClass()

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/util/CustomCredentialsProvider.java
```diff
@@ -0,0 +1,68 @@
...
+    } catch (NoSuchMethodException e) {
+      throw new IllegalArgumentException(
+          "Custom credentials class ["
+              + customAuthClassName
+              + "] must implement a constructor with single argument of type "
+              + Configuration.class.getName()
+              + ".",
+          e);
+    }
+
```
igorbernstein2, line 59: This and the newInstance should probably go into a static method of the BigtableOAuthCredentials interface/ abstract class

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/util/CustomCredentialsProvider.java
```diff
@@ -0,0 +1,68 @@
...
+    try {
+      return new BigtableCredentialsWrapper(
```
igorbernstein2, line 61: then you can inline BigtableCredentialsWrapper into the VeneerSettings and get rid of this class

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableCredentialsWrapper.java
```diff
@@ -0,0 +1,78 @@
...
+
+  public BigtableCredentialsWrapper(BigtableOAuthCredentials btCredentials) {
+    this.bigtableCredentials = btCredentials;
+  }
+
+  private final BigtableOAuthCredentials bigtableCredentials;
+
+  @Override
+  public String getAuthenticationType() {
+    return "OAuth2";
```
igorbernstein2, line 43: add a comment explaining where you got this magic string

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableCredentialsWrapper.java
```diff
@@ -0,0 +1,78 @@
...
+      return bigtableCredentials.getRequestMetadata(uri, MoreExecutors.directExecutor()).get();
+    } catch (Exception e) {
+      throw new RuntimeException("Failed to fetch credentials.", e);
+    }
+  }
+
+  public void getRequestMetadata(
+      final URI uri, Executor executor, final RequestMetadataCallback callback) {
+    executor.execute(() -> blockingGetToCallback(uri, callback));
+  }
```
igorbernstein2, line 58: This is quite messy and totally my fault. I think you were right that we should just use the blocking variant and be done with it....this is very hard to reason about.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOAuthCredentials.java
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * Get the current request metadata.
+   *
+   * <p>This should be called by the transport layer on each request, and the data should be
+   * populated in headers or other context.
+   *
+   * <p>The convention for handling binary data is for the key in the returned map to end with
+   * {@code "-bin"} and for the corresponding values to be base64 encoded.
+   *
+   * <p>This class should handle caching and refeshing of the metadata associated with the request.
```
igorbernstein2, line 56: Add a comment that the ideally the refresh would be async and preemptive

igorbernstein2 (changes requested):

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOAuthCredentials.java
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  /**
+   * Get the current request metadata.
+   *
+   * <p>This should be called by the transport layer on each request, and the data should be
+   * populated in headers or other context.
+   *
+   * <p>The convention for handling binary data is for the key in the returned map to end with
+   * {@code "-bin"} and for the corresponding values to be base64 encoded.
```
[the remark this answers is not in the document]
vermas2012: I am just copying what was in Credentials class. As this is pass through.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOAuthCredentials.java
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * A public interface that should be implemented by the users who want to inject a custom {@link
+ * com.google.auth.Credentials} implementation for auth purposes. Clients can't directly override
+ * the {@link com.google.auth.Credentials} class as it is shaded by Cloud Bigtable client.
+ *
+ * <p>Hence, customers should implement this class, which will be used for authentication. The
+ * authentication should be based on OAuth and must work by just including request metadata with
+ * each request at transport layer.
+ */
+public abstract class BigtableOAuthCredentials {
```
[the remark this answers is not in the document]
vermas2012: I want to keep the hbase configuration in the base class AND enforce the constructor with configuration. I think its ok to keep it as a abstract class.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOAuthCredentials.java
vermas2012: done.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOAuthCredentials.java
vermas2012: If we don't pass the configuration object, there is no way to configure the class. The child classes will have to hardcode everything. 

For example, you have some regional auth server, how do you configure the server URL? Becuse this will be initialized in the HBase region server and not user's services. The best way to configure is to add a string to hbase-site.xml

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableHBaseVeneerSettings.java
vermas2012: I don't think this is going to be a popular way to inject credentials. Customer will only use this method, when no other method will work. hence, its the last.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/util/CustomCredentialsProvider.java
vermas2012: done.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/util/CustomCredentialsProvider.java
vermas2012: agreed, with use of getClass, this method is simple enought to be a named construcutor itself.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/util/CustomCredentialsProvider.java
vermas2012: not sure what you mean by in-lining BigtableCredentialsWrapper into VeneerSettings?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableCredentialsWrapper.java
vermas2012: ack, reverting to the sync version only.

vermas2012:
@igorbernstein2 PTAL.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOAuth2Credentials.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  /**
+   * Get the current request metadata.
+   *
+   * <p>This should be called by the transport layer on each request, and the data should be
+   * populated in headers or other context.
+   *
+   * <p>The convention for handling binary data is for the key in the returned map to end with
+   * {@code "-bin"} and for the corresponding values to be base64 encoded.
```
igorbernstein2, line 55: This is not relevant for us. Instead this should explicitly name the keys that need to be specified for Bigtable service

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOAuth2Credentials.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  /**
+   * Creates a new instance of a child of @{@link BigtableOAuth2Credentials}.
+   *
+   * @param bigtableAuthClass the child class to be instantiated
+   * @param conf HBase configuration required to configure the @bigtableAuthClass
+   * @return a new instance of @bigtableAuthClass
+   */
+  public static Credentials newInstance(
```
igorbernstein2, line 77: can this be package private or at the very least it should be marked internal

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOAuth2Credentials.java
```diff
@@ -0,0 +1,106 @@
...
+    } catch (Exception e) {
+      throw new IllegalArgumentException(
+          "Failed to create object of custom Credentials class ["
+              + bigtableAuthClass.getName()
+              + "].",
+          e);
+    }
+  }
+
+  protected Configuration configuration;
```
igorbernstein2, line 105: Its weird that you have both a public getter and protected member variable for Configuration. I think it would be a lot simpler if you drop both and let endusers retain the ref from the constructor

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableCredentialsWrapper.java
```diff
@@ -0,0 +1,71 @@
...
+import java.net.URI;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A wrapper that implements @{@link Credentials} interface and delegates calls to the
+ * underlying @{@link BigtableOAuth2Credentials} object. This is required to decouple the users of
+ * shaded bigtable from the @{@link Credentials} class.
+ */
+public final class BigtableCredentialsWrapper extends Credentials {
```
igorbernstein2, line 30: Please mark this as internal api

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableCredentialsWrapper.java
```diff
@@ -0,0 +1,71 @@
...
+
+  public BigtableCredentialsWrapper(BigtableOAuth2Credentials btCredentials) {
+    this.bigtableCredentials = btCredentials;
+  }
+
+  private final BigtableOAuth2Credentials bigtableCredentials;
```
igorbernstein2, line 36: member variables should come before the ctor

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableCredentialsWrapper.java
```diff
@@ -0,0 +1,71 @@
...
+  public boolean hasRequestMetadataOnly() {
+    return true;
+  }
+
+  @Override
+  public void refresh() throws IOException {
+    // No-op. btCredentials should refresh internally as required.
+  }
+
+  public BigtableOAuth2Credentials getBigtableCredentials() {
```
igorbernstein2, line 68: Why do you need a getter?

igorbernstein2 (changes requested):

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOAuth2Credentials.java
vermas2012: What are those keys?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOAuth2Credentials.java
vermas2012: marked as internal. If I make it package private I i will need to move the wrapper, which is at its right place right now.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOAuth2Credentials.java
vermas2012: ok.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BigtableCredentialsWrapper.java
vermas2012: I was using it for testing, but yeah, we don't really need it. Removed.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableOAuth2Credentials.java
vermas2012: I have removed the comments about binary data.

vermas2012:
@igorbernstein2 PTAL

igorbernstein2 (approved):

PR #4140, opened 2023-07-31
fix: Add timeout to bulkMutation's close()

### LINKED ISSUE #4139: finishBundle() should timeout after a long time
I observe occasional hang in finishBundle() step for batch writes. Root cause unknown, but we should add timeout as a mitigation to prevent job from hanging forever.

kongweihan:

Fixes #4139 ☕️

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutatorHelper.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutatorHelper.java
@@ -80,14 +80,15 @@ public class BigtableBufferedMutatorHelper {
     this.adapter = adapter;
     this.settings = settings;
     this.dataClient = bigtableApi.getDataClient();
-    this.bulkMutation = dataClient.createBulkMutation(this.adapter.getTableId());
+    this.bulkMutation =
+        dataClient.createBulkMutation(
+            this.adapter.getTableId(), settings.getBulkMutationCloseTimeoutMilliseconds());
     this.operationAccountant = new OperationAccountant();
   }
 
   public void close() throws IOException {
     closedWriteLock.lock();
     try {
-      flush();
       bulkMutation.close();
       closed = true;
     } finally {
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/BigtableHBaseSettings.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/BigtableHBaseSettings.java
@@ -23,6 +23,7 @@ import com.google.api.core.InternalApi;
 import com.google.cloud.bigtable.hbase.BigtableOptionsFactory;
 import com.google.cloud.bigtable.hbase.util.Logger;
 import com.google.cloud.bigtable.hbase.wrappers.veneer.BigtableHBaseVeneerSettings;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import java.io.IOException;
 import java.util.Map;
@@ -39,6 +40,15 @@ public abstract class BigtableHBaseSettings {
   private final String instanceId;
   private final int ttlSecondsForBackup;
 
+  @VisibleForTesting
+  // This is used to override the default closing timeout of 10 minutes in the test so the test can
+  // run faster
+  public static final String BULK_MUTATION_CLOSE_TIMEOUT_MILLISECONDS =
+      "bulk.mutation.close.timeout.milliseconds";
+
+  // Must be non-negative. Set to 0 to disable timeout.
+  private final long bulkMutationCloseTimeoutMilliseconds;
+
   public static BigtableHBaseSettings create(Configuration configuration) throws IOException {
     return BigtableHBaseVeneerSettings.create(configuration);
   }
@@ -51,6 +61,10 @@ public abstract class BigtableHBaseSettings {
         configuration.getInt(
             BigtableOptionsFactory.BIGTABLE_SNAPSHOT_DEFAULT_TTL_SECS_KEY,
             BigtableOptionsFactory.BIGTABLE_SNAPSHOT_DEFAULT_TTL_SECS_VALUE);
+
+    this.bulkMutationCloseTimeoutMilliseconds =
+        configuration.getLong(BigtableHBaseSettings.BULK_MUTATION_CLOSE_TIMEOUT_MILLISECONDS, 0);
+    Preconditions.checkArgument(this.bulkMutationCloseTimeoutMilliseconds >= 0);
   }
 
   public Configuration getConfiguration() {
@@ -81,6 +95,10 @@ public abstract class BigtableHBaseSettings {
 
   // This is equivalent to allow server-side timestamp.
   public abstract boolean isRetriesWithoutTimestampAllowed();
+
+  public long getBulkMutationCloseTimeoutMilliseconds() {
+    return bulkMutationCloseTimeoutMilliseconds;
+  }
 
   protected String getRequiredValue(String key, String displayName) {
     String value = configuration.get(key);
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BulkMutationVeneerApi.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BulkMutationVeneerApi.java
@@ -25,6 +25,9 @@ import com.google.cloud.bigtable.metrics.BigtableClientMetrics.MetricLevel;
 import com.google.cloud.bigtable.metrics.Meter;
 import com.google.common.base.Preconditions;
 import java.io.IOException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 /** For internal use only - public for technical reasons. */
 @InternalApi("For internal usage only")
@@ -34,8 +37,14 @@ public class BulkMutationVeneerApi implements BulkMutationWrapper {
       BigtableClientMetrics.meter(MetricLevel.Info, "bulk-mutator.mutations.added");
   private final Batcher<RowMutationEntry, Void> bulkMutateBatcher;
 
-  BulkMutationVeneerApi(Batcher<RowMutationEntry, Void> bulkMutateBatcher) {
+  // If set to 0, timeout is disabled. Negative value is not accepted.
+  private final long closeTimeoutMilliseconds;
+
+  BulkMutationVeneerApi(
+      Batcher<RowMutationEntry, Void> bulkMutateBatcher, long closeTimeoutMilliseconds) {
     this.bulkMutateBatcher = bulkMutateBatcher;
+    Preconditions.checkArgument(closeTimeoutMilliseconds >= 0);
+    this.closeTimeoutMilliseconds = closeTimeoutMilliseconds;
   }
 
   /** {@inheritDoc} */
@@ -65,9 +74,16 @@ public class BulkMutationVeneerApi implements BulkMutationWrapper {
   @Override
   public synchronized void close() throws IOException {
     try {
-      bulkMutateBatcher.close();
-    } catch (InterruptedException e) {
+      ApiFuture future = bulkMutateBatcher.closeAsync();
+      if (closeTimeoutMilliseconds > 0) {
+        future.get(closeTimeoutMilliseconds, TimeUnit.MILLISECONDS);
+      } else {
+        future.get();
+      }
+    } catch (InterruptedException | ExecutionException e) {
       throw new IOException("Could not close the bulk mutation Batcher", e);
+    } catch (TimeoutException e) {
+      throw new IOException("Cloud not close the bulk mutation Batcher, timed out in close()", e);
     }
   }
 }
--- a/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
+++ b/bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
@@ -72,7 +72,13 @@ public class DataClientVeneerApi implements DataClientWrapper {
 
   @Override
   public BulkMutationWrapper createBulkMutation(String tableId) {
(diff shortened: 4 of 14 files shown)


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/BulkMutationVeneerApi.java
```diff
@@ -34,8 +37,17 @@ public class BulkMutationVeneerApi implements BulkMutationWrapper {
       BigtableClientMetrics.meter(MetricLevel.Info, "bulk-mutator.mutations.added");
   private final Batcher<RowMutationEntry, Void> bulkMutateBatcher;
 
+  private final long closeTimeoutSeconds; // If set to non-positive, never time out.
```
mutianf, line 40: Maybe move this comment to the constant definition

mutianf: I think we should make this > 0 = set timeout, 0 = disabled, < 0 is not allowed. And set a precondition check in BigtableHbaseVeneerSettings.

bigtable-dataflow-parent/bigtable-hbase-beam/src/test/java/com/google/cloud/bigtable/beam/FinishBundleTimeoutTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+public class FinishBundleTimeoutTest {
+
+  @Rule public TestPipeline writePipeline = TestPipeline.create();
+
+  @Rule public final TemporaryFolder workDir = new TemporaryFolder();
+
+  @Rule public ExpectedException thrown = ExpectedException.none();
+
+  @Test
```
mutianf, line 49: lets set a timeout of 20 seconds on this test.  `@Test(timeout = 20)`

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/SharedDataClientWrapper.java
```diff
@@ -52,6 +52,11 @@ public BulkMutationWrapper createBulkMutation(String tableId) {
     return delegate.createBulkMutation(tableId);
```
mutianf, line 52: return createBulkMutation(tableId, 0);

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutatorHelper.java
```diff
@@ -78,16 +78,18 @@ public class BigtableBufferedMutatorHelper {
   public BigtableBufferedMutatorHelper(
       BigtableApi bigtableApi, BigtableHBaseSettings settings, HBaseRequestAdapter adapter) {
     this.adapter = adapter;
-    this.settings = settings;
+    this.settings =
+        settings; // this settings should have all the configs from CloudBigtableTableConfiguration
```
mutianf, line 82: I think we should remove this comment here. CloudBigtableTableConfiguration is only relevant to dataflow. Adding the comment here feels like we only use bufferedMutator for dataflow which is not true.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -75,6 +75,11 @@ public BulkMutationWrapper createBulkMutation(String tableId) {
     return new BulkMutationVeneerApi(delegate.newBulkMutationBatcher(tableId));
```
mutianf, line 75: here call `new BulkMutationVeneerApi(deleget.newBulkMutationBatcher(tableId), 0)`.

mutianf (changes requested):

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutatorHelper.java
kongweihan: ok. But just curious what else this is designed for?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/BigtableHBaseSettings.java
```diff
@@ -51,6 +56,9 @@ public BigtableHBaseSettings(Configuration configuration) {
         configuration.getInt(
             BigtableOptionsFactory.BIGTABLE_SNAPSHOT_DEFAULT_TTL_SECS_KEY,
             BigtableOptionsFactory.BIGTABLE_SNAPSHOT_DEFAULT_TTL_SECS_VALUE);
+
+    this.bulkMutationCloseTimeoutSeconds =
+        configuration.getLong(BULK_MUTATION_CLOSE_TIMEOUT_SECONDS, 600);
```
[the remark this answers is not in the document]
kongweihan: In case the user sets a large operation timeout?
Even if we use it, should it longer than the operation timeout?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutatorHelper.java
mutianf: It implements hbase buffered mutator: https://hbase.apache.org/apidocs/org/apache/hadoop/hbase/client/BufferedMutator.html for batching requests, which are not always coming from dataflow

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/SharedDataClientWrapper.java
mutianf: I think it's cleaner when all the code goes through a single entry point when creating the object. https://stackoverflow.com/questions/581873/best-way-to-handle-multiple-constructors-in-java

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/BigtableHBaseSettings.java
```diff
@@ -39,6 +39,14 @@ public abstract class BigtableHBaseSettings {
   private final String instanceId;
   private final int ttlSecondsForBackup;
 
+  public static final String BULK_MUTATION_CLOSE_TIMEOUT_SECONDS =
+      "bulk.mutation.close.timeout.seconds";
+
+  // Must be non-negative. Set to 0 to disable timeout.
+  private final long bulkMutationCloseTimeoutSeconds;
+
+  private static final long DEFAULT_BULK_MUTATION_CLOSE_TIMEOUT_SECONDS = 600;
```
mutianf, line 48: Do we always want to force close the BufferedMutator, even when it's not used in dataflow? If not, we should set the default to 0, and in CloudBigtableConfiguration override it to 10 miutes.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/SharedDataClientWrapper.java
kongweihan: IIUC this stackoverflow question is saying different ctors should all call the same ctor? It didn't say we should only have 1 ctor.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutatorHelper.java
kongweihan: oh ok, so in hbase this is a separate class for batch mutation? Like it's not part of the regular hbase client?
Why do people have to create separate things for it instead of just adding the function to the client?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/BigtableHBaseSettings.java
kongweihan: You're right. Updated

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutatorHelper.java
mutianf: Sorry, not sure I follow. BufferedMutator is the hbase equivalent of our batcher

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -72,7 +72,12 @@ public class DataClientVeneerApi implements DataClientWrapper {
 
   @Override
   public BulkMutationWrapper createBulkMutation(String tableId) {
-    return new BulkMutationVeneerApi(delegate.newBulkMutationBatcher(tableId));
+    return new BulkMutationVeneerApi(delegate.newBulkMutationBatcher(tableId), 0);
```
mutianf, line 75: Sorry, one more nit, here you can also call `return createBulkMutation(tableId, 0);`

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/SharedDataClientWrapper.java
mutianf: I guess it doesn't matter too much

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableConfiguration.java
```diff
@@ -247,6 +248,15 @@ public Configuration toHBaseConfig() {
       }
     }
     setUserAgent(config);
+
+    ValueProvider<String> timeout =
+        configuration.get(BigtableHBaseSettings.BULK_MUTATION_CLOSE_TIMEOUT_SECONDS);
+    if (timeout == null) {
+      config.set(BigtableHBaseSettings.BULK_MUTATION_CLOSE_TIMEOUT_SECONDS, "600");
+    } else {
+      config.set(BigtableHBaseSettings.BULK_MUTATION_CLOSE_TIMEOUT_SECONDS, timeout.get());
```
mutianf, line 257: I don't think we need to use value provider here for timeout, and also we can't call valueProvider.get() at this point (even though in this case it won't throw error) because this is called at graph construction time. Value provider (RuntimeValueProvider more specificlly) is used in dataflow templates to defer the get of the parameters. I found some more explanation of ValueProvider here: https://stackoverflow.com/questions/62948039/google-dataflow-and-apache-beam-why-valueprovider.

I know you probably followed examples from project id etc. However those examples are bad 😂  and we should avoid making the same mistake here (even though it won't throw error). When we're constructing the pipeline, we don't call `getProjectId` etc, instead we need to call the `getProjectIdValueProvider` version. See https://github.com/googleapis/java-bigtable-hbase/blob/main/bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableIO.java#L421-L424.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/BigtableBufferedMutatorHelper.java
kongweihan: no worries I just wondered why the batcher needs to be a separate class, why not adding the batching logic to the client directly.

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableConfiguration.java
mutianf: nvm i was wrong. this is called at runtime. so we should be good

bigtable-dataflow-parent/bigtable-beam-import/src/main/java/com/google/cloud/bigtable/beam/TemplateUtils.java
```diff
@@ -44,7 +44,8 @@ public static CloudBigtableTableConfiguration buildImportConfig(
             .withProjectId(opts.getBigtableProject())
             .withInstanceId(opts.getBigtableInstanceId())
             .withTableId(opts.getBigtableTableId())
-            .withConfiguration(BigtableOptionsFactory.CUSTOM_USER_AGENT_KEY, customUserAgent);
+            .withConfiguration(BigtableOptionsFactory.CUSTOM_USER_AGENT_KEY, customUserAgent)
+            .withConfiguration(BigtableOptionsFactory.BIGTABLE_MUTATE_RPC_TIMEOUT_MS_KEY, "600000");
```
mutianf, line 48: revert this change?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/BigtableHBaseSettings.java
```diff
@@ -39,6 +39,12 @@ public abstract class BigtableHBaseSettings {
   private final String instanceId;
   private final int ttlSecondsForBackup;
 
+  public static final String BULK_MUTATION_CLOSE_TIMEOUT_MILLISECONDS =
```
mutianf, line 42: Can we add `@VisibleForTesting` and a comment on "this is used to override the default closing timeout of 10 minutes in the test so the test can run faster".

mutianf (approved):
LGTM after nits.

PR #4166, opened 2023-09-21
fix: Add paging to hbase client

ron-gal:


bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -230,41 +246,67 @@ protected void onCompleteImpl() {
...
+    private final Queue<Result> buffer;
+    private final Query.QueryPaginator paginator;
+    private final paginatorFunction wrapper;
+    private final int refillSegmentWaterMark;
+
+    RowResultScanner(Query.QueryPaginator paginator, paginatorFunction wrapper) {
+      this.paginator = paginator;
+      this.wrapper = wrapper;
+      this.buffer = new ArrayDeque<>();
+      this.refillSegmentWaterMark = (int) (PAGE_SIZE * WATERMARK_PERCENTAGE);
```
mutianf, line 270: You need to do Math.max(1, (int) (PAGE_SIZE * WATERMARK_PERCENTAGE) to avoid not reading all the data when the page size is small.

ron-gal: In the beam client it made sense. Here I am multiplying two private static final variables :)

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestFilters.java
```diff
@@ -2038,7 +2038,7 @@ public void testPageFilters() throws IOException {
     PageFilter pageFilter = new PageFilter(20);
     scan.setFilter(pageFilter);
     try (ResultScanner scanner = table.getScanner(scan)) {
-      Assert.assertEquals(20, Iterators.size(scanner.iterator()));
+      Assert.assertEquals(100, Iterators.size(scanner.iterator()));
```
mutianf, line 2041: I don't think this is right, the pageFilter sets the limit to 20, which should be respected.

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestFilters.java
```diff
@@ -2048,7 +2048,7 @@ public void testPageFilters() throws IOException {
             pageFilter);
     scan.setFilter(filterList);
     try (ResultScanner scanner = table.getScanner(scan)) {
-      Assert.assertEquals(20, Iterators.size(scanner.iterator()));
+      Assert.assertEquals(100, Iterators.size(scanner.iterator()));
```
mutianf, line 2051: same here.

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/TestScan.java
```diff
@@ -210,7 +210,7 @@ public void test100ResultsInScanner() throws IOException {
 
     Scan scan = new Scan();
     scan.withStartRow(rowKeys[0])
-        .withStopRow(rowFollowing(rowKeys[rowsToWrite - 1]))
+        .withStopRow(rowFollowing(rowKeys[rowsToWrite - 1]), true)
```
mutianf, line 213: I don't think this is right either, this changed the test behavior and made the stop key inclusive

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestDataClientVeneerApi.java
```diff
@@ -204,53 +206,58 @@ public void testReadRows() throws IOException {
...
     verify(serverStream, times(2)).iterator();
     verify(mockStreamingCallable, times(2))
-        .call(Mockito.eq(query), Mockito.any(GrpcCallContext.class));
+        .call(Mockito.any(Query.class), Mockito.any(GrpcCallContext.class));
   }
 
   @Test
-  public void testReadRowsCancel() throws IOException {
-
+  public void testRead100Rows() throws IOException {
```
mutianf, line 229: where did testReadRowsCancel() go? we still want the cancel test

mutianf (changes requested):
We need a few more tests to make sure the paginator works properly:
1. Test cancel the read
2. Test reading rows where the query limit is below PAGE_SIZE
3. test reading rows where the query limit is above PAGE_SIZE
4. Test reading query with multiple row ranges (if it's already covered we can skip this)
5. Test full table scan 
6. Test error handling

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestFilters.java
ron-gal: fixed

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/AbstractTestFilters.java
ron-gal: fixed

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/TestScan.java
ron-gal: fixed

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestDataClientVeneerApi.java
ron-gal: brought it back

ron-gal:
1. done
2. done
3. done
4. already exists (testMultiRangeFilter in AbstractTestFilters.java)
5. done
6. on it

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -136,8 +144,22 @@ public Result apply(Row row) {
 
   @Override
   public ResultScanner readRows(Query request) {
+    final RequestContext requestContext =
+        RequestContext.create("ProjectId", "InstanceId", "AppProfile");
+    int requestedPageSize = (int) request.toProto(requestContext).getRowsLimit();
```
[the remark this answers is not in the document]
ron-gal: This is handled by the 
[paginator](https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java#L390)

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -230,41 +255,75 @@ protected void onCompleteImpl() {
...
-    RowResultScanner(ServerStream<Result> serverStream) {
-      this.serverStream = serverStream;
-      this.iterator = serverStream.iterator();
+    private ServerStream<Result> serverStream;
+    private ByteString lastSeenRowKey = ByteString.EMPTY;
+    private Boolean hasMore = true;
+    private final Queue<Result> buffer;
+    private final Query.QueryPaginator paginator;
+    private final paginatorFunction wrapper;
+    private final int refillSegmentWaterMark;
```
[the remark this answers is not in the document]
ron-gal: Done

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -230,41 +255,92 @@ protected void onCompleteImpl() {
...
+
+    private void waitReadRowsFuture() {
+      if (this.serverStream == null) {
+        return;
+      }
+      Iterator<Result> iterator = this.serverStream.iterator();
+      while (iterator.hasNext()) {
+        Result result = iterator.next();
+        this.buffer.add(result);
+        if (result == null || result.rawCells() == null) {
```
mutianf, line 333: Is this check necessary? I don't think result would be null

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
```diff
@@ -298,13 +300,30 @@ public ResultScanner getScanner(final Scan scan) throws IOException {
...
+      if (scan.getCaching() == -1) {
+        final ResultScanner scanner = clientWrapper.readRows(hbaseAdapter.adapt(scan));
+        if (hasWhileMatchFilter(scan.getFilter())) {
+          return Adapters.BIGTABLE_WHILE_MATCH_RESULT_RESULT_SCAN_ADAPTER.adapt(scanner, span);
+        }
+        // TODO: need to end the span when stream ends
+        return scanner;
+      } else {
+        final Query request = hbaseAdapter.adapt(scan);
+        final RequestContext requestContext =
```
mutianf, line 314: request should already have the row limit, can we do QueryPaginator paginator = request.createPaginator(scan.getCaching) instead?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/SharedDataClientWrapper.java
```diff
@@ -110,7 +110,8 @@ public void close() throws IOException {
   }
 
   @Override
-  public void setCaching(int caching) {
+  public ResultScanner readRows(Query.QueryPaginator paginator, int requestedPageSize) {
     // unimplemented.
+    return null;
```
mutianf, line 115: can we add this method to ClientWrapper and call delegate.readRows(QueryPaginator) here?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
```diff
@@ -298,13 +300,30 @@ public ResultScanner getScanner(final Scan scan) throws IOException {
...
+            RequestContext.create("ProjectId", "InstanceId", "AppProfile");
+        int requestedPageSize = (int) request.toProto(requestContext).getRowsLimit();
+        if (requestedPageSize > scan.getCaching()) {
+          requestedPageSize = scan.getCaching();
+        }
+
+        Query.QueryPaginator paginator = request.createPaginator(requestedPageSize);
+        final ResultScanner scanner = clientWrapper.readRows(paginator, requestedPageSize);
+        if (hasWhileMatchFilter(scan.getFilter())) {
```
mutianf, line 326: can we extract this logic out?
ResultScanner resultScanner;
if (getCaching == -1) {
  resultScanner = clientWrapper.readRows(...);
} else {
   Query request = adapter.adapt(...);
   ...
   resultScanner = clientWrapper.readRows(..);
}

if (hasWhileMatchFilter(..)) {
 ..
}
return scanner;

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
```diff
@@ -298,13 +300,30 @@ public ResultScanner getScanner(final Scan scan) throws IOException {
...
+        final Query request = hbaseAdapter.adapt(scan);
+        final RequestContext requestContext =
+            RequestContext.create("ProjectId", "InstanceId", "AppProfile");
+        int requestedPageSize = (int) request.toProto(requestContext).getRowsLimit();
+        if (requestedPageSize > scan.getCaching()) {
+          requestedPageSize = scan.getCaching();
+        }
+
+        Query.QueryPaginator paginator = request.createPaginator(requestedPageSize);
+        final ResultScanner scanner = clientWrapper.readRows(paginator, requestedPageSize);
```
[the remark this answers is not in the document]
ron-gal: to calculate the size of the cache, that needs to be 10% of that.
It's annoying, I agree =\

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/SharedDataClientWrapper.java
ron-gal: Done

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
ron-gal: Done

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
ron-gal: in case request.getRowsLimit is set (like 20 in the test), and caching not set, we will end up with a request for PAGE_SIZE (scan.getCaching() is 0\-1, and is not allowed in the paginator), thus failing the test (and effectively asking for 100 rather than 20, against the user desire)

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
```diff
@@ -298,8 +302,24 @@ public ResultScanner getScanner(final Scan scan) throws IOException {
...
     try (Scope scope = TRACER.withSpan(span)) {
+      ResultScanner scanner;
+      if (scan.getCaching() == -1) {
+        scanner = clientWrapper.readRows(hbaseAdapter.adapt(scan));
+      } else {
+        final Query request = hbaseAdapter.adapt(scan);
+        final RequestContext requestContext =
+            RequestContext.create("ProjectId", "InstanceId", "AppProfile");
+        int requestedPageSize = (int) request.toProto(requestContext).getRowsLimit();
+        if (requestedPageSize > scan.getCaching()) {
```
mutianf, line 313: I'm not sure why line 313 to line 317 is necessary. This should be handled by QueryPaginator already. And why do we still have a default PAGE_SIZE ? PAGE_SIZE should always be value from scan.getCaching().

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
mutianf: Yeah, but QueryPagniator will only return 20 in this case. https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java#L388-L392

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
```diff
@@ -298,8 +302,24 @@ public ResultScanner getScanner(final Scan scan) throws IOException {
...
+          requestedPageSize = scan.getCaching();
+        }
+        if (requestedPageSize <= 0) {
+          requestedPageSize = PAGE_SIZE;
+        }
 
-      final ResultScanner scanner = clientWrapper.readRows(hbaseAdapter.adapt(scan));
+        Query.QueryPaginator paginator = request.createPaginator(requestedPageSize);
```
mutianf, line 320: requestedPageSize should be scan.getCaching(). The only thing we need to verify is scan.getCaching() is > 0.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -134,6 +140,18 @@ public Result apply(Row row) {
         MoreExecutors.directExecutor());
   }
 
+  @Override
+  public ResultScanner readRows(Query.QueryPaginator paginator, int requestedPageSize) {
```
mutianf, line 144: Maybe a better way to do this is add a getPageSize() method to QueryPagninator so we don't need to pass in the requestedPageSize here again. Can you make a change to https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java#L349 and mark the getPageSize() method to be `@InternalApi`?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -229,8 +250,96 @@ protected void onCompleteImpl() {
...
+      Iterator<Result> iterator = this.serverStream.iterator();
+      while (iterator.hasNext()) {
+        Result result = iterator.next();
+        this.buffer.add(result);
+        if (result == null || result.rawCells() == null) {
+          continue;
+        }
+        this.lastSeenRowKey = RESULT_ADAPTER.getKey(result);
+        this.currentByteSize += Result.getTotalSizeOfCells(result);
+        if (this.currentByteSize >= maxSegmentByteSize) {
```
mutianf, line 332: Can we also add a test to test the behavior when maxSegmentByteSize is reached? We can make maxSegmentByteSize accessible to test and set it manually to a lower value.

mutianf: And I think here we need to cancel the serverStream

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
ron-gal: https://github.com/googleapis/java-bigtable/pull/1987

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
ron-gal: Done

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
ron-gal: Done

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
ron-gal: Done

ron-gal: Done

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
ron-gal: Done

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -70,6 +72,10 @@ public class DataClientVeneerApi implements DataClientWrapper {
     this.clientOperationTimeouts = clientOperationTimeouts;
   }
 
+  interface paginatorFunction {
```
igorbernstein2, line 75: Why lower case?
Also why do you need a separate interface for this? Can it be `Function<Query.QueryPaginator, ServerStream<Result>>` ?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -134,6 +140,17 @@ public Result apply(Row row) {
...
 
+  @Override
+  public ResultScanner readRows(Query.QueryPaginator paginator) {
+    return new PaginatedRowResultScanner(
+        paginator,
+        (p) -> {
+          return delegate
+              .readRowsCallable(RESULT_ADAPTER)
+              .call(p.getNextQuery(), createScanCallContext());
+        });
```
igorbernstein2, line 151: nit: you dont need the curly brackets or return
```
p -> delegate
              .readRowsCallable(RESULT_ADAPTER)
              .call(p.getNextQuery(), createScanCallContext());
```

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -229,8 +249,97 @@ protected void onCompleteImpl() {
...
+    private static final double WATERMARK_PERCENTAGE = .1;
+    private static final int MIN_BYTE_BUFFER_SIZE = 100 * 1024 * 1024;
+    private static final double DEFAULT_BYTE_LIMIT_PERCENTAGE = .1;
+    private static final RowResultAdapter RESULT_ADAPTER = new RowResultAdapter();
 
+    private final Meter scannerResultMeter =
+        BigtableClientMetrics.meter(BigtableClientMetrics.MetricLevel.Info, "scanner.results");
+    private final Timer scannerResultTimer =
+        BigtableClientMetrics.timer(
+            BigtableClientMetrics.MetricLevel.Debug, "scanner.results.latency");
```
igorbernstein2, line 263: Im not sure we want to expand the drop wizard style metrics. I'm planning on dropping this soon

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -229,8 +249,97 @@ protected void onCompleteImpl() {
...
+
+    private ServerStream<Result> serverStream;
+    private ByteString lastSeenRowKey = ByteString.EMPTY;
+    private Boolean hasMore = true;
+    private final Queue<Result> buffer;
+    private final Query.QueryPaginator paginator;
+    private final paginatorFunction wrapper;
+    private final int refillSegmentWaterMark;
+
+    static long maxSegmentByteSize =
```
igorbernstein2, line 273: This should not be mutable. I would recommend the following:

```
private static final DEFAULT_MAX_SEGMENT_SIZE = Math.max(...);
private final maxSegmentByteSize;

PaginatedRowResultScanner(Query.QueryPaginator paginator, paginatorFunction wrapper) {
  this(paginator, wrapper, DEFAULT_MAX_SEGMENT_SIZE);
}

// Only used for testing
PaginatedRowResultScanner(Query.QueryPaginator paginator, paginatorFunction wrapper, long maxSegmentByteSize) {
   this.maxSegmentByteSize = maxSegmentByteSize;
....
}
```

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -229,8 +249,97 @@ protected void onCompleteImpl() {
...
+    private final Timer scannerResultTimer =
+        BigtableClientMetrics.timer(
+            BigtableClientMetrics.MetricLevel.Debug, "scanner.results.latency");
+
+    private ServerStream<Result> serverStream;
+    private ByteString lastSeenRowKey = ByteString.EMPTY;
+    private Boolean hasMore = true;
+    private final Queue<Result> buffer;
+    private final Query.QueryPaginator paginator;
+    private final paginatorFunction wrapper;
```
igorbernstein2, line 270: wrapper is fairly ambiguous here. I think you want something like `streamSegmentFactory`
(its java, no variable name is complete with a factory :) )

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -229,8 +249,97 @@ protected void onCompleteImpl() {
   }
 
   /** wraps {@link ServerStream} onto HBase {@link ResultScanner}. */
-  private static class RowResultScanner extends AbstractClientScanner {
+  static class PaginatedRowResultScanner extends AbstractClientScanner {
```
igorbernstein2, line 252: This definitely needs more doc and an explanation how its meant to be used. The audience of the documentation should be a client developer

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
igorbernstein2: Also, why do we need a closure to begin with? why not pass the client directly?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -229,8 +249,97 @@ protected void onCompleteImpl() {
...
+    @Override
+    public void close() {
+      if (this.serverStream != null) {
+        this.serverStream.cancel();
+        this.serverStream = null;
+      }
+    }
+
+    public boolean renewLease() {
+      throw new UnsupportedOperationException("renewLease");
```
igorbernstein2, line 318: why throw? Why not just always return true?

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -229,8 +249,97 @@ protected void onCompleteImpl() {
...
+    }
+
+    private void waitReadRowsFuture() {
+      Iterator<Result> iterator = this.serverStream.iterator();
```
igorbernstein2, line 322: I think it would be significantly more efficient if you use the async api to fetch the next batch. ie

```java
SettableFuture<List<Result>> resultsFuture = SettableFuture.create();

dataClient.readRowsCallable(myRowMerger).call(req,  new ResponseObserver() {
   List<Result> results = new ArrayList();
   onStart(Controller c) {
     this.controller = c;
   }
   onResponse(Result r) {
      results.add(r);
      if (thresholdReached) {
         this.controller.cancel();
      }
   }
   onComplete() {
     resultsFuture.set(results);
   }
   onError(e) {
      resultsFuture.setExceptionally(e);
});

List<Result> results = resultsFuture.get();
....
```

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
ron-gal: Agreed not to change anything

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
ron-gal: Done

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
ron-gal: done

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
ron-gal: Done

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
ron-gal: Didn't know about Function<...> :( Done this way.
Passing the client feels like it's giving the scanner too much power, when all I want it to do is call a specific function that given a paginator - gets the next page.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
ron-gal: Done

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
ron-gal: done

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
ron-gal: Ended up passing the client, to be able to use the async API

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestDataClientVeneerApi.java
```diff
@@ -74,6 +80,8 @@ public class TestDataClientVeneerApi {
   private static final String TABLE_ID = "fake-table";
   private static final ByteString ROW_KEY = ByteString.copyFromUtf8("row-key");
 
+  private static AtomicBoolean cancelled = new AtomicBoolean(false);
```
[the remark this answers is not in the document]
ron-gal: copy pasted from beam, removed

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -155,7 +170,8 @@ public void readRowsAsync(Query request, StreamObserver<Result> observer) {
         .call(request, new StreamObserverAdapter<>(observer), createScanCallContext());
   }
 
-  // Point reads are implemented using a streaming ReadRows RPC. So timeouts need to be managed
+  // Point reads are implemented using a streaming ReadRows RPC. So timeouts need
+  // to be managed
```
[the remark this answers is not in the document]
ron-gal: done

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -134,6 +143,12 @@ public Result apply(Row row) {
         MoreExecutors.directExecutor());
   }
 
+  @Override
+  public ResultScanner readRows(Query.QueryPaginator paginator, long maxSegmentByteSize) {
+    return new PaginatedRowResultScanner(
+        paginator, delegate, maxSegmentByteSize, () -> this.createScanCallContext());
```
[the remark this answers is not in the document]
ron-gal: done

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -228,9 +246,167 @@ protected void onCompleteImpl() {
...
+          future = fetchNextSegment();
+        }
+        if (this.buffer.isEmpty() && this.future != null) {
+          try {
+            this.waitReadRowsFuture();
+          } catch (IOException e) {
+            return null;
+          }
+        }
+        scannerResultMeter.mark();
```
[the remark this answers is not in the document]
ron-gal: No.
In the non paginated scanner (RowResultScanner) this is called only when we return a response, same logic applies here.

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/wrappers/veneer/DataClientVeneerApi.java
```diff
@@ -228,9 +246,167 @@ protected void onCompleteImpl() {
...
+      this.refillSegmentWaterMark =
+          (int) Math.max(1, paginator.getPageSize() * WATERMARK_PERCENTAGE);
+      this.createScanCallContext = createScanCallContext;
+      this.future = fetchNextSegment();
+    }
+
+    @Override
+    public Result next() {
+      try (Context ignored = scannerResultTimer.time()) {
+        if (this.buffer.size() < this.refillSegmentWaterMark && this.future == null && hasMore) {
```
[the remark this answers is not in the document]
ron-gal: Done

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
```diff
@@ -298,8 +299,14 @@ public ResultScanner getScanner(final Scan scan) throws IOException {
...
     try (Scope scope = TRACER.withSpan(span)) {
-
-      final ResultScanner scanner = clientWrapper.readRows(hbaseAdapter.adapt(scan));
+      ResultScanner scanner;
+      if (scan.getCaching() == -1) {
+        scanner = clientWrapper.readRows(hbaseAdapter.adapt(scan));
+      } else {
+        Query.QueryPaginator paginator =
+            hbaseAdapter.adapt(scan).createPaginator(scan.getCaching());
+        scanner = clientWrapper.readRows(paginator, -1);
```
mutianf, line 308: Maybe we can pass in the default to begin with?

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestDataClientVeneerApi.java
```diff
@@ -204,30 +234,153 @@ public void testReadRows() throws IOException {
...
+                return null;
+              }
+            })
+        .when(mockStreamingCallable)
+        .call(
+            Mockito.any(com.google.cloud.bigtable.data.v2.models.Query.class),
+            Mockito.any(ResponseObserver.class),
+            Mockito.any(GrpcCallContext.class));
+
+    ResultScanner resultScanner = dataClientWrapper.readRows(query.createPaginator(100), 3);
```
mutianf, line 349: same question here, why do we return an empty result here? I don't think that's gonna use any memory? Maybe you want to return another expected result and assert readRowsCallable is called twice.

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestDataClientVeneerApi.java
```diff
@@ -242,15 +395,56 @@ public void testReadRowsCancel() throws IOException {
...
 
-    // make sure that the scanner doesn't iteract with the iterator on close
+    // make sure that the scanner doesn't interact with the iterator on close
     verify(serverStream).cancel();
-    verify(mockIter, times(1)).hasNext();
-    verify(mockIter, times(1)).next();
+  }
+
+  @Test
+  public void testRead100Rows() throws IOException {
```
mutianf, line 416: Was this a new test that's added to test the paginator?

mutianf (changes requested):

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
mutianf: We could move the default value in this class?

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestDataClientVeneerApi.java
```diff
@@ -204,30 +234,153 @@ public void testReadRows() throws IOException {
...
+            Mockito.any(com.google.cloud.bigtable.data.v2.models.Query.class),
+            Mockito.any(ResponseObserver.class),
+            Mockito.any(GrpcCallContext.class));
+
+    resultScanner.close();
+
+    ResultScanner noRowsResultScanner = dataClientWrapper.readRows(query.createPaginator(100), -1);
+    assertNull(noRowsResultScanner.next());
+
+    verify(mockDataClient, times(2)).readRowsCallable(Mockito.<RowResultAdapter>any());
```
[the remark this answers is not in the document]
mutianf: Ok, I see it now. However I don't think this actually tested anything. I think we can set a smaller page size (for example, 2), return more rows (say 3), and verify that readRowsCallable are called 3 times.

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestDataClientVeneerApi.java
mutianf: Similar to above, I don't think this actually tested the memory buffer function. We should at least return another result, and verify that readRowsCallable are called twice (first time the request got cancelled becuase we filled up the buffer, and the second time we finish the read).

mutianf (changes requested):

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/TestScan.java
```diff
@@ -179,9 +179,22 @@ public void testGetScannerNoQualifiers() throws IOException {
...
+  public void testManyResultsInScanner_equalToPageSize() throws IOException {
+    testManyResultsInScanner(100);
+  }
+
+  @Test
+  public void testManyResultsInScanner_greaterThanPageSize() throws IOException {
+    testManyResultsInScanner(105);
+  }
+
+  private void testManyResultsInScanner(int rowsToWrite) throws IOException {
```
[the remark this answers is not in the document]
ron-gal: Done

bigtable-client-core-parent/bigtable-hbase-integration-tests-common/src/test/java/com/google/cloud/bigtable/hbase/TestScan.java
```diff
@@ -179,9 +179,22 @@ public void testGetScannerNoQualifiers() throws IOException {
...
+
+  @Test
+  public void testManyResultsInScanner_equalToPageSize() throws IOException {
+    testManyResultsInScanner(100);
+  }
+
+  @Test
+  public void testManyResultsInScanner_greaterThanPageSize() throws IOException {
+    testManyResultsInScanner(105);
+  }
```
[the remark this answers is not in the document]
ron-gal: Done

bigtable-client-core-parent/bigtable-hbase/src/main/java/com/google/cloud/bigtable/hbase/AbstractBigtableTable.java
ron-gal: Done

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestDataClientVeneerApi.java
```diff
@@ -194,6 +199,28 @@ public void testReadRowAsync() throws Exception {
         .futureCall(Mockito.eq(expectedRequest), Mockito.any(GrpcCallContext.class));
   }
 
+  @Test
+  public void testReadRows_Errors() throws IOException {
+    Query query = Query.create(TABLE_ID).rowKey(ROW_KEY);
+    when(mockDataClient.readRowsCallable(Mockito.<RowResultAdapter>any()))
+        .thenThrow(new RuntimeException())
```
igorbernstein2, line 206: this doesnt make sense. readRowsCallable() will never throw an exception. You want to test what happens when the server returns an error. I think something like this:

```java
  public void testReadRows_Errors() throws IOException {
    Query query = Query.create(TABLE_ID).rowKey(ROW_KEY);
    when(mockDataClient.readRowsCallable(any(RowResultAdapter.class)))
        .thenReturn(mockStreamingCallable);
    when(mockStreamingCallable.call(any(Query.class), any(GrpcCallContext.class)))
        .thenReturn(serverStream);
    when(serverStream.iterator())
        .thenReturn(new Iterator<Result>() {
          @Override
          public boolean hasNext() {
            return true;
          }
          @Override
          public Result next() {
            throw new InternalException("fake error", null, GrpcStatusCode.of(Code.INTERNAL), false);
          }
        })
        .thenReturn(ImmutableList.<Result>of().iterator());

    assertThrows(Exception.class, () -> dataClientWrapper.readRows(query).next());

    ResultScanner noRowsResultScanner = dataClientWrapper.readRows(query);
    assertNull(noRowsResultScanner.next());
    noRowsResultScanner.close();

    verify(mockDataClient, times(2)).readRowsCallable(Mockito.<RowResultAdapter>any());
    verify(serverStream, times(2)).iterator();
    verify(mockStreamingCallable, times(2))
        .call(any(Query.class), any(GrpcCallContext.class));
  }
```

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestDataClientVeneerApi.java
```diff
@@ -204,30 +234,153 @@ public void testReadRows() throws IOException {
...
+    Query query = Query.create(TABLE_ID).rowKey(ROW_KEY);
+    when(mockDataClient.readRowsCallable(Mockito.<RowResultAdapter>any()))
+        .thenReturn(mockStreamingCallable)
+        .thenReturn(mockStreamingCallable);
+
+    doAnswer(
+            new Answer<Void>() {
+              @Override
+              public Void answer(InvocationOnMock invocation) throws Throwable {
+                ((ResponseObserver) invocation.getArgument(1)).onResponse(Result.EMPTY_RESULT);
```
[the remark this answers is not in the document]
igorbernstein2: This test doesnt really make sense. You want to be testing what happens between pages. Something like this:

```java
private static Result createRow(String key) {
    return Result.create(
        ImmutableList.<Cell>of(
            new com.google.cloud.bigtable.hbase.adapters.read.RowCell(
                Bytes.toBytes(key),
                Bytes.toBytes("cf"),
                Bytes.toBytes("q"),
                10L,
                Bytes.toBytes("value"),
                ImmutableList.of("label"))));
  }
  @Test
  public void testReadPaginatedRows() throws IOException {
    Query query = Query.create(TABLE_ID).range("a", "z");
    when(mockDataClient.readRowsCallable(Mockito.<RowResultAdapter>any()))
        .thenReturn(mockStreamingCallable);

    // First Page
    doAnswer((args) -> {
      System.out.println("first page");
      ResponseObserver<Result> observer = args.getArgument(1);
      observer.onResponse(createRow("a"));
      observer.onResponse(createRow("b"));
      observer.onComplete();
      return null;
    })
        .when(mockStreamingCallable)
        .call(eq(Query.create(TABLE_ID).range("a", "z").limit(2)), any(), any());

    // 2nd Page
    doAnswer((args) -> {
      System.out.println("2nd page");
      ResponseObserver<Result> observer = args.getArgument(1);
      observer.onResponse(createRow("c"));
      observer.onResponse(createRow("d"));
      observer.onComplete();
      return null;
    })
        .when(mockStreamingCallable)
        .call(
            eq(
              Query.create(TABLE_ID)
                .range(ByteStringRange.unbounded().startOpen("b").endOpen("z")).limit(2)),
            any(), any());

    // 3rd Page
    doAnswer((args) -> {
      System.out.println("last page");
      ResponseObserver<Result> observer = args.getArgument(1);
      observer.onResponse(createRow("e"));
      observer.onComplete();
      return null;
    })
        .when(mockStreamingCallable)
        .call(
        eq(
          Query.create(TABLE_ID)
            .range(ByteStringRange.unbounded().startOpen("d").endOpen("z")).limit(2)), any(), any());

    // 3rd Page
    doAnswer((args) -> {
      System.out.println("empty page");
      ResponseObserver<Result> observer = args.getArgument(1);
      observer.onComplete();
      return null;
    })
        .when(mockStreamingCallable)
        .call(
            eq(
                Query.create(TABLE_ID)
                    .range(ByteStringRange.unbounded().startOpen("e").endOpen("z")).limit(2)), any(), any());

    ResultScanner resultScanner = dataClientWrapper.readRows(query.createPaginator(2), 1000);

    assertThat(resultScanner)
        .comparingElementsUsing(Correspondence.transforming((Result r) -> new String(r.getRow()), "row key"))
        .containsExactly("a", "b", "c", "d", "e");
  }
```

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestDataClientVeneerApi.java
```diff
@@ -226,8 +253,112 @@ public void testReadRows() throws IOException {
...
+    verify(mockDataClient, times(2)).readRowsCallable(Mockito.<RowResultAdapter>any());
+    verify(mockStreamingCallable, times(2))
+        .call(
+            Mockito.any(Query.class),
+            Mockito.any(ResponseObserver.class),
+            Mockito.any(GrpcCallContext.class));
+  }
+
+  @Test
+  public void testReadRowsLowMemory() throws IOException {
```
igorbernstein2, line 312: Similar to above, you want to be testing the rpc between pages. so something like this:

```java
 public void testReadRowsLowMemory() throws IOException {
    Query query = Query.create(TABLE_ID);
    when(mockDataClient.readRowsCallable(any(RowResultAdapter.class)))
        .thenReturn(mockStreamingCallable);

    StreamController mockController = Mockito.mock(StreamController.class);
    doAnswer(invocation -> {
          cancelled.set(true);
          return null;
        })
        .when(mockController)
        .cancel();


    // Generate
    doAnswer(
        (Answer<Void>) invocation -> {
          ResponseObserver<Result> observer = invocation.getArgument(1);
          observer.onStart(mockController);

          for(int i=0; i < 1000 && !cancelled.get(); i++) {
            observer.onResponse(createRow(String.format("row%010d", i)));
            Thread.sleep(10);
          }
          observer.onComplete();
          return null;
        })
        .doAnswer(
        (Answer<Void>) invocation -> {
          ResponseObserver<Result> observer = invocation.getArgument(1);
          observer.onComplete();
          return null;
        })
        .when(mockStreamingCallable)
        .call(any(), any(), any());

    ResultScanner resultScanner = dataClientWrapper.readRows(query.createPaginator(100), 3);
    // Consume the stream
    Lists.newArrayList(resultScanner);

    verify(mockStreamingCallable, times(2))
        .call(
            any(Query.class),
            any(ResponseObserver.class),
            any(GrpcCallContext.class));
    assertTrue(cancelled.get());
  }
```

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestDataClientVeneerApi.java
igorbernstein2: This test isnt really testing anything, I would just drop it

ron-gal: Done

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestDataClientVeneerApi.java
ron-gal: Done

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestDataClientVeneerApi.java
ron-gal: Done

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestDataClientVeneerApi.java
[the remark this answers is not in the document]
ron-gal: Done

bigtable-client-core-parent/bigtable-hbase/src/test/java/com/google/cloud/bigtable/hbase/wrappers/veneer/TestDataClientVeneerApi.java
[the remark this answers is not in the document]
ron-gal: Done

mutianf (approved):

igorbernstein2 (approved):

PR #4188, opened 2023-10-16
feat: use concrete classes for hbase.client.registry.impl (#4043)

igorbernstein2:
Previously we used runtime bytecode generation which prevented endusers from using hbase-site.xml to target bigtable in hbase 2 applications


rkaregar (commented):
Do we need any new tests to make sure everything works (e.g., passing the new BigtableConnectionRegistry class or using Async APIs) or they are already covered by the existing tests?

igorbernstein2:
I plan on adding cross version tests, but that will be a much bigger change. I will add them soon

rkaregar (approved):

mutianf (approved):

PR #4207, opened 2023-11-01
deps: unflatten transitive dependencies in bigtable-hbase

igorbernstein2:
Since bigtable-hbase is just a bridge between hbase-client and google-cloud-bigtable ecosystems, we should defer to those artifacts to dictate the transitive dependencies. This change  makes this explicit by only depending on bigtable & hbase artifacts

Please note that this includes #4205

### DIFF
--- a/bigtable-client-core-parent/bigtable-hbase/pom.xml
+++ b/bigtable-client-core-parent/bigtable-hbase/pom.xml
@@ -51,106 +51,7 @@ limitations under the License.
   </dependencyManagement>
 
   <dependencies>
-    <!-- List implementation deps first -->
-    <dependency>
-      <groupId>${project.groupId}</groupId>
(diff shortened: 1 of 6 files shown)


bigtable-client-core-parent/bigtable-hbase/pom.xml
```diff
@@ -181,6 +93,11 @@ limitations under the License.
       <artifactId>grpc-google-common-protos</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.api.grpc</groupId>
+      <artifactId>grpc-google-cloud-bigtable-admin-v2</artifactId>
```
mutianf, line 98: is this not included in google-cloud-bigtable?

mutianf (approved):

PR #4223, opened 2023-11-02
deps: unflatten deps for 1x shaded and hadoop

igorbernstein2:
* order deps as: hbase, then bigtable, then bigtable-hbase
* mark all annotation deps as provided
* mark conscrypt as provided, which is optional as provided, end users will need to add it themselves if it is needed
* some formatting changes
* ensure that dep tree of bigtable & hbase are separately respected

At this point both of the artifacts (w/o any other deps) can be dropped into the HBASE_CLASSPATH and allow the end user to connect to bigtable

### DIFF
--- a/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-hadoop/pom.xml
+++ b/bigtable-hbase-1.x-parent/bigtable-hbase-1.x-hadoop/pom.xml
@@ -30,11 +30,6 @@ limitations under the License.
     bigtable-hbase-1.x.
   </description>
 
-  <properties>
-    <!-- define a property that can be ignored by renovate -->
-    <hbase1-hadoop-slf4j.version>1.6.1</hbase1-hadoop-slf4j.version>
-  </properties>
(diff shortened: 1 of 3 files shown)


bigtable-hbase-1.x-parent/bigtable-hbase-1.x-hadoop/pom.xml
```diff
@@ -112,34 +96,31 @@ limitations under the License.
...
+      <plugins>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-enforcer-plugin</artifactId>
+          <executions>
+            <execution>
+              <id>enforce</id>
+              <!-- requireUpperBoundDeps, banDuplicateClasses -->
+              <configuration>
+                <skip>true</skip>
```
mutianf, line 109: are we still skipping this plugin? can it be cleaned up?

mutianf (approved):

PR #4242, opened 2023-11-15
chore: clean up dependency enforcement checks

igorbernstein2:
* remove version consistency checks, instead we now rely on dependency mirroring checks to ensure that deps are coherent
* exclude provided & test deps from the bannedDeps rule
* remove commons-codec & commons-compress from the banned list, we have no control over them: compress comes from beam and codec comes from hbase. Neither is used by this project
* fix class conflict in the import job for commons-logging
* switch to using shaded version of gcs-connector to avoid conflicts with beam
* disable all enforcement checks in -hadoop jars, they simply repackage -shaded jars so any enforcement failures are in the hbase-client domain and are out of our control
* fix version alignment in mirroring client integration tests

### DIFF
--- a/hbase-migration-tools/mirroring-client/bigtable-hbase-mirroring-client-2.x-parent/bigtable-hbase-mirroring-client-1.x-2.x-integration-tests/pom.xml
+++ b/hbase-migration-tools/mirroring-client/bigtable-hbase-mirroring-client-2.x-parent/bigtable-hbase-mirroring-client-1.x-2.x-integration-tests/pom.xml
@@ -262,7 +262,8 @@ limitations under the License.
     <dependency>
       <groupId>io.prometheus</groupId>
       <artifactId>simpleclient_httpserver</artifactId>
-      <version>0.16.0</version>
+      <!-- Should match prometheus version used by opencensus-exporter-stats-prometheus -->
+      <version>0.6.0</version>
     </dependency>
 
     <!-- Testing deps -->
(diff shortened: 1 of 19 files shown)


hbase-migration-tools/mirroring-client/bigtable-hbase-mirroring-client-2.x-parent/bigtable-hbase-mirroring-client-1.x-2.x-integration-tests/pom.xml
```diff
@@ -262,7 +262,8 @@ limitations under the License.
     <dependency>
       <groupId>io.prometheus</groupId>
```
mutianf, line 263: where is this used?

mutianf (approved):

hbase-migration-tools/mirroring-client/bigtable-hbase-mirroring-client-2.x-parent/bigtable-hbase-mirroring-client-1.x-2.x-integration-tests/pom.xml
igorbernstein2: Its used in a rule for tests metrics export:
https://github.com/googleapis/java-bigtable-hbase/blob/ff97ac926c00aa3c09f4ed9ceba6a017d898e695/hbase-migration-tools/mirroring-client/bigtable-hbase-mirroring-client-1.x-parent/bigtable-hbase-mirroring-client-1.x-integration-tests/src/test/java/com/google/cloud/bigtable/hbase/mirroring/utils/PrometheusStatsCollectionRule.java#L42-L45

I'm not entirely sure where that rule is actually used, but for now I just wanna stick to dependency handling as much as possible

PR #4270, opened 2023-12-06
test: push hbase minicluster out to a separate classpath

igorbernstein2:
This should allow for much more accurate testing with a dependency tree that matches the real world.
Previously the integration tests would pull in hbase testing utility, which dump the entire hbase & hadoop implementation on the classpath. With the new approach, the testing utility is isolated in a maven plugin classloader, which allows the test classpath to be a lot cleaner

### DIFF
--- a/bigtable-test/hbase-minicluster-maven-plugin/src/main/java/com/google/cloud/bigtable/hbase_minicluster/Controller.java
+++ b/bigtable-test/hbase-minicluster-maven-plugin/src/main/java/com/google/cloud/bigtable/hbase_minicluster/Controller.java
@@ -44,7 +44,7 @@ class Controller {
     try {
       miniHBaseCluster = helper.startMiniCluster();
     } catch (Exception e) {
-
+      throw new MojoExecutionException("Failed to start mini cluster", e);
     }
 
     int port =


bigtable-test/hbase-minicluster-maven-plugin/pom.xml
```diff
@@ -0,0 +1,107 @@
...
+        <artifactId>bigtable-test</artifactId>
+        <version>2.12.1-SNAPSHOT</version> <!-- {x-version-update:bigtable-client-parent:current} -->
+    </parent>
+
+
+    <artifactId>hbase-minicluster-maven-plugin</artifactId>
+    <packaging>maven-plugin</packaging>
+
+    <properties>
+        <maven.compiler.source>11</maven.compiler.source>
```
mutianf, line 17: do we need to set this?

bigtable-test/pom.xml
```diff
@@ -34,5 +34,6 @@ limitations under the License.
     <module>bigtable-emulator-maven-plugin</module>
     <module>bigtable-build-helper</module>
     <module>bigtable-internal-test-helper</module>
+      <module>hbase-minicluster-maven-plugin</module>
```
mutianf, line 37: nit: format

bigtable-test/hbase-minicluster-maven-plugin/pom.xml
igorbernstein2: nope, good catch...ide added it and I didnt notice

PR #4280, opened 2023-12-18
test: add tests for different bigtable-habse packaging

igorbernstein2:
bigtable-hbase has 2 primary configuration that are intended to work hbase-client and hbase-shaded-client respectively. This PR add integration test profiles to ensure they are both tested properly. In addition this adds a 3rd profile that shows how to run bigtable-hbase-1x-shaded with hbase-client while avoiding known CVEs in hbase-client

Other changes:
- remove tests for JWT since its no longer relevant to bigtable-hbase and is tests in java-bigtable/gax
- remove tests that bigtable specific surfaces (BIgtableFilter). These are deprecated as they expose shaded internals


mutianf (approved):

mutianf (approved):

bigtable-hbase-1.x-parent/bigtable-hbase-1.x-integration-tests/pom.xml
```diff
@@ -35,9 +35,292 @@ limitations under the License.
...
+
+            <!-- This is only necessary modules built in the java-bigtable-hbase repo. End user
+            applications don't need these exclusions.
+            Workaround MNG-5899 & MSHADE-206. Maven >= 3.3.0 doesn't use the dependency reduced
+            pom.xml files when invoking the build from a parent project. So we have to manually exclude
+            the dependencies. Note that this works in conjunction with the manually promoted dependencies
+            in bigtable-hbase-1.x-shaded/pom.xml -->
+            <exclusion>
+              <groupId>com.google.cloud.bigtable</groupId>
+              <artifactId>bigtable-hbase-1.x</artifactId>
```
mutianf, line 304: the name of this mode is cbt-shaded, should this be the shaded client?

igorbernstein2: no, it needs both :(

The -shaded one needs to be excluded for the reasons you are thinking (because it is a legitimate dependency). This is done on L293

The non-shaded one (this one) needs to be excluded because maven models are immutable and the shade plugin cant remove the shaded dep. In other words  `bigtable-hbase-1.x-shaded` depends on `bigtable-hbase-1.x`. During packaging, `bigtable-hbase-1.x` gets absorbed into `bigtable-hbase-1.x-shaded` however, the shade plugin can't remove it as a dependency inside the reactor. So within the reactor, another module (like this) when it depends on `bigtable-hbase-1.x-shaded` will get both the absorbed bytecode and the original dep on `bigtable-hbase-1.x`. 

This is only problem within the reactor, and end users dont need to worry about this.  (see the comment above about MNG-5899 & MSHADE-206)

PR #4319, opened 2024-03-28
feat: add logic in replication for handling rows/cells that may exceed thresholds

jhambleton:
adding capabilities to replication library to control the shape of data as it is replicated to bigtable. the controls are disabled by default, and when enabled, allow users to filter rows and cells which may block replication. The configurations have unit tests as well have been run through a series of tests for a live deployment.


jhambleton:
Reference, main branch PR: https://github.com/googleapis/java-bigtable-hbase/pull/4315

PR #4325, opened 2024-03-29
chore: fix maven version to 2.13.0

igorbernstein2:
This was accidentally released to maven as 2.12.1 because git & maven tagging went out of sync. Resetting the versions to force a release to maven manually

Change-Id: I5506672667adc8518b58c0d64471df7e838c1e18

### DIFF
--- a/versions.txt
+++ b/versions.txt
@@ -1,6 +1,6 @@
 # Format:
 # module:released-version:current-version
 
-bigtable-client-parent:2.12.1:2.13.0
-bigtable-hbase-replication:1.11.1:1.11.2-SNAPSHOT
-bigtable-hbase-mirroring:0.7.1:0.7.2-SNAPSHOT
+bigtable-client-parent:2.13.0:2.13.0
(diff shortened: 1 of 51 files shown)


versions.txt
```diff
@@ -1,6 +1,6 @@
 # Format:
 # module:released-version:current-version
 
-bigtable-client-parent:2.12.1:2.12.2-SNAPSHOT
+bigtable-client-parent:2.12.1:2.13.0
 bigtable-hbase-replication:1.11.1:1.11.2-SNAPSHOT
```
chingor13, line 5: are all these released at the same time? We might need to bump these as well otherwise some artifacts may be referencing -SNAPSHOT versions

chingor13 (changes requested):

chingor13 (approved):

versions.txt
igorbernstein2: fixed

mutianf (approved):

PR #4360, opened 2024-05-16
fix: Remove duplicated calls to proto oneof builders

### LINKED ISSUE #4359: OneOf is set multiple times with different values
These calls to OneOf setters in `BigtableTableAdminClientWrapper.java` overwrite previously set values:

https://github.com/googleapis/java-bigtable-hbase/blob/d3403e14cf0463d39e71d1af62a81f3c4110d125/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableTableAdminClientWrapper.java#L240

https://github.com/googleapis/java-bigtable-hbase/blob/d3403e14cf0463d39e71d1af62a81f3c4110d125/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableTableAdminClientWrapper.java#L258

This was noticed because of an [Error Prone static check](https://errorprone.info/bugpattern/RedundantSetterCall), which reports the following error for that code:

```
src/main/java/com/google/cloud/bigtable/grpc/BigtableTableAdminClientWrapper.java:240: error: [RedundantSetterCall] The oneof `Target` (set via setDeleteAllDataFromTable, setRowKeyPrefix) was called twice with . Setting the same field multiple times is redundant, and could mask a bug.
            .setRowKeyPrefix(rowKeyPrefix)
                             ^
    (see https://errorprone.info/bugpattern/RedundantSetterCall)
src/main/java/com/google/cloud/bigtable/grpc/BigtableTableAdminClientWrapper.java:258: error: [RedundantSetterCall] The oneof `Target` (set via setDeleteAllDataFromTable, setRowKeyPrefix) was called twice with . Setting the same field multiple times is redundant, and could mask a bug.
            .setRowKeyPrefix(rowKeyPrefix)
                             ^
    (see https://errorprone.info/bugpattern/RedundantSetterCall)
```

cushon:
This change removes duplicated calls to proto oneof builders where a previously set value is overwritten in the fluent chain.

Fixes #4359 ☕️

### DIFF
--- a/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableTableAdminClientWrapper.java
+++ b/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/BigtableTableAdminClientWrapper.java
@@ -236,7 +236,6 @@ public class BigtableTableAdminClientWrapper implements IBigtableTableAdminClien
     DropRowRangeRequest protoRequest =
         DropRowRangeRequest.newBuilder()
             .setName(instanceName.toTableNameStr(tableId))
-            .setDeleteAllDataFromTable(false)
             .setRowKeyPrefix(rowKeyPrefix)
             .build();
     delegate.dropRowRange(protoRequest);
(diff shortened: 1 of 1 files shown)


igorbernstein2 (approved):
thank you!

PR #4438, opened 2024-09-04
feat: add Lineage metrics for CloudBigtableIO

ad548:

### DIFF
--- a/bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/it/CloudBigtableBeamIT.java
+++ b/bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/it/CloudBigtableBeamIT.java
@@ -17,6 +17,8 @@ package com.google.cloud.bigtable.beam.it;
 
 import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_ADMIN_HOST_KEY;
 import static com.google.cloud.bigtable.hbase.BigtableOptionsFactory.BIGTABLE_HOST_KEY;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasItem;
 
 import com.google.cloud.bigtable.beam.CloudBigtableIO;
 import com.google.cloud.bigtable.beam.CloudBigtableScanConfiguration;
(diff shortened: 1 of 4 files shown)


bigtable-dataflow-parent/bigtable-beam-import/src/test/java/com/google/cloud/bigtable/beam/it/CloudBigtableBeamIT.java
```diff
@@ -142,8 +147,11 @@ public void testWriteToBigtable() throws IOException {
     DataflowPipelineOptions options = PipelineOptionsFactory.as(DataflowPipelineOptions.class);
     properties.applyTo(options);
     options.setAppName("testWriteToBigtable-" + System.currentTimeMillis());
+    options.setExperiments(Collections.singletonList("disable_runner_v2"));
```
ad548, line 150: runner v2 (which is used by default) doesn't have the support for retrieving lineage metrics yet.

ad548: We'll wait for the support to be added before moving forward with these changes. Confirmed that changes work fine otherwise.

ad548: The support has been added and we can remove this now.

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableIO.java
```diff
@@ -300,7 +303,6 @@ public void validate() {
      *
      * @param options The pipeline options.
      * @return The estimated size of the data, in bytes.
-     * @throws IOException
```
andre-sampaio, line 303: Is removing this intentional?

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableIO.java
```diff
@@ -342,8 +344,6 @@ protected long calculateEstimatedSizeBytes(PipelineOptions options) throws IOExc
     /**
      * Splits the region based on the start and stop key. Uses {@link Bytes#split(byte[], byte[],
      * int)} under the covers.
-     *
-     * @throws IOException
```
andre-sampaio, line 346: Is removing this intentional?

andre-sampaio:
Can you also add some brief documentation to https://github.com/googleapis/java-bigtable-hbase/blob/main/bigtable-dataflow-parent/bigtable-beam-import/README.md?

andre-sampaio:
IIUC this would break existing workloads for projects without the lineage API enabled, is this correct?

If so we need to provide this as an option (probably defaulted to `false`). It might still be a good idea to have this behind a flag anyway

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableIO.java
```diff
@@ -906,15 +916,26 @@ public String toString() {
           Bytes.toStringBinary(rangeTracker.getStartPosition().getBytes()),
           Bytes.toStringBinary(rangeTracker.getStopPosition().getBytes()));
     }
+
+    void reportLineageOnce() {
+      if (!reportedLineage) {
```
andre-sampaio, line 921: IIUC this is just an optmization right? i.e. if it does happen that this gets reported twice it will just be some wasted work?

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableIO.java
ad548: this may have been the from the format plugin. I added a description to the tag so it doesn't complain

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableIO.java
ad548: same here

ad548:
> Can you also add some brief documentation to https://github.com/googleapis/java-bigtable-hbase/blob/main/bigtable-dataflow-parent/bigtable-beam-import/README.md?

That's a good idea! I added some documentation.

bigtable-dataflow-parent/bigtable-hbase-beam/src/main/java/com/google/cloud/bigtable/beam/CloudBigtableIO.java
ad548: Yes that's right!

ad548:
> IIUC this would break existing workloads for projects without the lineage API enabled, is this correct?
> 
> If so we need to provide this as an option (probably defaulted to `false`). It might still be a good idea to have this behind a flag anyway

Good question. So this feature is implemented using Beam metrics that are always emitted by Beam (alongside any number of other metrics) and sent over to Dataflow backend. If lineage is not enabled Dataflow drops those metrics in backend. So it should be fine to always report these. Other systems/IOs are doing a similar thing.

andre-sampaio (approved):

PR #4451, opened 2024-10-18
fix: Bring SampleRowKeys default timeout closer in line to the Veneer…

### LINKED ISSUE #4450: Sample row keys timeout is too short
Sample row keys uses the same default timeouts as other unary operations, which is too short for large enough tables. This should be closer to the default settings for the Veneer client

andre-sampaio:
This brings the default timeout for SampleRowKeys to the same values as the Veneer client

Fixes #4450 ☕️


sushanb (approved):

sushanb:
please add a test into TestBigtableHBaseVeneerSettings.java file and then i think it should be good to be merged.

mutianf (approved):

PR #4606, opened 2026-04-30
feat(import): add script tool for multiple hbase snapshot imports)

tianlei2:

[b/429250716](https://b.corp.google.com/issues/429250716)

This is the second PR with the shell tool script. This a child of https://github.com/googleapis/java-bigtable-hbase/pull/4600, so the only file that needs to be reviewed is the script itself (bigtable-dataflow-parent/bigtable-beam-import/run-snapshot-import.sh). If this looks ok, I will merge this with the parent PR.

The script is originally from https://docs.google.com/document/d/1T7BQF-AYY8xGbdbhwkV7zPA11A_Fh5SRjA_7nkzeTos/edit?tab=t.twzz7rrh6jz7#heading=h.m7p3xy7m76u2

### DIFF
--- a/bigtable-dataflow-parent/bigtable-beam-import/SNAPSHOT_IMPORT_USAGE.md
+++ b/bigtable-dataflow-parent/bigtable-beam-import/SNAPSHOT_IMPORT_USAGE.md
@@ -79,6 +79,8 @@ bin/run-snapshot-import.sh 12 15 &
 bin/run-snapshot-import.sh 16 19 &
 ```
 
+> [!NOTE]
+> Unlike `--all` or `--restore-only`, the script does **not** print a cleanup reminder at the end of a manual range invocation. This omission is intentional to prevent premature deletion: you must manually delete the restore directory (`gsutil rm -r $RESTORE_DIR`) only after **all** parallel Dataflow jobs have successfully finished and data validation passes.
 ## Troubleshooting
 
 ### JDK Compatibility
--- a/bigtable-dataflow-parent/bigtable-beam-import/SNAPSHOT_IMPORT_USAGE.md
+++ b/bigtable-dataflow-parent/bigtable-beam-import/SNAPSHOT_IMPORT_USAGE.md
@@ -1,6 +1,6 @@
 # HBase Snapshot Import Helper Script Usage
 
-This document describes the environment variables used by the `run-snapshot-import.sh` script to automate HBase snapshot imports into Cloud Bigtable using Dataflow.
+This document describes the environment variables used by the `bin/run-snapshot-import.sh` script to automate HBase snapshot imports into Cloud Bigtable using Dataflow.
 
 ## Environment Variables
 
@@ -22,17 +22,31 @@ The script relies on the following environment variables. You should set them be
 | `NETWORK` | VPC Network name for Dataflow workers. | `your-network` |
 | `SUBNETWORK` | VPC Subnetwork name for Dataflow workers. | `regions/us-central1/subnetworks/your-subnetwork` |
 
+## Understanding Sharding
+
+An HBase snapshot consists of multiple physical **regions** (ordered chunks of table keys). To migrate extremely large tables efficiently, you can partition (shard) the regions across multiple independent Dataflow jobs.
+
+*   **How Sharding Works:** The import tool distributes regions uniformly across a total number of shards (`NUM_SHARDS`) by hashing each region's encoded name and taking it modulo `NUM_SHARDS`.
+*   **Targeted Shards:** A Dataflow job configured with a specific `shardIndex` only reads and imports the HBase regions assigned to that particular shard.
+*   **Why use Sharding:** Instead of running a single massive Dataflow job that might hit quota limits or VM memory constraints, you can split the import into multiple smaller, parallelizable jobs.
+
 ## Usage
 
 ### Run a specific shard range
+
+Runs a range of shard indices sequentially within the script execution.
+
+> [!NOTE]
+> Both `<start_shard>` and `<end_shard>` are **inclusive**. E.g., running `bin/run-snapshot-import.sh 0 5` will submit and execute Dataflow jobs for shards `0, 1, 2, 3, 4, and 5` sequentially.
+
 ```bash
-./run-snapshot-import.sh <start_shard> <end_shard>
+bin/run-snapshot-import.sh <start_shard> <end_shard>
 ```
-Example: `./run-snapshot-import.sh 0 5`
+Example: `bin/run-snapshot-import.sh 0 5`
 
 ### Run all shards (Auto-parallel mode)
 ```bash
-./run-snapshot-import.sh --all
+bin/run-snapshot-import.sh --all
 ```
 This mode will first run the restore step, and then launch background processes for all shards in parallel groups of 4 by default.
 
@@ -48,18 +62,18 @@ To run shards in parallel groups (e.g., assuming 20 shards total), you can run m
 Example for manual parallel execution:
 ```bash
 # 1. Run the blocking restore step first!
-./run-snapshot-import.sh --restore-only
+bin/run-snapshot-import.sh --restore-only
 
 # 2. Once the restore is complete, launch shards in parallel:
-./run-snapshot-import.sh 0 3 &
-./run-snapshot-import.sh 4 7 &
-./run-snapshot-import.sh 8 11 &
-./run-snapshot-import.sh 12 15 &
-./run-snapshot-import.sh 16 19 &
+bin/run-snapshot-import.sh 0 3 &
+bin/run-snapshot-import.sh 4 7 &
+bin/run-snapshot-import.sh 8 11 &
+bin/run-snapshot-import.sh 12 15 &
+bin/run-snapshot-import.sh 16 19 &
 ```
 
 ## Troubleshooting
 
 ### JDK Compatibility
 
-If you are running on a newer, unsupported JDK version (e.g. JDK 25+) and hit ByteBuddy errors, you can simply set the `JVM_OPTS` variable to `"-Dnet.bytebuddy.experimental=true"` in the Configurations section of the `run-snapshot-import.sh` script.
+If you are running on a newer, unsupported JDK version (e.g. JDK 25+) and hit ByteBuddy errors, you can simply set the `JVM_OPTS` variable to `"-Dnet.bytebuddy.experimental=true"` in the Configurations section of the `bin/run-snapshot-import.sh` script.
--- a/bigtable-dataflow-parent/bigtable-beam-import/SNAPSHOT_IMPORT_USAGE.md
+++ b/bigtable-dataflow-parent/bigtable-beam-import/SNAPSHOT_IMPORT_USAGE.md
@@ -6,21 +6,21 @@ This document describes the environment variables used by the `bin/run-snapshot-
 
 The script relies on the following environment variables. You should set them before executing the script.
 
-| Variable | Description | Example / Suggested Value |
-| :--- | :--- | :--- |
-| `PROJECT_ID` | The Google Cloud Project ID where the Bigtable instance and Dataflow jobs reside. | `your-project-id` |
-| `INSTANCE_ID` | The Bigtable Instance ID to import data into. | `your-instance-id` |
-| `BUCKET` | The GCS bucket name used for Dataflow staging, temp files, and default snapshot source path. | `your-gcs-bucket` |
-| `REGION` | The GCP region to run the Dataflow jobs in. | `us-central1` |
-| `TABLE_NAME` | The target Bigtable table name. | `your-table-name` |
-| `SNAPSHOT_NAME` | The name of the HBase snapshot to import. | `your-snapshot-name` |
-| `SNAPSHOT_SOURCE_DIR` | The GCS path where the HBase snapshot export is located. | `gs://your-gcs-bucket/snapshots` |
-| `SERVICE_ACCOUNT` | The service account email to run the Dataflow jobs. | `your-service-account@developer.gserviceaccount.com` |
-| `NUM_SHARDS` | The number of shards to split the import into for parallel processing. | `20` |
-| `MAX_INFLIGHT_RPCS` | Maximum number of inflight RPCs for Bigtable client. | `100` |
-| `BULK_MUTATION_CLOSE_TIMEOUT_MINUTES` | Timeout in minutes for closing bulk mutations. | `30` |
-| `NETWORK` | VPC Network name for Dataflow workers. | `your-network` |
-| `SUBNETWORK` | VPC Subnetwork name for Dataflow workers. | `regions/us-central1/subnetworks/your-subnetwork` |
+| Variable | Status | Description | Default / Example |
+| :--- | :--- | :--- | :--- |
+| `PROJECT_ID` | **Required** | The Google Cloud Project ID where the Bigtable instance and Dataflow jobs reside. | `your-project-id` |
+| `INSTANCE_ID` | **Required** | The Bigtable Instance ID to import data into. | `your-instance-id` |
+| `BUCKET` | **Required** | The GCS bucket name used for Dataflow staging, temp files, and default snapshot source path. | `your-gcs-bucket` |
+| `REGION` | **Required** | The GCP region to run the Dataflow jobs in. | `us-central1` |
+| `TABLE_NAME` | **Required** | The target Bigtable table name. | `your-table-name` |
+| `SNAPSHOT_NAME` | **Required** | The name of the HBase snapshot to import. | `your-snapshot-name` |
+| `SNAPSHOT_SOURCE_DIR` | **Required** | The GCS path where the HBase snapshot export is located. | `gs://your-gcs-bucket/snapshots` |
+| `NUM_SHARDS` | **Required** (except `--restore-only`) | The number of shards to split the import into for parallel processing. | `20` |
+| `SERVICE_ACCOUNT` | *Optional* | The service account email to run the Dataflow jobs. | `your-service-account@developer.gserviceaccount.com` |
+| `MAX_INFLIGHT_RPCS` | *Optional* | Maximum number of inflight RPCs for Bigtable client. | `100` (Default) |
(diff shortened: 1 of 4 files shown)


bigtable-dataflow-parent/bigtable-beam-import/run-snapshot-import.sh
```diff
@@ -0,0 +1,139 @@
...
+    echo "Usage: $0 <start_shard> <end_shard>"
+    echo "   Or: $0 --all"
+    exit 1
+fi
+
+START_SHARD=$1
+END_SHARD=$2
+
+# Configurations (Uses environment variables if set, otherwise defaults)
+export PROJECT_ID="${PROJECT_ID:-google.com:cloud-bigtable-dev}"
```
vermas2012: can't have defaults here.

bigtable-dataflow-parent/bigtable-beam-import/run-snapshot-import.sh
```diff
@@ -0,0 +1,139 @@
...
+export INSTANCE_ID="${INSTANCE_ID:-tianlei-test-inst}"
+export BUCKET="${BUCKET:-tianlei-beam-test-bucket}"
+export REGION="${REGION:-us-central1}"
+
+export TABLE_NAME="${TABLE_NAME:-validation_test}"
```
vermas2012: where is the doc that shows what are available env variables and what is expected to be set there?

bigtable-dataflow-parent/bigtable-beam-import/run-snapshot-import.sh
```diff
@@ -0,0 +1,142 @@
...
+# ----------------------------------------
+
+# Standard Range Mode
+for i in $(seq $START_SHARD $END_SHARD); do
+  echo "Submitting Dataflow job for shardIndex: $i"
+
+  # We skip restore for all shards if running via --all because Step 1 handled it.
+  # If running manually via ranges, shard 0 will perform restore.
+  SKIP_RESTORE="true"
+  if [ $i -eq 0 ]; then
```
vermas2012: How does this work? If you have --all and have alredy restored the snaphsot, won't this re-restore?

bigtable-dataflow-parent/bigtable-beam-import/bin/run-snapshot-import.sh
```diff
@@ -0,0 +1,142 @@
...
+    --stagingLocation=gs://${BUCKET}/dataflow/staging \
+    --tempLocation=gs://${BUCKET}/dataflow/temp \
+    --workerMachineType=n1-highmem-4 \
+    --diskSizeGb=500 \
+    --maxNumWorkers=10 \
+    --region=${REGION} \
+    --serviceAccount=${SERVICE_ACCOUNT} \
+    --usePublicIps=false \
+    --enableSnappy=true \
+    --skipRestoreStep=${SKIP_RESTORE} \
```
vermas2012: This is good, but how are we passing the restorePath?

vermas2012: I feel like we should have a custom restore path and the script (idempoent by adding timestamp etc) and use it for restore the path and pass it as the restorepath in every job. 

Also, with this model, who cleans up the restore path? is there a way to trigger a cleanup at the end of the script? or we have a tool that can be used? We can also say its a manual step. but then this script should output something to the tune of "the snapshot was imported, please cleanup $RESTORE_PATH once validation succeeds."

bigtable-dataflow-parent/bigtable-beam-import/SNAPSHOT_IMPORT_USAGE.md
```diff
@@ -0,0 +1,62 @@
...
+### Manual Parallel Execution
+
+To run shards in parallel groups (e.g., assuming 20 shards total), you can run multiple instances of this script.
+
+> [!IMPORTANT]
+> Shard 0 performs the restore step. You MUST run the first group (including shard 0) first and let it complete the restore step before launching other groups in parallel. Otherwise, they will fail because the restored files won't exist yet!
+
+Example for manual parallel execution:
+```bash
+./run-snapshot-import.sh 0 3 &  # Run this first!
```
vermas2012: In this model, how does the shard1. know htat it has to wait for restore to finish from shard 0? won't it run and fail becuase restore path is empty?

tianlei2: updating this to run a restore-only command first

bigtable-dataflow-parent/bigtable-beam-import/bin/run-snapshot-import.sh
tianlei2: Added some output for manual cleanups and passing restore path to every job.

bigtable-dataflow-parent/bigtable-beam-import/run-snapshot-import.sh
tianlei2: seems outdated

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
```diff
@@ -334,7 +334,6 @@ limitations under the License.
 		      <exclude>META-INF/*.RSA</exclude>
           <exclude>META-INF/**/pom.properties</exclude>
           <exclude>META-INF/**/pom.xml</exclude>
-          <exclude>META-INF/MANIFEST.MF</exclude>
```
tianlei2, line 337: Removing the MANIFEST.MF exclusion preserves the Multi-Release: true header required to load JDK 21-compatible dependency classes and prevent startup crashes, while remaining fully backward-compatible and safe for JDK 8.

bigtable-dataflow-parent/bigtable-beam-import/bin/run-snapshot-import.sh
```diff
@@ -0,0 +1,188 @@
+#!/bin/bash
```
mutianf, line 1: nit, maybe we put this file under a bin/ or scripts/ directory?

bigtable-dataflow-parent/bigtable-beam-import/run-snapshot-import.sh
```diff
@@ -0,0 +1,188 @@
...
+# export BULK_MUTATION_CLOSE_TIMEOUT_MINUTES="30"
+
+# --- Network Configurations ---
+# export NETWORK="your-network"
+# export SUBNETWORK="your-subnetwork"
+
+# ------------------------------------------------------------------------------
+# Usage
+# ------------------------------------------------------------------------------
+# Usage: ./run-snapshot-import.sh <start_shard> <end_shard>
```
mutianf: Both start and end shard are inclusive, right? Probably good to add a note about it?

bigtable-dataflow-parent/bigtable-beam-import/run-snapshot-import.sh
```diff
@@ -0,0 +1,188 @@
...
+    echo "   Or: $0 --all"
+    echo "   Or: $0 --restore-only"
+    exit 1
+fi
+
+START_SHARD=$1
+END_SHARD=$2
+
+# Configurations
+JAR_PATH="target/bigtable-beam-import-2.18.2-SNAPSHOT-shaded.jar"
```
mutianf: Nit: we shouldn't have the jar using a fixed snapshot version. Also, we should probably build the jar first?

bigtable-dataflow-parent/bigtable-beam-import/run-snapshot-import.sh
```diff
@@ -0,0 +1,188 @@
...
+RESTORE_DIR="gs://${BUCKET}/restore-${SNAPSHOT_NAME}"
+# If running the launcher on a newer, unsupported JDK version (e.g. JDK 25+),
+# you can set JVM_OPTS to "-Dnet.bytebuddy.experimental=true" to bypass
+# ByteBuddy class generation crashes during pipeline construction.
+JVM_OPTS=""
```
mutianf: But will this override the settings described in the comment?

bigtable-dataflow-parent/bigtable-beam-import/bin/run-snapshot-import.sh
```diff
@@ -0,0 +1,188 @@
...
+      --hbaseSnapshotSourceDir=${SNAPSHOT_SOURCE_DIR} \
+      --snapshots=${SNAPSHOT_NAME}:${TABLE_NAME} \
+      --stagingLocation=gs://${BUCKET}/dataflow/staging \
+      --tempLocation=gs://${BUCKET}/dataflow/temp \
+      --region=${REGION} \
+      --performOnlyRestoreStep=true \
+      --restorePath=${RESTORE_DIR} \
+      --jobName="restore-job" \
+      "${NETWORK_ARGS[@]}"
+    echo "✅ Restore completed."
```
mutianf, line 187: Does the java job has awaitPipelineFinish in the end? Cause otherwise it'll just submit the job and we need to track progress on the dataflow dashboard.

bigtable-dataflow-parent/bigtable-beam-import/run-snapshot-import.sh
mutianf: Also what is a shard?

bigtable-dataflow-parent/bigtable-beam-import/bin/run-snapshot-import.sh
tianlei2: sure

bigtable-dataflow-parent/bigtable-beam-import/run-snapshot-import.sh
tianlei2: added a note here.

bigtable-dataflow-parent/bigtable-beam-import/run-snapshot-import.sh
tianlei2: removing this fixed verision and added a simple dynamic version resolution

bigtable-dataflow-parent/bigtable-beam-import/bin/run-snapshot-import.sh
tianlei2: yeah it is blocking, we have it in ImportJob.ImportOptions

bigtable-dataflow-parent/bigtable-beam-import/run-snapshot-import.sh
tianlei2: modifying this so we can pass via env variable

mutianf (approved):

bigtable-dataflow-parent/bigtable-beam-import/bin/run-snapshot-import.sh
vermas2012: Please add it for all the invocations, I don't think it is printed on the shard 1-4.

bigtable-dataflow-parent/bigtable-beam-import/bin/run-snapshot-import.sh
```diff
@@ -0,0 +1,300 @@
...
+# export NETWORK="your-network"
+# export SUBNETWORK="your-subnetwork"
+
+# ------------------------------------------------------------------------------
+# Usage
+# ------------------------------------------------------------------------------
+# Usage: bin/run-snapshot-import.sh <start_shard> <end_shard>
+#        (Both start and end shard indices are inclusive)
+#   Or:  bin/run-snapshot-import.sh --all
+#        (Runs all shards in parallel groups of 4 by default)
```
vermas2012: This header comment is stale — it still describes the old "groups of 4" model, but the implementation (and SNAPSHOT_IMPORT_USAGE.md) now partition into exactly 4 sequential runner streams. The two aren't equivalent: "groups of 4" caps 4 jobs at a time repeatedly, whereas the current code splits shards into 4 fixed lanes. Please update this line to match the "4 parallel runner streams" model so a reader skimming the script doesn't get the wrong mental model.

bigtable-dataflow-parent/bigtable-beam-import/bin/run-snapshot-import.sh
```diff
@@ -0,0 +1,300 @@
...
+    echo "❌ Error: Shaded JAR not found in ${SCRIPT_DIR}/../target/"
+    echo "   Please compile and package the project first by running:"
+    echo "   mvn clean package -DskipTests"
+    exit 1
+fi
+RESTORE_DIR="gs://${BUCKET}/restore-${SNAPSHOT_NAME}"
+# If running the launcher on a newer, unsupported JDK version (e.g. JDK 25+),
+# you can set JVM_OPTS to "-Dnet.bytebuddy.experimental=true" to bypass
+# ByteBuddy class generation crashes during pipeline construction.
+JVM_OPTS="${JVM_OPTS:-}"
```
vermas2012, line 146: ${JVM_OPTS} is the one intentional unquoted expansion in an otherwise carefully quoted script (it relies on word-splitting to pass multiple flags, which is correct). But shellcheck will flag this as SC2086, and there's no # shellcheck disable pragma in the file — if CI runs shellcheck, these three lines will fail it. Suggest adding a targeted # shellcheck disable=SC2086 directly above each java invocation to document the intent and keep CI green.

vermas2012: I am not sure if we have this shellcheck tool, but good find by AI if it's there.

bigtable-dataflow-parent/bigtable-beam-import/bin/run-snapshot-import.sh
```diff
@@ -0,0 +1,297 @@
...
+    fi
+      
+    echo "Restore completed. Proceeding to data import."
+
+    # Step 2: Launch 4 parallel runner streams
+    echo "Step 2/2: Launching 4 parallel runner streams..."
+    MAX_CONCURRENCY=4
+    SHARDS_PER_RUNNER=$(( (NUM_SHARDS + MAX_CONCURRENCY - 1) / MAX_CONCURRENCY ))
+    
+    pids=()
```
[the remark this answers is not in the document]
vermas2012: [nice-to-fix] The "Terminating background runners" message overpromises. Killing the runner subshells won't reap already-submitted Dataflow jobs — those keep running server-side and can't be cancelled with a local SIGTERM. Consider softening the wording to make clear that any Dataflow jobs already launched will continue and may need manual cancellation via gcloud dataflow jobs cancel. Otherwise an operator may Ctrl-C and wrongly assume everything stopped.

bigtable-dataflow-parent/bigtable-beam-import/bin/run-snapshot-import.sh
```diff
@@ -0,0 +1,300 @@
...
+    --shardIndex="${i}" \
+    --jobName="${JOB}" \
+    "${NETWORK_ARGS[@]}" \
+    --maxInflightRpcs="${MAX_INFLIGHT_RPCS}" \
+    --bulkMutationCloseTimeoutMinutes="${BULK_MUTATION_CLOSE_TIMEOUT_MINUTES}" \
+    "${EXTRA_ARGS[@]}"; then
+      echo "❌ Error: Dataflow job submission failed for shardIndex: ${i}"
+      exit 1
+  fi
+
```
vermas2012, line 301: [should-decide] Both --restore-only and --all print the manual gsutil rm -r "${RESTORE_DIR}" cleanup reminder, but plain range mode does not. Anyone following the manual parallel-execution workflow in the docs (run-restore-only, then several range invocations) finishes their last batch with no cleanup reminder at all. Either print the same notice after this loop, or explicitly document that the reminder only appears in --all/--restore-only. Flagging so it's a conscious choice rather than an accidental gap.

bigtable-dataflow-parent/bigtable-beam-import/bin/run-snapshot-import.sh
```diff
@@ -0,0 +1,300 @@
...
+    if [ "$1" != "--all" ]; then
+        if [ "${START_SHARD}" -ge "${NUM_SHARDS}" ] || [ "${END_SHARD}" -ge "${NUM_SHARDS}" ]; then
+            echo "❌ Error: Shard indices (${START_SHARD}, ${END_SHARD}) must be less than NUM_SHARDS (${NUM_SHARDS})."
+            exit 1
+        fi
+    fi
+fi
+
+# Strip leading gs:// and trailing slashes from BUCKET for robust GCS path construction
+BUCKET="${BUCKET#gs://}"
```
vermas2012, line 105: BUCKET is nicely sanitized (strips gs:// and trailing slash), but SNAPSHOT_SOURCE_DIR is passed straight through to --hbaseSnapshotSourceDir. The asymmetry means a trailing slash or inconsistent prefix on SNAPSHOT_SOURCE_DIR is handled differently from BUCKET. Minor usability wart — consider matching the treatment or adding a comment noting it's intentional.

bigtable-dataflow-parent/bigtable-beam-import/pom.xml
vermas2012: [optional — verify] The rationale for un-excluding MANIFEST.MF (preserving Multi-Release: true for JDK 21 class loading) is sound. One sanity check before merge: since the shade plugin will now merge manifests, confirm the shaded jar doesn't inherit an unexpected Main-Class or stale entries from a dependency — a quick unzip -p target/*-shaded.jar META-INF/MANIFEST.MF should confirm it's clean.

vermas2012 (approved):
Noting that the auto-build behavior was deliberately reversed during the review (an earlier revision ran mvn package when the JAR was missing; it now errors out with build instructions instead). That's a reasonable call — silently building inside a migration launcher is surprising — just flagging that this answers the earlier "should we build the jar first?" thread with an explicit "no, we instruct the user to."

bigtable-dataflow-parent/bigtable-beam-import/bin/run-snapshot-import.sh
tianlei2: done

bigtable-dataflow-parent/bigtable-beam-import/bin/run-snapshot-import.sh
tianlei2: done

